PR #139, opened 2026-05-05
chore(flow-functionality): add @stable and harden curlApiGeneration test

daniellicnerski1:
## Summary

- Replace the weak `clipboardContent.length > 0` assertion with structural checks against the actual curl command (HTTP verb, `/api/v1/run/<UUID>?stream=false`, headers, JSON body keys).
- Force the macOS/Linux platform sub-tab so the generated snippet is deterministic across machines (`getOS()` was driving the previous flakiness — Chromium's `navigator.platform` was producing the PowerShell variant).
- Rename the test (`curl_api_generation` → sentence form), wrap each phase in `test.step()`, drop the misleading `@api` tag (this test does not exercise the REST API), add `@stable`.
- Add the missing spec doc under `docs/flow-functionality/curlApiGeneration.md`.
- Mark coverage as `[x]` in `QA-CHECKLIST.md` and `QA-SCENARIOS-GUIDE.md` scenario 5.1.

## Validation

- [x] Baseline (pre-fix): 3/3 green — but copying the PowerShell variant; the old assertion was hiding it.
- [x] Forced-failure check: replaced `"input_value": "Hello"` with `"ThisShouldFail"` → test fails for the right reason.
- [x] Stability after fix: 3/3 green with `--retries=0`, average 3.7s per run.
- [x] `npx eslint tests/tests-automations/regression/flow-functionality/curlApiGeneration.spec.ts` — clean.
- [x] `npx tsc --noEmit` — clean.
- [x] No `waitForTimeout`, no `if/else`, no `try/catch`; 8 expects per test.

## Test plan

- [ ] `weekly-stable.yml` picks up the test on the next scheduled run via the new `@stable` tag.
- [ ] CI `pr-validation.yml` (typecheck + lint) passes.


daniellicnerski1:
## Copilot review — addressed

The single low-confidence note pointed out a real discrepancy: `QA-SCENARIOS-GUIDE.md` §5.1 step 4 was prescribing `curl -X POST`, while the actual snippet generator and the test assertion use `curl --request POST`. Fixed in e779583 by switching the scenario text to assert the **HTTP method is POST** without pinning the long/short flag form, so the guide tracks whichever variant the Langflow snippet generator emits.

Branch was also merged with the latest `main` (now 204 commits ahead — all from upstream main, plus one new commit for the doc fix). CI should re-run.

### Full revalidation after the change

| Step | Result |
| ---- | ------ |
| `npx eslint curlApiGeneration.spec.ts` | clean (0 errors / 0 warnings) |
| Static checklist (imports, false-positives, selectors, structure) | pass |
| `playwright test … --workers=1 --retries=0` | 1 passed (6.0s) |
| Force-fail (broke `^curl --request POST` → `^curl --xyz-impossivel`) | failed exactly on the modified assertion, then reverted and re-ran clean |
| `--trace=on` | trace.zip generated, steps coherent (7.8s) |
| `🚨 Backend Error` audit | zero occurrences |

PR #193, opened 2026-05-08
fix(ci): isolate flow fixture for LANGFLOW_LOAD_FLOWS_PATH

lice-reis:
## Problem

`migration-upgrade-with-flows` was failing at **Wait for Langflow latest** with:

```
pydantic ValidationError: Field required
input_value={'flows': [{'name': 'Getting Started...'}]}
```

Root cause: the workflow was mounting the entire `tests/assets/flows/` directory as `/flows`. That directory contains `collection.json` which has a `{"flows": [...]}` wrapper format — not a valid individual flow file. `LANGFLOW_LOAD_FLOWS_PATH` tries to load every `.json` in the directory, so it hit this file and crashed.

## Fix

Added a **Prepare flow directory** step that copies only `flow.json` to `/tmp/test-flows/` before starting the containers. Both the `latest` and `nightly` containers now mount `/tmp/test-flows` instead of the full assets directory.

## State of the three migration workflows after this fix

| Workflow | Expected outcome |
|---|---|
| `migration-test` | Fails on flow execution (no `OPENAI_API_KEY` secret) — expected |
| `migration-fresh-install` | Fails on startup — known upstream bug (#150), nightly not fixed yet |
| `migration-upgrade-with-flows` | Should pass Phase 1 now; Phase 2 may expose `UniqueViolation` bug |


lice-reis:
@copilot resolve the merge conflicts in this pull request

PR #229, opened 2026-05-15
test(core-components): nested grouping regression (Issue #189)

### LINKED ISSUE #189: test(core-components): consolidate Nested / Grouping regression coverage
## Context

`core-components/` Core Components — section **3.7 Nested / Grouping** in `QA-CHECKLIST.md` has 2 `[-]`:

> - Nested component
> - Enter and exit grouped component

These may already be partially covered by existing specs in `core-components/` (`toolModeGroup.spec.ts`, `componentHoverAdd.spec.ts`, etc.), but no spec is `@stable` for grouping behavior today.

## Scope (investigation + spec)

**Phase 1 — investigate**
- Map existing tests under `tests/tests-automations/regression/core-components/` and `ui-ux/` that touch grouping
- Determine what's covered and what's missing for: create group, enter group (zoom into nested canvas), exit group, ungroup

**Phase 2 — deliver**

Either:
- (a) Promote an existing test to `@stable` if its coverage is sufficient, or
- (b) Create `nested-grouping-regression.spec.ts` covering: create group from selection, enter group, exit group, ungroup → components return to outer canvas with original positions.

## Tags

`@stable @regression @components` (and `@workspace` if grouping is treated as canvas management).

## Deliverables

- [ ] Investigation report posted as a comment
- [ ] Spec or `@stable` promotion delivered per Phase 2 outcome
- [ ] `docs/core-components/<name>.md` (if a new spec)
- [ ] `QA-CHECKLIST.md` — section 3.7 bullets switched from `[-]` to `[x]`

daniellicnerski1:
## Summary

Closes #189. Adds `nested-grouping-regression.spec.ts` covering the two `Nested / Grouping` bullets in QA-CHECKLIST section 3.7. The original Draft state was blocked by a closure-rebind race in `SelectionMenu`'s `onClick`; this race no longer reproduces on Langflow 1.10.x — verified empirically with **15/15 first-click passes across two batches** (10 with the historical retry loop, 5 without). The helper was simplified to a single click + `toBeHidden` assertion.

## Tests covered

|---|---|---|
| 1 | `box-selecting two connected non-IO components and clicking Group collapses them into a single Group node` | Selecting two connected non-IO components and clicking Group must replace both source nodes with a single `title-Group` node on the outer canvas. The original component titles must be gone — proves the components were nested inside the Group rather than renamed. If this regresses, grouping is broken at the SelectionMenu → `handleGroupNode` → `generateFlow` path. |
| 2 | `ungrouping a Group node restores the original components and the edge between them` | Right-clicking a Group node and triggering Ungroup must re-emit the encapsulated subflow back to the outer canvas — two original components + the edge between them are restored, the Group node is gone. If this regresses, `expandGroupNode` is losing wiring data or the right-click toolbar entry is broken. |

## How the tests were built

- **Custom 2-node flow asset (`tests/assets/flows/two-non-io-connected.json`) instead of a starter template.** The Group button gates on `validateSelection` (`src/frontend/src/utils/reactflowUtils.ts`), which rejects IO components and selections overlapping sticky notes — both present in `Basic Prompting.json`. The asset is a Prompt Template → Language Model subset of that starter with IO and sticky notes stripped and positions explicitly set to `(100, 200)` and `(600, 200)` so `selectAll → Group` is a valid single action.
- **Flow creation via `POST /api/v1/flows/` instead of sidebar drag-and-drop.** Adding two components via the sidebar stacks them on the same default position, making the box-select non-deterministic. API creation with explicit `position` values is the only deterministic path for a multi-node flow on a fresh canvas.
- **Navigation via the home dashboard, not `page.goto("/flow/{id}")`.** Deep-linking to `/flow/{id}` immediately after API creation hits a stale React cache and redirects to the flows list. Clicking the flow card by name forces a fresh owned-flows fetch.
- **`expect(...).toHaveCount(0)` on `.react-flow__node.selected` instead of `waitForTimeout(200)`.** Syncs against the actual selection-cleared state after the pane click rather than a fixed sleep.
- **`toBeHidden` on the Group button as the success signal**, not the click event itself. The button disappears only after `setNodes(...)` commits — that is the actual "grouping happened" signal.
- **Right-click + `group-button-modal` to trigger Ungroup.** Ungroup is only rendered for Group-typed nodes (`isGroup && <SelectItem value="ungroup">` in `nodeToolbarComponent/index.tsx`); the right-click toolbar dropdown is the only UI path to reach it in modern Langflow.

## Dependencies

- No env vars required (no LLM execution).
- Runs in default parallel mode within the file; the two tests are independent because each creates its own flow.
- `afterEach` deletes the created flow via `DELETE /api/v1/flows/{id}` regardless of outcome (silenced on cleanup failure).
- Uses the standard `getAuthToken` helper for the POST authentication.

## What this test does not cover

- Grouping selections that include IO nodes (ChatInput/ChatOutput) — explicitly rejected by `validateSelection` and out of scope.
- Grouping selections that overlap sticky notes — also rejected by `validateSelection`.
- Multi-step nesting (grouping a Group node inside another Group).
- Editing nodes *inside* a Group after collapsing — Langflow does not expose a separate nested-canvas view in 1.10.x; the only path is Ungroup → edit → re-group.
- Grouping more than 2 nodes — the round-trip property is identical and adding more nodes does not exercise additional code paths.

## Known limitations

- The Group button's onClick is a `useCallback` with `[lastSelection, ...]` deps (`PageComponent/index.tsx:303`) — historically prone to a closure-rebind race where the click hit a stale `lastSelection`. The race is not observed on Langflow 1.10.x (15/15 first-click passes) but the underlying React code is unchanged, so a future timing regression upstream could re-introduce it. If that happens, restore the retry approach documented in the helper comment.
- `selectAllNodesBoxDrag` uses the React Flow node bounding boxes + 40px padding to drive a Shift+drag — sensitive to changes in `.react-flow__node` rendering size or the position values in the flow asset.

## Validation

- [x] Ran the tests in isolation with `--trace=on` and verified each `test.step()` appears as a named block in the trace
- [x] Forced a failure (broke the `title-Group` assertion to an impossible testid) and confirmed the failure is attributed to the correct `test.step` ("Assert the canvas collapsed to a single Group node and the original titles are gone")
- [x] Confirmed zero `🚨 Backend Error` in the run output
- [x] Updated `QA-CHECKLIST.md` section 3.7 bullets `[-] → [x]` with spec path references (Coverage Summary table left untouched — auto-generated)
- [x] Added section 28 to `QA-SCENARIOS-GUIDE.md` with the two scenarios in human language (28.1 group, 28.2 ungroup)


daniellicnerski1:
## 📝 Handoff notes — picking this up later

### Where we stopped
Spec passes ~7-9/10 on first attempt against a local Langflow nightly. The Playwright retry layer (`retries: 3` locally, `2` in CI) recovers most of those failures, but the residual rate is too high for `@stable`. **Do not un-draft until the click race below is fixed.**

### Root cause (verified against frontend source)
The Group button (`data-testid="group-node"`) lives inside `SelectionMenuComponent`, rendered by `PageComponent` via `@xyflow/react`'s `<NodeToolbar>` portal. The button's `onClick` is a `useCallback` (`handleGroupNode` at `src/frontend/src/pages/FlowPage/components/PageComponent/index.tsx:302`) whose closure captures `lastSelection` (PageComponent local state).

Sequence:
1. Shift+drag → ReactFlow fires `onSelectionChange` (multiple times during drag) and `onSelectionEnd`.
2. `lastSelection` is set via `setLastSelection` inside `onSelectionChange`.
3. A `useEffect` sets `selectionMenuVisible = true` when `selectionEnded && lastSelection.nodes.length > 1`.
4. `SelectionMenu` renders with `onClick={handleGroupNode}`. The closure for that `handleGroupNode` reference captures whatever `lastSelection` value was current at the render boundary.
5. The button has its own internal `isTransitioning` state (50 ms after `isOpen` flips) and `disable` state (computed only on `[isOpen, setIsOpen]` change — **not** on `lastSelection` change).

Failure mode: even after waiting for `.react-flow__node.selected` count to reach 2, the SelectionMenu's `onClick` closure can be holding a stale reference where `lastSelection.nodes.length < 2`. Clicking fires the handler, `validateSelection` returns errors (or `lastSelection` is empty), and `handleGroupNode` either:
- calls `setErrorData(...)` (error path) without committing the grouping, or
- silently no-ops if `lastSelection` is empty.

Either way the SelectionMenu stays visible and the assertion `expect(groupBtn).toBeHidden(...)` fails.

### Approaches I tried (don't re-do these)
1. **`setupPlayground` + Shift+drag everything** — Group button appears but is disabled (`error-group-node` testid) because ChatInput/ChatOutput are IO nodes (rejected by `validateSelection`).
2. **Basic Prompting template + partial box-select around middle 2 nodes** — picks up the sticky notes too (they overlap the Language Model node), so `validateSelection` returns "Select only connected components" → button stays disabled.
3. **Basic Prompting + shift+click on each node** — `lastSelection` never reaches 2 nodes via shift+click in ReactFlow; the Group button never appears at all.
4. **Custom 2-node flow (Prompt → Language Model, no IO, no notes)** — what's currently in the spec. Selection works, Group button appears enabled, but the click occasionally hits the stale-closure case.
5. **`dispatchEvent('click')` instead of `locator.click()`** — bypasses hit-testing but still hits the same closure race; if anything, slightly less reliable than `click({ force: true })`.
6. **Keyboard shortcut `Mod+G` as fallback** — same `handleGroupNode` closure → same race.
7. **`waitForTimeout(400)` after `node.selected = 2`** — made it **worse** (~2/10 passing). Adding a delay before clicking lets some other selection event fire and re-bind the handler in a stale state, apparently.
8. **Retry loop (5 attempts) re-running `selectAllNodesBoxDrag` between failures** — partial improvement (~9/10) but still not stable enough.

### Recommended fix
Bypass the SelectionMenu entirely. Three viable paths in order of preference:

1. **Expose the React Flow Zustand store on `window` in the dev build** (or a test-only build flag in Langflow) so the spec can call `useFlowStore.getState().setNodes(...)` to mark selection directly. This eliminates the `onSelectionChange` → `lastSelection` → closure-rebind chain entirely. Would require a small Langflow PR.
2. **Invoke `handleGroupNode` through a deterministic path**: dispatch a real `pointerdown`/`pointerup` sequence on the button rather than `click`/`dispatchEvent`, ensuring React's synthetic event system fires after the latest `lastSelection` commit. Worth trying with `page.mouse.down/up` directly at the button's center bounding box.
3. **Use a different multi-select method that updates the store atomically** — e.g., select-all keyboard shortcut (`Mod+A`) if Langflow has one wired to the flow store, or programmatically set `selected: true` on nodes by calling React Flow's imperative API.

### Files & state to know about
- `tests/tests-automations/regression/core-components/nested-grouping-regression.spec.ts` — the spec (current state on this branch).
- `tests/assets/flows/two-non-io-connected.json` — 2-node flow asset, subset of `Basic Prompting.json` with IO and sticky notes stripped, positions explicitly set to `(100, 200)` and `(600, 200)`.
- `triggerGroupMutation()` helper has the retry loop — fix the click stability and you can simplify it back to a single click + `toBeHidden` assertion.
- `selectAllNodesBoxDrag()` already waits for `.react-flow__node.selected` count = 2 before returning — that's load-bearing for any subsequent click attempt; **don't remove it**.

### Local repro
```bash
cd /Users/daniellicnerski/orion-e2e/langflow-e2e
git checkout feat/nested-grouping-regression-189
# Run 10× to characterize current flakiness baseline:
for i in 1 2 3 4 5 6 7 8 9 10; do
  echo "=== Run $i ==="
  npx playwright test tests/tests-automations/regression/core-components/nested-grouping-regression.spec.ts \
    --reporter=line --workers=1 --retries=0 2>&1 | grep -E 'passed|failed'
done
```

### Still TODO (before un-drafting — also in PR body)
1. Stabilize the Group click (root-cause analysis above; pick one of the three recommended fixes).
2. `docs/core-components/nested-grouping-regression.md` (spec doc).
3. Flip `[-]` → `[x]` on the two bullets under section 3.7 in `QA-CHECKLIST.md`.
4. Post Phase 1 investigation comment on issue #189 (the summary from this PR body is good seed material).
5. Once `@stable` is justified, the weekly workflow will start running these — verify on the first weekly cycle that we don't regress.

PR #233, opened 2026-05-18
test(core-components): tag componentHoverAdd as @stable after refactor

daniellicnerski1:
## Summary

- Refactor `componentHoverAdd.spec.ts` from `waitForSelector` + always-true opacity assertion into a deterministic two-assertion contract: initial Plus-icon opacity is `0`, hover + click produces a `.react-flow__node`.
- Drop the post-hover opacity poll (Playwright `expect.poll` interleaves with the CSS transition and produced false negatives — the test name only promises hover + click, which the boundary assertions already cover).
- Add `@stable` to the test tags and add the matching spec doc under `docs/core-components/component-hover-add.md`.

## Validation pipeline (7 steps, all PASS)

1. `npm run typecheck` — clean
2. `npx eslint <spec>` — 0 errors, 0 warnings (was 5 warnings)
3. Anti-pattern checklist — imports, tags, selectors, no false positives
4. `--workers=1 --retries=0` run — PASS (~6 s, 1 test)
5. Force-fail — broke `.react-flow__node` → `.react-flow__nodeXYZ`, failed at the expected line, reverted, repassed
6. `--trace=on` — coherent trace, steps match
7. Backend errors audit — zero `🚨 Backend Error`

## Test plan

- [x] `npx playwright test tests/tests-automations/regression/core-components/componentHoverAdd.spec.ts --workers=1 --retries=0`
- [ ] Weekly stable workflow picks up the new `@stable` tag on next run

### DIFF
--- a/docs/core-components/component-hover-add.md
+++ b/docs/core-components/component-hover-add.md
@@ -1,0 +1,80 @@
+# Spec: Component Hover-to-Add — Plus Icon Affordance
+
+**Test file:** `tests/tests-automations/regression/core-components/componentHoverAdd.spec.ts`
+
+**Last validated:** Langflow 1.10.x
+
+---
+
+## What this test validates
+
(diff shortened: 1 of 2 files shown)


daniellicnerski1:
Closing — will be consolidated into a single bulk PR alongside the other @stable validations.

PR #345, opened 2026-06-01
ci: harden Playwright browser install against CDN download stalls

### LINKED ISSUE #344: CI: Playwright browser install hangs on CDN download, burning the full 90-min job budget
## Problem

The \`Install Playwright browsers\` step (\`npx playwright install chromium --with-deps\`) downloads the Chromium binary from \`cdn.playwright.dev\` on every run. When that download stalls, there is **no per-step timeout and no retry**, so the stalled download consumes the entire 90-minute job budget. The job is then cancelled with *"exceeded the maximum execution time"* — no tests run, no Playwright report is produced, and the real cause is masked.

### Observed

- Run [26760434237](https://github.com/oriontech-me/langflow-e2e/actions/runs/26760434237) — `weekly-stable`, hung downloading `chrome-linux64.zip`, cancelled at 1h30m. No tests executed.
- Run [26782003123](https://github.com/oriontech-me/langflow-e2e/actions/runs/26782003123) — re-run, hung at the **same step** for >20 min (cancelled manually). Confirms it is not a one-off flake.

Repo visibility (private → public) is unrelated: it changes billing/minutes, not the runner's network path to the Playwright CDN.

## Affected workflows

Every workflow that installs Playwright browsers shares the same latent bug:
- \`weekly-stable.yml\`
- \`nightly.yml\`
- \`manual.yml\` (two jobs)
- \`adaptive-impacted.yml\`
- \`migration-test.yml\` (Python toolchain)

## Proposed fix

1. **Cache** \`~/.cache/ms-playwright\` keyed on the Playwright version → skip the download on cache hit.
2. **Per-attempt \`timeout\` + retry** → a stalled download is killed (~5 min) and retried, recovering from transient CDN stalls.
3. **Fail fast** → if all attempts fail, the step errors in minutes with a clear message instead of burning 90 min.

Node workflows share a composite action (\`.github/actions/setup-playwright\`); the Python-based \`migration-test\` gets the same retry+timeout inline.

Victor-w-Madeira:
## Problem

The `Install Playwright browsers` step (`npx playwright install chromium --with-deps`) downloads the Chromium binary from `cdn.playwright.dev` on **every** run, with **no per-step timeout and no retry**. When that download stalls, it consumes the entire 90-minute job budget and the job is cancelled with *"exceeded the maximum execution time"* — no tests run, no Playwright report is produced, and the real cause is masked.

Observed twice in a row (so not a one-off flake):
- Run [26760434237](https://github.com/oriontech-me/langflow-e2e/actions/runs/26760434237) — `weekly-stable`, cancelled at 1h30m mid-download.
- Run [26782003123](https://github.com/oriontech-me/langflow-e2e/actions/runs/26782003123) — re-run, hung at the **same step** for >20 min.

## Fix

New reusable composite action **`.github/actions/setup-playwright`** that attacks the problem on three fronts:

1. **Cache** `~/.cache/ms-playwright` keyed on the Playwright version → the download is skipped entirely on a cache hit (only the OS-level apt deps are installed, since those are not part of the cached directory).
2. **Per-attempt `timeout 5m` + up to 3 retries** → a stalled download is killed and retried, recovering from transient CDN stalls. Composite-action steps don't support `timeout-minutes`, so the bound is enforced with the `timeout` shell command.
3. **Fail fast** → if all attempts fail, the step errors in minutes with a clear message instead of burning 90 min.

## Scope

| Workflow | Change |
|---|---|
| `weekly-stable.yml` | uses composite action |
| `nightly.yml` | uses composite action |
| `manual.yml` (both jobs) | uses composite action |
| `adaptive-impacted.yml` | uses composite action |
| `migration-test.yml` | Python toolchain — same `timeout`+retry applied **inline** (cannot reuse the Node action) |

## Validation

- All workflow YAML + the composite action parse cleanly.
- No raw, unguarded `playwright install` remains — every occurrence is now inside a timeout/retry wrapper.
- The composite action is referenced as `./.github/actions/...`; every workflow runs `actions/checkout` before the install step, so the local action is available.

Closes #344

### DIFF
--- a/.github/actions/setup-playwright/action.yml
+++ b/.github/actions/setup-playwright/action.yml
@@ -1,0 +1,60 @@
+name: "Setup Playwright"
+description: >-
+  Install Playwright Chromium with browser caching, a per-attempt timeout, and
+  retries so a stalled CDN download cannot burn the whole job budget.
+
+# Composite-action steps do not support `timeout-minutes`, so each attempt is
+# bounded with the `timeout` shell command instead. A stalled `playwright
+# install` is killed after 5 minutes and retried, recovering from transient
+# `cdn.playwright.dev` download stalls. See issue #344 for the failure this fixes.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Resolve Playwright version
+      id: pw
+      shell: bash
+      run: echo "version=$(node -p "require('@playwright/test/package.json').version")" >> "$GITHUB_OUTPUT"
+
+    - name: Cache Playwright browsers
+      id: pw-cache
+      uses: actions/cache@v4
+      with:
+        path: ~/.cache/ms-playwright
+        key: ${{ runner.os }}-playwright-${{ steps.pw.outputs.version }}
+
+    # Cache miss: download the browser binary AND install the OS-level
+    # dependencies (the apt packages from `--with-deps` are not part of the
+    # cached `~/.cache/ms-playwright` directory).
+    - name: Install Playwright browser and system deps
+      if: steps.pw-cache.outputs.cache-hit != 'true'
+      shell: bash
+      run: |
+        for attempt in 1 2 3; do
+          echo "::group::Playwright install (attempt $attempt/3)"
+          if timeout 5m npx playwright install chromium --with-deps; then
+            echo "::endgroup::"
+            exit 0
+          fi
+          echo "::endgroup::"
+          echo "::warning::Playwright install attempt $attempt failed or stalled; retrying in 5s..."
+          sleep 5
+        done
+        echo "::error::Playwright install failed after 3 attempts (see issue #344)"
+        exit 1
+
+    # Cache hit: the browser binary is restored, but the OS-level apt
+    # dependencies still need to be installed on the fresh runner.
+    - name: Install Playwright system deps (cache hit)
+      if: steps.pw-cache.outputs.cache-hit == 'true'
+      shell: bash
+      run: |
+        for attempt in 1 2 3; do
+          if timeout 5m npx playwright install-deps chromium; then
+            exit 0
+          fi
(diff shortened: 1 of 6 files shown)


Victor-w-Madeira:
## Independent review — applied (commit 77cb095)

Ran an independent, adversarial review of the first pass before merge. It found two real correctness issues, both now fixed:

- **🔴 Blocker — cache poisoning.** The original cache-hit path ran only `install-deps` and never validated the browser. With a fixed per-version cache key and `actions/cache` saving the path in its post-step *even when the job fails*, a download killed mid-way by `timeout` could persist a **partial** browser directory under the canonical key — poisoning every later run with a cryptic `failed to launch chromium` and no way to self-heal until the Playwright version bumped. Strictly worse than the pre-fix behavior.
- **🟠 Should-fix — `timeout` killing apt.** `timeout` wrapped `install --with-deps`, so a SIGTERM during the apt phase could leave dpkg half-configured and break the retries.

Both shared one root cause: the timeout wrapped too much, and the cache-hit path trusted the cache blindly.

**Fix applied:**
- `playwright install-deps chromium` runs once, **un-timed** (apt must not be interrupted mid-transaction).
- The browser-only `playwright install chromium` runs **inside the timed retry loop on both cache hit and miss** — a fast no-op on a valid cache, a re-download on a partial one, so a corrupt cache **self-heals**.
- Added `timeout --kill-after=30s` so a process ignoring SIGTERM is force-killed.
- The version step now fails if the resolved version is empty (avoids a poisoned `Linux-playwright-` shared cache key).
- Same split applied inline to `migration-test` (Python toolchain).

Verified OK by the review: errexit behavior in the retry loop (commands in `if` conditions are exempt from `set -e`), `npm ci` → action ordering in every workflow, `exit` semantics in composite/inline steps, and YAML/wiring.

PR #347, opened 2026-06-01
ci(weekly): run inside official Playwright image to bypass GCS browser download

### LINKED ISSUE #346: CI: GitHub runners cannot download Chromium from Google Cloud Storage — use official Playwright container image
## Problem

Follow-up to #344. The timeout+retry hardening (PR #345) works as designed — it now **fails fast** (~17 min) with clear warnings instead of hanging the full 90 min — but it cannot make the download succeed, because the underlying problem is a network issue between the GitHub-hosted runners and the binary host.

### Evidence (run [26784289589](https://github.com/oriontech-me/langflow-e2e/actions/runs/26784289589))

- `playwright install-deps` (apt) succeeded: `Fetched 21.1 MB in 0s (45.7 MB/s)` — general runner internet is fine.
- The browser download stalled at zero progress on **all 3 attempts**, each killed by the 5-minute `timeout`:
  ```
  attempt 1/3  Downloading Chrome for Testing 145.0.7632.6 ...  → killed after 5 min
  attempt 2/3  Downloading ...                                  → killed after 5 min
  attempt 3/3  Downloading ...                                  → killed after 5 min
  ::error:: failed after 3 attempts
  ```

### Root cause

`cdn.playwright.dev` returns a **307 redirect** to `https://storage.googleapis.com/chrome-for-testing-public/...` — the Chromium binary is served from **Google Cloud Storage**. From the public internet this chain is healthy (HEAD 200, ~3 MB/s download). On the runners (Azure-hosted) the connection to Google Cloud Storage opens but the transfer never progresses. So the runner→GCS path is the failing leg, not the test code.

Possibly correlated with the recent private→public visibility switch (public and private repos use separate runner pools with different egress), but unproven.

## Fix

Stop depending on the runner downloading the browser at all: run the job inside the **official Playwright image** `mcr.microsoft.com/playwright:v1.58.2-noble`, which ships Chromium + all OS deps pre-installed and is pulled from `mcr.microsoft.com` (Microsoft/Azure registry, reachable from the runners). Verified the tag exists and matches the pinned `@playwright/test` version (1.58.2).

Knock-on change: when the job runs in a container, the Langflow service container is reached by its service hostname, so `PLAYWRIGHT_BASE_URL` changes from `http://localhost:7860/` to `http://langflow:7860/`.

## Rollout (validation-first)

1. **This PR:** migrate `weekly-stable.yml` only and validate live.
2. **Follow-up:** once proven green, apply the same pattern to `nightly`, `manual` (both jobs), `adaptive-impacted`. `migration-test` (Python toolchain, unpinned) is tracked separately.

Victor-w-Madeira:
## Why

Follow-up to #344 / PR #345. The timeout+retry hardening works exactly as designed — it now **fails fast (~17 min)** with clear warnings instead of hanging the full 90 min — but it cannot make the download succeed. The real problem is infra:

- `cdn.playwright.dev` returns a **307 redirect** to **Google Cloud Storage** (`storage.googleapis.com/chrome-for-testing-public/...`).
- From the public internet that chain is healthy (HEAD 200, ~3 MB/s).
- On the runners, `apt` downloads fine (`Fetched 21.1 MB in 0s (45.7 MB/s)`), but the **browser download from GCS stalls at zero progress on all 3 attempts** ([run 26784289589](https://github.com/oriontech-me/langflow-e2e/actions/runs/26784289589)).

So the failing leg is runner→Google Cloud Storage, not the test code. Possibly correlated with the private→public switch (separate runner pools / egress), unproven.

## Fix

Run the job inside the official Playwright image **`mcr.microsoft.com/playwright:v1.58.2-noble`** — Chromium + OS deps pre-installed, pulled from the Microsoft/Azure registry (reachable from the runners). No GCS download at all. The tag matches the pinned `@playwright/test` version (1.58.2), so the browser revision lines up. The `Install Playwright browsers` step is dropped (browser is in the image).

**Knock-on change:** inside a job container the Langflow service is reached by its service hostname, so `PLAYWRIGHT_BASE_URL` changes `http://localhost:7860/` → `http://langflow:7860/`.

## Static review done

- `playwright.config.ts` honors `PLAYWRIGHT_BASE_URL` (env wins over the `localhost` fallback); no `webServer` block.
- No test hardcodes `localhost:7860` — the only reference (`mcp-server.spec.ts`) reads the env var with localhost merely as a fallback. The Python `localhost` refs belong to `migration-test`, untouched here.
- The networking change (`langflow:7860`) is the one thing only a live run can confirm — hence this PR is **weekly-stable only**.

## Rollout

1. **This PR:** `weekly-stable.yml` only → validate live.
2. **Follow-up:** apply the proven pattern to `nightly`, `manual` (both jobs), `adaptive-impacted`. `migration-test` (Python, unpinned) tracked separately. The `setup-playwright` composite action stays until those are migrated.

Closes #346

### DIFF
--- a/.github/workflows/weekly-stable.yml
+++ b/.github/workflows/weekly-stable.yml
@@ -83,6 +83,9 @@ jobs:
       # so forward localhost:7860 -> langflow:7860 and keep PLAYWRIGHT_BASE_URL
       # on http://localhost:7860 — exactly as on ubuntu-latest. See issue #346.
       - name: Forward localhost:7860 to the Langflow service
+        # Force bash: inside the container the default shell is `sh` (dash),
+        # which lacks the `disown` builtin used below.
+        shell: bash
         run: |
           apt-get update -qq && apt-get install -y -qq socat
           nohup socat TCP-LISTEN:7860,fork,reuseaddr TCP:langflow:7860 >/tmp/socat.log 2>&1 &
--- a/.github/workflows/weekly-stable.yml
+++ b/.github/workflows/weekly-stable.yml
@@ -21,6 +21,14 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 90
 
+    # Run inside the official Playwright image: Chromium + all OS deps are
+    # pre-installed and pulled from mcr.microsoft.com (reachable from the
+    # runners), so we no longer download the browser from Google Cloud Storage
+    # — the leg that the runners cannot complete (see issue #346). The tag MUST
+    # match the pinned @playwright/test version in package.json.
+    container:
+      image: mcr.microsoft.com/playwright:v1.58.2-noble
+
     services:
       langflow:
         image: langflowai/langflow-nightly:${{ inputs.langflow_image_tag || 'latest' }}
@@ -49,22 +57,23 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Install Playwright browsers
-        uses: ./.github/actions/setup-playwright
+      # No "Install Playwright browsers" step: Chromium ships in the container
+      # image. npm ci still installs the @playwright/test runner, whose version
+      # is pinned to match the image tag so the browser revision lines up.
 
       - name: Collect models
         run: npx playwright test tests/collect-models.spec.ts --reporter=line
         continue-on-error: true
         env:
           CI: "true"
-          PLAYWRIGHT_BASE_URL: "http://localhost:7860/"
+          PLAYWRIGHT_BASE_URL: "http://langflow:7860/"
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
 
       - name: Run @stable tests
         run: npx playwright test --grep "@stable" --pass-with-no-tests --reporter=html,github,json
         env:
           CI: "true"
-          PLAYWRIGHT_BASE_URL: "http://localhost:7860/"
+          PLAYWRIGHT_BASE_URL: "http://langflow:7860/"
           PLAYWRIGHT_JSON_OUTPUT_NAME: results.json
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
 
--- a/.github/workflows/weekly-stable.yml
+++ b/.github/workflows/weekly-stable.yml
@@ -76,19 +76,41 @@ jobs:
       # pinned to EXACTLY 1.58.2 in package.json to match the image tag above,
       # so the browser revision lines up. Bump both together when upgrading.
 
+      # The async Clipboard API (and other secure-context-gated browser APIs)
+      # only exist on a secure context. Chromium treats http://localhost as
+      # secure but NOT the service hostname http://langflow. Inside a job
+      # container the Langflow service is only reachable as http://langflow:7860,
+      # so forward localhost:7860 -> langflow:7860 and keep PLAYWRIGHT_BASE_URL
+      # on http://localhost:7860 — exactly as on ubuntu-latest. See issue #346.
+      - name: Forward localhost:7860 to the Langflow service
+        run: |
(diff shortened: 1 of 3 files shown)


Victor-w-Madeira:
## Independent review — applied (commit 0e6a56e)

Ran an independent, adversarial review of the container migration. It found two real issues, both now fixed:

- **🔴 Blocker — version drift.** `package.json` declared `@playwright/test`/`playwright` as `^1.57.0` (caret), **not pinned**, even though a comment claimed it was "pinned to match the image tag". The lockfile resolves to `1.58.2` today, so it matches — but a future `npm update` pulling `1.59.x` would make `npm ci` install a runner expecting a browser revision the `v1.58.2` image doesn't ship, and **every test fails at launch** with a cryptic "browser executable doesn't exist". PR validation (tsc + eslint) wouldn't catch it.
  - **Fix:** pin both deps to exactly `1.58.2` (minimal 2-line lockfile diff; `npm ci --dry-run` confirms sync) **and** add a guard step that fails fast with a clear message if the npm version ever diverges from the image tag.
- **🟠 Should-fix — redundant `setup-node`.** `actions/setup-node@v4` inside the container layered a second Node on top of the image's own tested toolchain.
  - **Fix:** dropped it; the job now uses the image's Node directly.

**Confirmed-fine by the review:** the `langflow:7860` service hostname is correct for a job container (service *key* = hostname; the old `localhost:7860` would have been wrong in-container); `playwright.config.ts` honors `PLAYWRIGHT_BASE_URL` with no `webServer` block; the `--reporter=html,...` override keeps `playwright-report/` populated for the artifact; secrets work identically in-container.

**Only confirmable by a live run (why this PR is weekly-stable only):** that `git` is present in the image (needed by checkout + the history-commit steps) and that `langflow:7860` resolves from the job container at step time.

Victor-w-Madeira:
## First live run (26785883674) — infra fix validated, one regression fixed

Ran the workflow on this branch without merging. Result: **167 passed · 9 failed · 3 flaky · 16.9m**.

### ✅ The container migration works
- `Initialize containers` (Playwright image + Langflow service) healthy.
- `checkout` succeeded → `git` is present in the image.
- Version guard passed; **no browser download** (the GCS leg is gone).
- The suite actually **ran** (16.9m) and 167 tests passed — first time the weekly suite got past install.

### 🔴 One regression I introduced — now fixed (commit 62856ab)
Switching `PLAYWRIGHT_BASE_URL` to `http://langflow:7860` changed the origin from a **secure context** (`http://localhost` is treated as secure by Chromium) to an insecure one, so `navigator.clipboard` became `undefined`. That broke 4 clipboard tests:
`webhook-component-regression:224`, `output-modal-copy-button:20`, `curlApiGeneration:4`, `pythonApiGeneration:4`.

**Fix:** revert `PLAYWRIGHT_BASE_URL` to `http://localhost:7860` and forward `localhost:7860 -> langflow:7860` (socat) inside the job container. Restores the exact secure-context behavior tests had on `ubuntu-latest`; `playwright.config.ts` untouched (local dev unaffected).

### 🟡 The other failures are pre-existing, out of scope for this PR
They surfaced for the first time only because the suite could finally run:
- **4 traces tests** (`traces-delete`, `traces-detail-single`, `traces-latency-tokens`, `traces-list-filters`) fail because the workflow sets `LANGFLOW_DEACTIVATE_TRACING=true` → trace endpoints return empty (`Expected: > 0`). Config predates this PR.
- A few **agent / prompt-template / memory** tests (`agent stop button` 120s timeout, `prompt-template-invalid-patterns`, `memory session isolation`) — likely Langflow nightly behavior; need separate triage.

These belong in a separate triage issue, not this infra PR.

Next: re-run on the branch to confirm the 4 clipboard tests pass; the remaining ~5 pre-existing failures are expected until triaged separately.

Issue #383, opened 2026-06-15
fix(api-request): POST/GET tests fail when httpbin.org returns 5xx — @stable removed (weekly #382)


Victor-w-Madeira:
## Context

Surfaced by the weekly `@stable` run [27546448127](https://github.com/oriontech-me/langflow-e2e/actions/runs/27546448127) (2026-06-15). Triaged from #382.

**Test:** `core-components/api-request-component-regression.spec.ts:247`
`API Request component — POST method executes POST verb and returns 200`
Previous tags: `@stable @release @regression @components`

## What happened

- **Hard failure (all retries).** The POST test timed out in the `runAndOpenOutput` helper (`spec:54`) waiting for the output dialog textbox — the run never produced a readable output.
- The sibling **GET test (line 217)** failed the same window with the explicit signal:
  `expect(output).toContain("200")` → received `"status_code": 503` and body `503 Service Temporarily Unavailable` with `server: awselb/2.0`.

## Root cause (primary): external dependency

`httpbin.org` returned **HTTP 503 (AWS ELB)** during the run. Both api-request tests target `httpbin.org/get` and `httpbin.org/post`, so when httpbin is degraded the component returns the upstream 5xx and the assertions/output rendering fail. This is **not** a test logic bug and **not** a confirmed Langflow regression — it is a transient outage of an external service the suite depends on.

Recurrence pattern (the suite is chronically coupled to httpbin):
- weekly 2026-06-08 ([27133642480](https://github.com/oriontech-me/langflow-e2e/actions/runs/27133642480)): GET (217), PUT (276), non-2xx (369) hard-failed.
- weekly 2026-06-15: POST (247) hard-failed + GET (217) flaky.

## Caveat — 1.10.0 cycle

We are at the end of the 1.10.0 development cycle. A legitimate product change in the API Request component's output rendering cannot be fully ruled out from the logs alone, and a real bug/regression is also not excluded. Confirm against the current nightly before restoring `@stable`.

## Decision

- `@stable` **removed** to stop weekly noise (done in the triage PR — see below).
- Before restoring `@stable`: **harden the external dependency** — e.g. retry on 5xx, tolerate transient httpbin outages, or move to a more reliable / self-hosted echo endpoint.

Triaged from #382. `@stable` removal: see the linked PR.

Victor-w-Madeira:
### 🎯 Objective of this issue

[...]

PR #412, opened 2026-06-22
fix(api-request): migrate echo endpoint to postman-echo.com (#407)

### LINKED ISSUE #409: Make the @stable API Request specs reliable: decouple from the unreliable httpbin.org echo endpoint
## Problem

The `@stable` API Request specs (`tests/tests-automations/regression/core-components/api-request-component-regression.spec.ts`) execute real HTTP round-trips against the public `httpbin.org` echo endpoint. `httpbin.org` sits behind an AWS ELB that intermittently returns 5xx, and has gone fully down for entire weekly windows. This has caused repeated `@stable` weekly failures that are **not** Langflow regressions and **not** test-logic bugs:

- weekly 2026-06-08 — GET/PUT/non-2xx returned 503
- #383 (2026-06-15) — POST/GET hard-failed on httpbin 503; a retry-on-5xx was added as a band-aid
- #407 (2026-06-22) — GET returned 503 on all 3 retry attempts (httpbin fully down for the window)

The 3-attempt retry from #383 only survives a transient blip *within* a single test window; a sustained outage exhausts every attempt and fails the run. **The root cause is the dependency on an unreliable external echo endpoint — that is what needs to be fixed.**

## Scope

- **In scope:** the `@stable` API Request spec only (`core-components/api-request-component-regression.spec.ts`). Per internal convention, only `@stable` tests run in the weekly workflow and gate triage, so they are the ones that must be made reliable.
- **Out of scope:** the legacy `@release` / `@regression` specs (`api/flows/api-component-regression.spec.ts`, `api/flows/api-request-component-ui.spec.ts`), which still hardcode `httpbin.org` but do **not** run in the weekly workflow. They receive the same treatment only *if/when* they are promoted to `@stable`.

## Strategy — open

No specific implementation is prescribed. (The earlier proposal to self-host `go-httpbin` as a CI service container has a known blocker: the API Request component's URL validator rejects single-label hostnames, so an endpoint reachable only by its bare service label — e.g. `http://httpbin:8080` — is unusable.) Any approach that durably decouples the `@stable` spec from an unreliable echo endpoint is acceptable, e.g.:

- point the spec at a more reliable public echo service (e.g. `postman-echo.com`), or
- self-host an echo endpoint reachable *by Langflow* via a routable hostname (plus `LANGFLOW_SSRF_ALLOWED_HOSTS` if on a private IP), or
- another decoupling the implementer judges durable.

The spec already reads the endpoint from an env-var override, so the change is primarily about choosing a reliable default and proving it holds.

## Validation criterion

The `@stable` API Request spec passes across consecutive weekly runs with **no failure caused by an echo-endpoint outage**, and with **no transient-5xx retry** triggered by the chosen endpoint.

## Notes

- Surfaced by triage of #407 (now closed).
- PR #412 is one candidate solution attacking this problem.

daniellicnerski1:
## Context

Closes #409. The weekly `@stable` run of 2026-06-22 ([27952036517](https://github.com/oriontech-me/langflow-e2e/actions/runs/27952036517)) hard-failed:

> `API Request component — GET request returns 200 and output Data contains all required fields`

with `httpbin.org/get` returning **HTTP 503 (AWS ELB) on all 3 retry attempts**.

This is **not a Langflow regression and not a test-logic bug** — it is the **third recurrence** of the same external dependency outage:

| Date | Issue | What happened |
|---|---|---|
| 2026-06-08 | — | GET/PUT/non-2xx 503 |
| 2026-06-15 | #383 | `@stable` removed, then restored with retry-on-5xx |
| 2026-06-22 | #407 | retry-on-5xx exhausted — httpbin down the whole window |

The retry added in #383 only survives a transient blip *within* a single test window; a sustained outage exhausts every attempt and fails the run.

## What this PR does

Moves the default echo endpoint from `httpbin.org` to **`postman-echo.com`** (more reliable public echo). Verified empirically as a near drop-in before any code change:

- `/get`, `/post`, `/put`, `/patch`, `/delete` each return `200` only for the matching verb
- `/status/{code}` is a deliberate status endpoint (`/status/404` → 404)
- query params echo in `args`; the body echoes `host`/`url`

**One behavioral difference:** a wrong verb returns **404** on postman-echo vs **405** on httpbin. This is immaterial — the verb tests assert the output contains `200`, and `404` is just as much "not 200" as `405`, so the per-verb guarantee is preserved.

### Changes
- Default `ECHO_BASE` → `https://postman-echo.com`
- Rename constants `HTTPBIN_BASE`/`HTTPBIN_HOST` → `ECHO_BASE`/`ECHO_HOST`; override env var `ECHO_BASE_URL` (legacy `HTTPBIN_BASE_URL` still honored as a fallback)
- Update the spec doc, preserving the #383 history and adding the #407 record

No assertion logic changed; no `@stable` tag changed.

## Validation

Run against a live **Langflow 1.11.0** instance:

- ✅ All 8 endpoint-dependent tests pass: GET, POST, PUT, PATCH, DELETE, non-2xx (404), query params, cURL-execute
- ✅ typecheck clean; lint clean (0 errors)
- ✅ force-fail: breaking the echoed-host assertion failed exactly on that line; output showed `"source": "https://postman-echo.com/get"`, `"status_code": 200`; reverted and re-passed
- ✅ zero `🚨 Backend Error` occurrences

## Out of scope

A separate canvas-bootstrap flake (`sidebar-search-input` detaching during the Templates-modal dismiss) was observed **only on the local vite dev server**; it did not flake in the CI weekly run (#407 ran api-request 15/15 clean except the httpbin assertion). Hardening `addApiRequestComponent` is tracked separately to keep this PR focused on the endpoint migration.

### DIFF
--- a/docs/core-components/api-request-component-regression.md
+++ b/docs/core-components/api-request-component-regression.md
@@ -9,7 +9,7 @@
 Validates the **API Request** component end-to-end via 15 scenarios grouped into five categories:
 
 1. **Canvas rendering and inspector fields** — the node renders with the correct URL/API Response handles, and the inspector accepts URL + HTTP method input.
-2. **Execution per HTTP verb** — GET, POST, PUT, PATCH, DELETE each round-trip to an echo endpoint that **only** accepts that verb (any other verb returns 405). The output Data is asserted to contain `200`, the echoed URL host, and the expected structural keys (`source`, `status_code`, `response_headers`, `result`). The endpoint defaults to public `httpbin.org` (overridable via `HTTPBIN_BASE_URL`); a transient upstream 5xx is absorbed by a retry in `runAndOpenOutput` so it does not fail the suite — see "Validation criterion" and issue #383.
+2. **Execution per HTTP verb** — GET, POST, PUT, PATCH, DELETE each round-trip to an echo endpoint that **only** accepts that verb (any other verb returns a non-2xx — `404` on postman-echo, `405` on httpbin; either way it is not `200`, which is what the verb tests assert). The output Data is asserted to contain `200`, the echoed URL host, and the expected structural keys (`source`, `status_code`, `response_headers`, `result`). The endpoint defaults to `postman-echo.com` (overridable via `ECHO_BASE_URL`); a transient upstream 5xx is absorbed by a retry in `runAndOpenOutput` so it does not fail the suite — see "Validation criterion" and issues #383, #407.
 3. **Error and edge paths** — invalid URL shows the build-error notification, non-2xx responses (404) propagate as `status_code` without raising, query parameters embedded in the URL are sent and echoed back.
 4. **Inspector tables and cURL mode** — the headers key-value table accepts both key and value cell entries via the View Text editor; the body key-value table is accessed through the Controls modal (the body field is `advanced=True`) and accepts both key and value cell entries via the same editor; the cURL tab switches mode and the cURL parser auto-populates the URL field with the URL extracted from the cURL command, after which the run executes the GET successfully.
 5. **Persistence to flow JSON** — configuring URL, method and a headers row triggers an autosave that persists into the flow's saved state (verified by polling `GET /api/v1/flows/{id}`), and a full page reload rehydrates the inspector with the same values.
@@ -39,7 +39,7 @@ Every test starts with `addApiRequestComponent(page)` which:
 - Asserts exactly one node on the canvas (`react-flow__node` count === 1).
 
 ### 2. `inspector fields accept configured values`
-- Fills `popover-anchor-input-url_input` with `https://httpbin.org/get` and asserts the value.
+- Fills `popover-anchor-input-url_input` with `https://postman-echo.com/get` and asserts the value.
 - Opens the method dropdown, selects "POST" and asserts the displayed value matches.
 
 ### 3. `invalid URL is accepted by field and run shows error notification`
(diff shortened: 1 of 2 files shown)


Victor-w-Madeira (approved):
## LGTM ✅

Reviewed the diff and validated the endpoint migration empirically — clean, well-scoped change, approving it.

**Verification (local, against live postman-echo + the `@stable` spec):**
- `/status/404` returns a real HTTP **404** (body `{"status":404}`) → the non-2xx test's `status_code: 404` + `not.toContain('"error"')` assertions hold.
- Wrong verb returns **404** (vs httpbin's 405) — immaterial: the verb tests assert the output contains `200`, and `404` is just as much "not 200".
- `/get?...` echoes query params under `args`, and `url` carries the full host/path → the query-param and echoed-host assertions hold.
- Env fallback `ECHO_BASE_URL ?? HTTPBIN_BASE_URL ?? postman-echo.com` is backward-compatible; all `HTTPBIN_*` → `ECHO_*` renames are consistent with no orphan references.
- `@stable` preserved; ESLint + TypeScript checks green.

**One follow-up before merge (PR description, not code):**
The body says `Closes #407`, but #407 is the auto-generated weekly-failure ticket and is already closed. The canonical issue for this work is now #409, which has been re-scoped to *"make the `@stable` API Request specs reliable by decoupling from the unreliable httpbin.org echo endpoint"* (strategy-open, `@stable`-only). Please update the description from `Closes #407` to `Closes #409` so the merge closes the right issue.

**Context (non-blocking):** this swaps one public echo endpoint for a more reliable one rather than eliminating the public dependency entirely — postman-echo can still have outages. That residual is exactly what the re-scoped #409 tracks (success criterion = stable across consecutive weekly runs), so no change is needed here; just making the linkage explicit.

Nice work — the empirical near-drop-in verification and the documented 404-vs-405 note made this easy to validate.

Issue #396, opened 2026-06-16
MCP v2 registration returns 500 instead of 409/404 — regression watchdog added; report upstream + restore @stable on fix


Victor-w-Madeira:
## Context

Surfaced while investigating Group 1 of #384. Registering an **already-existing** MCP server returns **HTTP 500** instead of a proper `409 Conflict`. We hardened our test to no longer depend on this behavior (commit 3362d33 — robust pre-clean via `auto_login`), but the backend behavior itself is worth a closer look before we decide whether to report it upstream.

## Reproduction (pure API, against `langflowai/langflow-nightly:latest`, `1.10.0.dev69`)

```bash
TOKEN=$(curl -s http://localhost:7860/api/v1/auto_login | python3 -c "import sys,json;print(json.load(sys.stdin)['access_token'])")

# First registration → 200 OK
curl -s -w "\n%{http_code}\n" -X POST "http://localhost:7860/api/v2/mcp/servers/http-form-server" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"url":"http://localhost:1/mcp"}'
# {"url":"http://localhost:1/mcp"} 200

# Second registration (same name) → 500 (expected 409)
curl -s -w "\n%{http_code}\n" -X POST "http://localhost:7860/api/v2/mcp/servers/http-form-server" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"url":"http://localhost:1/mcp"}'
# {"detail":"Server already exists."} 500
```

## Backend source

The 500 is raised deliberately in the upstream Langflow backend:

`src/backend/base/langflow/api/v2/mcp.py:340`
```python
if check_existing and server_name in server_list["mcpServers"]:
    raise HTTPException(status_code=500, detail="Server already exists.")
```

A duplicate-resource condition should be a `409 Conflict` (a client error), not a `500 Internal Server Error`. Returning 500 also trips our suite's backend-error monitor.

## Why it matters to us

Any leftover/duplicate MCP server turns a benign "already registered" state into a hard test failure. Our test no longer triggers it (clean pre-clean state), but the 500 is the kind of thing that bites again under load/races.

## Investigation goals (to decide: report upstream or not)

- [ ] Confirm it is a **regression** vs. long-standing behavior — when was line 340 introduced / last changed? Check across recent Langflow releases.
[...]

Issue #384, opened 2026-06-15
investigate: weekly flaky set (7 tests) + nightly backend 500 instability (weekly #382)


Victor-w-Madeira:
## Context

Surfaced by the weekly `@stable` run [27546448127](https://github.com/oriontech-me/langflow-e2e/actions/runs/27546448127) (2026-06-15). Triaged from #382.

Single investigation issue for the **7 flaky tests** of this run (all passed on retry — absorbed by the retry budget, so no individual `@stable` removal). Grouping them because they share a likely common cause and may expose addressable fragility now.

## Flaky tests (passed on retry)

| Test | File:line |
|---|---|
| API Request — GET returns 200 | `core-components/api-request-component-regression.spec.ts:217` |
| Chat Input → Chat Output — inspector-attached file rendered | `core-components/chat-input-files-field-regression.spec.ts:233` |
| Prompt Template — replacing a variable updates handles | `core-components/prompt-template-component-regression.spec.ts:109` |
| Agent Component Regression — agent interaction suite | `core-functionality/llm-agents/agent-component-regression.spec.ts:145` |
| Export and Import Flow — export to JSON success toast | `flow-functionality/export-import-flow.spec.ts:65` |
| Flow Rename via Header | `flow-functionality/flow-rename-header.spec.ts:13` |
| MCP Client — configure via HTTP form + verify registration | `mcp/client/mcp-client-regression.spec.ts:260` |

## Common backdrop — nightly backend instability

The run logged heavy backend `500 Internal Server Error` on:
- `/api/v1/flows/`
- `/api/v1/validate/prompt`
- `/api/v1/build/<id>/flow`

This instability is the most likely shared driver of the flakiness (UI timeouts, detached elements, build races). `agent-component:145` and the MCP specs also recur as flaky across previous weeks and keep self-healing.

## Goal

Investigate now while context is fresh:
1. Is the nightly backend 500 storm a known upstream issue at the 1.10.0 tail, or environment-specific?
2. Are there shared fragilities we can harden today (waits/selectors/build-readiness) regardless of backend noise?
3. Note the GET (217) flake here is the same external httpbin dependency tracked in the api-request issue.

`@stable` **kept** on all 7 for now (per the lifecycle table, a first-occurrence flake is absorbed by retries). Re-evaluate on the next weekly.

Triaged from #382.

PR #414, opened 2026-06-22
test(#397): standardize MCP specs auth on shared getAuthToken

### LINKED ISSUE #397: standardize MCP specs auth on shared getAuthToken (auto_login)
## Context

Follow-up from #384 (Group 1 hardening). The `mcp-client-regression.spec.ts` test authenticated its MCP-server pre-clean / verification API calls via a hardcoded `POST /api/v1/login` (`langflow/langflow`). Under the suite's standard `LANGFLOW_AUTO_LOGIN=true` mode this can silently return no token, so the follow-up DELETE runs unauthenticated and the pre-clean never removes a leftover server. We fixed that one file by routing through the shared `getAuthToken(page.request)` helper (which uses `GET /api/v1/auto_login`).

The sibling MCP specs still use the old hardcoded pattern and have the same latent fragility:

- `tests/tests-automations/regression/mcp/server/mcp-server.spec.ts:1035`
- `tests/tests-automations/regression/mcp/client/mcp-client-agent.spec.ts:144`
- `tests/tests-automations/regression/mcp/client/mcp-client-agent.spec.ts:182`

## Why it doesn't currently break CI

In CI each run gets a fresh, ephemeral Langflow backend, so there is no leftover MCP server for the pre-clean to remove — the first registration is always clean. The bug bites on a **persistent** backend (e.g. a local dev instance), where a leftover server makes the next registration hit a pre-existing name and the backend returns HTTP 500 ("Server already exists." — tracked in #396).

## Goal

Standardize MCP-spec auth on the shared `getAuthToken(page.request)` helper:
- [ ] `mcp-server.spec.ts:1035`
- [ ] `mcp-client-agent.spec.ts:144`
- [ ] `mcp-client-agent.spec.ts:182`
- [ ] Confirm `@stable` tests still pass after the change.

Low priority — consistency/robustness cleanup, not a CI blocker.

## References
- Pattern fixed in `mcp-client-regression.spec.ts` (branch `fix/384-group2-autosave-race`, #384).
- Backend 500 investigation: #396.

Victor-w-Madeira:
Closes #397

## What

Routes the MCP-server pre-clean / cleanup `DELETE` calls in the sibling MCP specs through `getAuthToken(page.request)` (`GET /api/v1/auto_login`) instead of a hardcoded `POST /api/v1/login` (`langflow`/`langflow`), matching the fix already applied in `mcp-client-regression.spec.ts` (#384).

## Why

Under the suite's standard `LANGFLOW_AUTO_LOGIN=true` mode the hardcoded login returns no token, so the `DELETE` ran unauthenticated and the pre-clean never removed a leftover server (HTTP 500 on a persistent backend, see #396).

## Files

- `mcp/server/mcp-server.spec.ts` (cleanup)
- `mcp/client/mcp-client-agent.spec.ts` (afterEach + registration step)

Net simplification: 8 insertions / 19 deletions.

### DIFF
--- a/tests/tests-automations/regression/mcp/client/mcp-client-agent.spec.ts
+++ b/tests/tests-automations/regression/mcp/client/mcp-client-agent.spec.ts
@@ -13,6 +13,7 @@ import {
 import type { ProviderRecord } from "../../../../helpers/provider-setup/collect-models";
 import { cleanAllFlows } from "../../../../helpers/flows/clean-all-flows";
 import { adjustScreenView } from "../../../../helpers/ui/adjust-screen-view";
+import { getAuthToken } from "../../../../helpers/auth/get-auth-token";
 
 if (!process.env.CI) {
   dotenv.config({ path: path.resolve(__dirname, "../../../../.env") });
@@ -140,14 +141,9 @@ for (const { label, options, skipReason } of targets) {
   test.describe(`MCP Client – Agent using MCPTools [${label}]`, () => {
     test.afterEach(async ({ page }) => {
       try {
-        const token = await page.request
-          .post("/api/v1/login", {
-            form: { username: "langflow", password: "langflow" },
-          })
-          .then((r) => r.json())
-          .then((d) => d.access_token as string);
+        const authHeader = await getAuthToken(page.request);
         await page.request.delete(`/api/v2/mcp/servers/${MCP_SERVER_NAME}`, {
-          headers: { Authorization: `Bearer ${token}` },
+          headers: { Authorization: authHeader },
         });
       } catch {
         // best-effort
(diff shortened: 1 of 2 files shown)


Victor-w-Madeira:
## Validation — local run against Langflow Nightly 1.11.0.dev12

Ran the two affected specs on this PR branch against a live 1.11.x instance, plus a direct check of the auth premise.

**Static (PR CI):** ESLint ✅ · TypeScript ✅

**Affected specs:**
- `mcp/server/mcp-server.spec.ts` — ✅ 4 passed, 2 flaky (recovered on retry; pre-existing UI instability unrelated to this change). The cleanup-bearing test passed.
- `mcp/client/mcp-client-agent.spec.ts` — OpenAI target ✅ passed end-to-end, exercising the modified `afterEach` + registration cleanup. The Google target failed, but in `helpers/provider-setup/setup-google.ts` (model-toggle setup) **before** reaching any code this PR touches — tracked separately in #415.

**Auth premise (direct curl against the instance):** `GET /api/v1/auto_login` returns a valid `access_token`, so the cleanup `DELETE` now authenticates correctly. (This particular instance also accepts the old hardcoded `POST /api/v1/login`, so it would not reproduce #396 by itself — but the new path is confirmed working.)

**Backend errors during runs:** none.

Conclusion: the code changed by this PR is validated. Merging.

Issue #589, opened 2026-07-08
Finish cleanAllFlows deprecation: migrate remaining non-validated specs + evaluate per-worker isolation


Victor-w-Madeira:
## Context

Follow-up to #515. The PR that closes #515 migrates the **five validated (`@stable`) specs** off the global \`cleanAllFlows()\` to scoped, per-flow teardown. To keep it 1-issue/1-PR, the broader cleanup items from the #515 body were intentionally deferred here.

## Remaining work

1. **Migrate the two non-validated specs** that still call the global helper to scoped teardown:
   - `core-functionality/playground/playground-session-clear.spec.ts`
   - `core-functionality/project-management/user-progress-track.spec.ts`
2. **Narrow or deprecate `tests/helpers/flows/clean-all-flows.ts`** once no parallel-suite spec depends on it — or delete it if it has no legitimate remaining use. Also remove the now-dead `MainPage.cleanAllFlows()` wrapper (no callers).
3. **Evaluate per-worker user isolation** as the more robust structural fix (each Playwright worker gets its own Langflow user), which would make a global wipe safe again and remove the shared-user contention entirely. Decide whether it is worth the infra cost vs. scoped teardown.

## Reusable pattern

The migration in #515 established the pattern to follow: capture the flow id from the **creation POST response** (the canvas URL id is a transient client-side handle on current Langflow and 404s on delete), then delete only that id via `deleteFlow(page.request, id, { headers: { Authorization } })` with an explicit auth token (page.request is anonymous under AUTO_LOGIN). See `setup-blank-flow.ts` and `load-template-by-name.ts`.

Related: #515, #588.

Issue #588, opened 2026-07-08
Harden flow creation/open against the parallel POST /api/v1/flows/ 500 race


Victor-w-Madeira:
## Context

Follow-up surfaced while implementing the scoped-teardown migration for #515.

Removing the global \`cleanAllFlows()\` from the validated specs exposed a pre-existing backend race that the global wipe had been masking (it deleted everything each run, and the retry hid the failure). Now that flows are only created/deleted per-test, the race is directly observable.

## Symptom

Under concurrent flow creation, \`POST /api/v1/flows/\` intermittently returns **500 "An internal error occurred while creating the flow"** — the unique-name suffixing (`New Flow (N)`) is not transaction-safe upstream. A related flake is the **flow-open by name on the home list** racing when many flows are present.

Observed locally when stressing three flow-heavy specs together (`agent-component`, `webhook-component`, `loop-component`) with `--repeat-each=2`: `loop` test 4 and `webhook copy-button` failed on the **home flow-list** (not the #515 empty-state). `loop` test 4 failed identically before any change — i.e. it is not caused by the teardown migration.

## Root cause

Upstream Langflow: non-transaction-safe unique-name handling on flow creation. We cannot fix the backend here, but we **can** make the suite resilient.

## Proposed mitigation (to decide)

- Add a defensive **retry-on-500** wrapper around flow creation in the shared helpers: `setup-blank-flow.ts` (`setupBlankFlow`), the loop spec's `createFlowFromAsset`, and the new `openBlankFlow` helper in `mcp-client-regression.spec.ts`.
- Harden flow-open-by-name on the home list (scope/retry the card lookup).
- Evaluate whether unique per-test flow names (already mostly the case) fully sidestep the suffix race.

Related: #515 (the migration that exposed this).

Victor-w-Madeira:
Concrete repro from #590's CI: `agent-component-regression.spec.ts` "renders on canvas with default fields and handles" failed under the parallel `Run impacted E2E specs` run, immediately after `🚨 Backend Error: 500 Internal Server Error - /api/v1/flows/`. The same test **passes when run isolated/serial locally** — confirming this is the parallel-creation race, not a spec bug. Good candidate test to validate the retry-on-500 mitigation against.

Victor-w-Madeira:
## New occurrence — daily 2026-07-10 (umbrella #629)

[...]

Issue #553, opened 2026-07-07
[Flaky] llm-invalid-api-key: playground input remains usable after API error — recurring (daily #548)


Victor-w-Madeira:
## Source

Recurring flaky surfaced by the daily `@stable` workflow — most recently in daily #548 ([run 28860759001](https://github.com/oriontech-me/langflow-e2e/actions/runs/28860759001)), 2026-07-07, against `langflowai/langflow-nightly:latest`.

## Test

`tests/tests-automations/regression/core-functionality/llm-agents/llm-invalid-api-key-ui.spec.ts:77` — "playground input remains usable after API error (mocked)" (`@stable @release @workspace @regression @agents @playground`).

## Recurrence (why this is not one-off noise)

Flaky (failed-then-passed on retry) on **two** daily runs:

- **2026-07-06** — 3 attempts
- **2026-07-07** — 2 attempts

Per the recurring-flaky policy (2+ runs), this warrants an issue while **keeping `@stable`**.

## Prior history

This exact test was fixed once before in #444 ("playground build/run request path changed", closed 2026-06-29). The new recurrence may be a **regression of that fix** or a different, timing-related cause — the investigation should start by comparing against #444.

## Investigation scope — suspect BOTH product and test

Do **not** assume the fault is in the test. Investigate both sides:

- **Product:** After a mocked API error, does the playground input reliably return to a usable state on nightly? Any regression in error handling / input re-enable timing?
- **Test:** Is the assertion using a web-first, auto-retrying wait, or a synchronous snapshot / fixed short timeout that races the UI recovering? Is the mock intercept applied deterministically before the run?
- Reproduce locally and inspect the retry traces from the flaky runs to pinpoint which step races.

## Deliverables (required to close this issue)

1. Root-cause conclusion stating whether the flake is **product** or **test** driven.
2. A stabilization fix (test hardening and/or an upstream product bug report) as applicable.
3. Confirmation the test is no longer flaky across subsequent daily runs. `@stable` is **kept** throughout — no restoration PR needed unless a later hard failure removes it.

Surfaced by daily #548.

Issue #547, opened 2026-07-07
Add a shared deleteFlow() cleanup helper that surfaces failed flow deletions


Victor-w-Madeira:
Surfaced by GitHub Copilot's review on #546.

## Problem

Flow-teardown code across the suite calls:

```ts
await page.request.delete(`/api/v1/flows/${id}`);
```

Playwright's `APIRequestContext.delete()` returns an `APIResponse` even on 4xx/5xx and **does not throw**. Because the result is never checked, a failed deletion passes silently — the flow is *not* removed and the suite can quietly start re-accumulating leftover flows (exactly the buildup #545 set out to stop, ~1.5k flows observed on a long-lived local instance).

This is a **repo-wide convention**, not specific to any one spec — the unchecked pattern appears in ~53 files (all the `afterEach` flow cleanups, plus `setup-blank-flow.ts`, `setup-playground.ts`, `clean-all-flows.ts`).

## Proposed change

Introduce a shared helper and adopt it everywhere the raw delete is used:

```ts
// tests/helpers/flows/delete-flow.ts
export async function deleteFlow(page: Page, id: string): Promise<void> {
  const res = await page.request.delete(`/api/v1/flows/${id}`);
  if (!res.ok()) {
    throw new Error(`Flow cleanup failed: ${res.status()} — ${await res.text()}`);
  }
}
```

Benefits:
- Cleanup failures become **visible** instead of silently regressing the anti-accumulation guarantee.
- Single place to evolve the behavior — e.g. add one retry if a transient teardown 5xx turns out to introduce flake.

## Trade-off to decide during implementation

Throwing inside `afterEach` couples the test result to teardown success: a transient `DELETE` 5xx under load would fail an otherwise-green test. Options to weigh:
- hard-throw (max visibility, small teardown-flake risk), or
- one internal retry then throw (balances both).

Pick one and apply consistently.

## Scope / notes

- Transversal change touching ~53 files — keep it mechanical and reviewable; consider doing it in one focused PR.
- No product/UI behavior change; teardown-only.
- Retains `@stable`; no spec-doc changes.
- Relates to #465 / #515 (flow accumulation) and follows #545.

PR #590, opened 2026-07-08
fix(flows): scope flow teardown off global cleanAllFlows (#515)

### LINKED ISSUE #515: Suite-wide flake hazard: global cleanAllFlows() deletes concurrent tests' flows
## Summary

Follow-up to #465. While investigating that flake we found the true root cause: `cleanAllFlows(page)` (`tests/helpers/flows/clean-all-flows.ts`) deletes **every** flow of the user, not just the flow the calling test created. Because all workers share a single `auto_login` user and the suite runs `fullyParallel`, any spec that calls `cleanAllFlows` during another test's build window **deletes that other test's flow mid-run** — the output never renders and the run "does not settle", then passes on retry.

This is a latent, cross-file flake source affecting the whole suite, not just the playground DataFrame test fixed in #465.

## Evidence

- Serial (`--workers=1 --repeat-each=3`): 3/3 pass.
- Parallel (`--repeat-each=3`): fails, and the failing snapshot is the bare home empty-state ("Welcome to Langflow") — i.e. the flow was deleted underneath the test, not a slow build.
- `webhook-component-regression.spec.ts` already hit this exact class of problem and worked around it by forcing `mode: serial` (see the comment at the top of that file) — prior art that the collision is real.

## Validated (`@stable`) specs still calling the global cleanAllFlows

| Spec | How it calls | Perpetrator against other tests? |
|---|---|---|
| `mcp/client/mcp-client-regression.spec.ts` | `afterEach` | yes |
| `mcp/client/mcp-client-agent.spec.ts` | `afterEach` | yes |
| `core-components/agent-component-regression.spec.ts` | `afterEach` | yes |
| `core-components/loop-component-regression.spec.ts` | inline in test | yes |
| `core-components/webhook-component-regression.spec.ts` | inline (runs serial internally) | yes — `mode: serial` only serializes its own tests; across files/workers it still deletes other files' flows |

(Non-validated specs `playground-session-clear.spec.ts` and `user-progress-track.spec.ts` also use it but are out of scope — not canonical.)

## Action

- [ ] Investigate each validated spec above and confirm whether its `cleanAllFlows` usage can race concurrent tests.
- [ ] Evaluate moving off methods that can compromise other tests: prefer **scoped teardown** (delete only the flow(s) the test created, by id — the pattern already used in `create-runnable-chat-flow-via-api.ts`) over the global `cleanAllFlows`.
- [ ] Decide whether `cleanAllFlows` should be deprecated/narrowed at the helper level, or whether per-worker user isolation is the more robust fix.
- [ ] Migrate the affected specs accordingly.

The scoped-teardown fix for the playground DataFrame test is handled separately in #465.

Victor-w-Madeira:
## What & why

Closes #515.

`cleanAllFlows(page)` deletes **every** flow of the shared `auto_login` user. In the `fullyParallel` suite it wipes flows other workers are actively building mid-run — the victim's canvas 404s ("Flow not found"), the run "does not settle", and it only goes green on retry. `mode: serial` does not help (it only orders tests within one file; the global delete still reaches across workers).

This migrates the **five validated (`@stable`) specs** to **scoped teardown**: each test captures the id of the flow(s) it created and deletes only those — the pattern already proven in `playground-output-data.spec.ts` (#465), `stop-building.spec.ts`, and `load-template-by-name.ts` (#553).

## Changes

- **`SimpleAgentTemplatePage.load()`** now returns the flow id from `loadTemplateByName` (`void` → `string`, backwards-compatible).
- **agent / loop / webhook / mcp-client-regression / mcp-client-agent**: track created ids; delete per-id in `afterEach`; dropped the global `cleanAllFlows` calls. (Loop tests 1–3 previously leaked their flows — now cleaned up too.)
- Spec docs for loop/webhook updated to describe the scoped teardown.

## Hardening learned during verification

- The **canvas URL id is a transient client-side handle** on current Langflow and does not match the persisted flow — deleting it 404s and silently leaks the real one. All UI-`blank-flow` paths now capture the id from the **creation `POST /api/v1/flows/` response** instead.
- **`page.request` is anonymous under AUTO_LOGIN** — teardown passes an explicit `Authorization` header (via `getAuthToken`) and no longer swallows `deleteFlow` failures, so a failed cleanup surfaces (#547).

## Verification

- `npm run typecheck` ✅ / `npm run lint` ✅ (0 errors).
- Isolated (serial): create/delete nets to **zero** — no leaks.
- Combined parallel `--repeat-each=2` across the three flow-heavy specs: **no empty-state wipe** (the #515 hazard is gone); the pre-existing 5 unrelated flows survive.
- Remaining failures under that artificial over-concurrency are the pre-existing parallel-create backend 500 race + clipboard flakiness — tracked separately, out of scope.

## Follow-ups

- #588 — harden flow creation/open against the parallel `POST /api/v1/flows/` 500 race (exposed, not caused, by this change).
- #589 — migrate the 2 remaining non-validated specs and evaluate per-worker user isolation / `cleanAllFlows` deprecation.

### DIFF
--- a/docs/core-components/loop-component-regression.md
+++ b/docs/core-components/loop-component-regression.md
@@ -54,7 +54,7 @@ If any of these tests fails, the Loop component is broken in the product: either
 11. Assert the message **contains** "title" (case-insensitive) via `toContainText(/title/i, { timeout: 240000 })`. `toContainText` re-evaluates as tokens stream in, so it never samples a partially-streamed response — this is what makes the assertion robust against the streaming race tracked in #356 (the earlier code read the text once, on the first streamed token, and intermittently saw no "title" yet). Matching ≥ 1 occurrence confirms at least one complete loop iteration (Parser → LLM → done)
 
 **Test 4 — stops after exhausting input DataFrame and emits aggregated done**
-1. `awaitBootstrapTest` and `cleanAllFlows`
+1. `awaitBootstrapTest` (each created flow's id is tracked and deleted in `afterEach` — scoped teardown, #515 — never a global `cleanAllFlows`, which races concurrent workers)
 2. Build a flow body in-memory from `tests/assets/flows/loop-exit-condition.json` with the Create List node's `texts.value` set to a 3-element list and a randomized flow name
 3. `POST /api/v1/flows/` with the body to create the flow server-side
 4. Navigate to "/" and click the flow card matching the randomized name (the home navigation triggers a full owned-flows re-fetch, avoiding the deep-link cache race)
--- a/docs/core-components/webhook-component-regression.md
+++ b/docs/core-components/webhook-component-regression.md
@@ -38,8 +38,8 @@ If any of these tests fails, the Webhook component is broken in one of its core 
 ## Step by step *(required)*
 
 **Setup helper — `addWebhookComponent`** (shared by tests 1–9)
-1. Call `awaitBootstrapTest` and `cleanAllFlows` to ensure a clean state
-2. Click `blank-flow`
+1. Call `awaitBootstrapTest`
+2. Click `blank-flow`, then capture the created flow id from the `/flow/{id}` URL so `afterEach` can delete only that flow (scoped teardown, #515 — never a global `cleanAllFlows`, which races concurrent workers)
 3. Search "webhook" in the sidebar and wait for `input_outputWebhook` to appear
 4. Hover over the result and click `add-component-button-webhook`
 5. Call `adjustScreenView`
--- a/docs/core-components/webhook-component-regression.md
+++ b/docs/core-components/webhook-component-regression.md
@@ -39,7 +39,7 @@ If any of these tests fails, the Webhook component is broken in one of its core 
 
 **Setup helper — `addWebhookComponent`** (shared by tests 1–9)
 1. Call `awaitBootstrapTest`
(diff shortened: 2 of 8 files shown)


Victor-w-Madeira:
## Merging despite the red "Run impacted E2E specs" check — rationale

The `Run impacted E2E specs` check (added recently in #585) diffs the PR's changed `tests/**/*.spec.ts` and runs **those full specs against a Langflow nightly**. Because this PR touches five specs, the check ran all of them — and surfaced **pre-existing nightly breakage in those specs, none of it a regression from this change**.

**Evidence this PR's teardown migration is clean:**
- No `deleteFlow` leak in any local run — flow count returns to baseline (create N → delete N) in isolated runs.
- None of my capture/guard throws fired (`returned no flow id`, `Flow cleanup failed`).
- The failing assertions are in **test bodies and a shared helper this PR does not modify**.

**Each CI failure maps to an existing, unrelated issue:**

| Failing test | Cause | Issue | Verified |
|---|---|---|---|
| `agent` — renders on canvas | parallel `POST /flows/` 500 race | #588 | **Passes isolated locally**; only fails under parallel create |
| `loop` test 3 — ArXiv template | `model_model` click intercepted in `setup-language-model-openai.ts:108` (a helper this PR doesn't touch) | #580 | **Reproduces isolated & serial locally** — independent of this PR |
| `mcp-client-agent` / `mcp-client-regression` — echo tool | `npx server-everything` registration / echo timing | #463 / #552 | Inferred from error type + open issues (not reproduced locally; needs npx) |

Since the check runs pre-existingly-failing specs against nightly and is non-blocking, and the change itself is verified correct, merging. Follow-ups #588 and #589 track the surrounding work.

Issue #597, opened 2026-07-09
model-provider-model-toggle: 'llm-toggle-gemini-3.5-flash' never appears in Model Providers settings


Victor-w-Madeira:
## Symptom

`@stable` test **\"disabling a model removes it from a component model dropdown\"** (`model-provider-model-toggle.spec.ts:162`) hard-failed on the 2026-07-09 daily run, in the step *\"disable the target model in Settings → Model Providers\"*.

\`\`\`
TimeoutError: locator.waitFor: Timeout 10000ms exceeded.
  - waiting for getByTestId('llm-toggle-gemini-3.5-flash') to be visible
\`\`\`

The per-model toggle for `gemini-3.5-flash` never renders in Settings → Model Providers, so the test cannot disable it.

## History (most persistent of the batch)

| Date | Result |
|---|---|
| 2026-07-06 | flaky |
| 2026-07-07 | **hard fail** |
| 2026-07-08 | **hard fail** |
| 2026-07-09 | **hard fail** |

Three consecutive daily hard failures — hence high priority.

## Likely cause (shallow)

The `gemini-3.5-flash` model isn't present in the provider's model list on the nightly (name/list change or the provider list not loading), so the `llm-toggle-gemini-3.5-flash` control is never rendered. Shares the `gemini-3.5-flash` theme with #552 (mcp-client-agent no echo for the same model) — worth checking for a common root cause.

## Triage action

Per the daily-failure procedure, `@stable` is being removed from this test in a dedicated PR (the auto-remove step never ran this cycle). Restoring `@stable` requires a human check once the root cause is addressed.

Victor-w-Madeira:
**Deliverable:** one of the acceptance criteria for closing this issue is **restoring the `@stable` tag** to `model-provider-model-toggle.spec.ts:162` (*disabling a model*), which was removed in the triage PR as part of the daily-failure procedure. Restoration is intentionally human-gated: re-add `@stable` only after the root cause is fixed and the test is confirmed green.

Issue #596, opened 2026-07-09
language-model: Google provider flow never reaches 'built successfully'


Victor-w-Madeira:
## Symptom

`@stable` test **\"language model must respond with Google provider\"** (`language-model-regression.spec.ts:110`) hard-failed on the 2026-07-09 daily run.

\`\`\`
TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.
  - waiting for locator('text=built successfully') to be visible
\`\`\`

The Language Model component configured with the **Google** provider never emits the `built successfully` toast within 30s — the build/execution does not complete.

## History (degrading signal)

| Date | Result |
|---|---|
| 2026-07-07 | flaky |
| 2026-07-08 | flaky |
| 2026-07-09 | **hard fail** |

It flaked two days in a row and turned into a hard failure today.

## Likely cause (shallow)

Google provider model availability / latency on the nightly, or a change in the selected Gemini model. Needs a focused look at whether the flow errors out or just runs past 30s.

## Triage action

Per the daily-failure procedure, `@stable` is being removed from this test in a dedicated PR (the auto-remove step never ran — the scheduled run was cancelled at the queue stage and the fix run went through `workflow_dispatch`). Restoring `@stable` requires a human check once the root cause is addressed.

Victor-w-Madeira:
**Deliverable:** one of the acceptance criteria for closing this issue is **restoring the `@stable` tag** to `language-model-regression.spec.ts:110` (*Google provider*), which was removed in the triage PR as part of the daily-failure procedure. Restoration is intentionally human-gated: re-add `@stable` only after the root cause is fixed and the test is confirmed green.

Issue #594, opened 2026-07-09
daily-stable: ollama service pull denied — permissions block missing packages: read


Victor-w-Madeira:
## Summary

The `daily-stable.yml` scheduled run fails at **Initialize containers** because the `ollama` service image cannot be pulled from GHCR:

\`\`\`
docker pull ghcr.io/oriontech-me/langflow-e2e/ollama-e2e:llama3.2-1b
Error response from daemon: denied
\`\`\`

With `Initialize containers` failing, checkout / install / **Run @stable tests** are all skipped, and the history-append + commit steps fail too — so no entry is written to `reports/daily-history.jsonl`.

## Root cause

`daily-stable.yml` declares an explicit `permissions:` block:

\`\`\`yaml
permissions:
  issues: write
  contents: write
\`\`\`

When a workflow sets an explicit `permissions:` block, every scope not listed defaults to `none`. The `ollama` service pulls a **private** GHCR image using `GITHUB_TOKEN`, which therefore has `packages: none` → the pull is denied.

`nightly.yml` has **no** `permissions:` block, so its `GITHUB_TOKEN` inherits the repository default (which includes `packages: read`) and pulls the same image fine.

## Why it surfaced now

The `ollama` service was added in #583 (commit `1a1190f`) on 2026-07-08 16:45 BRT. The last scheduled daily run before that (run `28935001447`) ran at 2026-07-08 10:14 UTC — **before** the ollama service existed, so its Initialize step passed. The 2026-07-09 run is the first scheduled daily-stable to include ollama, and it failed. This will recur on every scheduled run until fixed.

## Fix

Add `packages: read` to the `permissions:` block in `.github/workflows/daily-stable.yml`:

\`\`\`diff
 permissions:
   issues: write
   contents: write
+  packages: read
\`\`\`

## Notes

- Today's gap in `reports/daily-history.jsonl` (2026-07-09) will only be filled once the workflow runs again with the fix in place (next scheduled run, or a rerun after merge).
- The 2026-07-09 run first attempt was separately cancelled at the queue stage (GitHub failed to allocate a runner) — a transient infra issue, not addressed here.

Issue #580, opened 2026-07-08
loop-component: model_model click intercepted by api_key popover on nightly (setup fails)


daniellicnerski1:
## Failing test — tracking

`loop-component-regression.spec.ts` — the `@stable` test that configures OpenAI via
`setupLanguageModelOpenAI` (the test whose body calls the helper) **hard-fails during
setup** on the current Langflow nightly.

- **File:** `tests/tests-automations/regression/core-components/loop-component-regression.spec.ts` (the `setupLanguageModelOpenAI` test, ~line 111)
- **Env:** `langflowai/langflow-nightly:latest` (1.11.0.dev36)

## Symptom

`setupLanguageModelOpenAI` fails at the final model selection:

```
TimeoutError: locator.click: Timeout 20000ms exceeded.
  - waiting for getByTestId('model_model')
  - <div data-testid="anchor-popover-anchor-input-api_key" ...> ... subtree intercepts pointer events
```

The `model_model` combobox is visible and enabled, but the node's **`api_key` input
popover** (`anchor-popover-anchor-input-api_key`) overlays it and steals the pointer
event, so the click never lands and times out.

## Pre-existing / not a regression

Confirmed this reproduces **identically on `origin/main`** (baseline, without any recent
changes) — it is a pre-existing failure surfaced by the current nightly UI, not caused
by PR #579. PR #579 only touches the `memory-history` test + the shared helper's
toggle-enabling / model-pin logic; the `model_model` interception is upstream of that.

Surfaced while working on #569 / PR #579 (the shared helper `setupLanguageModelOpenAI`
is used by both specs).

## Likely cause

The Agent/model node renders an `api_key` input popover that is positioned over the
`model_model` combobox. When the credential is not yet bound (or the popover is open),
Playwright's actionability check sees the popover intercepting pointer events on
`model_model`.

## Action

- [ ] Reproduce and capture the node state when the popover overlaps `model_model`
- [ ] Harden `setupLanguageModelOpenAI` model selection against the `api_key` popover
  (dismiss/blur the popover before clicking `model_model`, or ensure the credential is
  bound so the popover does not open)
- [ ] Consider whether the memory-history `setAgentModelViaApi` model-pin approach
  (PR #579) is the more robust path for `loop-component` too
- [ ] Confirm no `🚨 Backend Error` on a clean run

## Watch list

- A second test in the same file (`run without connections shows build failed
  notification`, ~line 80) **passed in isolation** but **failed within the full-file
[...]

PR #598, opened 2026-07-09
test(triage): remove @stable from 4 tests hard-failing on 2026-07-09 daily

Victor-w-Madeira:
## What

Removes `@stable` from the four tests that hard-failed on the **2026-07-09** daily run:

| Test | File | Tracking issue |
|---|---|---|
| Loop — Research Translation Loop template | `loop-component-regression.spec.ts:150` | #580 |
| language model — Google provider | `language-model-regression.spec.ts:110` | #596 |
| disabling a model removes it from a component model dropdown | `model-provider-model-toggle.spec.ts:162` | #597 |
| select + bulk actions | `bulk-actions.spec.ts:8` | #420 |

Only the `@stable` tag is removed from each `test()`; no other tags, test logic, or spec docs change.

## Why

This is the **manual equivalent of the daily workflow's auto-remove step**. This cycle the scheduled `daily-stable` run was cancelled at the queue stage (GitHub failed to allocate a runner), and the fix run (#594) executed via `workflow_dispatch`, whose auto-remove/issue steps are gated on `schedule` — so nothing fired automatically. Each failure was triaged and tracked:

- **#580** (existing) — loop `model_model` click intercepted by the api_key popover. Recurrence logged.
- **#596** (new) — Google provider flow never reaches `built successfully` (flaky 07-07/08 → hard fail 07-09).
- **#597** (new, high priority) — `llm-toggle-gemini-3.5-flash` never renders in Model Providers settings (3 consecutive daily hard fails).
- **#420** (existing) — bulk-actions empty-state/welcome-panel; new facet folded in.

## Deliverable / restoration

Restoring `@stable` to each test is human-gated and tracked as a deliverable on the respective issue — re-add the tag only after the root cause is fixed and the test is confirmed green.

## Notes

- No spec-doc change: per the tag policy, temporary `@stable` removals are tracked via a GitHub issue, not a spec-doc edit.
- `QA-CHECKLIST.md` Phase 0 / stable block is auto-generated — `update-coverage-summary.yml` regenerates it on merge with `[skip ci]`; not hand-edited here.

Refs #580, #596, #597, #420

### DIFF
--- a/tests/tests-automations/regression/core-components/loop-component-regression.spec.ts
+++ b/tests/tests-automations/regression/core-components/loop-component-regression.spec.ts
@@ -149,7 +149,7 @@ test(
 
 test(
   "Loop component — Research Translation Loop template: full wiring and iterates over 2 ArXiv papers",
-  { tag: ["@stable", "@release", "@components", "@templates", "@playground"] },
+  { tag: ["@release", "@components", "@templates", "@playground"] },
   async ({ page }) => {
     test.skip(
       !process.env.OPENAI_API_KEY,
--- a/tests/tests-automations/regression/core-functionality/llm-agents/language-model-regression.spec.ts
+++ b/tests/tests-automations/regression/core-functionality/llm-agents/language-model-regression.spec.ts
@@ -109,7 +109,7 @@ test.describe("Language Model Component Regression", () => {
 
   test(
     "language model must respond with Google provider",
-    { tag: ["@stable", "@release", "@components", "@model-provider"] },
+    { tag: ["@release", "@components", "@model-provider"] },
     async ({ page }) => {
       test.skip(
         !process?.env?.GOOGLE_API_KEY,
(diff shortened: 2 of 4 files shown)


Victor-w-Madeira:
Merging despite the red **Run impacted E2E specs** check — this is expected and is precisely the reason for this PR.

That check re-runs the changed specs against the nightly, and these four specs are exactly the tests that hard-failed on the 2026-07-09 daily run. Their failure here confirms the very instability this PR responds to. TypeScript, ESLint and Detect changed specs all pass; the only red check is the failing tests themselves.

The PR does not fix those tests — it removes `@stable` from them (the manual equivalent of the daily workflow's auto-remove step, which didn't fire this cycle) so they stop gating the stable suite while their root causes are tracked in #580, #596, #597 and #420. Restoring `@stable` is human-gated on each fix.

PR #625, opened 2026-07-10
test(observability): cascade span-delete regression for clear-all traces

Victor-w-Madeira:
Adds a standalone API regression spec `traces-delete-cascade.spec.ts` guarding the bulk trace delete (`DELETE /api/v1/monitor/traces?flow_id=...`, the wire behind **Clear All**) against a re-introduction of the span→trace cascade gap fixed upstream in `langflow-ai/langflow#13955` (*"Langflow traces - Clear all does not worked"*).

**Root cause of the upstream bug:** `span.trace_id` was declared without `ondelete="CASCADE"` (unlike `trace.flow_id`). The handler deletes traces with a single raw `sa.delete(TraceTable)` that bypasses the ORM cascade, so it relied on a DB-level cascade that did not exist — a trace with spans made the delete violate `span_trace_id_fkey`, return **500**, and leave every trace behind. The fix (upstream PR #13976, `release-1.11.0`) adds the cascade.

## What the test does

- Seeds a flow whose run emits a trace with a **populated span tree** (`ChatInput → ChatOutput`, completes successfully, no provider key needed).
- Anchors on `spanCount > 0` — this is what makes the DELETE exercise the span→trace FK, distinguishing it from the plain contract test.
- Asserts `DELETE` returns **204** + empty body, the trace list is emptied (`total: 0`), and the trace detail is **404**.

Deliberately scoped to the **cascade / data-integrity** dimension. The delete endpoint's status-code / ownership contract stays owned by `traces-delete.spec.ts` — no duplicated coverage (boundary documented in the spec doc's *What this test does not cover*).

## Validation

Internal validation guide (CONTRIBUTING.md), run against **both sides of the fix with `foreign_keys` enforced**:

- **Red on the bug** — Langflow 1.10.0 (buggy) → `DELETE` 500 → test fails on the 204 assertion only; `beforeAll` (anchor, run 200, spans present) and `afterAll` (cleanup 200/200) all clean. Confirmed no false positive via full trace.
- **Green on the fix** — Langflow 1.11.0 (fixed) → `DELETE` 204 → `1 passed`. Same test, same FK-enforced condition; only the Langflow version changes.
- **Backend/flow logs** — none; note this is an API-only test (`request` fixture), so the fixtures' `page`-based backend-error monitor does not observe the call — the explicit status assertion is the sole (by-design) detection signal.
- **typecheck** clean; **lint** 0 errors.
- **Checklist** — section 8.1 bullets added; Coverage Summary + Phase 0 blocks regenerated via `npm run coverage:summary`.

## Notes

- ⚠️ **The bug is only observable when the DB enforces foreign keys.** Postgres always does; SQLite requires `PRAGMA foreign_keys=ON`, which Langflow omits by default — against an FK-unenforced SQLite SUT the buggy delete "succeeds" (204, orphaned spans) and the test passes for the wrong reason. Documented in the spec doc's *External dependencies*.
- Consequently, on the current nightly/daily SUT (vanilla SQLite, FK off) this test passes trivially. **Making it bite in CI requires enabling the pragma in `scripts/start-langflow-docker.sh`** — a shared-infrastructure change intentionally kept out of this PR and left as a dedicated follow-up.
- `@stable` per policy (spec ships stable; lifecycle handled by triage).

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -433,6 +433,7 @@
 - [x] Single-trace API returns populated tokenUsage + modelName on the LLM span (OpenAI) → `traces-detail-llm-span-populated.spec.ts`
 - [x] Bulk delete traces API returns 404 for an unknown flow_id → `traces-delete.spec.ts`
 - [x] Bulk delete traces API clears all traces for the flow (204 + empty list) → `traces-delete.spec.ts`
+- [x] Bulk delete of a trace with a populated span tree cascades (204, no FK violation) — regression #13955 → `traces-delete-cascade.spec.ts`
 - [x] Trace list filter `?status=error` returns only the failing trace; `?status=<unknown>` returns 422 → `traces-list-filters.spec.ts`
 - [x] Trace list filter `?status=ok` returns only the successful trace → `traces-list-filters.spec.ts`
 - [x] Trace list filter `?start_time` pins the >= lower bound (past hits, future misses) → `traces-list-filters.spec.ts`
@@ -760,7 +761,7 @@
 | `core-functionality/knowledge-ingestion/` | 8 | 0 | 4 | 0 | 4 |
 | `core-functionality/llm-agents/` | 40 | 30 | 2 | 0 | 8 |
 | `core-functionality/model-provider/` | 33 | 31 | 0 | 0 | 2 |
-| `core-functionality/observability-monitoring/` | 23 | 15 | 7 | 0 | 1 |
+| `core-functionality/observability-monitoring/` | 24 | 16 | 7 | 0 | 1 |
 | `core-functionality/playground/` | 48 | 43 | 3 | 1 | 1 |
 | `core-functionality/project-management/` | 11 | 4 | 6 | 1 | 0 |
 | `core-functionality/templates/` | 41 | 2 | 39 | 0 | 0 |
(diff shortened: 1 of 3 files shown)


Victor-w-Madeira:
Pushed 4257fb6 addressing the review + the impacted-E2E failure.

**Impacted-E2E failure (root cause):** the check runs this spec against the *nightly* image, and it failed in `beforeAll` — not on the bug assertion. My `ChatInput → ChatOutput` fixture emitted spans locally but not reliably on the nightly SUT, so the span-count anchor timed out. Switched the seed to `basic-prompting-trace-fixture.json` — the same provider-less fixture used by `traces-delete.spec.ts` and the `traces-detail-*` `@stable` specs, which prove it emits `spans.length > 0` against the nightly SUT daily. Run now accepts `[200, 500]`.

**Copilot comments (all addressed):**
- Added `@release` to the tag set (parity with the module's other `@stable` specs) — spec + doc Tags.
- Corrected the spec doc `Last validated` to `1.11.x` (green on the fixed release; red on 1.10.0) — it was misleadingly `1.10.x`, a version where the test is expected to be red.

Re-validated green locally on Langflow 1.11.0 with FK enforced; typecheck + lint clean.

Victor-w-Madeira:
**Merging despite the red "Run impacted E2E specs" check — it is a non-blocking, environment limitation, not a defect.**

Root cause: the impacted-E2E job disables tracing (`LANGFLOW_DEACTIVATE_TRACING: "true"` in `pr-validation.yml` and `adaptive-impacted.yml`), so no traces/spans are emitted and this trace-dependent spec times out in `beforeAll` before ever reaching the delete assertion. This spec is the first trace-dependent test to exercise that path; the sibling `traces-detail-*` specs only run on `daily-stable`, which deliberately keeps tracing **ON** (see its inline comment) and where they pass daily.

- Blocking checks (TypeScript, ESLint) are green.
- The spec passes on the daily SUT (tracing on) and was validated red-on-1.10.0 / green-on-1.11.0 locally with FK enforced.

Follow-up (separate PR): make the impacted workflows tracing-aware for observability/traces specs (enable tracing for them, or exclude them from the tracing-off run).

Issue #501, opened 2026-07-02
Promote collect-models.spec.ts to @stable — key validation + model collection + Save/Replace


Victor-w-Madeira:
**Wave 1 deliverable** — see `ROADMAP.md` · Milestone: Wave 1 — Agents & providers (2026-07-14)

**Type:** Promote existing spec to @stable
**Spec file:** `tests/collect-models.spec.ts`
**Spec doc:** `docs/collect-models.md` (to create)

**Checklist bullets (`QA-CHECKLIST.md`):**
- §7.1 — Validate API keys of all providers via real call
- §7.1 — Collect available models per provider via UI
- §7.1 — Configure provider API key via Save Configuration (first setup)
- §7.1 — Replace provider API key via Replace Configuration (existing key)

**Notes:** Spec already exists (`docs/collect-models.md`). This generates `models.json` — foundation for the agent specs.

**Done when:** spec written + `@stable` + spec doc + checklist bullet(s) flipped to `[x]`.

PR #648, opened 2026-07-10
fix(collect-models): candidate-model fallback + settled-model promotion + anti-skip gate (#570)

### LINKED ISSUE #570: collect-models: single inaccessible lead model disables whole provider (16 OpenAI agent tests silently skipped)
## Summary

`collect-models` probes **only the first** model per provider, so a single
inaccessible lead model disables the whole provider for the suite. On the
2026-07-08 daily run this silently skipped **16 OpenAI-variant agent tests**:
Langflow nightly's provider panel now lists `gpt-5.5-pro` as the first OpenAI
model, and the CI OpenAI project (`proj_vtkyNBz9ytVmv1HpewDvfUaL`) has no access
to it, so OpenAI was marked `inactive` and every OpenAI-variant agent test
`test.skip()`-ed itself.

Skip reason observed on all 16:

```
Provider "openai" inactive — Project `proj_vtkyNBz9ytVmv1HpewDvfUaL` does not have access to model `gpt-5.5-pro`
```

## Mechanism

1. `tests/collect-models.spec.ts` (CI "Collect models" step) scrapes the visible
   model toggles from the Settings → providers panel into `models.json`.
2. `collect-models.ts` → `collectProviders()` validates each provider with a
   **real** chat completion, using only `firstModelFor(models, "openai")`
   (`collect-models.ts:124–135`).
3. On nightly the first OpenAI toggle is `gpt-5.5-pro`; `validateOpenAI` gets an
   OpenAI access error → provider recorded `status: "inactive"`.
4. Each agent spec builds `Provider "openai" inactive — <error>` and skips the
   OpenAI parametrization (e.g. `agent-system-prompt.spec.ts:53`).

## Impact — silently skipped OpenAI agent coverage (run 28935001447)

16 tests, all `@stable`/`@release`:

- `agent-component-regression.spec.ts` (1)
- `agent-empty-refusal-response.spec.ts` (2)
- `agent-input-sources.spec.ts` (2)
- `agent-max-iterations.spec.ts` (2)
- `agent-max-tokens.spec.ts` (2)
- `agent-multimodal-image-input.spec.ts` (2)
- `agent-system-prompt.spec.ts` (2)
- `agent-tool-error-handling.spec.ts` (1)
- `agent-tool-name-validation.spec.ts` (2)

Chronic: 2026-07-07 already reported ~18 skips, consistent with the same 16
OpenAI skips (a skip is not a failure, so it never trips the daily-failure gate
— it just erodes coverage quietly).

## Proposed fix (a) — validate a candidate list, not just the first

Make `collectProviders`/`validateOpenAI` try multiple candidate models and pick
the first that actually validates, instead of `firstModelFor` only. This mirrors
the fallback already in `setup-language-model-openai.ts`
(`selectPreferredChatModel` iterates a preferred list). A gated lead model
(preview/pro) then no longer disables the whole provider.

Alternatives considered:
- (b) Grant the CI OpenAI project access to `gpt-5.5-pro` — account change, and
  doesn't prevent the next preview model from re-triggering this.
- (c) Pin `MODEL_TEST_ID` to an accessible model (`gpt-5.4` is `active` in the
  committed `providers.json`) — quick unblock but static; loses "test whatever
  nightly exposes".

## Acceptance criteria

- [ ] A gated lead model no longer marks the whole provider `inactive` when
  another accessible model exists
- [ ] The 16 OpenAI-variant agent tests run (not skip) against nightly
- [ ] `collect-models` logs which model it settled on per provider
- [ ] `npm run typecheck` / `npm run lint` pass

Surfaced during daily-failure triage #566 (skipped-tests analysis). Related: #568, #569.

rafaelgiln:
Fixes #570 (`qa-infra` / `daily-failure` — approved exception issue, from triage #566).

## Root cause

`collectProviders` probed **only** `firstModelFor()` per provider. One gated/preview lead model (nightly listed `gpt-5.5-pro` first; the CI OpenAI project has no access) marked the whole provider `inactive`, and every spec parametrized on it silently `test.skip()`-ed — 16 OpenAI-variant agent tests on run 28935001447, invisible to the daily-failure gate because a skip is not a failure. The same class hit Google (`gemini-omni-flash-preview` leads the catalog and only supports the Interactions API).

## Fix (proposed option (a) + two gaps found during validation)

1. **`validateProviderWithFallback`** — try every collected model in catalog order, settle on the first that validates. Failed probes are rejected before inference (zero token cost); only the successful probe consumes ~1 token, so exhausting the catalog costs seconds, not money. Missing env key short-circuits. Logs the settled model and each skipped candidate (acceptance criterion).
2. **`promoteSettledModels`** — models.json is persisted with each provider's settled model first. Spec parametrization takes the FIRST model per provider, so without this a provider validated via fallback would still be *tested* against its unrunnable lead model — converting #570's silent skips into hard failures (proven live: google went active but the sample spec targeted `gemini-omni-flash-preview` and failed).
3. **Anti-skip gate in `collect-models.spec`** — an env-keyed provider that still ends `inactive` now FAILS the spec, and with it the CI "Collect models" step → daily-failure issue the same day. With the full-catalog fallback, that state means a genuine key/account/provider problem; failing loudly is the guarantee that provider-inactive skips can never erode coverage silently again. (The `collectAll` helper itself stays tolerant/never-throws — the gate lives in the spec, preserving the #501 contract split.)

## Validation (nightly 1.11.0.dev38)

- Live run: `google: candidate "gemini-omni-flash-preview" failed — This model only supports Interactions API.` → `settled on "gemini-3.5-flash"` → **all 3 providers active** (google/openai had been inactive locally for days).
- models.json leads after promotion: `openai → gpt-5.6`, `anthropic → claude-sonnet-5`, `google → gemini-3.5-flash`.
- Downstream proof: `agent-tool-name-validation` ran **2/2 green** on the settled Google model; the OpenAI target **executed instead of skipping** (issue acceptance).
- **Force-fail executed:** loop capped to 1 candidate → reproduced the old behavior exactly (google ❌ inactive, `all 1 candidate model(s) failed validation`). Revert proven (`grep FF-MUTATION` → 0) + green re-run.
- `npm run typecheck` / `npm run lint`: 0 errors. Orphan flows: 0.
- Note: the final cap-removal (5 → full catalog) and the new spec gate were typechecked but not re-run live per reviewer's call — behavior-identical for the green path (google settles at candidate 2); the gate triggers only on the inactive-with-key state exercised by the force-fail above.

## Real finding surfaced (separate issue candidate — not papered over)

`gpt-5.6` validates on the raw API probe (~1-token completion) but Langflow's **Agent** returns an empty reply with it (Playground renders "Message empty."), failing answer-content asserts. Different class (agent × model integration, not collect-models). Will open a dedicated issue on request.

### DIFF
--- a/tests/collect-models.spec.ts
+++ b/tests/collect-models.spec.ts
@@ -78,5 +78,26 @@ test(
         }
       }
     });
+
+    await test.step("every env-keyed provider is ACTIVE — inactive silently skips its whole parametrization", async () => {
+      // The #570 guarantee: a provider whose key is configured but that ends
+      // "inactive" silently test.skip()s every spec parametrized on it (16
+      // OpenAI-variant agent tests on 2026-07-08) — and a skip never trips
+      // the daily-failure gate. With the candidate fallback in collectAll,
+      // inactive-with-key now means ALL probed candidates failed: a real
+      // key/account/provider problem that must fail this spec (and the CI
+      // "Collect models" step) loudly instead of eroding coverage quietly.
+      const providers = JSON.parse(fs.readFileSync(PROVIDERS_PATH, "utf-8")) as ProviderRecord[];
+      for (const p of providers) {
+        const envKeys = providerConfigMap[p.provider as Provider]?.envKeys ?? [];
+        const hasEnvKey = envKeys.some((k: string) => !!process.env[k]);
+        if (hasEnvKey) {
+          expect(
+            p.status,
+            `env-keyed provider "${p.provider}" must probe active (error: ${p.error})`,
+          ).toBe("active");
+        }
+      }
+    });
   },
 );
--- a/tests/helpers/provider-setup/collect-models.ts
+++ b/tests/helpers/provider-setup/collect-models.ts
@@ -121,17 +121,65 @@ async function validateGoogle(model: string): Promise<ProviderRecord> {
   }
 }
 
-function firstModelFor(models: ModelRecord[], provider: string): string {
-  return models.find((m) => m.provider === provider)?.model ?? "";
+function modelsFor(models: ModelRecord[], provider: string): string[] {
+  return models.filter((m) => m.provider === provider).map((m) => m.model);
+}
+
+// A single gated/preview lead model must not disable the whole provider
+// (#570: nightly listed gpt-5.5-pro first, the CI project had no access to
+// it, and 16 OpenAI-variant agent tests silently skipped). Try EVERY
+// collected model in catalog order and settle on the first that validates —
+// failed probes are rejected before inference (zero token cost), and only
+// the single successful probe consumes ~1 token, so exhausting the catalog
+// costs time (~1s per candidate, once per run), not money. "inactive" then
+// genuinely means nothing the provider exposes works with this key.
+async function validateProviderWithFallback(
+  provider: string,
+  candidates: string[],
+  validate: (model: string) => Promise<ProviderRecord>,
+): Promise<ProviderRecord> {
+  if (candidates.length === 0) {
+    return {
+      provider,
+      model: null,
+      status: "inactive",
+      error: "no models collected from the providers panel",
+      checkedAt: new Date().toISOString(),
+    };
+  }
+
+  const tried: string[] = [];
+  let last: ProviderRecord | null = null;
+  for (const model of candidates) {
+    const result = await validate(model);
+    if (result.status === "active") {
+      if (tried.length > 0) {
+        console.log(
+          `   ${provider}: settled on "${model}" after skipping ${tried.length} gated/unavailable candidate(s): ${tried.join(", ")}`,
+        );
+      }
+      return result;
+    }
+    // A missing key fails identically for every candidate — stop immediately.
+    if (result.error?.endsWith("not set")) return result;
+    console.log(`   ${provider}: candidate "${model}" failed — ${result.error}`);
+    tried.push(model);
+    last = result;
+  }
+
+  return {
+    ...(last as ProviderRecord),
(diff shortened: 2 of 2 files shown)


rafaelgiln:
## Downstream validation — the 9 affected spec files, CI-mirror (one model per provider, all 3 providers)

Ran the full affected-spec batch on the post-merge state (main + this PR) before approval. Result drove one additional commit (`67bab35`).

**Round 1 — catalog-order candidates:** OpenAI settled on `gpt-5.6` (raw probe passes) and **8/8 non-multimodal spec files failed on the openai target** — the Agent returns an empty reply with `gpt-5.6` on the 1.11 nightly ("Message empty."). The raw probe validates *access*, not *agent compatibility*.

**Fix added:** `rankCandidates()` — probe known agent-compatible models first (pref regexes mirroring `resolveGptModel`/`resolveGeminiModel` + anthropic), raw catalog order as tail. Settled: `gpt-4o-mini` / `claude-sonnet-5` / `gemini-2.5-flash`.

**Round 2 — with ranking:** **35/39 tests green across all 3 providers.** The 4 failures are unrelated to this PR:
- `agent-max-iterations [openai]` + `agent-max-tokens [openai]` — one-off flakes (canvas-load timeout / wording), both **2/2 green on isolated re-run**.
- `agent-input-sources [google]` + `agent-tool-error-handling [anthropic]` — the Agent intermittently returns an **empty reply** ("Message empty."), observed today across **4 different models** (gpt-5.6 deterministic; gemini-2.5-flash, gemini-3.5-flash, claude-sonnet-5 intermittent). Product-side pattern on nightly 1.11.0.dev38, hits the suite regardless of this PR — dedicated issue to follow.

PR #736, opened 2026-07-14
fix(history): record error_signature on flaky entries

### LINKED ISSUE #728: history appender does not record error_signature for flaky entries — blocks CONTRIBUTING flake-recurrence criterion
## Gap

`reports/daily-history.jsonl` records `error_signature` for **hard failures** (`.failures[]`) but **not for flaky entries** (`.flaky[]`, whose keys are only `test, file, line, tags, attempts`).

Confirmed while triaging #704: the flaky rows carry no signature.

## Why it matters

`CONTRIBUTING.md` → *Triage protocol* keys the flake-recurrence decision on the **`error_signature`** ("same signature within a 30-day window → recurrent → dedicated issue + remove `@stable`"). Because flaky rows omit the signature, an analyst **cannot mechanically apply that criterion** — recurrence can only be judged by test identity across days, which cannot distinguish "same sustained cause" from "two unrelated blips on the same test." The history file is meant to be the source of truth for exactly this decision.

Real consequence in #704: two recurrent flakes (`agent-tool-error-handling`, `edit-flow-name`) had to be judged on test identity + healthy-day recurrence rather than on signature, and the triage summary had to carry an explicit caveat.

## Proposed fix

Record `error_signature` on flaky entries too, in `scripts/append-weekly-history.mjs` (the shared appender). The signature already exists for the failing attempt of a flaky test (it failed before passing on retry) — surface the first failed attempt's error message through the same normaliser used for hard failures. Keep it backwards-compatible (additive field → no schema `version` bump per `reports/README.md`).

## Acceptance

- New flaky rows in `daily-history.jsonl` include `error_signature`.
- The 30-day same-signature recurrence query in `CONTRIBUTING.md` works against `.flaky[]` as well as `.failures[]`.
- `reports/README.md` notes the field is now present on flaky rows.

Victor-w-Madeira:
## What

The history appender (`scripts/append-weekly-history.mjs`) recorded `error_signature` only for **hard failures** (`.failures[]`), leaving **flaky** rows (`.flaky[]`) without it. This blocked the flake-recurrence criterion in `CONTRIBUTING.md`, which keys on **same signature within a 30-day window**: without a signature on flaky rows, recurrence could only be judged by test identity — unable to distinguish one sustained cause from two unrelated blips on the same test (the exact caveat that #704 triage had to carry).

## Changes

- **`scripts/append-weekly-history.mjs`** — flaky entries now record `error_signature`, taken from the **first failed attempt** (the one that triggered the retry that later passed), reusing the same `firstErrorMessage` normaliser used for hard failures. Falls back to `"unknown"` when no message is available. Additive field → **no schema `version` bump**, per `reports/README.md`.
- **`reports/README.md`** — documents the field on `flaky[]` rows; clarifies that `failures[]` uses the **last** failed result while `flaky[]` uses the **first**, and that both arrays now carry `error_signature` so the 30-day criterion applies mechanically to either.
- **`CONTRIBUTING.md`** — no change needed: the triage query already used `(.failures + .flaky)[]` with a `// "flaky"` fallback. The fallback still covers old (pre-fix) rows; new rows now carry the real signature.

## Verification

- Ran the appender against a synthetic report: flaky row got the first-attempt signature (`TimeoutError: ... Timeout 5000ms exceeded`), hard failure kept its last-result signature.
- The 30-day same-signature recurrence query now groups flakes by signature in `.flaky[]` (count ≥ 2 → recurrent).
- ESLint passes.

## Acceptance (from #728)

- [x] New flaky rows include `error_signature`.
- [x] The 30-day same-signature recurrence query works against `.flaky[]` as well as `.failures[]`.
- [x] `reports/README.md` notes the field is now present on flaky rows.

Closes #728

### DIFF
--- a/reports/README.md
+++ b/reports/README.md
@@ -58,7 +58,8 @@ The series is **not continuous**. Document any missing weeks here rather than ba
       "file": "...",
       "line": 78,
       "tags": [...],
-      "attempts": 2                          // result entries; >1 means at least one retry happened
+      "attempts": 2,                         // result entries; >1 means at least one retry happened
+      "error_signature": "..."               // first line of the FIRST failed attempt (before the passing retry)
     }
   ]
 }
@@ -67,8 +68,8 @@ The series is **not continuous**. Document any missing weeks here rather than ba
 ### Field semantics
 
 - `tags` reflects the **state at the moment of the run**, not the current state in the repo. A test that was `@stable` at run time and has since had `@stable` removed will still show `@stable` in its historical entries.
-- `error_signature` is the first non-empty line of the last failed result's error message, truncated to 240 chars. Stack frames and locator details are stripped — enough to cluster recurring failures, not enough to debug from history alone.
-- `failures` are tests where Playwright's final `test.status === "unexpected"`. `flaky` are tests where final status is `"flaky"` (failed and then passed on retry).
+- `error_signature` is the first non-empty line of a failed result's error message, truncated to 240 chars. Stack frames and locator details are stripped — enough to cluster recurring failures, not enough to debug from history alone. For `failures[]` it is taken from the **last** failed result (the one that made the test fail for good); for `flaky[]` it is taken from the **first** failed attempt (the one that triggered the retry that later passed). Falls back to `"unknown"` when no message is available.
+- `failures` are tests where Playwright's final `test.status === "unexpected"`. `flaky` are tests where final status is `"flaky"` (failed and then passed on retry). Both carry `error_signature`, so the 30-day same-signature flake-recurrence criterion in `CONTRIBUTING.md` applies mechanically to either array.
 
 ---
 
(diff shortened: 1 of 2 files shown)


Victor-w-Madeira:
Addressed both Copilot review comments in 79f03d1 (fixed forward in the implementation rather than relaxing the docs):

1. **Flaky signature selection** — no longer relies on `status !== "passed"` (which could latch onto an `interrupted`/no-message attempt). It now picks the first result that actually carries a message, so a no-message attempt preceding the real failure is skipped.
2. **`firstErrorMessage` vs. doc wording** — the normaliser now returns the first **non-empty** line (trimmed), matching the "first non-empty line" wording in `reports/README.md` and skipping leading blank lines instead of emitting an empty signature.

Verified with an edge-case synthetic report (leading blank lines + a no-message `interrupted` attempt before the failed one): the signature comes out clean. ESLint passes.

PR #793, opened 2026-07-16
fix(ci): stop migration-test jobs from closing each other's failure issues

### LINKED ISSUE #770: Langflow Migration Test Failed (latest → nightly)
## Run #100 — 2026-07-15

docker-compose migration test failed. Source: `langflowai/langflow:latest` → Target: `langflowai/langflow-nightly:latest`.
Compose file: pinned to upstream main of `langflow-ai/langflow/docker_example/docker-compose.yml`.

[Workflow Run](https://github.com/oriontech-me/langflow-e2e/actions/runs/29403094622)

/cc @lice-reis

lice-reis:
## Summary
- `migration-test` (API-script job) and `migration-test-compose` (docker-compose job) both open/close their failure-tracking issue using the same `migration-test` label.
- Because the two jobs run in parallel, a pass in one job's "Close issue on success" step can close the issue the *other* job just filed on failure — silently marking a real regression as resolved.
- Confirmed on [run #101](https://github.com/oriontech-me/langflow-e2e/actions/runs/29485977654): `migration-test-compose` failed at 09:10:46 UTC with an outdated-component `HTTP 400` on the migrated witness flow and filed/updated issue #770; `migration-test` passed at 09:12:20 UTC and its "Close issue on success" step closed #770 right after, even though the compose failure was never addressed.
- Fix: give each job its own tracking label (`migration-test-api` / `migration-test-compose`, both created in the repo) while keeping `migration-test` as a shared label on new issues for cross-job queries. Each job's success/failure logic now only ever touches its own issue.

## Test plan
- [ ] Next scheduled `migration-test.yml` run: confirm the compose-job failure (if it recurs) opens/keeps open its own issue independent of the API job's result
- [ ] Confirm a genuine pass in both jobs still closes both issues correctly

Note: the compose job's underlying failure (outdated OpenAI component blocking flow execution post-migration) is a separate, real issue — the flow_validation guard in `lfx/utils/flow_validation.py` intentionally blocks running components whose code hash differs from the server's, and nightly rebuilds daily. That's left out of scope for this PR, which only fixes the issue-tracking collision.

### DIFF
--- a/.github/workflows/migration-test.yml
+++ b/.github/workflows/migration-test.yml
@@ -279,7 +279,7 @@ jobs:
             const { data: existing } = await github.rest.issues.listForRepo({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              labels: 'migration-test',
+              labels: 'migration-test-api',
               state: 'open',
             });
             if (existing.length > 0) {
@@ -320,7 +320,7 @@ jobs:
             const { data: existing } = await github.rest.issues.listForRepo({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              labels: 'migration-test',
+              labels: 'migration-test-api',
               state: 'open',
             });
 
@@ -345,9 +345,9 @@ jobs:
               await github.rest.issues.create({
                 owner: context.repo.owner,
                 repo: context.repo.repo,
-                title: `Langflow Migration Test Failed (latest → nightly)`,
+                title: `Langflow Migration Test Failed — API script (latest → nightly)`,
                 body,
-                labels: ['migration-test', 'automated'],
+                labels: ['migration-test-api', 'migration-test', 'automated'],
               });
             }
 
@@ -696,7 +696,7 @@ jobs:
             const { data: existing } = await github.rest.issues.listForRepo({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              labels: 'migration-test',
+              labels: 'migration-test-compose',
               state: 'open',
             });
             if (existing.length > 0) {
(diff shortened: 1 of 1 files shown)


Victor-w-Madeira:
Surfaced this PR while auditing stale branches — it is 12 days without review, and the
collision it fixes is **still firing on `main`**, so here is the evidence to unblock it.

`migration-test.yml` is scheduled once a day (`cron: "0 7 * * *"`), so any failure issue that
is opened and closed inside the same hour cannot be "a later run fixed it" — it can only be
the sibling job. Two examples from the `migration-test` label:

| Issue | Created | Closed | Lifetime |
|---|---|---|---|
| #815 | 2026-07-18 08:45:01Z | 2026-07-18 08:46:03Z | **62 s** |
| #801 | 2026-07-17 09:05:54Z | 2026-07-17 09:07:25Z | **91 s** |

One job filed a real failure and the other job's *Close issue on success* step closed it about
a minute later, exactly as described in the summary. `main` still carries the single
`migration-test` label at all four call sites, so nothing has fixed this in the meantime.

A second, smaller symptom worth folding in if you touch the file again: both jobs also write
the **same issue title** (`Langflow Migration Test Failed (latest → nightly)`), so the label
collision aside, the open issues do not say which job failed. This PR already fixes that for
the API job (`— API script`); the compose job would benefit from the same treatment.

One practical note: the branch is 177 commits behind `main` and GitHub reports
`mergeable: UNKNOWN`, so it likely needs a rebase before it can be merged. The four PR checks
that ran are green (the impacted-E2E job correctly skipped — workflow-only change).

Issue #867, opened 2026-07-21
[Daily #864] expandFocusedNode more-options-modal menu times out under single-backend load — 3 Chat Input specs (transient-saturation)


rafaelgiln:
Spun out of daily-failure triage #864 (run [29822019255](https://github.com/oriontech-me/langflow-e2e/actions/runs/29822019255), 2026-07-21).

## Symptom

Chat Input component canvas interactions time out on a `locator.click` — the element is present but the click never resolves within 5s:

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests-automations/regression/core-components/chat-input-output-component-regression.spec.ts:101` ("Chat Input component — renders on canvas with Message output handle and Input Text field") | canvas element click | `Error: locator.click: Timeout 5000ms exceeded.` |
| `tests-automations/regression/core-components/chat-input-files-field-regression.spec.ts:121` ("Chat Input — toggling `showfiles` exposes the Files inspector field") *(non-actionable this run — same click-5000ms signature, noted for context)* | inspector toggle click | `Error: locator.click: Timeout 5000ms exceeded.` |

## Flake signal

Confirmed recurrent, same signature: `chat-input-output:101` failed as a flake on dailies **2026-07-20** and **2026-07-21** (`same_signature: true`). As prevention, `@stable` is being removed at triage so it stops running in the daily until this issue is worked:

- `tests-automations/regression/core-components/chat-input-output-component-regression.spec.ts` (test at line 102)

Restoring `@stable` after the fix is a deliverable of this issue. The sibling `chat-input-files-field:121` shares the exact same click-5000ms signature but was a single-occurrence flake this run — included here as context because a shared Chat Input canvas root cause is plausible.

## Preliminary read (descriptive — NOT a verdict)

Two Chat Input specs hit an identical `locator.click: Timeout 5000ms` on the canvas within the same run. This is consistent with a recent Chat Input component UI change (a moved/renamed element or a click target now covered by an overlay), but that must be confirmed, not assumed. Both recovered on retry (flaky, not final).

## Investigation directive

Investigate all paths independently, **product (Langflow) as prime suspect first**: on the current nightly, drop a Chat Input on the canvas and confirm the target element is clickable (not covered/renamed). Only after ruling out a product/DOM change, consider: (a) testid/selector drift → update the locator; (b) timing — the element renders but is not yet interactable → harden the wait before the click.

## Deliverables (Done when)

[...]

Issue #833, opened 2026-07-18
Shard the @stable daily suite across N standard runners


rafaelgiln:
Follow-up to #817 (diagnosis). #817 concluded that the durable, cost-free fix for the daily-stable CPU contention is **sharding the `@stable` suite across N standard `ubuntu-latest` runners** — each shard with its own dedicated `langflow`/`ollama`/`go-httpbin` stack. Standard runners are free/unlimited on this public repo; larger runners are paid; sharding removes the real root cause (one shared `langflow` serving contending workers).

## Scope
- [ ] Add a matrix (`--shard=i/N`) to `daily-stable.yml`; each shard job spins up its own service stack.
- [ ] Set **`workers=1` per shard** (parallelism comes from shard count, not workers) — no cross-worker backend contention.
- [ ] **Merge the N blob reports** into a single `results.json` before the downstream steps.
- [ ] Make the downstream steps run **post-merge**: auto-remove-`@stable`, `append-weekly-history.mjs` (daily history), QA-Platform POST, and the failure-issue creation all currently assume one `results.json`.
- [ ] **`@database` / state-sharing tests must stay within one shard** — shard by file with affinity; never split a dependent pair across shards.
- [ ] Pick N from the measured wall-clock (target well under the 90-min job timeout with headroom for growth).

## Non-goals
- Not the paid 4-core stopgap (Option A in #817) — that's a separate reversible one-liner if relief is needed before this lands.

## Acceptance
- Daily wall-clock drops ~N×; timeout-class failure rate drops materially; history/auto-remove/QA-POST/failure-issue all still work off the merged report.

Ref: `ISSUE-817-CI-RUNNER-SIZING.md`, PR #832.

rafaelgiln:
Root-cause note from #817 direct measurement (run 29658914382): the runner is already **4-core/16 GB** and **not CPU/mem-starved** — timeout failures do **not** correlate with VM load. So sharding's value here is **one dedicated `langflow` backend per shard** (removes the single-backend request serialization), **not** more CPU. Also note #817's cheaper Option 0 (raise the langflow service's own backend concurrency) should be tried first; this sharding work is the durable fix if that's insufficient.

rafaelgiln:
[...]

PR #910, opened 2026-07-23
test(chat-input): drop @stable from files-field showfiles toggle (#867)

rafaelgiln:
## What

Removes `@stable` from `chat-input-files-field-regression.spec.ts:121` ("Chat Input — toggling `showfiles` exposes the Files inspector field") as **triage prevention** so it stops running in the daily until #867 is worked.

## Why

Recurrent flake, same signature `locator.click: Timeout 5000ms exceeded` on **2026-07-21, 07-22, 07-23**. 07-21 was the **only clean (non-guard-tripped) daily** in the window (failed=3), so this reproduces off the mass-failure days — a durable signal, not saturation collateral. It was previously carried as context on #867; the 2026-07-23 triage graduated it to actionable.

## Scope / caveat

- This spec is `test.describe.configure({ mode: "serial" })` and keeps `@stable` on two later tests (`:229`, `:285`). Confirmed the removed test is **not** a setup dependency for those — they don't consume its state — but the daily's remaining `@stable` serial subset should be watched on the next run.
- Only the tag array changes; no test logic touched.

## Follow-up

Restoring `@stable` after the fix is a **deliverable of #867** (re-validate per `CONTRIBUTING.md` before restoring).

Refs #867

### DIFF
--- a/tests/tests-automations/regression/core-components/chat-input-files-field-regression.spec.ts
+++ b/tests/tests-automations/regression/core-components/chat-input-files-field-regression.spec.ts
@@ -120,7 +120,7 @@ function chatInputNodeScope(page: Page) {
 
 test(
   "Chat Input — toggling `showfiles` exposes the Files inspector field",
-  { tag: ["@stable", "@regression", "@components"] },
+  { tag: ["@regression", "@components"] },
   async ({ page }) => {
     // Scope to the Chat Input node body. The `input-file-component` testid is also
     // mounted by the advanced-options side panel even before the toggle, so an


rafaelgiln:
Closing — the `@stable` decision for `chat-input-files-field-regression.spec.ts:121` is folded into the #867 fix instead of a separate remove-now/restore-later cycle. #867's root cause is the shared `expandFocusedNode` helper (the node toolbar's `more-options-modal` intermittently unmounts and the helper never re-selects the node); its fix touches the helper and must re-validate every caller, including this Chat Input spec. `:121` is currently green locally (4/4 at --workers=1) and the daily is now sharded (#833), so the interim flake risk is low. Restoring/keeping `@stable` on `:121` is handled in the #867 fix PR.

Issue #909, opened 2026-07-23
[Daily #906] Lock flow not persisted — saved-state assertion mismatch (recurrent, incl. clean daily)


rafaelgiln:
Spun out of daily-failure triage #906 (run [29998922242](https://github.com/oriontech-me/langflow-e2e/actions/runs/29998922242), 2026-07-23, `langflowai/langflow-nightly:latest`).

## Symptom

A `@stable` flow-functionality spec flaked (recovered on retry) — the saved lock state does not match what was set:

| Spec (line) | Assertion | Signature |
|---|---|---|
| `tests-automations/regression/flow-functionality/lock-flow.spec.ts:43` ("user must be able to lock a flow and it must be saved") | saved lock state equality | `Error: expect(received).toBe(expected) // Object.is equality` |

## Flake signal

Confirmed recurrent, same signature (07-21, 07-23). **Crucially, 07-21 was the only clean (non-guard-tripped) daily in the window** (failed=3) — so this reproduces off the mass-failure days and is a **durable** signal, not saturation collateral.

As prevention, `@stable` removal is proposed at triage so it stops running in the daily until this issue is worked (see the triage's separate `@stable`-removal authorization — the removal PR will be linked here once opened):

- `tests-automations/regression/flow-functionality/lock-flow.spec.ts` (test at ~line 43)

Restoring `@stable` after the fix is a deliverable of this issue.

## Preliminary read (descriptive — NOT a verdict)

The saved lock state read back does not equal the value set. Consistent with a **persistence/save-timing** issue (the lock toggle not flushed before the assertion, or a product regression in lock persistence), but that must be confirmed. Recovered on retry (flaky, not final).

## Investigation directive

Investigate all paths independently, **product (Langflow) as prime suspect first**: on the current nightly, lock a flow, reload, and confirm the lock persists. Only after ruling out a product regression, consider: (a) save-timing — assert after the persist round-trip completes (deterministic observable), not immediately after the toggle; (b) selector drift on the lock control/state read.

## Deliverables (Done when)

- [ ] Root cause confirmed (product persistence regression vs. save-timing vs. selector drift) with evidence on the current nightly.
- [ ] Spec passes reliably (multiple clean `--retries=0` runs).
- [ ] **`@stable` restored** in the fix PR — the tag is removed at triage as prevention (recurrent incl. a clean daily); re-validate per `CONTRIBUTING.md` before restoring.

rafaelgiln:
[...]

PR #911, opened 2026-07-23
test(lock-flow): drop @stable from lock-persisted spec (#909)

rafaelgiln:
## What

Removes `@stable` from `lock-flow.spec.ts:43` ("user must be able to lock a flow and it must be saved") as **triage prevention** so it stops running in the daily until #909 is worked.

## Why

Recurrent flake, same signature `expect(received).toBe(expected) // Object.is equality` (saved lock state mismatch) on **2026-07-21 and 07-23**. 07-21 was the **only clean (non-guard-tripped) daily** in the window (failed=3), so this reproduces off the mass-failure days — a durable signal, not saturation collateral.

## Scope

Only the tag array changes; no test logic touched.

## Follow-up

Restoring `@stable` after the fix is a **deliverable of #909** (re-validate per `CONTRIBUTING.md` before restoring).

Refs #909

### DIFF
--- a/tests/tests-automations/regression/flow-functionality/lock-flow.spec.ts
+++ b/tests/tests-automations/regression/flow-functionality/lock-flow.spec.ts
@@ -42,7 +42,7 @@ test.afterEach(async ({ page }) => {
 
 test(
   "user must be able to lock a flow and it must be saved",
-  { tag: ["@stable", "@release", "@components", "@workspace"] },
+  { tag: ["@release", "@components", "@workspace"] },
   async ({ page }) => {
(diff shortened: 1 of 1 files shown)


rafaelgiln:
Closing as obsolete. #909 was resolved by #915 (merged), which fixed the false lock-persist assertion and hardened the edge-count waits while **keeping** `@stable`. On `main`, `lock-flow.spec.ts` already carries `@stable` — so this removal PR would strip the tag from a fixed, validated test. The prevention it was opened for no longer applies.

Issue #946, opened 2026-07-24
Validate: Settings and UI Configuration — Settings page, editor shortcuts (§15.10)


rafaelgiln:
Promote §15.10 remaining `[-]` items (theme-toggle, shortcut-edit, api-keys-tz already `[x]`).

**Items:**
- Access Settings page
- Message history settings (`ui-ux/settings-message-history.spec.ts`)
- Keyboard shortcuts work in editor
- All documented shortcuts work (`[~]` — scope which)

**Exit:** validated, spec doc, `@stable`, checklist flipped.

Roadmap: Wave 4.

PR #769, opened 2026-07-15
fix(loop-component): select working OpenAI model for Research Translation Loop template (#722)

Victor-w-Madeira:
**Fixes #722.**

## Problem
`Loop component — Research Translation Loop template` hard-failed on 3 dailies (2026-07-08, 07-09, 07-13). The Playground AI message stayed on the placeholder-dots (empty) state and `toContainText(/title/i)` timed out at 240s.

Root cause (reproduced live on `1.11.0.dev38`, not a product regression): the Language Model component moved to the unified `ModelInput` widget, whose curated dropdown lists only the latest model per provider and defaults to the Anthropic reasoning model `claude-sonnet-5`. That default rejects the `temperature` param the component sends → `HTTP 400 "temperature is deprecated for this model"` → build fails → empty message. The shared `setupLanguageModelOpenAI` helper was stale: none of its `PREFERRED_CHAT_MODELS` (`gpt-4o-mini`, …) exist in the curated dropdown, so the provider-agnostic fallback selected the first option — `claude-sonnet-5`.

## Fix
- `setupLanguageModelOpenAI`: refresh `PREFERRED_CHAT_MODELS` with the curated `gpt-5.6` family and prefer a `gpt-*` (OpenAI) option over the first-match fallback; return the selected model name.
- Add `waitForLanguageModelPersisted`: the UI selection writes the correct model object (with `metadata.model_class`, which the backend needs to instantiate the model) but the Playground build races its autosave. Gate on the selection persisting, then reload so the build uses the saved graph — mirrors the Agent's `setAgentModelViaApi` + reload pattern. **Rejected alternative:** synthesizing the value via API — the persisted `model.options` are empty right after setup, and a hand-built value missing `model_class` fails to instantiate (verified: still produced the empty message).
- Loop spec Test 3: gate + `page.reload()` after model setup.
- Spec doc: document the ModelInput/temperature root cause and the new flow.

## Scope note
No product/behavior change. The Loop component itself is healthy (the non-LLM exit-condition test passes unchanged). Tests 1/2/4 are untouched and share no changed code path. `@stable` kept (the guard left it in place). No `QA-CHECKLIST.md` change (the 4 Loop bullets were already `[x]`).

## Validation (nightly 1.11.0.dev38, `--retries=0`)
- typecheck ✅ · eslint ✅ (0 errors)
- Test 3: 3 clean runs (20.7s / 24.2s isolated; 41.0s in the full 4-test file) ✅
[...]

Issue #751, opened 2026-07-14
[Daily #744] agent/flow execution does not complete — div-chat-message / 'built successfully' never render (3 specs)


Victor-w-Madeira:
Spun out of daily-failure triage #744 (run [29323509096](https://github.com/oriontech-me/langflow-e2e/actions/runs/29323509096), 2026-07-14). Surfaced while tracing serial-skip collateral during the skip review.

## Symptom

Three `@stable` tests hard-failed (final status, after retries) with the same shape — **the flow/agent execution never completes**, so the expected output never renders:

| Spec (line) | Waits for | Signature |
|---|---|---|
| `core-functionality/llm-agents/agent-component-regression.spec.ts:145` ("agent interaction suite") | `getByTestId('div-chat-message')` | `expect(locator).toBeVisible() failed` |
| `core-functionality/llm-agents/agent-input-sources.spec.ts:235` ("input via ChatInput handle drives the agent response") | `getByTestId('div-chat-message')` | `expect(locator).toBeVisible() failed` |
| `core-components/loop-component-regression.spec.ts:358` ("Loop component — stops after exhausting input DataFrame and emits aggregated done") | `text=built successfully` | `waitForSelector 60000ms timeout` |

All three load a template/pre-built flow and run it; the AI/agent message (or "built successfully") never appears within the timeout.

The serial siblings `agent-input-sources:271` and (unrelated) others were auto-skipped as a consequence.

## Preliminary read (descriptive — NOT a verdict)

This surfaced on a **mass-failure day** (27 hard + 27 flaky, guard tripped) and the "execution never completes" symptom is consistent with **provider latency / instance saturation under load** — i.e. possible wave collateral. But these are genuine final hard failures (they did not recover on retry), and the day-level environmental signal must **not** be assumed as the cause. Note: `loop-component`'s "Research Translation" variant is tracked separately in #722; this is a **different** loop test.

## Investigation directive

Investigate all paths independently, **product as prime suspect first**: on the current nightly, confirm whether these flows actually execute to completion (agent responds / loop builds) before concluding provider/latency. Only after ruling out a regression, consider: (a) transient saturation → prove on 3 environments and restore; (b) wait-strategy fragility → harden the completion wait to a deterministic observable.

## Deliverables (Done when)

- [ ] Root cause confirmed per spec (regression vs. provider/latency vs. wait-strategy) with evidence on the current nightly.
[...]

Issue #722, opened 2026-07-13
[Daily #704] loop-component Research Translation Loop hard-fails (AI message never resolves) — recurring 07-08/07-09/07-13


Victor-w-Madeira:
Spun out of the daily-failure triage #704 (run [29244611612](https://github.com/oriontech-me/langflow-e2e/actions/runs/29244611612)).

## Failure

- **Spec:** `tests/tests-automations/regression/core-components/loop-component-regression.spec.ts:151`
- **Test:** `Loop component — Research Translation Loop template: full wiring and iterates over 2 ArXiv papers`
- **Date / version:** 2026-07-13 · Langflow `1.11.0.dev41` (`langflowai/langflow-nightly:latest`)
- **Result:** hard failure (all 3 attempts)

```
expect(locator).toContainText(expected) failed
Locator: locator('[data-testid^="chat-message-AI-"]').last()
Timeout: 240000ms
Expected pattern: /title/i
Received string: "······················ (placeholder dots — AI message never resolved)
```

## Recurrence (from reports/daily-history.jsonl)

Hard failure on **3 separate dailies**: **2026-07-08, 2026-07-09, 2026-07-13**. This is why it warrants a dedicated issue rather than being folded into the day's environmental cluster.

Note: 07-08 was itself a mass-failure day (28 hard fails) and 07-13's run tripped the mass-failure guard, so instance health is a plausible contributor — but the 07-09 occurrence landed on a low-failure day, which weakens a pure "bad-instance" explanation.

## Descriptive observations (not a verdict)

- Today the final AI chat message stayed on the placeholder-dots state and never rendered text matching `/title/i` within the 240s timeout — i.e. the loop/playground run did not complete.
- Whether that is the Loop component itself, the ArXiv fetch, the template wiring, or instance-level playground execution is **undetermined here**.

## Investigation (all paths independent — suspect the product first)

1. **Regression?** Reproduce the Research Translation Loop template live on the nightly and watch whether the loop iterates over the 2 papers and emits a titled message. Confirm the failure is (or is not) reproducible outside CI.
2. If the product works manually, look at timing/wait design in the spec (240s already generous) and any flakiness in the ArXiv dependency.
3. Fix per the validation guide, prove `--retries=0` clean, and **restore `@stable`** (guard left it in place; no auto-removal happened).

Victor-w-Madeira:
## Investigation update — root cause confirmed, but paused on two blockers

[...]

PR #982, opened 2026-07-27
fix(provider-setup): self-configure OpenAI when the ModelInput dropdown offers none (#961)

### LINKED ISSUE #961: Harden setupLanguageModelOpenAI: self-configure OpenAI when the ModelInput dropdown offers no gpt-* option
## Context

Follow-up extracted from PR #769, which is being closed as superseded: its parent issue #722 was resolved by the merged PR #813 (AI-message timeout raised 240s → 360s, root cause diagnosed as CI saturation) rather than by the ModelInput/`temperature` theory PR #769 was built on.

Two hardenings in that PR's `tests/helpers/provider-setup/setup-language-model-openai.ts` diff are **independent of #722** and are still not covered by `main`. This issue tracks porting only those, validated fresh against the current nightly.

Reference branch (not merged, kept for the diff): `fix/issue-722-loop-modelinput-temperature`.

## Gap 1 — provider unconfigured under the unified `ModelInput` widget

`main`'s `setupLanguageModelOpenAI` only handles the "no `model_model` dropdown → click **Setup Provider**" path. Under the 1.11 unified `ModelInput` widget the dropdown *does* render, but it lists models only from **configured** providers. So when the instance holds no OpenAI credential — e.g. the `Run impacted E2E specs` PR job, which runs without a `collect-models` step — no `gpt-*` option is offered and the helper proceeds anyway, leaving the node on whatever the dropdown listed first.

Proposed: when no `gpt-*` option is visible, configure the OpenAI credential from `OPENAI_API_KEY` through the dropdown's "Manage Model Providers" panel, then reopen the dropdown and select as usual. Skip the detour when a `gpt-*` option is already offered (daily-stable configures OpenAI via `collect-models`). PR #769 implemented this as `configureOpenAIProviderFromDropdown`.

## Gap 2 — the provider-agnostic fallback can hand back a non-OpenAI model

`selectPreferredChatModel`'s last-resort fallback is "the first option that is neither `NON_CHAT_MODEL` nor `AVOID_MODEL`". In a mixed-provider dropdown that can be the Anthropic default, so a helper named `...OpenAI` silently configures a non-OpenAI model whenever `PREFERRED_CHAT_MODELS` drifts behind the curated list.

Proposed: prefer any `gpt-*` chat option before that last-resort fallback.

## Explicitly out of scope — do not port from #769

- **The `PREFERRED_CHAT_MODELS` rewrite.** PR #769 dropped `main`'s `gpt-5.x` nano/mini/`5.4`/`5.5` entries in favour of a `gpt-5.6` family. That list was authored on 1.11.0.dev38 (2026-07-15) and must be re-verified against the current nightly before anyone touches it; reasoning-model exclusion is already handled by `isReasoningOption` (#569 / #610).
- **`waitForLanguageModelPersisted` + `page.reload()` in `loop-component-regression.spec.ts` test 3.** #813 fixed that test differently and it has been green in the dailies since 2026-07-16.

## Validation

Shared helper — the change reaches every spec that imports it:

- `core-components/loop-component-regression.spec.ts` (test 3)
- `core-functionality/llm-agents/memory-history-regression.spec.ts`

Both must run clean at `--retries=0` in a burst, plus a force-fail on the new branch of `setupLanguageModelOpenAI` (unset the OpenAI credential so Gap 1's path actually executes) to prove the added code is load-bearing rather than dead.

Victor-w-Madeira:
**Closes #961.**

## Problem

`setupLanguageModelOpenAI` is shared by `loop-component-regression.spec.ts` (test 3) and `memory-history-regression.spec.ts` (all 3 tests). On 1.11+ the Language Model node uses the unified `ModelInput` widget, whose dropdown lists models **only from enabled providers** — and the node renders that dropdown (instead of the legacy `Setup Provider` button) as soon as **any** provider is enabled.

So on an instance where another provider is enabled but OpenAI is not, the helper found no `gpt-*` option and `selectPreferredChatModel`'s provider-agnostic last-resort fallback ("first option that is neither non-chat nor avoided") silently picked the first entry in a mixed dropdown. Reproduced live on `langflow-nightly:latest` (1.12.0.dev7) with Anthropic + Google enabled and OpenAI not: 49 options, none from OpenAI, and the helper **completed successfully** leaving the node on `claude-opus-5`. A helper named `...OpenAI` configured Anthropic, and the downstream failure ("Message empty.", build timeouts) pointed nowhere near the cause.

Two additional findings while scoping this, both corrections to the issue's premise:

- With **zero** providers configured the node still renders the legacy `Setup Provider` button (`model_model visible=false | Setup Provider buttons=1`), which the pre-existing path already handles. The gap is specifically the *other-provider-enabled* state.
- The `Run impacted E2E specs` job does run `collect-models` whenever a changed spec references `provider-setup` (`pr-validation.yml`, from #946), so the real trigger is OpenAI ending up inactive (key/quota outage, or a run that configured only Anthropic/Google) — not a missing collect-models step.

## Fix

Ported the two hardenings from the superseded PR #769 that are independent of #722:

- **Gap 1** — when the open dropdown offers no OpenAI model, configure the credential from `OPENAI_API_KEY` through the dropdown's "Manage Model Providers" panel (`manage-model-providers` → `provider-item-OpenAI` → `sk-...` → Save → enable one model toggle → Close), then reopen the dropdown and select as usual. Skipped when OpenAI is already offered, so the configured path (daily-stable, which imports credentials via `collect-models`) is untouched. Re-saving an already-stored credential is skipped as well — it 400s on `PATCH /variables` and would trip the backend-error monitor (#751).
- **Gap 2** — prefer any `gpt-*` chat option before the provider-agnostic last-resort fallback, so a drifted `PREFERRED_CHAT_MODELS` degrades to another OpenAI model instead of to a different provider.

Three deviations from #769, each driven by the live 1.12 surface rather than by the branch diff:

| #769 | This PR | Why |
|---|---|---|
| detected OpenAI by model name only (`gpt-*`) | detects by **provider in the option testid** (`/^openai-/i`), name as fallback | the frontend emits `data-testid="${provider}-${model}-option"` (`ModelList.getModelOptionTestId`); confirmed live: `Anthropic-claude-opus-5-option`, `Google Generative AI-gemini-2.5-flash-option` |
| `getByRole("button", { name: /^(Save\|Retry)$/i })` | `/^(Save\|Retry Save)$/` | the real label is **"Retry Save"** (`modelProviders.retrySaveButton`) — the original regex could never match it |
| `.catch(() => {})` on the model-toggle wait | hard wait + explicit error | swallowing it re-created the exact silent fallback to another provider this branch exists to prevent |

Signature stays `Promise<void>`: #769 changed it to return the selected model name for `waitForLanguageModelPersisted`, which is explicitly out of scope here.

## Scope note

No `.spec.ts` was touched, no test added or retagged, and no coverage change — so no spec doc and no `QA-CHECKLIST.md` edit. Both out-of-scope items from #769 were deliberately **not** ported: the `PREFERRED_CHAT_MODELS` rewrite (needs re-verification against the current nightly; reasoning-model exclusion is already handled by `isReasoningOption`, from #569 / #610) and `waitForLanguageModelPersisted` + `page.reload()` in the loop spec (#813 fixed that differently and it has been green in the dailies since 2026-07-16).

## Validation (nightly 1.12.0.dev7, `--retries=0`, `--workers=1`)

- `npm run typecheck` ✅ · `npm run lint` ✅ (0 errors)
- Burst 2× of both importing specs → **7 passed (1.1m)** each run ✅
- End-to-end run of `loop-component-regression` test 3 **in the Gap 1 state** (OpenAI credential deleted from Langflow; the preflight confirms it is absent) → **1 passed (23.7s)**, and the helper recreated the `OPENAI_API_KEY` global variable itself ✅
- Zero `🚨 Backend Error` ✅
- Flow cleanup: every flow the specs created was removed by their id-scoped `afterEach`; the only flows left in the default project are the `awaitBootstrapTest` bootstrap artifacts ✅
- `--trace=on` not run — known local hang on heavy-template specs (#490 / #518). Covered instead by the `--retries=0` bursts and the force-fail battery below.

### Force-fail (executed, one mutation per mechanism)

| Mutation | Observed |
|---|---|
| `isOpenAIOffered` forced to `true` (Gap 1 detection off) | helper completes silently on `claude-opus-5` — the reported bug |
| `manage-model-providers` testid corrupted | `locator.click: Timeout 20000ms exceeded` |
| control: unmutated helper, same state | `gpt-4o-mini` |
| whole `PREFERRED_CHAT_MODELS` invalidated, Gap 2 clause present | `gpt-5.6-sol` (still OpenAI) |
| same, Gap 2 clause removed | `claude-opus-5` |
| invalid `OPENAI_API_KEY` in the Gap 1 state | explicit `saving OPENAI_API_KEY in the Model Providers panel did not load any OpenAI model. Check that OPENAI_API_KEY is valid…` |

Revert verified: 0 mutation markers left, final diff `1 file changed, 128 insertions(+)`.

### DIFF
--- a/tests/helpers/provider-setup/setup-language-model-openai.ts
+++ b/tests/helpers/provider-setup/setup-language-model-openai.ts
@@ -1,4 +1,4 @@
-import { type Page, expect } from "@playwright/test";
+import { type Locator, type Page, expect } from "@playwright/test";
 
 // Cheap, fast chat models in priority order. `gpt-4o-mini` is kept first so older
 // Langflow builds still match; the `gpt-5.x` entries cover newer builds (1.11.0+)
@@ -55,9 +55,17 @@ async function selectPreferredChatModel(page: Page): Promise<void> {
     // that fallback used to hand back a non-OpenAI model from a helper named
     // `...OpenAI` whenever PREFERRED_CHAT_MODELS drifted behind the build's
     // curated list (issue #961).
+    // A multi-line label carries a badge rendered inside the option (e.g.
+    // "Deprecated"): skip it. Unlike the exact-match PREFERRED lookup above,
+    // this pattern-based branch would otherwise put the badge text into
+    // `chosen` and break the trigger assertion below — and a deprecated model
+    // is not what a regression run wants either.
     labels.find(
       (label) =>
-        /^gpt-/i.test(label) && !NON_CHAT_MODEL.test(label) && !AVOID_MODEL.test(label),
+        !label.includes("\n") &&
+        /^gpt-/i.test(label) &&
+        !NON_CHAT_MODEL.test(label) &&
+        !AVOID_MODEL.test(label),
     ) ||
     labels.find((label) => !NON_CHAT_MODEL.test(label) && !AVOID_MODEL.test(label));
 
@@ -189,9 +197,23 @@ export async function setupLanguageModelOpenAI(page: Page): Promise<void> {
   // persisted the credential).
   if (!(await isOpenAIOffered(page))) {
     await configureOpenAIProviderFromDropdown(page);
-    await modelDropdown.dispatchEvent("click");
 
-    if (!(await isOpenAIOffered(page))) {
+    // Gate on an OpenAI OPTION appearing — never on a single sample of the list.
+    // Closing the panel refreshes every ModelInput *asynchronously*
+    // (`refreshAllModelInputs`, fired after the pending model toggles flush), so
+    // the reopened dropdown can still be rendering the pre-refresh list; sampling
+    // it once reports "no OpenAI model" for a provider that was just configured
+    // fine. The second dispatch covers the opposite state: if the panel left the
+    // popover open, the first dispatch toggled it shut instead of opening it.
+    const openAIOption = page.getByTestId(/^openai-.+-option$/i).first();
+    await modelDropdown.dispatchEvent("click");
+    let offered = await optionAppeared(openAIOption);
+    if (!offered) {
+      await modelDropdown.dispatchEvent("click");
+      offered = await optionAppeared(openAIOption);
+    }
+
+    if (!offered) {
       await page.keyboard.press("Escape");
       throw new Error(
         "setupLanguageModelOpenAI: the model dropdown still offers no OpenAI model after " +
@@ -204,17 +226,28 @@ export async function setupLanguageModelOpenAI(page: Page): Promise<void> {
   await selectPreferredChatModel(page);
 }
 
+// Resolves true when the option becomes visible within the wait, false on timeout.
+async function optionAppeared(option: Locator): Promise<boolean> {
+  return option
+    .waitFor({ state: "visible", timeout: 15000 })
+    .then(() => true)
+    .catch(() => false);
+}
+
 // True when the open ModelInput dropdown offers at least one OpenAI model.
 // Each option carries `data-testid="${provider}-${model}-option"` (frontend
 // `ModelList.getModelOptionTestId`), so the provider is authoritative straight
 // from the DOM; the `gpt-*` label check is a fallback for builds whose option
 // testid is not provider-prefixed. Returns false on the empty dropdown ("No
 // Models Enabled"), which is what an unconfigured instance renders.
+// The initial wait matches selectPreferredChatModel's own 15s budget for the same
+// list: a saturated runner that takes >5s to populate it must be waited out, not
+// mistaken for "OpenAI is missing" (which would trigger a pointless panel detour).
 async function isOpenAIOffered(page: Page): Promise<boolean> {
   const options = page.locator('[data-testid$="-option"]');
   await options
     .first()
-    .waitFor({ state: "visible", timeout: 5000 })
+    .waitFor({ state: "visible", timeout: 15000 })
     .catch(() => {});
 
   const testIds = await options.evaluateAll((els) =>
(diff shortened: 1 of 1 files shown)


Victor-w-Madeira:
## Review follow-up — 3 fixes applied (`480d04a`)

An independent pass over the diff found three defects; all are fixed and re-validated on nightly 1.12.0.dev7.

**1. The post-configuration check sampled the option list once (the serious one).** `ModelProviderModal.handleClose` refreshes every ModelInput **asynchronously** — `onClose()` → `await flushPromise` → `refreshAllModelInputs({ silent: true })`. The reopened dropdown could therefore still be rendering the *pre-refresh* list, and the helper would throw "still offers no OpenAI model" for a provider it had just configured correctly. On a saturated runner that is a misleading hard failure in two `@stable` specs.

Now it waits for an actual OpenAI **option** and re-dispatches the trigger once, which also covers the opposite state (panel left the popover open → the first dispatch toggled it shut). The testid shape was confirmed live: `OpenAI-gpt-5.6-sol-option`, `OpenAI-gpt-4o-mini-option`, … — 41/41 options match `/^openai-.+-option$/i`.

**2. `isVisible({ timeout: 2000 })` never waited.** `playwright-core/types/types.d.ts:13940` documents the option as ignored, so the "already configured" branch was an instantaneous sample. A premature reading re-saves a stored credential → `PATCH /variables` 400 → backend-error monitor (the #751 class the code comment claims to avoid). Replaced with a real wait on the form's submit button plus a branch on the label it renders, scoped to `getByRole("dialog")` so no same-named canvas button can match.

**3. The `gpt-*` preference could pick a label containing a badge.** `allInnerTexts()` includes the `Deprecated` badge rendered inside the option (`ModelList.tsx:60-70`). Unlike the exact-match PREFERRED lookup, this pattern-based branch would have put the badge text into the selected name and broken the trigger assertion — and selected a deprecated model. Multi-line labels are now skipped. No model in the current catalog carries the flag (`GET /api/v1/models` → 0 deprecated), so this is a guard against a live code path with no live trigger today.

Two smaller items from the same pass: the failed-save error now reports the submit button's **actual** label instead of asserting "Retry Save" — the force-fail proved that wrong, the button stays on `"Save"` after a rejected key (an assumption inherited from #769) — and the option-list wait was aligned to the 15s `selectPreferredChatModel` already uses for the same list, so a slow runner is waited out instead of triggering a pointless panel detour.

### Re-validation (nightly 1.12.0.dev7, `--retries=0`, `--workers=1`)

- typecheck ✅ · eslint 0 errors ✅
- Burst 2× of both importing specs → **7 passed** (2.8m, 1.4m) ✅
- Gap 1 path with the OpenAI credential deleted from Langflow → helper self-configures and selects `gpt-4o-mini` ✅
- Zero `🚨 Backend Error` ✅

Force-fail, one mutation per mechanism (7 executed, all reverted — 0 markers left):

| Mutation | Observed |
|---|---|
| new OpenAI-option gate regex corrupted | explicit "still offers no OpenAI model" error |
| submit-button name regex corrupted | `locator.waitFor: Timeout 10000ms exceeded` |
| control: fixed helper, same state | `gpt-4o-mini` |
| `PREFERRED_CHAT_MODELS` invalidated, `gpt-*` clause present | `gpt-5.6-sol` (still OpenAI) |
| same + badge guard inverted | `claude-opus-5` |
| same + `gpt-*` clause removed | `claude-opus-5` |
| `isOpenAIOffered` forced to `true` | `claude-opus-5` |
| invalid key in the Gap 1 state | `…did not load any OpenAI model (submit button reads "Save")` |

### Known risk, stated explicitly

The Gap 1 branch **never executes in CI**: `daily-stable` and the impacted-specs job both run `collect-models`, so OpenAI is always configured and the first check short-circuits. The new code only runs in a degraded state, which means no continuous coverage — a future regression in it would stay invisible until the day it is actually needed. Validated here only by the local runs above.

Issue #966, opened 2026-07-27
[Daily #962] run-flow — welcome panel predicate times out in openNewFlowTemplatesModal (recurrent)


Victor-w-Madeira:
Spun out of daily-failure triage #962 (run [30261409427](https://github.com/oriontech-me/langflow-e2e/actions/runs/30261409427), 2026-07-27, `langflowai/langflow-nightly:latest`).

## Symptom

Recovered on retry (flaky, not a final hard failure), but **recurrent on the same signature**. After the "New Flow" entry point, **neither** the `FlowBuilderWelcome` overlay **nor** the templates modal surfaced within 30s, so the shared reconciliation poll never turned true.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/flow-functionality/run-flow.spec.ts:12` ("user should be able to use Run Flow without any issues") | `expect.poll(() => overlayOrModalAppeared(page, 0), { timeout: 30000 }).toBe(true)` — `tests/helpers/flows/open-new-flow-templates-modal.ts:51` | `Error: expect(received).toBe(expected) // Object.is equality` — Expected `true`, Received `false`; `Timeout 30000ms exceeded while waiting on the predicate` |

The failure is inside the shared helper `dismissWelcomeOverlayAndWaitForModal`, before the modal-title wait at line 61 — i.e. the entry point produced neither expected end state.

## Preliminary read (descriptive — NOT a verdict)

- The guard did **not** trip (2 hard failures vs. a threshold of 6+) — this is a clean, non-guarded daily, so the signal is not mass-failure collateral.
- The helper already reconciles two known post-1.10 entry-point behaviors (overlay vs. modal). A poll that exhausts 30s means **a third state** was reached — the page navigated somewhere else, the overlay rendered attached-but-hidden past the budget, or the click never registered. Which one it is has to be read off a trace, not inferred.
- `tests/helpers/flows/open-new-flow-templates-modal.ts` is a **shared helper**: any change here touches every spec that opens the templates modal, so the blast radius must be assessed before editing it.
- The same test also flaked on **2026-07-13** with a *different* signature (`TimeoutError: locator.click: Timeout 20000ms exceeded.`), which is not counted toward this recurrence but is worth reading as prior context on the same entry point.

## Flake signal

[...]

Issue #948, opened 2026-07-24
Validate: MCP Server remainder — server tab, add via modal, starter project, protocol endpoint, resource URI (§14.1)


rafaelgiln:
Promote §14.1 MCP Server `[-]` items to `@stable` (not already promoted in Wave 3 #829). Prompt-template item is not-implementable on 1.11.x (`prompts/list` → `[]`).

**Items:**
- MCP Server tab in flow
- Add MCP server via modal
- Starter project with MCP
- Flow exposed as MCP server — verify generated endpoint → `mcp/server/mcp-server-protocol.spec.ts`
- Execute MCP server tool via MCP protocol
- Resource exposed by server accessible via URI → `mcp/server/mcp-server-resources.spec.ts`

**Exit:** validated, spec doc, `@stable`, checklist flipped.

Roadmap: Wave 4.

Issue #907, opened 2026-07-23
[Daily #906] Groq/Mistral provider config+run — waitForSelector 30s timeout (2 specs, recurrent)


rafaelgiln:
Spun out of daily-failure triage #906 (run [29998922242](https://github.com/oriontech-me/langflow-e2e/actions/runs/29998922242), 2026-07-23, `langflowai/langflow-nightly:latest`).

## Symptom

Two `@stable` model-provider specs hard-failed (final status, after retries) with the same shape — a `page.waitForSelector` never resolves while configuring the provider's API key and executing the flow:

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests-automations/regression/core-functionality/model-provider/groq-provider.spec.ts:92` ("the Groq component configures the API key and executes the flow") | `page.waitForSelector` | `TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.` |
| `tests-automations/regression/core-functionality/model-provider/mistral-provider.spec.ts:92` ("the MistralAI component configures the API key and executes the flow") | `page.waitForSelector` | `TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.` |

**Same-shape sibling this run (today-only, not a separate cluster):** `model-provider/ollama-provider.spec.ts:172` ("the Ollama component lists the local model live and executes the flow") hit the identical `page.waitForSelector: Timeout 30000ms` — but only on this run for this signature (no cross-day recurrence of the same shape), so it is folded here as context rather than filed independently.

## Preliminary read (descriptive — NOT a verdict)

Both Groq and Mistral are cross-day-recurrent on the same signature (07-22 and 07-23). Neither is Google-parameterized, so this is **distinct from #898** (missing `langchain-google-genai`). Both recurrence dates are themselves **guard-tripped mass-failure days** (07-22: 22 hard; 07-23: 8 hard, guard threshold 5) — the wider window shows 8 of the last 9 dailies tripping the guard — so the `waitForSelector 30s` shape is consistent with **provider latency / single-backend saturation under load**, but that must be ruled out, not assumed. No occurrence on the one clean daily in the window (07-21).

## Investigation directive

[...]

Issue #898, opened 2026-07-22
Nightly missing langchain-google-genai → all Google-parameterized @stable specs fail (agents, model-provider, Google-embedding KB)


rafaelgiln:
Dedicated follow-up dispatched from the #879 daily triage (run [29911701167](https://github.com/oriontech-me/langflow-e2e/actions/runs/29911701167), 2026-07-22, `langflowai/langflow-nightly:latest`).

## Symptom
On the daily (`--grep "@stable"`, one-per-provider = openai + anthropic + **google**), every **Google-parameterized** run fails. Playwright surfaces it late — e.g. `[data-testid^="chat-message-AI-"]` / `built successfully` never appears — so it reads as a generic "agent didn't respond" / "not visible" timeout.

## Root cause (confirmed live on 1.12.0.dev0)
The nightly image ships **without `langchain-google-genai`**. Langflow's Google model class does a hard top-level import (`lfx/base/models/google_generative_ai_model.py`: `from langchain_google_genai import ChatGoogleGenerativeAI`), so building any Google model — chat **or embedding** — raises:

```
ImportError: Could not import 'lfx.base.models.google_generative_ai_model' for model class
'ChatGoogleGenerativeAIFixed'. Install the missing package (e.g. uv pip install langchain-google-genai).
```

Verified in the running container:
```
$ docker exec langflow-e2e-runner python -c "import langchain_google_genai"
ModuleNotFoundError: No module named 'langchain_google_genai'
```
The agent/model never builds → no final message → the spec times out. `collect-models` marks google **active** because it validates the raw provider key with a direct API call (not through Langflow), so nothing upstream of the run flags the gap.

## Affected @stable specs (Google variant / Google embeddings)
Cross-day-recurrent (durable, not guard collateral):
- `llm-agents/agent-current-date-tool.spec.ts` (4×: 07-13/14/15/22)
- `llm-agents/agent-multi-tool-selection.spec.ts`, `llm-agents/agent-context-id-isolation.spec.ts` (3×)
- `llm-agents/agent-component-regression.spec.ts`, `llm-agents/agent-input-sources.spec.ts`, `llm-agents/agent-multimodal-image-input.spec.ts` (2×)
- `model-provider/google-provider.spec.ts` (2×: 07-14/22)
[...]

PR #984, opened 2026-07-27
fix(regressions): harden the ledger generator, promote LE-2019, guard the indicator in CI

### LINKED ISSUE #983: Regression Ledger: register the product regressions the suite catches (REGRESSIONS.md)
## Why

The suite's core value is catching **real Langflow regressions** — product breakage a green run would never prove. That value is currently invisible: the evidence is scattered across three places that each lose part of it.

- Ad-hoc markdown at the repo root (`LANGFLOW-BUG-*.md`, `ISSUE-*.md`) — no shared schema, no index, no count.
- `reports/daily-history.jsonl` — records run outcomes only; it cannot tell a real product regression apart from a flake or a saturation failure.
- The deterministic pipeline — classifies a `langflow-regression` verdict per issue, but that signal stays trapped in one issue's runtime state and is never aggregated.

Nothing answers the question that matters most: *how many real Langflow regressions has this suite caught, and what is their status?*

## What ships

`REGRESSIONS.md` at the repo root — a curated, append-only ledger that is both the ROI indicator and the canonical registry of confirmed product regressions, each linked to its upstream ticket and evidence report.

- **Qualifying bar — all three:** a confirmed `langflow-regression` verdict, adversarial (refute-first) validation, and a filed upstream ticket. Confirmed but unticketed goes to **Candidates**; a finding validation downgrades to a non-user-facing gap is listed nowhere.
- **Scope:** regressions the *suite* caught — every row traces to a spec failure or spec-validation run recorded in a repo issue. Regressions QA confirms by hand stay on the Jira board.
- **Generated indicator** between `<!-- REGRESSIONS:START/END -->` markers via `npm run regressions:summary` (`scripts/regressions-summary.ts`), mirroring the `coverage-summary.ts` pattern: idempotent, fails loudly on a malformed row, never hand-edited.
- **Curation is mandatory**, documented in `CONTRIBUTING.md`: confirming a regression and filing the ticket includes adding the row.

## Seeded state

4 regressions caught — 1 open, 3 fixed. `High 2 · Medium 2`, `model-provider 2 · auth 1 · mcp 1`.

| Area | Regression | Upstream | Status |
|---|---|---|---|
| mcp | `resources/read` crashes with `AttributeError: 'str' object has no attribute 'hex'` | LE-2012 | Open |
| model-provider | Groq / Mistral / Ollama components silently hidden from the sidebar | LE-1987 | Fixed |
| model-provider | Nightly ships without `langchain-google-genai` — ~17 `@stable` specs down | LE-1974 | Fixed |
| auth | Logout does not terminate the session | LE-1850 | Fixed |

## Off-wave

Wave 4 is Canvas UI/UX & MCP; this is process/infra tooling and does not belong to it. Opening this as the exception issue the PR links, per `ROADMAP.md` → Intake and the PR review checklist.

rafaelgiln:
Closes #983. Review round on Victor's feedback — **re-scoped**.

## Why the diff changed shape

The ledger itself (`REGRESSIONS.md`, the generator, the `CONTRIBUTING.md` rule, the decision record) reached `main` through #987 while this PR was open — byte-identical to what this branch carried. Keeping the original commits would have replayed content already on `main`, so the branch was reset onto `origin/main` and now carries only the delta: the review fixes plus one new regression row.

## Blocking items

**1. Dangling evidence references — fixed, and the root cause was staleness.** This branch was cut from `0c1deaa`; `main` had moved four commits ahead. The New Flow evidence *is* committed — as `docs/upstream-bugs/UPSTREAM-BUG-new-flow-dead-click.md`, added by #987 after the branch point. The ledger pointed at the pre-commit draft name. Repointed.

The two bulk-delete filenames are gone: neither exists in the tree, on disk, or anywhere in history (they were "do not commit" peer-review drafts). The counter-example keeps its full reasoning and now says plainly that the write-ups were never committed, so the record is the entry plus the Jira board. Also documented that `Report` holds an evidence file when one exists and the carrying issue otherwise — both shapes allowed, an unresolvable reference is not.

**2. Stale-on-arrival decision record — rewritten.** Status is now "Active", the branch housekeeping is gone, and the "push and open the PR" close is replaced by the one thing that genuinely still ages: `Fixed in` carries a PR on three rows until a spec re-runs green on a build with the fix.

## New: LE-2019 promoted from Candidate to a row

Filed 2026-07-27, reporter Rafael, and the ticket itself records *"Found by the E2E regression suite: #966"*. It clears all three requirements — the investigation established **1.10.0 as last-known-good** and **1.10.1 (langflow#12575) as first-affected**, refuting the initial "1.12 regression" reading by comparing the nightly and 1.11.x frontend trees byte-for-byte, and isolating the trigger to a click landing before the flows list paints its cards.

Indicator: **5 caught — 2 open · 3 fixed** · `High 2 · Medium 3` · `model-provider 2 · auth 1 · flows 1 · mcp 1`. Candidates is now empty, which is the intended lifecycle rather than an exception.

## Item 4 — four ways the generator could publish a wrong number

Each reproduced against a broken copy before fixing:

| Input | Old behaviour |
|---|---|
| Markers reversed (`END` before `START`) | passed the count check, **exit 0**, block duplicated and the prose between the markers deleted |
| `Area / Test` without the ` · ` separator | `**By area:** … auth/logout-flow.spec.ts 1` |
| Pasted duplicate row | `5 regression(s)` from 4 rows, silently |
| Emptied `## Ledger` table | silent `0 regression(s)`, while the design doc claimed a damaged section aborts |

All four now abort with a specific message. The empty-table case throws rather than the doc being softened — the file is append-only, so an empty table means damage.

## Item 3 — CI drift guard

`--check` recomputes the block and exits 1 with the expected content when the committed one disagrees, exposed as `npm run regressions:check` and wired as a step of the existing **typecheck** job (no extra `npm ci`).

The tension you named is now written down in both `CLAUDE.md` and the decision record rather than left implicit: committing a generated block in-PR is what `check-checklist-guard.mjs` forbids for `QA-CHECKLIST.md`, and the ledger is a deliberate exception because it changes a handful of times per quarter and never on another PR's lines — with the auto-regen job named as the resolution if that rate ever rises.

## Item 5 + minors

- `CLAUDE.md` gains a `REGRESSIONS.md` section: the bar, the suite-scope rule, never hand-edit the block, every abort condition, and the in-PR-commit rationale.
- `CONTRIBUTING.md`: "has it **been** adversarially validated"; suite-scope rule added; the CI check documented.
- `README.md` links the ledger next to the `CONTRIBUTING.md` pointer.
- `escapeRe()` helper replaces the duplicated regex-escape expression; `(SEVERITIES as readonly string[]).includes(...)` replaces the unsound cast.

**Not taking one minor:** `docs/product-regression-core/` stays. That path was Rafael's explicit choice when the record was filed, not an oversight — moving it is a call for him, not this review round.

## Verification

- `npm run regressions:summary` → `5 regression(s)`, matches the table by hand-count; second run → *already up to date*, no diff.
- `npm run regressions:check` → in sync, exit 0. Against a hand-edited count → exit 1 and prints the expected block.
- Reversed markers / emptied table / 8-column row / `Critical` severity / separator-less area / duplicated row → all abort with their specific error.
- `npm run typecheck` and `npm run lint` → 0 errors.

No spec files touched.

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -24,6 +24,12 @@ jobs:
       - name: TypeScript check
         run: npm run typecheck
 
+      # The REGRESSIONS.md indicator block is generated and committed by hand.
+      # Fail the PR when a row was added without regenerating it, so the
+      # headline number can never silently disagree with the table.
+      - name: Regression Ledger indicator in sync
+        run: npm run regressions:check
+
   lint:
     name: ESLint
     runs-on: ubuntu-latest
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -188,6 +188,14 @@ The JSONL files are **machine-written and human-read only** — never hand-edit.
 
 Triage rules driven by this history (when to remove `@stable`, when to open an issue for a recurring flake) live in `CONTRIBUTING.md` under "Tag @stable — validated tests".
 
+## REGRESSIONS.md — the Regression Ledger
+
+Curated, append-only registry of the real Langflow regressions **the suite caught**, and the source of the ROI indicator at the top of the file. A row is added only when all three hold: a confirmed `langflow-regression` verdict, adversarial (refute-first) validation, and a filed upstream ticket (Jira `LE-####` or a `langflow-ai/langflow` issue). Confirmed but unticketed goes under **Candidates**; a finding validation downgrades to a non-user-facing gap is listed nowhere. Regressions the team confirms by hand, outside the suite, stay on the Jira board — every row must trace to a spec failure or spec-validation run recorded in a repo issue.
+
+The block between `<!-- REGRESSIONS:START -->` and `<!-- REGRESSIONS:END -->` is **auto-generated** — never hand-edit it, and never hand-edit the counts. Edit the `## Ledger` table only, then run `npm run regressions:summary` to regenerate the block and commit both. Logic lives in `scripts/regressions-summary.ts`; it is idempotent and aborts rather than emitting a wrong count (bad marker order, missing/empty `## Ledger` section, wrong column count, invalid `Severity`/`Status`, an `Area / Test` cell without the `area · spec-file` separator, or two rows with the same `Upstream` ticket).
+
+Unlike `QA-CHECKLIST.md`, the generated block here **is** committed in the PR that adds the row — the file changes a handful of times per quarter, so the collision risk that forced the checklist guard (issue #741) does not apply. The `pr-validation.yml` typecheck job runs `npm run regressions:check`, which fails the PR when the committed block disagrees with the table.
+
 ## QA-CHECKLIST.md
 
 Two passages in `QA-CHECKLIST.md` are **auto-generated**:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -652,11 +652,16 @@ regressions the suite has caught, and the source of the ROI indicator at its
 top. It is the suite's headline evidence that it finds product breakage a green
 run never proves — so keeping it current is not optional.
 
-When a resolution confirms a **`langflow-regression`** verdict, has it
-**adversarially validated** (refute-first across source/API/UI where
-applicable), and a **filed upstream ticket** exists (DataStax Jira `LE-####` or
-a `langflow-ai/langflow` issue), **adding a row to `REGRESSIONS.md` is a
-mandatory step of the report/resolution** — not a follow-up.
+When a resolution confirms a **`langflow-regression`** verdict, has it **been
+adversarially validated** (refute-first across source/API/UI where applicable),
+and a **filed upstream ticket** exists (DataStax Jira `LE-####` or a
+`langflow-ai/langflow` issue), **adding a row to `REGRESSIONS.md` is a mandatory
+step of the report/resolution** — not a follow-up.
+
+Only regressions the **suite** caught earn a row: every row traces to a spec
+failure or a spec-validation run recorded in a repo issue. A regression the team
+confirms by hand — a manual Desktop session, an API investigation outside the
+suite — is filed upstream but stays off the ledger.
 
 - Confirmed but not yet ticketed → add it under **Candidates**, not the ledger.
 - A finding that adversarial validation downgrades to a non-user-facing
@@ -664,7 +669,9 @@ mandatory step of the report/resolution** — not a follow-up.
   `REGRESSIONS.md`).
 - After editing the table, run `npm run regressions:summary` to regenerate the
   indicator block, then commit. **Never hand-edit** the block between the
-  `<!-- REGRESSIONS:START -->` / `<!-- REGRESSIONS:END -->` markers.
+  `<!-- REGRESSIONS:START -->` / `<!-- REGRESSIONS:END -->` markers. The
+  `pr-validation.yml` typecheck job runs `npm run regressions:check` and fails
+  the PR if the committed block disagrees with the table.
 
 ### Monitoring rules driven by run history
 
--- a/README.md
+++ b/README.md
@@ -275,3 +275,5 @@ See [`QA_CHECKLIST.md`](./QA_CHECKLIST.md) for the full coverage map.
 ## Contributing
 
 See [`CONTRIBUTING.md`](./CONTRIBUTING.md) for the complete guide on how to create tests, validate coverage and respond to file-watcher issues.
+
+See [`REGRESSIONS.md`](./REGRESSIONS.md) for the ledger of real Langflow regressions this suite has caught — each one adversarially validated and tracked upstream.
--- a/REGRESSIONS.md
+++ b/REGRESSIONS.md
@@ -31,18 +31,23 @@ Never hand-edit the block between the markers.
 Langflow version once the suite re-runs green against a build carrying that fix
 — the PR is what we can verify today, the version is what we will have verified.
 
+`Report` is the committed evidence file when one exists, and otherwise the repo
+issue that carries the evidence. Both shapes are allowed; what is not allowed is
+a reference that does not resolve.
+
 <!-- REGRESSIONS:START -->
-**Regressions caught:** 4 — **Open:** 1 · **Fixed:** 3
+**Regressions caught:** 5 — **Open:** 2 · **Fixed:** 3
 
-**By severity:** High 2 · Medium 2 · Low 0
+**By severity:** High 2 · Medium 3 · Low 0
 
-**By area:** model-provider 2 · auth 1 · mcp 1
+**By area:** model-provider 2 · auth 1 · flows 1 · mcp 1
 <!-- REGRESSIONS:END -->
 
 ## Ledger
 
 | Found | Area / Test | Regression | Severity | Detected by | Upstream | Status | Fixed in | Report |
 |-------|-------------|------------|----------|-------------|----------|--------|----------|--------|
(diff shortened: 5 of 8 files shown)


Victor-w-Madeira (changes requested):
## Overview

Docs/tooling only — no test code, no CI change: `REGRESSIONS.md` (4-row ledger + generated indicator block), `scripts/regressions-summary.ts` (149 lines) with the `regressions:summary` script, the mandatory-curation section in `CONTRIBUTING.md`, and the decision record.

The idea is sound and the qualifying bar (verdict + adversarial validation + upstream ticket) is the right call — a puncturable count would be worse than none. Process checklist is satisfied: `Closes #983`, off-wave exception issue labelled `qa-infra`, no spec touched so no `@stable` / spec-doc requirement applies.

**Verified independently** (ran the generator against this branch's file in a sandbox): the committed block matches the table exactly, a second run reports *already up to date* with zero diff, and the seeded data checks out — #948 / #907 / #898 / #808 exist with matching titles and labels, #966 is open. `tsconfig.json` includes `scripts/**/*.ts`, so the typecheck claim holds. `npm run lint` is `eslint tests/` only, so the new script is not linted — consistent with the existing scripts, not a defect.

## Blocking

### 1. Three of the ledger's evidence links do not exist in the repo

The ledger's value is that every row traces to evidence. Three references dangle:

- `REGRESSIONS.md:58` → `docs/upstream-bugs/UPSTREAM-BUG-new-flow-creates-without-navigating.log` — not in the repo and not added by this PR (`docs/upstream-bugs/` contains only the mcp log).
- `REGRESSIONS.md:69-70` → `LANGFLOW-BUG-bulk-flow-delete-sqlite-lock.md` and `bulk-flow-delete-500-validation.md` — bare filenames, no path, and neither file exists anywhere in the tree (checked repo-wide; not gitignored).

Only the LE-2012 row's `.log` resolves. Either commit the missing evidence files alongside this PR, or replace the references with the issue that carries them (#966 for the candidate). Shipping the registry with unresolvable pointers undercuts the "every claim is verifiable" framing on its first read.

### 2. The decision record is stale on arrival

`docs/product-regression-core/regression-ledger.md` states `Status: … not merged, no PR yet`, "**Local only: not pushed, no PR, not merged**", and closes with "**Open decision for the team:** push and open the PR" — all false the moment this merges. Same for "The original `feat/regression-ledger` branch is superseded", which is transient branch housekeeping inside a permanent doc (it already lives in the PR body, which is the right place). Rewrite the status header to reflect the merged state.

## Worth fixing here (a handful of lines each)

### 3. Nothing catches indicator drift

Regeneration is local-only by decision and the block is committed by hand. If a contributor adds a row and forgets `npm run regressions:summary`, the headline number silently disagrees with the table and no job notices. Cheap fix: a `--check` mode that recomputes and exits non-zero on mismatch, wired as a step in the existing `typecheck` job of `pr-validation.yml` (no extra `npm ci`).

Worth naming the tension: committing generated blocks in PRs is exactly the pattern `check-checklist-guard.mjs` exists to forbid (the #741 conflicts). At ~4 rows per quarter the collision risk is genuinely low, so hand-committing is defensible — but state that in the decision doc rather than leaving the inconsistency implicit, and note the auto-regen job as the eventual resolution.

### 4. Three fail-loudly gaps in the generator

The script correctly throws on a wrong column count and on bad `Severity` / `Status` (both reproduced). Three cases slip through silently and produce a wrong headline number — the one thing the design says must never happen. All three reproduced:

- `scripts/regressions-summary.ts:128-137` — the count check passes if the markers appear in the **wrong order**. With `END` before `START` the script exits 0, duplicates the block, and deletes the prose between the markers. One line fixes it: `if (md.indexOf(END) < md.indexOf(START)) throw …`.
- `:89-90` — an `Area / Test` cell without ` · ` makes the whole cell the area: `**By area:** … auth/logout-flow.spec.ts 1 …`. Throw when the separator is absent, since the schema mandates `area · spec-file`.
- `:83` — duplicate rows inflate the count (reproduced: 5 from a pasted row). A dedup guard on the `Upstream` cell makes the headline number tamper-evident against copy-paste.

Also: an empty-but-present `## Ledger` section emits `0 regression(s)` silently, while the decision doc claims a missing section aborts. Low priority (append-only file, never legitimately empty post-seed), but the doc overstates the guarantee.

### 5. `CLAUDE.md` not updated

`CLAUDE.md` is where the never-hand-edit rules for generated blocks live for agents (`QA-CHECKLIST.md`, `reports/*.jsonl`). `CONTRIBUTING.md` covers humans; an agent editing `REGRESSIONS.md` will not see it. Add a short subsection: the marker block is generated, `npm run regressions:summary` regenerates it, never hand-edit.

## Minor

- `CONTRIBUTING.md:653` — "has it **adversarially validated**" → "has been adversarially validated".
- `docs/product-regression-core/` — a new top-level directory matching neither the `docs/` convention (it mirrors `regression/` test paths) nor a ROADMAP wave axis (no such axis in `ROADMAP.md`). Existing design docs live at the root (`ISSUE-833-SHARDING-DESIGN.md`). Consider the root or a `docs/decisions/` convention.
- `README.md` does not link `REGRESSIONS.md`. For an artifact whose purpose is ROI *visibility*, one line next to the `CONTRIBUTING.md` pointer (README:277) costs nothing.
- The `Report` column mixes shapes: a `.log` path on one row, `#907`-style issue refs on the other three. Pick one, or document that both are allowed.
- `:77` / `:83` — `SEVERITIES.includes(severity as Severity)` casts unsoundly to satisfy the readonly tuple; `(SEVERITIES as readonly string[]).includes(severity)` is the safer idiom.
- `:128-129` — the regex-escape expression is duplicated; a two-line `escapeRe()` helper reads better. The markers are literals, so `split(START).length - 1` also works.

## Risk / security / performance

No spec files touched, so no suite impact and nothing to run. The script reads and writes a single fixed in-repo path with no user input, network, or shell — no security surface. Runtime is one file read plus string ops.

## Verdict

Requesting changes on 1 and 2 only — both are quick edits. 3 and 4 are a handful of lines and worth landing here, but reasonable as follow-ups if you want the ledger merged today.

rafaelgiln:
@Victor-w-Madeira — all five items addressed, plus one thing the review could not have known. Reproduced every claim before touching anything; all four generator gaps were real, including the reversed-marker one that corrupts the file instead of failing.

**Branch was re-scoped.** The ledger reached `main` via #987 while this PR was open, byte-identical to what this branch carried. Reset onto `origin/main`; the diff is now only the review fixes.

**On item 1 — you were right, and the cause was staleness.** The branch was cut from `0c1deaa` and `main` had moved four commits ahead. The New Flow evidence *is* committed, as `UPSTREAM-BUG-new-flow-dead-click.md` — added by #987 after the branch point. The ledger pointed at the draft's pre-commit name. The two bulk-delete filenames are gone entirely: neither exists in the tree or in history.

**LE-2019 changes the ledger.** Filed 2026-07-27; the ticket records *"Found by the E2E regression suite: #966"*, with 1.10.0 as last-known-good and 1.10.1 (langflow#12575) as first-affected — established by comparing the nightly and 1.11.x frontend trees, which refutes the "1.12 regression" reading. That clears both gaps you and I flagged on the candidate, so it is a row now: **5 caught — 2 open · 3 fixed**, Candidates empty.

**Item 4 — all four abort now**, each reproduced first. The empty-table case throws rather than the doc being softened: the file is append-only, so an empty table means damage, not a legitimate zero.

**Item 3 — `--check` + a step in the typecheck job.** The tension you named is written down in `CLAUDE.md` and the decision record instead of left implicit, with the auto-regen job named as the resolution if the row rate ever rises.

**One minor I am not taking:** `docs/product-regression-core/` stays. That path was Rafael's explicit choice when the record was filed — moving it is his call, not this round's.

Issue #327, opened 2026-05-25
Bootstrap timeout: mainpage_title not visible within 30s in nightly 1.10.0


daniellicnerski1:
## Summary

In Langflow nightly `1.10.0` the 4 `playground-session-*` specs all fail at the very first step (`setupPlayground` → `awaitBootstrapTest`) with the same signature:

```
TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.
  - waiting for locator('[data-testid=\"mainpage_title\"]') to be visible
    at helpers/other/await-bootstrap-test.ts:15
```

Issue auto-opened by `weekly-stable.yml` is #324; this issue tracks the bootstrap-specific cluster.

## Notes on root cause

- The `mainpage_title` testid still exists in upstream Langflow source (`src/frontend/src/pages/MainPage/components/header/index.tsx:121` and several other files) — the selector was not renamed
- All 4 failures hit the same line (`await-bootstrap-test.ts:15`), at the start of every retry, suggesting a deterministic render/timing regression rather than a flake
- This is distinct from issue #326 (LanguageModelComponent) — these tests build a ChatInput → ChatOutput flow only, no Language Model component involved
- May be related to a slower first-paint, an extra redirect on `/`, or a frontend regression that delays `mainpage_title` mount

Manual verification against the upstream `langflowai/langflow-nightly:latest` is needed to confirm whether the first-paint regression reproduces outside CI (e.g. local Docker run with the same image).

## Affected specs (4)

All currently tagged `@stable`:

- [ ] `tests/tests-automations/regression/core-functionality/playground/playground-session-clear.spec.ts:77`
- [ ] `tests/tests-automations/regression/core-functionality/playground/playground-session-id.spec.ts:20`
- [ ] `tests/tests-automations/regression/core-functionality/playground/playground-session-nav.spec.ts:20`
- [ ] `tests/tests-automations/regression/core-functionality/playground/playground-session-rename.spec.ts:30`

## History context

None of these 4 specs appear in `reports/weekly-history.jsonl` as failed or flaky in prior weekly runs — all are 1st-occurrence hard failures.

## Mitigation

Per CONTRIBUTING.md (1st-occurrence hard failure, root cause likely upstream and timing of upstream fix unknown): a separate PR removes `@stable` from all 4 specs and references this issue. Tags will be restored after confirming the bootstrap regression is fixed and re-validating each spec via the 5-step guide.

## Next steps

[...]

Issue #637, opened 2026-07-10
[Daily #629] Flaky: playground session rename — renamed label not shown; correlated 404 on rename endpoint


Victor-w-Madeira:
Surfaced in the 2026-07-10 daily triage (umbrella #629). **Flaky** (recovered on retry), so `@stable` is retained per the daily triage policy — this issue tracks the investigation.

## Failing test
`tests/tests-automations/regression/core-functionality/playground/playground-session-rename.spec.ts:106`
— "rename option must be available and functional for a session with messages"
- flaky 2026-07-02, 2026-07-10 — recurring.

## Run context
- Date: 2026-07-10 · Langflow: `langflowai/langflow-nightly:latest`
- Run: https://github.com/oriontech-me/langflow-e2e/actions/runs/29087610808

## Observed evidence (not diagnosis)
- Assertion: `expect(getByTestId('session-selector').getByText('my renamed session')).toBeVisible()` timed out (5s) — the renamed session label did not appear.
- **Correlated backend error in the same run:**
  `404 Not Found - /api/v1/monitor/messages/session/New%20Session%200?new_session_id=my+renamed+session`
  i.e. the session-rename request itself 404'd. This correlation is the reason this test gets a dedicated issue rather than being folded into the flow-open race (#588).

### Axis 1 — is it a Langflow product bug?
Rule this out FIRST:
- Is the `PATCH/PUT /api/v1/monitor/messages/session/{old}?new_session_id={new}` rename endpoint intermittently returning 404? Under what condition (session not yet persisted? name-encoding of "New Session 0"? race between message creation and rename)? A rename that 404s is a real product defect.
- Does the 404 line up 1:1 with the assertion failure across runs? Confirm causation, not coincidence.

### Axis 2 — is the TEST exercising Langflow correctly?
- Does the test rename **before** the session is durably persisted (racing session creation)? If so, the 404 may be a test-timing artifact rather than a product bug — but that must be proven, not assumed.
- Is the session identified/encoded the way the current build expects (the `New Session 0` path segment + `new_session_id` query)? An upstream contract change would make our rename call target a stale route.
- Is 5s an adequate budget for the label to reflect, or is the test racing a re-render?

[...]

Issue #549, opened 2026-07-07
[Triage] tool-mode "use components as tool": node_duration_url never appeared (daily #548)


Victor-w-Madeira:
## Source

`@stable` was auto-removed in daily #548 (run [28860759001](https://github.com/oriontech-me/langflow-e2e/actions/runs/28860759001)), 2026-07-07, against `langflowai/langflow-nightly:latest`.

## Failing test

`tests/tests-automations/regression/core-components/tool-mode.spec.ts:5` — "User should be able to use components as tool" (`@release @components`).

## Failure signature

```
expect(getByTestId('node_duration_url')).toBeVisible() failed
Timeout: 30000ms — element(s) not found
at tool-mode.spec.ts:142 (after clicking button_run_url)
```

## Pre-verification (2026-07-07)

- The same run had **5 hard fails + 13 flaky** with heavy backend noise: `500` on `DELETE /api/v1/flows/`, repeated `500` on `/api/v1/validate/prompt`, `500` on `/api/v1/monitor/messages/...` ("0 rows matched"), and many `404` on `flows/{id}/events`.
- The `node_duration_<name>` test id is **still present** in the latest Langflow frontend source — not renamed.
- `node_duration_url` only renders **after** the URL component finishes running, so the run likely did not complete.

## Investigation scope — suspect BOTH product and test

Do **not** assume the fault is in the test. Investigate both sides:

- **Product:** Did the URL component actually run on nightly? Is component execution regressing (possible link to the repeated `500` on `/api/v1/validate/prompt` in the same run)?
- **Test:** Is the 30s budget enough under nightly load? Any race between clicking run and the duration badge? Does the test tolerate transient backend instability?
- Reproduce locally against a clean nightly; if it passes cleanly and consistently, the failure was transient.

## Deliverables (required to close this issue)

1. Root-cause conclusion stating whether the fault is in the **product** or the **test** (or transient nightly instability).
2. The corresponding fix — a test change and/or an upstream product bug report — as applicable.
3. **Restore the `@stable` tag** on the test via PR (spec `test()` + spec doc). This is an explicit deliverable: the issue is not done until `@stable` is back and passing.

Derived from #548.

Victor-w-Madeira:
[...]

PR #1001, opened 2026-07-28
fix(playground): redesign playground-session-id against the current session surface (#994)

### LINKED ISSUE #994: playground-session-id: session_id input absent from the playground — product change or regression? (verdict before fix)
`playground-session-id.spec.ts:21` — "session ID input accepts a custom value" — lost `@stable` on 2026-06-01 (`290038d`, the 13-spec block attributed to #327) and has had no live tracker since. It now fails for a **different reason than the one it was quarantined for**, which is why this needs a verdict before a fix.

## Reproduces

Langflow Nightly **1.12.0.dev7**, `--workers=1 --retries=0`:

```
TimeoutError: locator.clear: Timeout 20000ms exceeded.
Call log:
  - waiting for getByTestId('popover-anchor-input-session_id')
  at playground-session-id.spec.ts:42
```

The element never appears. The test assumes the `session_id` input is present in the playground by default:

```ts
const sessionInput = page.getByTestId("popover-anchor-input-session_id");
await sessionInput.clear();
```

## Answer this before writing any fix

Elsewhere in the suite the same field is **not** available by default — `agent-n-messages-limit.spec.ts` has to expose it explicitly via `inspector-add-session_id` before it can be filled. So there are two very different verdicts:

1. **Intended product change** — the field moved behind the inspector (or out of the playground). Then this test's premise is dead and it needs **redesign**, not repair, and the redesign must assert whatever the new surface for setting a session id is.
2. **Product regression** — the field should be there and is not. Then this is a Langflow defect to **report upstream**, and the test is correct as written.

Do not widen the timeout or bolt on an `inspector-add-session_id` call before deciding: option 1 changes what the test covers, and option 2 means a test-side workaround would delete real coverage. The #974 audit exists because that distinction was skipped repeatedly.

Note the original quarantine cause was different and is resolved: #327 concluded the 2026-06 cluster was the `cleanAllFlows` race, *not* upstream — and its three sibling specs (`playground-session-nav`, `-rename`, `-clear`) all pass today. This one is a separate, newer break.

## Deliverables (Done when)

- [ ] Verdict recorded: intended product change (redesign) or product regression (report upstream).
- [ ] If a regression: the upstream issue is filed and linked, and this issue stays open until the fix lands and the test is re-validated.
- [ ] If intended: the test is redesigned against the current surface, `@stable` restored in the same PR, spec doc updated.
- [ ] `Last validated` bumped on `docs/core-functionality/playground/playground-session-id.md`.

## Reproduce

```bash
npx playwright test tests/tests-automations/regression/core-functionality/playground/playground-session-id.spec.ts \
  --workers=1 --retries=0
```

Surfaced by the #974 audit.

Victor-w-Madeira:
**Closes #994.**

## Problem

`playground-session-id.spec.ts` timed out on `locator.clear` waiting for
`getByTestId('popover-anchor-input-session_id')` — the element never appears. The test
lost `@stable` on 2026-06-01 (`290038d`, the 13-spec block attributed to #327, whose cause
was the `cleanAllFlows` race) and, with `nightly.yml` disabled since 03-2026, nothing has
run it since. The issue required a **verdict before any fix**: intended product change, or
Langflow regression?

**Verdict: intended product change — not a regression, nothing to report upstream.**
Evidence recorded on [#994](https://github.com/oriontech-me/langflow-e2e/issues/994#issuecomment-5099222713):

1. `popover-anchor-input-session_id` is not a Playground element. It comes from the generic
   parameter renderer (`parameterRenderComponent/components/inputComponent`,
   ``id={`popover-anchor-${id}`}``), so it renders wherever a component's `session_id` field
   renders. On `ChatInput` (`lfx/components/input_output/chat.py`) that field is
   `advanced=True` and always has been — which is exactly why `agent-context-id-*.spec.ts`
   must call `inspector-add-session_id` first.
2. The current Playground has no session-ID input at all. A live dump of every `data-testid`
   with the playground open on 1.12.0.dev7 returns `new-chat`, `session-selector`,
   `session-<id>-more-menu`, `chat-header-more-menu`, `playground-close-button`,
   `input-chat-playground`, `button-send` — no `popover-anchor-*`. Grepping the whole
   `components/core/playgroundComponent/` subtree confirms it: the only session-related input
   that exists is `session-rename-input`.
3. It is not a recent break. The Playground became a sliding sidebar in upstream
   `126c037aa0` (2026-02-05, pre-1.8 main fork), and its session model is identical in
   `1.10.0`, `release-1.11.0` and `main` — `pages/FlowPage/index.tsx` renders the same
   `FlowPageSlidingContainerContent` in all three. The free-text input lived in the old modal
   playground (`modals/IOModal/*`, still in the tree but no longer mounted), so the spec was
   already asserting against a dead surface when it was last rewritten (`bc72c33`).

## Fix

Redesigned against the surface that exists today — **New chat + Rename** — and moved the
validation criterion from `toHaveValue` on an input to the stored data:

| 1 | `a session renamed in the playground is the session its messages are stored under` | After renaming the session to `e2e-session-<uuid>`: `GET /api/v1/monitor/messages?flow_id=` returns **0** messages under the pre-rename session ID and **all** under the custom one; the message sent afterwards is persisted with `session_id === <custom>`; `GET /api/v1/monitor/messages/sessions?flow_id=` equals exactly `["<custom>"]` |

Rejected alternatives, per the issue's explicit instruction:

- **Widening the timeout / adding `inspector-add-session_id`** — the field is a node
  parameter, not a Playground control; either would have made a dead premise "pass" and
  deleted the coverage.
- **Asserting the renamed label in the sidebar** — that is
  `playground-session-rename.spec.ts`'s job (affordance + label). Reading the database
  instead keeps the two specs distinct: a rename that only repainted the sidebar passes there
  and fails here.
- **Covering the ChatInput node's `session_id` override** (`chat.py`:
  `self.session_id or self.graph.session_id`) — a genuine uncovered gap, but a
  `core-components` concern; noted in the spec doc's *does not cover* section rather than
  smuggled into a Playground spec.

The persistence gate before the rename is kept from `playground-session-rename` (#637): the
rename endpoint reads the DB and 404s while the message is only in the client cache, and on
404 the frontend silently keeps the old name.

## Scope note

`@stable` restored in this PR, so the test is back under `daily-stable.yml` — the silence
mechanism described in the issue is closed, not just the assertion. Spec doc rewritten with
`Last validated: 1.12.x`. The `QA-CHECKLIST.md` bullet already referenced this spec as `[x]`
and is untouched; the generated blocks regenerate on merge. No helper, POM or fixture changed —
only this spec and its doc.

## Validation (nightly 1.12.0.dev7, `--workers=1 --retries=0`)

- typecheck ✅ · eslint ✅ (0 errors) · `check:checklist-coverage` ✅ · `validate:specs` ✅
- 4/4 clean runs, no flake ✅ — 10.4s / 11.9s / 11.3s / 11.6s
- `--trace=on` ✅ — 11.2s, no hang, 172 frames, all six `test.step`s present
- zero `🚨 Backend Error` ✅ · zero orphan flows after the runs (`E2E Playground` = 0) ✅
- force-fail ✅ — three executed mutations, all reverted (`grep` for markers = 0 hits, final run green):
  - rename writes `<custom>-FFMUTATION` (the substring `getByText` still matches, so the label
    step still passes) → fails the re-key poll: `expected > 0, received 0`
  - `sendMessage("second message")` commented out → fails the follow-up poll:
    `expected "e2e-session-49abbce1", received "not persisted yet"`
  - an extra `new-chat` + message leaked before the last step → fails `toEqual`:
    `Received + 1` in the sessions array

### DIFF
--- a/docs/core-functionality/playground/playground-session-id.md
+++ b/docs/core-functionality/playground/playground-session-id.md
@@ -1,12 +1,14 @@
-# Playground — Session ID Input
+# Playground — Session ID
 
-**Last validated:** Langflow 1.10.x
+**Last validated:** Langflow 1.12.x
 
 ---
 
 ## What this test validates *(required)*
 
-Validates that the Playground session ID input field is editable and correctly reflects a custom value typed by the user. If this breaks, users cannot control which session context their messages are sent to — affecting multi-session workflows.
+Validates that the session ID a user sets from the Playground is the session the backend actually stores the conversation under — both for the messages already in the session and for every message sent afterwards. If this breaks, users cannot control which session context their messages are sent to, which affects every multi-session workflow.
+
+The Playground has no free-text `session_id` input. The modal playground that rendered the Chat Input node's parameters was replaced by the sliding-sidebar playground (upstream `126c037aa0`, pre-1.8), whose session surface is the session list — **New chat** plus **Rename**. The `session_id` field itself survives only as an `advanced=True` parameter on the Chat Input node (exposed through `inspector-add-session_id`) and as a REST parameter. This spec therefore covers the surface that exists today (verdict recorded on #994).
 
 ---
 
@@ -19,9 +21,13 @@ Validates that the Playground session ID input field is editable and correctly r
 ## Step by step *(required)*
 
 1. Use `setupPlayground(page)` to create a blank flow with ChatInput → ChatOutput connected and capture the created `flowId`
-2. Click `playground-btn-flow-io` and wait for `input-chat-playground` to be visible (confirms playground is open)
-3. Clear `popover-anchor-input-session_id` and fill it with `session-${Date.now()}`
-4. Assert the field's value equals the filled string via `toHaveValue`
+2. Open the Playground (`playground-btn-flow-io`) and wait for `input-chat-playground` to be visible
+3. Click `new-chat` and send `first message`; wait for the send to settle
+4. Poll `GET /api/v1/monitor/messages?flow_id={flowId}` until a message carrying a `session_id` other than `flowId` is queryable, and record that auto-generated session ID. The rename endpoint reads the database and 404s while the message is only in the client cache (same persistence gate as `playground-session-rename`, #637)
+5. Rename the session to `e2e-session-<uuid>` through the session more-menu → `rename-session-option` → `session-rename-input` → `Enter`
+6. Assert over the API that every message of the flow now carries the custom session ID and none carries the pre-rename one
+7. Send `second message` in the same (renamed) session
+8. Assert the new message is persisted under the custom session ID as well, and that `GET /api/v1/monitor/messages/sessions?flow_id={flowId}` returns exactly `["<custom>"]`
 
 `afterEach` navigates to `/` and deletes the flow created by the helper via `DELETE /api/v1/flows/{id}`.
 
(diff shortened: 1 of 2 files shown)


Victor-w-Madeira:
Applied both review findings in `7cbf134` (new commit, no amend).

**1. Poll-safe reads.** `readMessages()` asserted `response.ok()` from inside three
`expect.poll` generators, and Playwright runs the generator outside its retry try/catch
(`matchers/expect.js` → `pollMatcher`, `const value = await poll.generator()` precedes the
`try`). One transient 500 therefore aborted the test instead of retrying it — the exact
saturation signature of #549/#553, on a spec re-entering the daily as `@stable`. It now
returns `null` on a failed read and the polls read that as "nothing yet", the shape
`playground-session-rename` already uses. The two reads outside a poll go through
`readMessagesOrFail()`, where a failed GET must stay fatal.

**2. A dead run now names itself.** `sendMessage()` proved nothing about execution: the chat
input is cleared optimistically on send and `toBeHidden` on the stop button passes when it
never rendered. It now waits for one more `div-chat-message` — rendered exclusively by
`bot-message.tsx`, so the bubble *is* the flow's answer.

Also replaced the non-null assertion on the first persisted message with an explicit
empty-string check.

### Force-fail evidence for the two mechanisms

Both are behavioral (mutate the mechanism, observe the effect), all reverted — `grep` for
markers = 0 hits:

- `FF-D:` first 3 reads forced to non-200 → **still passes** (13.9s): the polls absorb the
  transients.
- `FF-D2:` same injection with the pre-fix `expect(ok).toBe(true)` shape restored → **fails at
  10.7s**, `Expected: true, Received: false`, well inside the 20s poll budget. That is the
  aborted-poll behaviour, proven.
- `FF-E:` `POST /api/v2/workflows` (the endpoint the playground actually sends through —
  confirmed by request probe, not `/api/v1/build`) aborted → **fails inside `sendMessage`**
  with `toHaveCount Expected: 1, Received: 0` instead of downstream at the persistence poll.
  The failure now points at the run, not at the database.

### Re-validation (nightly 1.12.0.dev7, `--workers=1 --retries=0`)

- typecheck ✅ · eslint ✅ (0 errors; 2 `no-conditional-in-test` warnings, same pattern the
  sibling `playground-session-rename` carries)
- 3/3 clean runs — 13.3s / 11.6s / 11.7s ✅
- zero orphan flows after the runs ✅

Issue #1008, opened 2026-07-28
api: frontend requests GET /api/v1/projects/undefined after the last project is deleted (422)


rafaelgiln:
Found while validating #965 (the touched-spec burst in the deterministic pipeline aborts on `backendErrors=true`). **Not caused by that PR** — see *Attribution* below.

## Symptom

`tests/tests-automations/regression/core-functionality/project-management/folder-deletion-integrity.spec.ts` logs a backend error even though all 4 tests pass:

```
🚨 Backend Error: 422 Unprocessable Content - http://localhost:7860/api/v1/projects/undefined?page=1&size=12&is_component=false&is_flow=true&search=
   Response: {"detail":[{"type":"uuid_parsing","loc":["path","project_id"],"msg":"Input should be a valid UUID, invalid character: found `u` at 1","input":"undefined","ctx":{"error":"invalid character: found `u` at 1"}}]}
```

The frontend issues the paginated flows query with the literal string `undefined` in place of the selected project id.

## Scope — isolated to one test

Running each test alone (`--grep`, `--workers=1 --retries=0`), counting `🚨 Backend Error` lines:

| Test | 422 emitted |
|---|---|
| `deleting a folder should update the folder list immediately` | 0 |
| `deleting one folder should not affect other folders` | 0 |
| `creating a new folder after deletion should work correctly` | 0 |
| `deleting every folder lands on the empty project screen` | **1** |

So the trigger is the state where **no project remains**: the UI keeps a selected-project id that is `undefined` and still fires the flows query for it. A user reaching the empty-project screen hits the same request.

## Attribution — pre-existing, product-side

- Reverting the #965 diff for this file (`git stash push -- <spec>`) and re-running reproduces the identical 422 with `4 passed`. The change under review only swaps a cleanup `request.delete` for the `deleteProject` helper, in a `finally` branch that this path does not even enter.
- The failing request is a `GET` issued by the frontend, not by any test helper.
- Reproduced on `langflowai/langflow-nightly:latest` at **1.12.0.dev7** and **1.12.0.dev8**.
- No open issue in this repo mentions `projects/undefined`.

## Reproduce

```bash
docker run -d --name langflow-e2e-runner -p 7860:7860 \
  -e LANGFLOW_AUTO_LOGIN=true -e LANGFLOW_SUPERUSER=langflow \
  -e LANGFLOW_SUPERUSER_PASSWORD=langflow123 -e LANGFLOW_WORKERS=1 \
[...]

Issue #965, opened 2026-07-27
[Daily #962] api-folders DELETE returns 500 instead of 204 (recurrent)


Victor-w-Madeira:
Spun out of daily-failure triage #962 (run [30261409427](https://github.com/oriontech-me/langflow-e2e/actions/runs/30261409427), 2026-07-27, `langflowai/langflow-nightly:latest`).

## Symptom

Recovered on retry (flaky, not a final hard failure), but **recurrent on the same signature**. `DELETE /api/v1/projects/{id}` answered **HTTP 500** where the contract expects **204 No Content**.

| Spec (line) | Assertion | Signature |
|---|---|---|
| `tests/tests-automations/regression/api/flows/api-folders-crud.spec.ts:82` ("DELETE removes folder and it no longer appears in listing") | `expect(deleteRes.status()).toBe(204)` (line 101) | `Error: expect(received).toBe(expected) // Object.is equality` — Expected `204`, Received `500` |

## Preliminary read (descriptive — NOT a verdict)

- The guard did **not** trip (2 hard failures vs. a threshold of 6+) — this is a clean, non-guarded daily, so the signal is not mass-failure collateral.
- A **500** on a well-formed delete of a folder the test just created is a server-side error surface, not a client-side rejection. Whether it is an unhandled backend exception, a cascade/foreign-key condition on the flows attached to the project, or a race with a concurrent worker's cleanup is open — the response body and the backend logs for the failing call are the first thing to capture, and neither is available from the JSON report alone.
- **Relationship to #932** (open, same spec file): #932 tracks `api-folders-crud.spec.ts:115` ("moving flow between folders via PATCH `folder_id` updates association") — a **different test** with a **different symptom** (an equality mismatch on the association after a PATCH, no 5xx). Same file, distinct root paths, so they are tracked separately. If the investigation shows a shared backend cause, fold them.

## Flake signal

This test is confirmed recurrent — failed on dailies **2026-07-22** and **2026-07-27** with the same `error_signature` (`expect(received).toBe(expected) // Object.is equality`), inside the 30-day window.

As prevention, it was **quarantined** at triage — `@stable` removed **and** `test.fixme` added — so it stops running in **every** context (daily, PR impacted-specs gate, full suite) until this issue is worked:

- `tests/tests-automations/regression/api/flows/api-folders-crud.spec.ts` (test at line 82)

Lifting the quarantine after the fix (remove `test.fixme` + restore `@stable`) is a deliverable of this issue.

[...]

Issue #932, opened 2026-07-24
[Daily #930] api-folders PATCH folder_id — flaky association assertion (recurrent)


rafaelgiln:
Spun out of daily-failure triage #930 (run [30085452003](https://github.com/oriontech-me/langflow-e2e/actions/runs/30085452003), 2026-07-24, `langflowai/langflow-nightly:latest`).

## Symptom

| Spec (line) | Assertion | Signature |
|---|---|---|
| `tests-automations/regression/api/flows/api-folders-crud.spec.ts:115` ("moving flow between folders via PATCH folder_id updates association") | `expect(received).toBe(expected)` | `Error: expect(received).toBe(expected) // Object.is equality` |

Recovered on retry (flaky, not a final hard failure), but **recurrent on the same signature**.

## Preliminary read (descriptive — NOT a verdict)

Recurrent same-signature on **07-15 and 07-24** (actionable per the recurrence window). The 07-24 daily did **not** trip the mass-failure guard. An equality mismatch on the folder association after a `PATCH folder_id` is consistent with either a backend eventual-consistency window (association not yet reflected when re-read) or a product change to the PATCH response — to be confirmed, not assumed.

## Investigation directive

**Product-first**: on the current nightly, `PATCH` a flow's `folder_id` and re-read; confirm whether the association reflects immediately or needs a settle. Rule out: (a) product change to `PATCH folder_id` semantics/response; (b) read-after-write race in the test (missing poll/settle before assert) → harden to a deterministic wait; (c) test data/isolation leak.

## Flake signal

This test is confirmed recurrent (failed on dailies **2026-07-15, 2026-07-24**). As prevention, it was **quarantined** at triage — `@stable` removed **and** `test.fixme` added (quarantine PR linked in a comment below) — so it stops running in **every** context (daily, PR impacted-specs gate, full suite) until this issue is worked:

- `tests-automations/regression/api/flows/api-folders-crud.spec.ts` (test at line 115)

Lifting the quarantine after the fix (remove `test.fixme` + restore `@stable`) is a deliverable of this issue.

## Deliverables (Done when)

- [ ] Root cause confirmed (product `PATCH` semantics vs. read-after-write race vs. isolation) with evidence on the current nightly.
- [ ] Spec passes reliably (multiple clean `--retries=0` runs).
- [ ] **Quarantine lifted** in the fix PR — remove `test.fixme` **and** restore `@stable`. Re-validate per `CONTRIBUTING.md`.

[...]

Issue #741, opened 2026-07-14
Guard QA-CHECKLIST generated blocks from PR edits (stop recurring merge conflicts)


Victor-w-Madeira:
## Problem

Spec PRs keep hitting merge conflicts on `QA-CHECKLIST.md` even though the rule already says they shouldn't touch the generated blocks.

The convention (`CLAUDE.md`, `langflow-e2e` skill VERIFY step) is: a PR edits **only** the manual Part II bullet (`- [x] <item> → <spec>.spec.ts`); the auto-generated blocks — Coverage Summary table (`**TOTAL**`, %s), Coverage Summary Note, and the `Phase 0 — Validated` block (count + bulleted list) — are regenerated automatically by `update-coverage-summary.yml` on push to `main` (`chore(checklist): auto-update coverage summary [skip ci]`). Leaving the counts to the merge job is what allows many PRs in flight without colliding.

In practice this rule is being violated: recent spec PRs run `npm run coverage:summary` locally and commit the regenerated blocks. Since every `@stable` add/remove rewrites the **same** count lines (TOTAL row, "N test() calls…" in Phase 0), any two such PRs conflict on `QA-CHECKLIST.md` — a guaranteed, avoidable collision.

Observed collisions: #740 vs #683 (this is what prompted the issue); #683 and #670 both committed regenerated blocks.

## Proposal

**(a) Reinforce the rule where it's actionable.**
- `CONTRIBUTING.md` (PR review checklist): add an explicit line — *"PR diff to `QA-CHECKLIST.md` must be bullet-only; it must NOT change the Coverage Summary table, the Coverage Summary Note, or the `Phase 0 — Validated` block. Those regenerate on merge via `update-coverage-summary.yml`."*
- Make sure the `langflow-e2e` / `langflow-e2e-issues` skills say the same (don't run `coverage:summary` for the PR; edit the bullet only).

**(b) Add a CI guard (pr-validation.yml) that fails a PR touching the generated blocks.**
- New lightweight job: check out the PR diff of `QA-CHECKLIST.md`; if any changed line falls inside the generated regions (the Coverage Summary table between its markers, the Coverage Summary Note, or the `Phase 0 — Validated` block), fail with a message pointing to the bullet-only rule.
[...]

PR #1009, opened 2026-07-28
fix(flows): stop leaking projects when DELETE 500s under contention (#965)

rafaelgiln:
Works [#965](https://github.com/oriontech-me/langflow-e2e/issues/965) (`daily-failure`, off-wave exception). **`Refs`, not `Closes`** — the root cause is a product defect and the issue's own criteria keep it open until the upstream fix lands on the nightly.

## Verdict — `langflow-regression` (product defect confirmed live), not a test defect

`DELETE /api/v1/projects/{id}` is declared `status_code=204` but answers **`500`** whenever another write transaction is in flight, and **the project is not deleted**:

```json
{"detail":"(sqlite3.OperationalError) database is locked\n[SQL: DELETE FROM folder WHERE folder.id = ?]\n[parameters: ('0913fdcdc2bd4f68bfa26d8ed3f0fc83',)]..."}
```

Why it is the endpoint and not "SQLite being SQLite" — same P=2 contention, 12 rounds:

| Endpoint | Result |
|---|---|
| `POST /api/v1/projects/` | 24/24 → `201` |
| `POST /api/v1/flows/` | 24/24 → `201` |
| `DELETE /api/v1/flows/{id}` | 24/24 → `200` |
| `DELETE /api/v1/projects/{id}` | 13/24 → `204`, **11/24 → `500`** |

`busy_timeout` **is** set (30 000 ms) with WAL active, yet failures return in ~0.03 s — the busy handler never engages on this path. `delete_project` funnels every exception into `HTTPException(500, str(e))` and only `DeploymentGuardError` is retried.

**Not a new defect, but ~7× worse on 1.12** (A/B/A/B, one arm at a time, 3 alternations, P=2, 30 deletes/round, orphans purged between rounds):

| Build | `204` | `500` | median latency of a failure |
|---|---|---|---|
| stable `1.10.3` | 66/90 | **5 (6 %)** | 1.80 s |
| nightly `1.12.0.dev7` | 50/90 | **40 (44 %)** | 0.03 s |

`delete_project` and `services/database/service.py` are byte-identical between the two builds, as are SQLAlchemy 2.0.51 / aiosqlite 0.22.1 / SQLite 3.46.1 — the rate change is reproducible but **not explained at code level**, and the report says so instead of guessing.

**Is it user-facing?** Tested, because a sibling SQLite-lock finding was previously downgraded in `REGRESSIONS.md` on the grounds that a client retry masks it. At 2 background writers that defence holds — 6/6 UI deletes succeeded, 2 via a masked retry. At 4 writers the retry budget is exhausted: the project stays in the sidebar with **no toast and an empty notification centre**, only a console error. The silent no-op is the user-facing part.

**#932 settled — separate causes, both stay open.** Under contention `PATCH /api/v1/flows/{id}` also 500s (7/20 at P=2, 27/32 at P=4), but **every** `PATCH` that returned `200` had persisted the new `folder_id` (13/13 and 5/5). #932's symptom (`200` + stale association) is not produced by contention.

## What this PR changes

The broken assertion is **untouched** — `expect(deleteRes.status()).toBe(204)` stays bare and the test stays `test.fixme` without `@stable`. What is fixed is our own fallout: five cleanup sites issued `request.delete(...)` and ignored the status, so every failed teardown left a permanent orphan project.

- **`tests/helpers/flows/delete-project.ts`** (new) — mirrors `delete-flow.ts`: `2xx`/`404` is the idempotent end state, a deterministic `4xx` throws immediately, a transient `5xx` is retried on a 250 ms backoff (3 attempts), and each retry is logged so run artifacts show how often the defect fired.
- **Adopted at the 5 sites**: `api-folders-crud`, `folder-crud`, `bulk-actions`, `flow-navigation-between-folders`, `folder-deletion-integrity`. Cleanup failures are logged, never thrown, so they cannot mask an assertion that already ran.
- **Docs**: the spec doc gains the defect, the A/B table, the #932 verdict and a concrete quarantine-lift criterion (0 of 90 deletes non-`204` under the same script). `QA-CHECKLIST.md` marks the two quarantined bullets `[!]` with their issue pointers.
- **Evidence**: `docs/upstream-bugs/UPSTREAM-BUG-project-delete-500-under-contention.md` + the 6 reproduction scripts under `docs/upstream-bugs/scripts/`, and a `REGRESSIONS.md` **Candidates** row (no ticket filed yet, so not counted in the indicator).

## Validation

Langflow `1.12.0.dev7` (burst) and `1.12.0.dev8` (re-checks) — `langflowai/langflow-nightly:latest`, fresh container, `LANGFLOW_WORKERS=1`.

- **VALIDATE burst** (`--retries=0 --workers=1`, 3×): `api-folders-crud` 3/3, `bulk-actions` 3/3, `flow-navigation-between-folders` 3/3, `folder-crud` 3/3 — all `unexpected=0 flaky=0 backendErrors=false`. `typecheck=0 lint=0`, QA-CHECKLIST guard ok.
- **FORCE_FAIL — 10/10.** Every active `test()` in all five specs was mutated one assertion at a time (marked `// FF-MUTATION`), proven red at `--retries=0 --workers=1`, then restored byte-for-byte and re-run green.
- **Behavioral cleanup force-fail.** With `deleteProject()` no-op'd, `api-folders-crud` still passed but left **2 orphan projects** (`Test Folder …`, `List Folder …`); with the helper restored, **0**. The cleanup is load-bearing.
- **Instance clean afterwards**: `GET /api/v1/projects/` → 0, `GET /api/v1/flows/` → 0.

### Known blocker, pre-existing and now tracked

`folder-deletion-integrity.spec.ts` logs `🚨 Backend Error: 422` for `GET /api/v1/projects/undefined` (the frontend fires the flows query with an `undefined` project id once the last project is deleted). All 4 tests pass, but the deterministic pipeline treats `backendErrors=true` as a hard stop, so **#965's pipeline is parked at VALIDATE by design** and this PR was opened outside the gate, with the user's authorization.

Attribution is measured, not assumed: reverting this PR's diff for that file (`git stash push -- <spec>`) reproduces the identical 422 with `4 passed`; the failing request is a frontend `GET`, not a helper call; and it reproduces on both `dev7` and `dev8`. Isolated per test, only *"deleting every folder lands on the empty project screen"* emits it (1/1; the other three 0/1). Filed as [#1008](https://github.com/oriontech-me/langflow-e2e/issues/1008).

## Follow-ups

- **#965 stays open** until the upstream fix lands on `langflowai/langflow-nightly:latest`; then re-validate, lift `test.fixme`, restore `@stable`, and flip the checklist bullet from `[!]` to `[x]`.
- Jira card for the product defect still to be filed; the `REGRESSIONS.md` row moves from **Candidates** to the **Ledger** once it has a `LE-####`.
- #1008 owns the 422 that blocks the gate for any PR touching `folder-deletion-integrity.spec.ts`.

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -100,6 +100,11 @@
 - [x] POST with invalid API key → returns 401/403 → `api-invalid-key.spec.ts`
 - [x] POST to non-existent flow → returns 404 → `api/flows/api-run-flow.spec.ts`
 
+#### 1.3.1 Folder (Project) CRUD via API
+- [x] POST `/api/v1/projects/` → creates folder, returns id and name → `api/flows/api-folders-crud.spec.ts`
+- [x] GET `/api/v1/projects/` → lists folders including the created one → `api/flows/api-folders-crud.spec.ts`
+- [!] DELETE `/api/v1/projects/{id}` → returns 204 and the folder leaves the listing — quarantined (#965): under concurrent writes the endpoint answers **500** (`sqlite3.OperationalError: database is locked`) and the folder survives; measured 44% of deletes on `1.12.0.dev7` vs 6% on stable `1.10.3` at the same 2-client contention. Product defect, upstream card pending; the `204` assertion is unchanged → `api/flows/api-folders-crud.spec.ts`
+
 #### 1.4 Components via API
 - [x] GET `/api/v1/all` → lists all available components → `api/flows/api-custom-component-creation.spec.ts`
 - [x] POST `/api/v1/custom_component` → creates custom component → `api/flows/api-custom-component-creation.spec.ts`
@@ -630,7 +635,7 @@
 #### 12.5 Flow Operations
 - [x] Lock flow — prevents editing → `flow-functionality/lock-flow.spec.ts`
 - [x] Unlock flow → `flow-functionality/flow-lock.spec.ts`
-- [x] Move flow between folders via API → `api/flows/api-folders-crud.spec.ts`
+- [!] Move flow between folders via API — quarantined (#932): `PATCH /api/v1/flows/{id}` returns `200` but the association read back is intermittently stale. Separate root cause from #965 — under contention the `PATCH` can 500, yet every `200` observed persisted the new `folder_id` (18/18) → `api/flows/api-folders-crud.spec.ts`
 - [x] Publish flow → `flow-functionality/publish-flow.spec.ts`
 - [x] Save flow components as template → `core-components/saveComponents.spec.ts`
 
--- a/REGRESSIONS.md
+++ b/REGRESSIONS.md
@@ -60,9 +60,16 @@ moment a ticket is filed. Not counted in the indicator.
 
 | Found | Area / Test | Regression | Severity | Report |
 |-------|-------------|------------|----------|--------|
+| 2026-07-27 | api · api-folders-crud.spec.ts | `DELETE /api/v1/projects/{id}` answers `500` (`sqlite3.OperationalError: database is locked`) instead of `204` while any other write is in flight, and the project survives. Not new — stable 1.10.3 emits the same instant `500` — but 1.12 raises the rate ~7× (6 % → 44 % at 2 concurrent clients, A/B/A/B) and flips the mode: 1.10.3 blocks and mostly honours the contract, 1.12 gives up in 0.03 s. Sibling write endpoints (`POST /projects`, `POST /flows`, `DELETE /flows`) survive the identical contention | Medium | docs/upstream-bugs/UPSTREAM-BUG-project-delete-500-under-contention.md |
 
-*(none — the New Flow dead click was promoted to the Ledger on 2026-07-27 when
-LE-2019 was filed.)*
+Note on the row above, so the earlier adjudication is not re-litigated blindly:
+the *Bulk-flow-delete SQLite-lock 500* entry below was downgraded because a
+client-side retry masked the failure. That defence was re-tested here and holds
+**only at light contention** — 6/6 UI deletes succeeded at 2 background writers,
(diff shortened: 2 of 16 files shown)


rafaelgiln:
### First-attempt flake on `folder-deletion-integrity.spec.ts:193` — diagnosed, pre-existing, now tracked

The first CI run showed `creating a new folder after deletion should work correctly` failing (~1.0 min) and passing on retry; the re-run with the amended commit is green across all five jobs.

Not from this PR's diff: the change to that file touches only the `finally` of the **first** test, and the failing test has no `finally` and never calls `deleteProject`.

The cause is the file's own destructive sibling — `deleting every folder lands on the empty project screen` deletes **every folder of the shared superuser**, and `fullyParallel: true` + `workers: 2` in CI lets it run beside its file-siblings. Reproduced the mechanism locally at `--workers=2 --retries=0` (3 runs, all absorbed but the collision visible): `404 Not Found` on `/api/v1/flows/<id>`, `/events` and `/note_translations` for flow ids a sibling test was still using — the signature of destructive cross-test cleanup, same class as #553. Clean at `--workers=1`, which is why single-worker validation never sees it.

Per `references/triage-verdicts.md` this is fixed on the **wiper**, not the victim, and the fix needs design (per-test user, a low-concurrency lane, or re-scoping the assertion — the blast radius also covers `folder-crud`, `bulk-actions` and `flow-navigation-between-folders`, so serial mode on the file would only be a partial mitigation). Filed as #1010 rather than widened into this PR.

Also fixed here: the `QA-CHECKLIST guard` failure on the first run. The new bullet had landed **below** the `## Coverage Summary` anchor, i.e. inside an auto-generated block (issue #741's guard). It now lives in the manual Part II under a new `#### 1.3.1 Folder (Project) CRUD via API` heading, with the two passing bullets `[x]` and the quarantined DELETE as `[!]`.

Issue #690, opened 2026-07-13
Validate & promote track-user-progress — §8.3


Victor-w-Madeira:
Wave 2 work item — pointer to a decided `QA-CHECKLIST.md` backlog bullet.

- **Section:** observability §8.3 User State
- **Intent:** Validate & promote to @stable
- **Spec:** to be located (user-progress track)
- **Bullet(s):** Track user progress

Part of **ROADMAP.md → Wave 2 — Components, RAG, flows & observability** (validation-heavy wave; convergence ~59% → ~74%). Merge with all tests tagged `@stable` and a spec doc under `docs/` mirroring the path.

PR #1021, opened 2026-07-28
fix(project-management): confine the account-wide folder wiper to a destructive lane (#1010)

### LINKED ISSUE #1010: project-management: the destructive 'deleting every folder' test wipes sibling tests' folders and flows under fullyParallel
Split out while working #965 (PR #1009). The flake below is **pre-existing** and unrelated to that PR's diff — it is an isolation defect, so per `references/triage-verdicts.md` it belongs to the **wiper**, not to its victims.

## The wiper

`tests/tests-automations/regression/core-functionality/project-management/folder-deletion-integrity.spec.ts:296` — *"deleting every folder lands on the empty project screen"* — deletes **every folder of the shared superuser** in order to reach the empty-project screen. The file already admits the hazard in a comment:

> NOTE: destructive — this test deletes EVERY folder of the shared superuser. Under fullyParallel it can race with sibling folder tests (delete their folders / be prevented from reaching zero). Proper isolation (a per-test user) is tracked separately; CI retries currently absorb the rare collision.

This issue is that "tracked separately".

## Observed victim

On PR #1009, `Run impacted E2E specs` failed on the first attempt at
`folder-deletion-integrity.spec.ts:193` — *"creating a new folder after deletion should work correctly"* (~1.0 min, i.e. a `waitFor` timeout) and passed on retry, so the job ended green. That test creates `folder-one` via the UI, deletes it, then creates `folder-two`; if the wiper runs concurrently, the folder it is waiting for is deleted underneath it.

The victim is reachable because `playwright.config.ts` sets `fullyParallel: true` with `workers: 2` in CI, and `fullyParallel` distributes tests **within a file** across workers — so the wiper and its file-siblings can run at the same time.

## Local reproduction of the mechanism

Three runs of that spec at `--workers=2 --retries=0` against `1.12.0.dev8` (all `4 passed`, i.e. absorbed, but the collision is visible in the backend-error log):

```
run 1: 4 passed (15.2s) — clean
run 2: 4 passed (54.1s)
  🚨 Backend Error: 404 Not Found - /api/v1/flows/00b8a94f-.../events?since=...
  🚨 Backend Error: 404 Not Found - /api/v1/flows/00b8a94f-.../note_translations
  🚨 Backend Error: 404 Not Found - /api/v1/flows/00b8a94f-...   (x3)
run 3: 4 passed (53.7s)
  🚨 Backend Error: 404 Not Found - /api/v1/flows/a99efb35-...   (x2)
  🚨 Backend Error: 404 Not Found - /api/v1/flows/a99efb35-.../events?since=...
```

`404` on a flow id a *sibling test is still using* is the signature of destructive cross-test cleanup (the same class as #553). At `--workers=1` the spec is clean, which is why single-worker validation never sees it.

## Blast radius is wider than one file

The wiper deletes every folder of the shared superuser, so it can also take out folders created by the sibling specs that run in the same impacted-specs job:

- `core-functionality/project-management/folder-crud.spec.ts`
- `core-functionality/project-management/bulk-actions.spec.ts`
- `core-functionality/project-management/flow-navigation-between-folders.spec.ts`

That is why `test.describe.configure({ mode: 'serial' })` on the file is only a partial mitigation: it removes the intra-file race (the one that produced the failure above) but not the cross-file one.

## Directions (to be decided when the issue is worked)

1. **Per-test user isolation** — run this test as its own user so "every folder" means "every folder of a user nobody else touches". Closest to correct; needs a seeding helper and `getAuthToken` for a non-superuser.
2. **Low-concurrency lane** — tag it so it runs where nothing else runs concurrently (see the `langflow-e2e-infra` skill, "low-concurrency lane").
3. **Re-scope the assertion** — assert the empty-project screen from a state the test itself owns instead of deleting the whole account's folders. Cheapest, but weakens what the test proves, so it needs an explicit call.

## Deliverables (Done when)

- [ ] The wiper can no longer delete a folder or flow it does not own (whichever direction is chosen, stated with its trade-off).
- [ ] `folder-deletion-integrity.spec.ts` runs clean at `--workers=2` with **no** `404` on sibling-owned ids, 3× in a row.
- [ ] The NOTE comment in the spec is updated to point at the resolution instead of "tracked separately".
- [ ] Not fixed by relaxing the victim's timeouts, and not by removing `@release` from anything.

rafaelgiln:
**Closes #1010.**

## Problem

`folder-deletion-integrity.spec.ts` — *"deleting every folder lands on the empty project screen"* — deletes **every folder of the shared superuser**. `playwright.config.ts` runs `fullyParallel: true` with `workers: 2`, and `fullyParallel` distributes tests **within a file** across workers, so the wiper ran beside its own file-siblings and beside `folder-crud.spec.ts`, `bulk-actions.spec.ts` and `flow-navigation-between-folders.spec.ts` in the same job. The file already admitted the hazard in a comment and deferred it to "tracked separately" — this is that.

The mechanism is sharper than the `waitFor` timeout the CI failure on PR #1009 looked like:

```
folder-deletion-integrity.spec.ts:193 › creating a new folder after deletion should work correctly
  Error: expect(received).toBe(expected)   Expected: true  Received: false
  Timeout 30000ms exceeded while waiting on the predicate
    at dismissWelcomeOverlayAndWaitForModal (helpers/flows/open-new-flow-templates-modal.ts:106)
    at addFlowToTestOnEmptyLangflow (helpers/flows/add-flow-to-test-on-empty-langflow.ts:10)
    at awaitBootstrapTest (helpers/other/await-bootstrap-test.ts:24)
```

The wiper empties the account mid-flight, so the victim's `awaitBootstrapTest` takes the **empty-instance branch** and waits 30 s for a welcome overlay that state never produces. Scheduling decides who loses — when the wiper happens to run last the whole set is green, which is why the daily sees a rare flake and `--workers=1` validation never reproduces it.

Per `references/triage-verdicts.md` this is verdict 5, cross-worker destructive cleanup: **fix the wiper, never the victim**. No victim spec is edited, no timeout relaxed, no tag removed.

## Fix — a low-concurrency lane

The wiper keeps deleting for real and keeps asserting the real empty-project screen. What changes is *when* it may run.

```ts
// playwright.config.ts
const DESTRUCTIVE_LANE = !!process.env.PW_DESTRUCTIVE;
grepInvert: DESTRUCTIVE_LANE ? undefined : /@destructive/,
fullyParallel: DESTRUCTIVE_LANE || process.env.PW_SHARD_FILE_LEVEL ? false : true,
workers: DESTRUCTIVE_LANE ? 1 : process.env.CI ? 2 : undefined,
```

- **`grepInvert`, not `grep`** — deliberate: a CLI `--grep` (the daily's `@stable`, nightly's optional filter) *overrides* `config.grep` but leaves `config.grepInvert` in place, so no caller can bypass the exclusion with a filter of its own. Verified: `--grep @release` matches all 4 tests and `--list` still reports 3.
- **`workers: 1` lives in config**, not on the command line, so a caller cannot forget it and let two destructive tests wipe each other.
- **CI runs the lane as a separate step after the normal run** in `nightly.yml`, `pr-validation.yml`, `adaptive-impacted.yml` and the shared `run-e2e` action (used by `manual.yml`) — each with `--pass-with-no-tests` (no-op when nothing destructive is in scope), `--reporter=github` (writes no HTML, so it cannot clobber the main run's `playwright-report/`) and its own `--output` directory.
- **The exclusion announces itself.** Every plain run prints the tag it left out and the exact command that runs it, so this can never become a silent coverage cap.

### Rejected alternatives, with the reason

| Direction | Why not |
|---|---|
| **Per-test user isolation** | Measured, not assumed: with `LANGFLOW_AUTO_LOGIN=true` (every workflow + `start-langflow-docker.sh`) injecting another user's token into the browser cookies is undone on mount — the app fires `GET /api/v1/auto_login`, overwrites the token, `whoami` reports `langflow` again and the probe user's own project is invisible in the sidebar. Matches the #690 finding. API-level isolation *does* work, but this test's subject is the UI. |
| **Re-scope with a `page.route` mock** | Leaves CI untouched, but the deletion→empty-screen transition stops being end-to-end and becomes "the frontend renders an empty list". |
| **Cross-process lockfile mutex** | Keeps the assertion, but requires every sibling folder spec to take a shared lock and silently regresses the moment a future folder spec forgets. |
| **`mode: "serial"` on the file** | Partial only — removes the intra-file race, not the cross-file one. The wiper deletes folders belonging to three other spec files. |

## Also fixed — an orphan-flow leak in the same spec

Found while checking the orphan count (standing rule: every touched spec cleans up its own flows id-scoped). Tests 2 and 3 open the "Basic Prompting" starter template to reach the folder view through the real navigation path — and opening a template **creates a flow**, which nothing deleted. The local instance had **47 flows, 20 of them stray `Basic Prompting` copies**. They now capture the id from the creation `POST` and delete exactly that flow in an `afterEach` (#515 pattern, never a global wipe).

## Validation (nightly `1.12.0.dev8`, `--retries=0`)

**The criterion** — the four folder-touching specs run together at `--workers=2`:

| | Failures | `404` on a **project** id (the wiper's fingerprint) |
|---|---|---|
| **Before** (3 runs) | **1** | 3, across 2 of the 3 runs |
| **After** (6 runs) | **0** | **0** |

- **Residual, reported not silenced:** 1 of those 6 runs logged five `404`s on one *flow* id, with the `/events?since=` + `/note_translations` signature of an open editor polling a deleted flow. Not the wiper — it is excluded from those runs and the `422` it emits disappeared with it — and not reproducible in pairs (`folder-deletion-integrity` alone, `+bulk-actions`, `+folder-crud`: 0 `404`s each). A separate collider; it gets its own issue rather than being folded in here.
- **3× `--workers=1 --retries=0`** on the spec: `expected=3 unexpected=0 flaky=0 backendErrors=false`. The `3` (not 4) is the lane working. **`backendErrors=false` is a side effect worth naming:** the #1008 `422` came from the wiper, so this file no longer trips the pipeline gate that #1008 was blocking — #1008 itself stays open, the `422` now appears only in the lane.
- **The lane:** `PW_DESTRUCTIVE=1 npx playwright test <spec> --grep @destructive` → `Total: 1 test`, `1 passed (6.4s)`, reaching the real empty-project screen (sidebar message + `new_project_btn_empty_page`), with the expected #1008 `422` there.
- **Force-fail:** all **4** tests mutated red one at a time and restored byte-for-byte, 0 `FF-MUTATION` markers left. The 4th ran **inside** the lane — without `PW_DESTRUCTIVE` it would have been a vacuous pass-with-no-tests, which is itself a check that the gating is real.
- **Orphans:** local instance purged back to its 28-flow baseline before the runs; nothing left after them.
- typecheck ✅ · eslint ✅ · QA-CHECKLIST guard ✅ · spec↔doc↔checklist coverage guard ✅ · all four edited YAML files parse ✅

## Known gap, documented rather than implied

`daily-stable.yml` filters `--grep "@stable"` and has **no destructive lane**. It needs none today (the wiper is not `@stable`), but that means a test carrying **both** tags would be excluded from the daily and silently never run. `@stable` and `@destructive` are recorded as mutually exclusive in the `CLAUDE.md` tag table and in `CONTRIBUTING.md` until the daily grows a lane of its own.

## Docs

The spec had **no doc** under `docs/` — one is added (`docs/core-functionality/project-management/folder-deletion-integrity.md`) with the four tests, the lane, the measured baseline, the rejected directions and the daily gap. The two existing `[-]` checklist bullets (*Integrity after deletion*, *Create folder after deleting all folders*) get the spec reference the coverage guard requires, plus a new bullet for the wiper. They stay `[-]`, not `[x]` — the spec is not `@stable` while #1008 is open.

## Reproduce locally

```bash
D=tests/tests-automations/regression/core-functionality/project-management

# default lane — 3 tests, the destructive one excluded (the run says so)
npx playwright test $D/folder-deletion-integrity.spec.ts --workers=1 --retries=0

# the destructive lane — 1 test, alone (DELETES every project of the local superuser)
PW_DESTRUCTIVE=1 npx playwright test $D/folder-deletion-integrity.spec.ts --grep "@destructive" --retries=0

# the criterion — the scenario that used to break
npx playwright test $D/folder-deletion-integrity.spec.ts $D/folder-crud.spec.ts \
  $D/bulk-actions.spec.ts $D/flow-navigation-between-folders.spec.ts --workers=2 --retries=0
```

### DIFF
--- a/.github/actions/run-e2e/action.yml
+++ b/.github/actions/run-e2e/action.yml
@@ -52,6 +52,32 @@ runs:
         echo "Running: npx playwright test ${ARGS[*]}"
         npx playwright test "${ARGS[@]}"
 
+    # Destructive lane (#1010). `@destructive` tests mutate account-wide state, so
+    # playwright.config.ts excludes them from the run above and pins this lane to
+    # workers=1. It runs after that run, ANDed with whatever filters the dispatch
+    # asked for, and no-ops when nothing destructive matches.
+    - name: Run destructive lane
+      if: always()
+      shell: bash
+      env:
+        CI: "true"
+        PW_DESTRUCTIVE: "1"
+        PLAYWRIGHT_BASE_URL: ${{ inputs.base_url }}
+        TEST_TAG: ${{ inputs.test_tag }}
+        TEST_GREP: ${{ inputs.test_grep }}
+        OPENAI_API_KEY: ${{ inputs.openai_api_key }}
+      run: |
+        # Same one-grep constraint as above: AND the filters into a single
+        # lookahead regex, always including @destructive.
+        FILTER="(?=.*@destructive)"
+        if [ -n "$TEST_TAG" ]; then FILTER="${FILTER}(?=.*${TEST_TAG})"; fi
+        if [ -n "$TEST_GREP" ]; then FILTER="${FILTER}(?=.*${TEST_GREP})"; fi
+        # --reporter=github only: keeps this lane from overwriting the HTML report
+        # the step above produced and the artifact below uploads.
+        echo "Running destructive lane: --grep $FILTER"
+        npx playwright test --pass-with-no-tests --reporter=github \
+          --output=test-results-destructive --grep "$FILTER"
+
     - name: Upload Playwright report
       if: always()
       uses: actions/upload-artifact@v4
--- a/.github/workflows/adaptive-impacted.yml
+++ b/.github/workflows/adaptive-impacted.yml
@@ -202,6 +202,27 @@ jobs:
             npx playwright test $SPECS --reporter=github
           fi
 
+      # Destructive lane (#1010). Excluded from the run above by
+      # playwright.config.ts and pinned to workers=1 here; runs after it, scoped to
+      # the same subset, and no-ops when nothing destructive is in scope.
+      - name: Run destructive lane
+        if: always()
+        env:
+          CI: "true"
+          PW_DESTRUCTIVE: "1"
+          PLAYWRIGHT_BASE_URL: "http://localhost:7860/"
+          RUN_FULL: ${{ needs.compute-impact.outputs.run_full }}
+          SPECS: ${{ needs.compute-impact.outputs.specs }}
+        run: |
+          if [ "$RUN_FULL" = "true" ]; then
+            npx playwright test --grep "@destructive" --pass-with-no-tests \
+              --reporter=github --output=test-results-destructive
+          else
+            # shellcheck disable=SC2086
+            npx playwright test $SPECS --grep "@destructive" --pass-with-no-tests \
+              --reporter=github --output=test-results-destructive
+          fi
+
       - name: Upload Playwright report
         uses: actions/upload-artifact@v7
         if: always()
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -96,6 +96,31 @@ jobs:
           CI: "true"
           PLAYWRIGHT_BASE_URL: "http://localhost:7860/"
 
+      # Destructive lane (#1010). `@destructive` tests mutate account-wide state
+      # (deleting every project of the shared superuser), so playwright.config.ts
+      # excludes them from the run above and pins this lane to workers=1. It runs
+      # AFTER that run, and `--pass-with-no-tests` makes it a no-op when the
+      # optional grep filter above selected nothing destructive.
+      #
+      # `--reporter=github` replaces the whole configured reporter list, so this
+      # step writes no HTML and cannot clobber the main run's `playwright-report/`
+      # — at the cost of the lane not reaching Flakiness.io. `--output` keeps its
+      # traces/screenshots out of the main run's `test-results/` too. The GitHub
+      # annotations are the lane's signal.
+      - name: Run destructive lane
+        if: always()
+        run: |
+          GREP="${{ github.event.inputs.test_grep }}"
+          if [ -n "$GREP" ]; then
+            npx playwright test --grep "(?=.*@destructive)(?=.*${GREP})" --pass-with-no-tests --reporter=github --output=test-results-destructive
+          else
+            npx playwright test --grep "@destructive" --pass-with-no-tests --reporter=github --output=test-results-destructive
+          fi
+        env:
+          CI: "true"
+          PW_DESTRUCTIVE: "1"
+          PLAYWRIGHT_BASE_URL: "http://localhost:7860/"
+
       - name: Upload Playwright report
         uses: actions/upload-artifact@v7
         if: always()
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -276,6 +276,30 @@ jobs:
           # shellcheck disable=SC2086
           npx playwright test $SPECS --reporter=github
 
+      # Destructive lane (#1010). `@destructive` tests wipe account-wide state, so
+      # playwright.config.ts keeps them out of the run above and pins this lane to
+      # workers=1. Scoped to the same impacted $SPECS, so it is a no-op (via
+      # --pass-with-no-tests) unless a destructive spec is actually impacted.
+      # --reporter=github replaces the reporter list, so no HTML is written here and
+      # the main run's playwright-report/ stays intact.
+      - name: Run destructive lane
+        if: always()
+        env:
+          CI: "true"
+          PW_DESTRUCTIVE: "1"
+          PREFLIGHT_SKIP_CREDENTIALS: ${{ needs.detect-specs.outputs.needs_models == 'true' && '0' || '1' }}
+          PLAYWRIGHT_BASE_URL: "http://localhost:7860/"
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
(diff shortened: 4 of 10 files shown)


rafaelgiln:
Correction to the Validation table in the PR body: the "after" column is **1 project-id `404` in 1 of 6 runs**, not zero. I counted the post-fix `404`s with `head -3` on a `uniq -c` and the project line fell outside what I read. Full breakdown from the same logs:

| | Failures | `404` on a project id | `404` on a flow id |
|---|---|---|---|
| Before (3 runs) | 1 | 3, across 2 of 3 runs | 1 |
| After (6 runs) | 0 | **1, in 1 of 6 runs** | 5, in that same run |

The conclusion the PR draws is unchanged — 0 failures across 6 runs, and the wiper is no longer a collider — but the claim "zero project-id 404s" was wrong and is worth correcting on the record.

The surviving one is a different shape from the wiper's. It was logged under `folder-crud.spec.ts:22` (*creates, renames and deletes an empty project folder via the UI*), which reported exactly 1 backend error, and the URL is the flows-in-folder query carrying that same folder id — i.e. the UI refetching the folder the test itself had just deleted, not another test's folder being wiped. Attribution is from the run log, not proven by id ownership.

Both residuals (this one and the five flow-id 404s with the `/events?since=` + `/note_translations` editor-polling signature) are filed as a follow-up: #REPLACE

rafaelgiln:
Follow-up filed: #1023 (both residuals, with the corrected numbers above).

PR #1034, opened 2026-07-28
chore(triage): render and validate the dedicated-issue body from code

### LINKED ISSUE #1032: Enforce the dedicated-issue template in code, not prose — verbatim signatures and a stated grouping argument
## Context

The team decided that failure issues are opened **per root cause**, not per failing test.
That decision is already the de facto model: `triage-core.mjs` computes
`provider_wide_clusters`, `CONTRIBUTING.md` gates the recurrent-flake issue on the same
`error_signature` within 30 days, and the triage skill already carries a canonical
template at `.claude/skills/langflow-e2e-triage/references/issue-templates.md`.

**Scope note — which issues this is about.** Two different issues get opened around a
red daily, and only one is in scope here:

- The **umbrella** issue is created by `daily-stable.yml` through
  `github.rest.issues.create`. The GitHub API bypasses issue templates entirely, so no
  template can govern it. Out of scope.
- The **dedicated per-cause issues** are created by the triage skill (Phase 7 of
  `langflow-e2e-triage`, driven by `triage-dispatch.yml`) via `gh issue create`. These
  are what this issue is about.

## The real problem: prose is the only enforcement

The canonical template exists and is mature — provenance line, `## Symptom` table,
`## Preliminary read (descriptive — NOT a verdict)`, `## Investigation directive`,
`## Deliverables (Done when)`, the quarantine block, the Enrich-vs-Create rule. What it
does **not** have is any mechanism: it is a Markdown reference read by an LLM. Since the
primary author of these issues is Claude Code, "the template says so" is the weakest
possible guarantee — it degrades silently under a long context or an unusual cluster
shape, and nothing fails when it does.

The issue body already has a machine consumer: the QA Platform parses backticked
spec paths out of it to render the `tracked · #NNN` chip on run pages. So the format is
load-bearing beyond human reading, which makes prose-only enforcement worse than it looks.

## Explicitly NOT in scope: a verdict field at dispatch time

An earlier draft of this proposal added a required **Verdict** dropdown and a required
**Root cause** field. Both are rejected: they contradict the established doctrine that a
dispatch-time triage issue is **descriptive, not conclusive** (`issue-templates.md` →
*"Do not conclude cause — that is the investigation's job"*). The verdict, and the proof
across environments behind it, is a **deliverable of the dedicated issue**, not an input
to opening it. Nothing here should push the triage agent to conclude cause earlier.

### 1. Make the structure code, not prose — `scripts/lib/triage-core.mjs`

Two new exported functions, unit-tested next to the existing `triage-core.test.mjs` (both
already gate every PR via `npm run test:scripts`):

- **`renderDedicatedIssueBody(cluster)`** — deterministically renders the canonical body
  from the triage dataset the skill already builds. The agent fills a data object and
  calls this, instead of composing Markdown from memory. Section order, heading text and
  the provenance line stop being a judgement call.
- **`assertDedicatedIssueBody(body)`** — validates a body before `gh issue create`:
  required headings present, at least one backticked repo-relative spec path (the QA
  Platform join), a non-empty signature block, and no placeholder text left in. Fails
  loud rather than opening a malformed issue.

Phase 7 of the skill then calls the renderer instead of describing the format.

### 2. Three content gaps in `references/issue-templates.md`

- **Signatures must be verbatim.** The `## Symptom` table's `Signature` column is
  specified loosely ("the exact error message or timeout signature") and the worked
  example carries paraphrases (`expect(locator).toBeVisible() failed`,
  `waitForSelector 60000ms timeout`). Those do not match the `error_signature` values in
  `reports/daily-history.jsonl`, so recurrence cannot be matched mechanically. Require
  the value copied unedited from the history row, and record `"unknown"` where the run
  recorded it rather than substituting a description.
- **Add a "Why these failures are one cause" section.** Today there are aggressive
  grouping *rules* (guard-day) but no field stating the linking argument for a given
  cluster. Without it, a root-cause issue is indistinguishable from a batch of unrelated
  failures filed together. This stays descriptive — what ties the observations together
  (shared timing, shared shard, shared dependency, common step) — so it does not
  conflict with the no-verdict rule above.
- **Key the Enrich-vs-Create rule on the normalized signature.** It currently matches on
  symptom prose ("Same failure symptom (e.g., 'div-chat-message timeout')"). Route it
  through `normalizeSignature()` against the open `daily-failure` issues so the same
  cause reliably lands as a comment. Without this, root-cause grouping degrades into one
  new issue per day for the same cause. Add the inverse too: two open issues found to
  share a signature cluster get merged and cross-linked.

### 3. `.github/ISSUE_TEMPLATE/failure-root-cause.yml` — thin mirror, lowest priority

A GitHub issue form only applies to the web "New issue" UI; `gh issue create --body-file`
ignores it. So this is **not** the enforcement mechanism — it exists so a human opening a
cause issue by hand produces the same shape the renderer produces. It must mirror the
canonical section list, not invent a parallel one.

## Out of scope (separate issues if we want them)

- **`normalizeSignature()` is too weak to be a matching key.** It only strips ANSI,
  collapses whitespace and lowercases, over the raw first error line (240 chars). So
  `Timeout 30000ms exceeded` and `Timeout 60000ms exceeded` are distinct keys, and any
  message embedding a generated flow name is unique per run — recurrence under-fires
  exactly on the tests carrying dynamic data. Fix shape: keep `error_signature` (raw,
  human-facing) and add a masked `signature_key` for matching, with `"unknown"` excluded.
  This blocks item 2's third bullet from being fully reliable, but not from being an
  improvement over prose matching.
- **The two payload producers disagree on flakes.** `scripts/append-weekly-history.mjs`
  writes `error_signature` on `flaky[]` rows; `scripts/build-run-payload.mjs` writes it
  only on `failures[]`. Since the flake rule is entirely signature-keyed, the QA Platform
  cannot currently reproduce the triage verdict for flakes.

## Intake

Off-wave exception (`qa-infra` + `follow-up`): tooling follow-up of the triage machinery
shipped around #744 / #1012. No Wave 4 item covers it.

Victor-w-Madeira:
## Type

---

- [x] **Exception** — off-wave work backed by an issue: a follow-up, a `daily-failure` triage issue, or a `community`-labeled regression. Issue #1032

---

## What this PR does

The team decided failure issues are opened **per root cause**, not per failing test. The canonical format for those issues lived only in `.claude/skills/langflow-e2e-triage/references/issue-templates.md` — Markdown read by an LLM. Since `gh issue create` bypasses `.github/ISSUE_TEMPLATE` entirely and Claude Code is the primary author of these issues, prose was the only thing holding the structure together.

**Scope:** this covers the **dedicated per-cause issues** created by the triage skill in Phase 7. The umbrella issue is created by `daily-stable.yml` through the GitHub API, which no template can govern — out of scope by construction.

### 1. The body is now rendered and validated by code

Three functions in `scripts/lib/triage-core.mjs`:

- `renderDedicatedIssueBody()` — builds the body from triage data. Fixed section order, provenance line, `**Upstream:**` slot, canonical deliverables, optional flake block. Refuses a cluster whose test carries no `error_signature`; escapes a literal `|` so a signature cannot break the table. Signatures otherwise pass through untouched, **including the literal `unknown`** — substituting a description makes the next run's occurrence unmatchable.
- `renderDedicatedIssueTitle()` — refuses a title built from the run id instead of the umbrella number. Both are bare integers in the dataset and swapping them yields a plausible title that links nowhere.
- `assertDedicatedIssueBody()` — validates before creation, covering hand-written and enriched bodies: required sections, provenance, the Upstream slot, at least one backticked repo-relative spec path, checkbox deliverables, leftover placeholders.

Phase 7 of `SKILL.md` now calls them instead of describing the format.

### 2. Two bugs found while building it

**`ANSI_RE` dropped the ESC byte.** It was `/\[[0-9;]*m/g` — no ESC — so it stripped the `[2m` and left the bare control byte:

```
stripAnsi('Error: <ESC>[2mexpect(')  ->  'Error: <ESC>expect('
```

`normalizeSignature()` inherits that, so the same cause recorded once with ANSI and once without does **not** compare equal — and the same-signature 30-day rule is what decides whether a flake earns an issue. **150 of the 437 entries** in `reports/daily-history.jsonl` carry a raw ESC, so roughly a third of every recurrence comparison was affected. `scripts/build-run-payload.mjs` already had the correct pattern; this copy had drifted.

**`npm run test:scripts` never reached `.claude/skills/**`.** It was `node --test scripts/`. The 28 tests in `triage-core.test.mjs` existed, passed locally, and ran in no PR — the lane reported 22. The lane now enumerates files the way `test:units` already does, with the same empty-result guard. **22 → 71 tests.**

### 3. Three content gaps in the reference doc

- **Signatures are copied, not written.** The Signature column was specified as "the exact error message or timeout signature" and the worked example carried paraphrases, which do not match `daily-history.jsonl`. Now: copied unedited, `unknown` written as `unknown`.
- **New section "Why these failures are one cause"** — the linking argument. The doc had aggressive grouping *rules* for guard days but no field stating why a given cluster is one cause, which leaves a root-cause issue indistinguishable from a batch of unrelated failures. Kept descriptive, so it does not cross the existing no-verdict line.
- **Enrich-vs-Create keyed on `normalizeSignature()`**, not on symptom prose, plus the inverse merge rule, plus a warning that generic signatures (`toBeVisible failed`, `unknown`) are too weak to match on alone.

### 4. The `**Upstream:**` slot

This issue layer tracks the **failure**; the treatment is worked on the Jira board. That split only holds if the layers are joinable, and the card key previously appeared only inside a `REGRESSIONS.md` deliverable checkbox — an acceptance criterion, not a link. Always rendered now, `_not filed_` when no card exists yet, and the validator fails when the line is missing.

---

## Tests covered

|---|---|---|
| 1 | `stripAnsi leaves no orphan ESC byte behind` | The regression that broke signature comparison: codes stripped, control bytes left behind |
| 2 | `stripAnsi does not eat bracketed text that is not an escape sequence` | The fix does not over-strip legitimate `[2m]` in prose |
| 3 | `renderDedicatedIssueBody preserves the literal "unknown" signature` | A message-less failure keeps a matchable key instead of a guess |
| 4 | `renderDedicatedIssueBody escapes a pipe so the table survives` | A signature containing `\|` cannot silently corrupt the Symptom table |
| 5 | `renderDedicatedIssueBody refuses a test with no signature` | A cluster cannot be filed without the key recurrence matches on |
| 6 | `renderDedicatedIssueTitle rejects a non-issue-number umbrella` | The run-id-for-umbrella swap cannot produce a title that links nowhere |
| 7 | `assertDedicatedIssueBody rejects a spec named in prose only` | The QA Platform's `tracked · #NNN` join cannot be silently dropped |
| 8 | `renderDedicatedIssueBody always carries the Upstream slot` | The seam to the Jira layer survives even when no card exists yet |

Plus emission order, the flake block, canonical deliverables, placeholder detection, and code-span tolerance — 21 new tests, 71 in the lane.

---

- **The umbrella issue.** Created via `github.rest.issues.create` in `daily-stable.yml`; templates do not apply to the API path.
- **A `Verdict` or `Root cause` field at dispatch time.** Deliberately rejected: it contradicts the established doctrine that a triage issue is descriptive, not conclusive (`issue-templates.md` → *"Do not conclude cause — that is the investigation's job"*). The verdict is a deliverable of the issue, not an input to opening it.
- **The treatment layer.** What gets done about a failure is worked on the Jira board; this layer tracks the failure and links out to it.
- **A masked `signature_key`.** `normalizeSignature()` still only strips ANSI, collapses whitespace and lowercases, so `Timeout 30000ms` and `Timeout 60000ms` remain distinct keys and a signature embedding a generated flow name is unique per run. Now an enhancement on a working comparison rather than a fix for a broken one; worth its own issue.
- **The producer asymmetry.** `scripts/append-weekly-history.mjs` writes `error_signature` on `flaky[]` rows; `scripts/build-run-payload.mjs` writes it only on `failures[]`, so the QA Platform cannot reproduce the triage verdict for flakes. Untouched here.

---

## Known limitations

**Enforcement is one step short of airtight.** The renderer and validator are deterministic and fail loud — but *calling* them is still an instruction in `SKILL.md`, so an agent that ignores Phase 7 and hand-writes a body can still create a malformed issue. The airtight version is a workflow on `issues.opened` (label `daily-failure`) that runs `assertDedicatedIssueBody` against the created body and labels/comments when it fails. Not in this PR; the honest description of what changed is "prose describing a format" → "prose invoking a deterministic enforcer", which is a real improvement in kind but not a hard gate.

**The renderer can now fail an unattended path.** Phase 7 runs headless from `triage-dispatch.yml`. A cluster shape the renderer rejects (missing signature, empty test list) will now throw where the previous prose path would have produced a degraded issue. Failing loud is the intended trade, but it is a behavior change in a path with no human watching.

**The issue form YAML is unvalidated locally.** No YAML parser is installed in the repo (`yaml`, `js-yaml`, `pyyaml` all absent). GitHub validates issue forms on push and surfaces errors on the "New issue" page — worth a look after merge.

---

## Related issue

Closes #1032

---

## Validation

- [x] `npm run test:scripts` — 71 passing (was 22 before the lane fix)
- [x] `npm run lint` — 0 errors
- [x] Rendered and validated a body from the real `2026-07-27` history row: both hard failures emitted, `unknown` signature preserved verbatim, validator returned no problems
- [x] Confirmed the ANSI fix against real data — 150 of 437 history entries carry a raw ESC

### DIFF
--- a/.claude/skills/langflow-e2e-triage/SKILL.md
+++ b/.claude/skills/langflow-e2e-triage/SKILL.md
@@ -239,11 +239,18 @@ plan and present it again — the gate re-applies to the revised plan too.
 
 Only after the user approves the plan:
 
-1. For each **create** row: `gh issue create` using the dedicated-issue
-   template and area-label mapping in
+1. For each **create** row: build the body with
+   `renderDedicatedIssueBody()` from `scripts/lib/triage-core.mjs` — do **not**
+   compose the Markdown by hand. The renderer owns section order, the
+   provenance line and the canonical deliverables; it refuses a cluster with a
+   missing `error_signature`, and `renderDedicatedIssueTitle()` refuses a title
+   built from the run id instead of the umbrella number. Then run
+   `assertDedicatedIssueBody(body, { throwOnError: true })` before
+   `gh issue create`, and apply the area-label mapping in
    `references/issue-templates.md`.
 2. For each **enrich** row: `gh issue comment` on the matched existing issue,
-   per the same reference's *Enrich vs Create Rule*.
+   per the same reference's *Enrich vs Create Rule*. Match on the **normalized
+   signature**, not on a description of the symptom.
 3. **Quarantines — separate authorization, one at a time.** For each recurrent
    flake in the removals block: show the **exact diff** (drop `@stable` **and**
    wrap the test as `test.fixme(<reason + issue #>)`, at `spec:line`) and ask
--- a/.claude/skills/langflow-e2e-triage/references/issue-templates.md
+++ b/.claude/skills/langflow-e2e-triage/references/issue-templates.md
@@ -44,6 +44,25 @@ Example: `[Daily #744] agent/flow execution does not complete — div-chat-messa
 Spun out of daily-failure triage #<umbrella> (run [<run_id>](<run_url>), <date>).
 ```
 
+**Upstream (second line)**
+
+```
+**Upstream:** LE-1234
+```
+
+The seam to the treatment layer. **This issue tracks the failure, not the fix** —
+what gets done about it is worked on the Jira board, so the card key has to be a
+field the body always carries. It is rendered as `_not filed_` when no card exists
+yet (the normal state at triage time, since the card follows the investigation),
+and the renderer never omits the line: an always-present slot is what makes the
+failure → issue → card walk possible, and what lets the link be swept for
+mechanically. Accepts a DataStax Jira key (`LE-####`) or a
+`langflow-ai/langflow#N` issue.
+
+Do **not** rely on the `REGRESSIONS.md` deliverable checkbox to carry this. That
+checkbox is an acceptance criterion — it records that a row is owed, not where the
+card is, and it vanishes from any sweep if nobody ticks it.
+
 **Sections**
 
 #### Symptom
--- a/.claude/skills/langflow-e2e-triage/references/issue-templates.md
+++ b/.claude/skills/langflow-e2e-triage/references/issue-templates.md
@@ -2,6 +2,27 @@
 
 This document defines the canonical template and rules for dedicated-issue triage in the `langflow-e2e-triage` skill.
 
+> **The body is rendered from code, not composed by hand.** `renderDedicatedIssueBody()`
+> and `assertDedicatedIssueBody()` in `scripts/lib/triage-core.mjs` own the structure;
+> Phase 7 calls them before `gh issue create`. This document explains *what each section
+> is for* and *what belongs in it* — the renderer enforces that the sections exist, that
+> every affected test carries a signature, and that at least one backticked spec path is
+> present. Editing the shape here without editing the renderer changes nothing.
+>
+> Why code and not prose: `gh issue create` bypasses `.github/ISSUE_TEMPLATE` entirely,
+> and the primary author of these issues is an LLM. A Markdown reference is advice; the
+> renderer is a guarantee. (The GitHub issue form at
+> `.github/ISSUE_TEMPLATE/failure-root-cause.yml` mirrors this same shape for humans
+> opening a cause issue by hand through the web UI.)
+
+## Scope — which issues this covers
+
+Two issues get opened around a red daily; only one is templated:
+
+- The **umbrella** is created by `daily-stable.yml` through the GitHub API
+  (`github.rest.issues.create`), which no template can govern. Not this.
+- The **dedicated per-cause issues** are created here, in Phase 7. These.
+
 ## Dedicated-Issue Template
 
 When opening a new issue to track a daily-failure cluster, use this structure:
@@ -44,7 +65,25 @@ Include:
 - Line number where the test fails
 - Human-readable test name (from `test()` or `test.describe()`)
 - The locator/selector being waited on
-- The exact error message or timeout signature
+- The `error_signature` **copied verbatim** from the run's row in
+  `reports/daily-history.jsonl`
+
+**The signature is copied, never written.** It is the key the same-signature
+recurrence rule matches on (`CONTRIBUTING.md` → *Monitoring rules driven by run
+history*), and `normalizeSignature()` only strips ANSI, collapses whitespace and
+lowercases — it does not understand paraphrase. So:
+
+- Copy the value out of the history row unedited. Do not shorten it, do not
+  reword it, do not merge two tests' signatures into one description.
+- If the run recorded `"unknown"` — which happens when the failure carried no
+  error message — write `unknown`. Substituting a description of what you think
+  happened makes the next run's occurrence unmatchable, and the flake rule then
+  reads a recurrence as a first occurrence.
+- ANSI codes are stripped by the renderer; a literal `|` is escaped so it cannot
+  break the table. Both are safe to leave in the value you copy.
+
+A paraphrased signature is the single most likely way per-cause issues quietly
+degrade back into one issue per day for the same cause.
 
 **Name every spec you are claiming, in backticked repo-relative form, on every
 `daily-failure` issue** — in this table at triage time, and in follow-up comments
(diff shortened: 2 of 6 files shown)


Victor-w-Madeira:
The outstanding limitation in *Known limitations* — the gate that would close the loop — is now tracked in #1035.

Worth flagging for review: the naive design I described there (a workflow on `issues.opened`) **does not work** for the agent path. The `execute` job runs `gh issue create` as `GITHUB_TOKEN`, and GitHub suppresses workflow runs for `GITHUB_TOKEN`-raised events — the same constraint that already forced `triage-dispatch.yml` onto a `workflow_run` trigger for the umbrella. An `issues.opened` gate would cover only hand-opened issues and never fire on the ones it exists for, while looking green. #1035 records that and the two triggers that do fire.

Victor-w-Madeira:
## Independent review — four findings, all fixed in `0adbd16`

Reviewed the diff by probing edge cases rather than reading the happy path.

| # | Finding | Severity |
|---|---|---|
| 1 | `run.date` unvalidated; provenance regex `\(run .+\)` accepted `(run 123, undefined)` — **shipped a broken join and passed validation** | high |
| 2 | Placeholder detector case-insensitive: `\bTODO\b` matched ordinary prose. The table *quotes* the test title rather than fencing it, so the code-span strip does not protect it — a test titled "todo list renders" **aborts issue creation** in the unattended Phase 7 path | high |
| 3 | `renderDedicatedIssueBody` did not guard `umbrella` while `renderDedicatedIssueTitle` did. `buildDataset` legitimately returns `umbrella_issue: null`, which rendered `#null` and surfaced as "malformed provenance line" — the wrong cause, unattended | medium |
| 4 | Validator checked shape, not content: every heading present-and-empty plus empty Signature cells passed clean. It is documented as covering hand-written and enriched bodies — precisely the ones that can be hollow — and an unchecked Signature column contradicts the PR's own premise | medium |

Findings 1 and 2 were the ones worth blocking on: one publishes wrong data, the other kills a triage run for a legitimate cluster.

Each fix carries a regression test named after the finding. Lane: 71 → 80. Re-verified against the real `2026-07-27` history row — validator clean.

Also folded in two cosmetics the review turned up: a `"` in a test title could close the quoted cell early, and the `eslint-disable-next-line no-control-regex` went dead when `ANSI_RE` moved to a `\u001b` escape (invisible in CI, since `npm run lint` only covers `tests/` — **no file in this PR is linted**).

PR #1036, opened 2026-07-28
ci(triage): gate the dedicated-issue contract outside the agent (#1035)

### LINKED ISSUE #1035: Gate the dedicated-issue contract outside the agent — note that issues.opened never fires for agent-created issues
## Context

#1032 / PR #1034 moved the dedicated-issue format from prose into code:
`renderDedicatedIssueBody()` builds the body and `assertDedicatedIssueBody()`
validates it, both in `.claude/skills/langflow-e2e-triage/scripts/lib/triage-core.mjs`,
with Phase 7 of the skill wired to call them.

That is a real improvement in kind — a deterministic enforcer instead of a format
description — but it is **not a hard gate**. Calling the renderer is still an
instruction in `SKILL.md`. An agent that hand-writes a body, or a human who opens a
cause issue without the form, can still create a malformed issue: no signature, a
spec named in prose without its path (invisible to the QA Platform's `tracked · #NNN`
join), a dropped `**Upstream:**` slot, a missing grouping argument.

The gate has to run **outside** the agent that writes the body. That is the whole
point: a check the author can skip is not a check.

## The trap: `issues.opened` does not fire for these issues

The obvious design — a workflow on `issues.opened` filtered to `daily-failure` —
does not work for the agent path, and would ship looking like it did.

The `execute` job in `.github/workflows/triage-dispatch.yml` runs the skill with
`GH_TOKEN: ${{ github.token }}`, so `gh issue create` acts as `GITHUB_TOKEN`.
**GitHub suppresses workflow runs for `GITHUB_TOKEN`-raised events.** The
`triage-dispatch.yml` header already documents this for the umbrella issue, which is
why that workflow uses `workflow_run` rather than an `issues` trigger.

So an `issues.opened` gate would silently cover only the hand-opened case and never
fire on the agent-created issues it exists for — the worst failure mode, since a
green-looking gate that never runs reads as "validated".

## Two paths that actually fire

Both are viable; pick one or combine.

1. **Post-step in the `execute` job.** After the skill returns, a deterministic step
   in the same job lists the `daily-failure` issues created in that run and pipes each
   body through `assertDedicatedIssueBody`. Runs outside the agent, in the same job, so
   there is no token problem and no event to suppress. Cheapest, and the verdict lands
   in the run log where triage is already read.
2. **`workflow_run` on Triage Dispatch completing.** `workflow_run` is exempt from the
   `GITHUB_TOKEN` suppression (same reason `triage-dispatch.yml` uses it). Sweeps the
   `daily-failure` issues opened since the run started and validates them. Decoupled
   from the skill's job, at the cost of a second run and a time-window query.

Keep a plain `issues.opened` trigger **in addition**, for issues opened by a human
through `.github/ISSUE_TEMPLATE/failure-root-cause.yml` — that path does fire normally.

## Must not fire on the umbrella

`daily-stable.yml` creates the umbrella with `['daily-failure', 'needs-triage']` and a
completely different body shape. Validating it against the dedicated-issue contract
would produce a permanent false failure every red day.

Discriminate on the title, which is already how the codebase tells them apart
(`matchUmbrella` / `findNewestUmbrella` in `triage-core.mjs`):

- umbrella — `[Daily Failure] ...`
- dedicated — `[Daily #<n>] ...`

## Scope

- A validation entrypoint that reads an issue body and exits non-zero with the problem
  list. `assertDedicatedIssueBody` already returns exactly that array; it needs a thin
  CLI wrapper (`scripts/`-style `.mjs`, unit-tested — note the lane now covers
  `.claude/skills/**` after PR #1034).
- Wiring per the paths above, with the umbrella excluded by title.
- **Report, do not block.** A malformed issue should get a comment naming the missing
  fields and a `needs-triage` (or a new `malformed`) label — not be closed or rejected.
  A half-filled cause issue still carries evidence someone needs; the gate's job is to
  make the gap visible and attributable, not to discard the report.
- Decide whether `issues.edited` re-validates. Leaning yes for the human path, so
  fixing the body clears the flag; watch for comment spam on repeated edits.

## Out of scope

- Enforcing the format on the umbrella issue (different contract, created via the API).
- The masked `signature_key` work and the `error_signature`-on-flakes producer
  asymmetry, both recorded in #1032.

## Intake

Off-wave exception (`qa-infra` + `follow-up`): direct follow-up of PR #1034, which
names this as its outstanding limitation.

Victor-w-Madeira:
## Type

---

- [x] **Exception** — off-wave work backed by an issue: a follow-up, a `daily-failure` triage issue, or a `community`-labeled regression. Issue #1035

---

## What this PR does

PR #1034 moved the dedicated-issue format from prose into code. But *calling* the renderer is still an instruction in `SKILL.md` — an agent that hand-writes a body, or a human who opens a cause issue without the form, bypasses it entirely. A check the author can skip is not a check.

This runs the same `assertDedicatedIssueBody()` against the issue **as it exists on GitHub**, from outside whoever wrote it.

### Two callers, because one trigger cannot cover both paths

The obvious design — a workflow on `issues.opened` — does not work for the agent path, and would ship looking like it did:

> `triage-dispatch.yml`'s `execute` job runs `gh issue create` as `GITHUB_TOKEN`, and **GitHub suppresses workflow runs for `GITHUB_TOKEN`-raised events**. That is the same constraint that already forced this workflow onto a `workflow_run` trigger for the umbrella. An `issues`-triggered gate would never fire for the issues it exists to gate — while showing green.

So:

- **Agent path** — an inline post-step of `triage-dispatch.yml`'s `execute` job, in `since` mode. Outside the agent, inside its job: no event to suppress, no token problem. `if: always()`, since a triage that died mid-way can still have created issues.
- **Human path** — `issue-contract-guard.yml` on `issues` `opened`/`edited`, which does fire because a user account raises it. `edited` closes the loop: fixing the body re-runs the check.

Shared logic lives in `.github/actions/guard-dedicated-issue`, following the repo's existing `.github/actions/` convention, so the two callers differ only in how they select issues.

### Reports, never blocks

A malformed issue gets a comment listing the missing fields plus `needs-triage` — never a closure. A half-filled cause issue still carries evidence someone needs; the gate's job is to make the gap visible and attributable, not to discard the report. The comment is upserted via an HTML marker, so editing the body updates it in place instead of appending another.

### The umbrella is excluded by title

`daily-stable.yml` opens the umbrella with the same `daily-failure` label but a completely different body. Enforcing the dedicated contract on it would produce a permanent false failure every red day. `classifyIssueTitle()` makes the split (`[Daily Failure] …` vs `[Daily #N] …`) testable and shares the semantics `matchUmbrella()` already uses.

### A verdict the guard cannot produce is not a pass

Found while exercising the shell: a crashed validator leaves an empty verdict, and the naive `.checked != true` branch reported that as a benign `skipped`. **A gate that silently passes when it is itself broken is the exact failure mode this issue exists to prevent.** It now emits `::error`, counts into `guard_error_count`, and exits non-zero.

---

## Tests covered

|---|---|---|
| 1 | `classifyIssueTitle separates umbrella, dedicated and everything else` | Both umbrella shapes `daily-stable.yml` actually emits are recognised, so the guard never enforces on them |
| 2 | `the umbrella is never validated against the dedicated contract` | The permanent-false-failure-every-red-day scenario |
| 3 | `an unrelated title carrying the label is left alone` | A manual note with the label is not a contract breach |
| 4 | `a compliant dedicated body passes` | No false positive on renderer output |
| 5 | `a hand-written dedicated body is caught` | The bypass this PR exists to close |
| 6 | `a body that drops the spec path is caught even when it looks complete` | The QA Platform's `tracked · #NNN` join cannot be silently lost |
| 7 | `renderVerdictComment is null when there is nothing to report` | No comment spam on healthy issues |
| 8 | `renderVerdictComment lists the problems and does not threaten closure` | Report-don't-block, and the upsert marker is present |
| 9 | `parseArgs reads the flags the workflow passes` | The action/CLI contract |

Lane: 80 → **89**.

---

## Dependencies

Depends on #1034 (merged) — `assertDedicatedIssueBody()` and the widened `test:scripts` lane that runs `.claude/skills/**` come from it.

---

- **Retroactive sweeps.** Only issues opened/edited from now on, and issues created by a triage run. Pre-existing malformed issues are not swept.
- **A dedicated `malformed` label.** Uses the existing `needs-triage` rather than creating a label without authorization. A dedicated one would sweep better — one line in `action.yml`.
- **Blocking.** By design: report only.

---

## Known limitations

**The workflow YAML has no local proof.** No YAML parser, `actionlint`, or `shellcheck` is installed in the repo, so structure is unvalidated locally — stated plainly rather than implied green. What *is* proven is the risky part: the shell logic, exercised against a stubbed `gh` (below).

**Final proof is the next run.** Human path: open any issue labelled `daily-failure` and watch `Issue Contract Guard`. Agent path: the next `pode abrir` on an umbrella, watching the `Guard — issues created by this run…` step in `triage-dispatch`.

**The `since` window uses runner clock minus one minute** for skew against the API. A triage that runs longer than the issue-list `--limit 100` window could in principle miss an issue; not realistic at current volumes.

---

## Related issue

Closes #1035

---

## Validation

- [x] `npm run test:scripts` — **89 passing** (was 80)
- [x] `npm run lint` — 0 errors; `npx eslint` on the new files — clean
- [x] `bash -n` on the extracted action shell — valid
- [x] **Shell logic exercised against a stubbed `gh`**, four scenarios:
  - sweep with one compliant + one malformed + the umbrella → `2 checked, 1 invalid`, umbrella untouched, comment+label only on the malformed one
  - guard comment already present → `PATCH` in place, no duplicate
  - broken validator → `::error`, exit 1, **zero mutations**
  - scratch files land in `mktemp -d`, not the checkout
- [x] CLI smoke-tested against a body rendered from the real `2026-07-27` history row
- [ ] Workflow YAML structure — **not validated locally**, see Known limitations


Victor-w-Madeira:
## Correction on the validation claim

I added a `workflow_dispatch` entry believing it would let the guard be proven from this branch before merge. **It does not.** GitHub requires a `workflow_dispatch` workflow to exist on the **default branch** before it can be dispatched at all:

```
HTTP 404: workflow issue-contract-guard.yml not found on the default branch
```

The `triage-dispatch.yml` precedent I reasoned from works because that file was already on `main` when *changes* to it were tested. A brand-new workflow has no such path.

So, plainly: **the event wiring in this PR cannot be validated before merge.** The workflow comment now says that instead of overstating it.

### What that leaves unproven, and the blast radius

Unproven: the two `on:` blocks firing, the label pre-filter, and the post-step's placement in `triage-dispatch.yml`. Everything downstream of "the step ran" is proven — the composite action's shell was exercised against a stubbed `gh` across six scenarios (sweep, umbrella exclusion, comment upsert, broken-validator fail-loud, dry-run, non-dry-run), and the validator itself has 89 unit tests.

The downside of merging and observing is bounded:

- `issue-contract-guard.yml` fails in isolation — nothing depends on it, and no other workflow reads its result.
- The `triage-dispatch` post-step is `continue-on-error: true`, so a broken guard cannot take the triage down with it.
- The guard is report-only. Worst realistic outcome is a comment that does not appear, not a wrong action taken on an issue.

### Post-merge verification

```bash
# 1. Human path — dry-run against a known-malformed issue, no mutation
gh workflow run issue-contract-guard.yml --repo oriontech-me/langflow-e2e \
  -f issue=<N> -f dry_run=true
# expect: '[dry-run] would comment and apply the needs-triage label'

# 2. Umbrella must be skipped, not flagged
gh workflow run issue-contract-guard.yml --repo oriontech-me/langflow-e2e \
  -f issue=<umbrella N> -f dry_run=true
# expect: 'skipped (umbrella)'
```

3. **Agent path** — watch the `Guard — issues created by this run…` step on the next `pode abrir`. Expect `Contract guard: N checked, 0 invalid, 0 unevaluable` when the skill used the renderer.

The dispatch entry still earns its place from the next change onward: every later edit to this guard *is* testable from its own branch, and it doubles as an on-demand re-check.

Victor-w-Madeira:
Deferred verification of the **agent path** is tracked in #1037 — it can only be exercised on the first `pode abrir` after this merges, and `continue-on-error: true` means a sweep that selects nothing looks identical to a clean run. The human path is verified immediately after merge via `workflow_dispatch` and needs no tracking.

PR #1042, opened 2026-07-28
fix: gate provider-hardcoded specs on provider health, not env-var presence

### LINKED ISSUE #1029: Provider-hardcoded specs gate on the env var, not on provider health — a drained key wedges the shard
## Problem

`collect-models` already records a provider as `inactive` in `providers.json` when its key is dead. Specs that are **parametrized** by provider honor that through `getProviderSkipReasons()`. Specs that **hardcode** a provider do not — they gate on the mere presence of the env var:

```ts
test.skip(!process?.env?.GOOGLE_API_KEY, "GOOGLE_API_KEY required to run this test");
```

(`language-model-regression.spec.ts:134` and `:217`)

A key that exists but is dead therefore runs the test. On run [30374528125](https://github.com/oriontech-me/langflow-e2e/actions/runs/30374528125) the Google key had exceeded its monthly spending cap — `collect-models` recorded `google` as `inactive (transient billing/quota)` — and the two Google tests in `language-model-regression.spec.ts` ran anyway. Each blocked a backend request past gunicorn's 300s timeout, killing the shard's single Langflow worker.

Six kill/restart cycles, ~313s apart (15:55:53, 16:01:06, 16:06:20, 16:11:33, 16:16:46, 16:21:59), left shard 2's backend effectively dead and produced **14 collateral failures** with `apiRequestContext.get: Timeout 20000ms exceeded` across unrelated specs (playground message edit and session nav, traces detail and delete-cascade, folder-crud, canvas-copy-paste). Shard 2 took 37.9 min against 12–13 min on the other three shards.

Correlating each wedge window with the tests in flight, `language-model-regression.spec.ts:130` / `:213` appear in nearly every one, at 260–322s per attempt.

## Why the existing pre-flight issues do not cover this

- #900 (build-probe the model class in `collect-models`) and #976 (resolve a live key before the suite) are both **pre-flight**. They make provider status more trustworthy; they do not make a spec consume it.
- The gap is spec-side: the status already exists in `providers.json` and the spec ignores it.

## Proposal

Make provider-hardcoded specs gate on provider **health**, not env-var presence — reading the same source the parametrized specs already read, so a provider recorded `inactive` produces a `test.skip` quoting the collected reason instead of a live call against a dead key.

Sweep the suite: `language-model-regression.spec.ts` is the one that bit us, but the `test.skip(!process.env.<PROVIDER>_API_KEY)` pattern must be audited everywhere a spec hardcodes a provider.

## Done when

- [ ] Provider-hardcoded specs skip when `providers.json` records the provider `inactive`, quoting the collected reason
- [ ] Every spec gating on raw `<PROVIDER>_API_KEY` presence is audited; each one either consumes provider health or documents why it must not
- [ ] Verified on a run where at least one provider is inactive: those specs skip and no shard shows a backend wedge attributable to them

## Notes

- Complementary to #900 / #976, not a substitute: they make the status trustworthy, this makes specs obey it.
- The blast radius of any request that *does* hang mid-run is tracked in #1030.
- Trigger of the day: the Google key hit its monthly spending cap between the 2026-07-27 and 2026-07-28 dailies — on 07-27 the Google-parametrized specs still ran normally with `gemini-2.5-flash`.

Victor-w-Madeira:
Closes #1029

## Problem

`collect-models` already records a provider as `inactive` in `providers.json` when its key is dead. Provider-**parametrized** specs honor that. Provider-**hardcoded** specs gated on the mere presence of the env var, so a key that exists but is drained still made the live call.

On run [30374528125](https://github.com/oriontech-me/langflow-e2e/actions/runs/30374528125) the Google key had exceeded its monthly spending cap. `models.json` mirrors the Langflow **catalog**, not the validation, so all 36 Google models were still listed, `resolveGeminiModel()` returned one, and the two Google tests in `language-model-regression.spec.ts` called it anyway. Each blocked a backend request past gunicorn's 300s timeout: six worker kill/restart cycles ~313s apart left shard 2's backend dead and produced **14 collateral timeouts** in specs that never touch Google.

## Approach

`tests/helpers/provider-setup/provider-health.ts` is the single implementation of the rule, reading the same `providers.json` the parametrized specs read.

- **Precedence:** missing env key first (more actionable — and for an unset key `collect-models` records only `"<KEY> not set"` anyway), then the recorded `inactive` reason, quoted verbatim into the skip.
- **Fails open** when `providers.json` is absent or unparseable. A fresh clone has no file (it is gitignored) and CI is allowed to run with a failed `Collect models` step (#980), so "no signal" must never skip the suite.
- `IGNORE_PROVIDER_HEALTH=1` overrides a stale local file (`collect-models` is not part of a targeted local run).
- The decision matrix is split from the I/O so it is unit-testable without a fixture on disk.

## The sweep

Every `test.skip` whose condition reads a provider env var was audited — **22 sites across 21 specs**. Twelve gates in nine specs now consume health:

| Spec | `@stable` | Provider |
|---|---|---|
| `language-model-regression` x3 | yes | openai / google / both |
| `rag-pipeline` (file-level) | yes | google |
| `settings-message-history` | yes | openai |
| `memory-history-regression` x2 | yes | openai |
| `general-bugs-agent-images-playground` | yes | openai |
| `chatInputOutputUser-shard-{0,1,2}` | no | openai |
| `general-bugs-agent-sum-duplicate-message-playground` | no | anthropic |
| `refresh-dropdown-list` | no | anthropic |

…and ten more that a **first, incomplete pass missed** (see *Audit method* below):

| Spec | `@stable` | Provider | Live call |
|---|---|---|---|
| `core-components/loop-component-regression` | yes | openai | **two sequential completions, 8-min budget** |
| `knowledge-ingestion-management/vector-store-index-query` | yes | **google** | file-level; embeds every chunk |
| `observability-monitoring/traces-detail-llm-span-populated` | yes | openai | *"A real LLM call is the whole point of this spec"* |
| `flow-functionality/freeze-path` | no | openai | three builds |
| `flow-functionality/general-bugs-shard-3836` | no | openai | one build |
| `flow-functionality/generalBugs-shard-1` | no | openai | one build |
| `flow-functionality/generalBugs-shard-10` | no | openai | two builds + playground reply |
| `flow-functionality/generalBugs-shard-3` | no | openai | playground send |

`vector-store-index-query` is the sharpest miss: **Google**, the provider whose spend cap wedged shard 2, and a direct sibling of `rag-pipeline.spec.ts` which the first pass did convert. `loop-component-regression` is worse than the case that motivated the issue — two hung calls under an 8-minute test timeout.

The switch test in `language-model-regression` drives two providers, so a dead key on either one gates it.

### Five gates deliberately keep env-var presence

Each documented in place, after checking what the test actually executes:

- `composio` — a tool provider, not an LLM provider; `collect-models` never probes it, so there is no health record to consume, and no completion is driven.
- `voice-assistant` — the enclosing `test.skip(...)` makes the body unreachable.
- `model-provider-api-key` — asserts the key-edit **surface**; a configured-but-drained key is a state it must still cover. No completion; its instance-side variable check is already stronger than env-var presence.
- `lock-flow` (`@stable`) — locks/unlocks a flow and deletes edges; never runs the model.
- `general-bugs-shard-3909` — asserts the run buttons *render*; never clicks one.

The last two read as **vestigial**: nothing in either test consumes `OPENAI_API_KEY`. Removing the gate would make them run where they currently skip — a behaviour change this sweep deliberately does not make.

`refresh-dropdown-list` gates on `ANTHROPIC_API_KEY` but configures the node with `initialGPTsetup` (**OpenAI**). The mismatch predates this PR; preserved and flagged in a comment rather than swapped silently, since the spec is unvalidated, carries no `@stable`, and a provider change there deserves its own review.

### Audit method

The first pass covered **12 of 22** sites: its grep was truncated with `head -60`, and a follow-up pass still missed one, because `process?.env?.X` (optional chaining) does not match a `process\.env` pattern. The complete set was found by parsing every `test.skip` condition rather than grepping lines. Re-verifiable:

```
providerSkipGate call sites: 22 across 19 spec files
env-var gates remaining:      4  (composio, voice-assistant, lock-flow, general-bugs-shard-3909)
```

## Evidence

Validated against a live Langflow (Nightly **1.12.0.dev7**) whose `providers.json` had `google: inactive` with the spend-cap error — the exact state of the run that motivated the issue — with `GOOGLE_API_KEY` present and confirmed configured by the preflight:

| Scenario | Result |
|---|---|
| `language-model-regression --grep Google` | **2 skipped** (previously ran and wedged the worker) |
| `rag-pipeline` (file-level gate) | **1 skipped** |
| `language-model-regression --grep "OpenAI provider"` — negative control | **1 passed (18.4s)**, real completion |
| `providers.json` removed | `readProviderHealth() = null` → gate returns RUN (fail-open) |

Zero flows touched in the hour after those runs — id-scoped cleanup leaked nothing.

Local gates: `typecheck` clean · `lint` 0 errors · `test:units` 61/61 (18 new) · `test:scripts` 89/89 · checklist coverage + guard · `regressions:check`.

```bash
npm run typecheck && npm run lint && npm run test:units && npm run test:scripts

npx playwright test tests/tests-automations/regression/core-functionality/llm-agents/language-model-regression.spec.ts \
  --grep "Google" --workers=1 --retries=0 --reporter=line   # → 2 skipped

npx playwright test tests/tests-automations/regression/core-functionality/llm-agents/language-model-regression.spec.ts \
  --grep "OpenAI provider" --workers=1 --retries=0 --reporter=line   # → 1 passed
```

## What only CI can prove

The issue's third *Done when* — verified on a run where at least one provider is inactive, with no shard wedge attributable to those specs — closes on the next `daily-stable`. If the Google cap is still in force, watch for: the Google tests reported **skipped** with the collected reason, no worker kill/restart cycle in the shard log, and none of the `apiRequestContext.get: Timeout 20000ms exceeded` collateral.

## Out of scope

Nineteen specs still carry an inline copy of `getProviderSkipReasons()`, plus a fourth hardcoded variant in `mcp-client-agent-gemini-tool-regression`. They already honor health correctly, so folding them into this helper is cleanup, not a fix — tracked separately.

### DIFF
--- a/.claude/skills/langflow-e2e-infra/references/failure-modes.md
+++ b/.claude/skills/langflow-e2e-infra/references/failure-modes.md
@@ -34,14 +34,18 @@ path). This skill never removes `@stable` itself.
 
 **Symptom:** whole provider's agent specs silently skip, or fail with 403; a
 single inaccessible lead model disables the provider; PR impacted-specs gate picks
-an inaccessible model; a raw API key probes OK but the model isn't Langflow-buildable.
+an inaccessible model; a raw API key probes OK but the model isn't Langflow-buildable;
+a spec runs a live call against a provider already recorded `inactive` and hangs
+the shard's worker.
 
 **Levers:** build-probe the provider's model class (not just the raw key) ·
 collect models **before** the impacted-specs gate · set `PREFLIGHT_SKIP_CREDENTIALS`
-on PR collect-models · resolve model via alias/settled, never a hardcoded dated id.
+on PR collect-models · resolve model via alias/settled, never a hardcoded dated id ·
+gate provider-hardcoded specs on recorded **health**, never on env-var presence
+(`providerSkipGate` in `tests/helpers/provider-setup/provider-health.ts`).
 
-**Docs:** `docs/collect-models.md`.
-**Issues/PRs:** #570 · #873 · #900 · #886 · #892 · PR #878 · PR #887 · PR #893 · PR #901.
+**Docs:** `docs/collect-models.md` (→ *Who consumes the recorded health*).
+**Issues/PRs:** #570 · #873 · #900 · #886 · #892 · #1029 · PR #878 · PR #887 · PR #893 · PR #901.
 
 ## 3. External-dependency hard-fail
 
--- a/docs/collect-models.md
+++ b/docs/collect-models.md
@@ -68,6 +68,28 @@ A provider with a key that genuinely fails its probe (e.g. a model the
 account cannot access) is a legitimate `inactive` — recorded, logged, not a
 test failure.
 
+### Who consumes the recorded health
+
+Writing `inactive` is only half the mechanism — a spec has to obey it. Two kinds
+of spec do:
+
+- **Provider-parametrized** specs (the `agent-*` family, `mcp-client-agent`) build
+  their target list from `models.json` and drop a target whose provider is
+  `inactive`, quoting the recorded reason.
+- **Provider-hardcoded** specs gate through
+  `providerSkipGate(...)` in `tests/helpers/provider-setup/provider-health.ts`
+  (#1029). That helper is the single implementation of the rule: missing env key
+  first, then the recorded `inactive` reason. It **fails open** when
+  `providers.json` is absent or unparseable — a fresh clone has no file (it is
+  gitignored) and CI is allowed to run with a failed `Collect models` step (#980),
+  so "no signal" must never skip the suite. `IGNORE_PROVIDER_HEALTH=1` overrides a
+  stale local file.
+
+Before #1029 the hardcoded specs gated on env-var presence, so a key that existed
+but was drained still made the live call. On run 30374528125 that hung two Google
+tests past gunicorn's 300s timeout, killed shard 2's Langflow worker six times, and
+produced 14 collateral timeouts in specs that never touch Google.
+
 ### Env-keyed provider must be ACTIVE — with one exception
 
 A provider whose key IS configured but that ends `inactive` silently
--- a/docs/core-components/loop-component-regression.md
+++ b/docs/core-components/loop-component-regression.md
@@ -41,7 +41,9 @@ If any of these tests fails, the Loop component is broken in the product: either
 5. Verify that `button_run_loop` is still accessible and `title-Loop` is still visible with a single node on the canvas
 
 **Test 3 — Research Translation Loop template: full wiring and iterates over 2 ArXiv papers**
-1. Skip the test if `OPENAI_API_KEY` is not set in the environment
+1. Skip the test when OpenAI cannot serve a live call — `OPENAI_API_KEY` unset, or
+   the provider recorded `inactive` in `providers.json`
+   (`providerSkipGate("openai")`, #1029)
 2. Navigate to "All Templates" and wait for the `template-research-translation-loop` card
 3. Click the template and wait for `title-Loop` to appear
 4. Verify that there are edges on the canvas (confirms template wiring)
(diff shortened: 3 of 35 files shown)


Victor-w-Madeira:
Follow-up for the out-of-scope cleanup: #1043

Victor-w-Madeira:
## Review pass — audit was incomplete, now closed

A skeptical re-read of this PR found the sweep covered **12 of 22** env-gated `test.skip` sites. The original audit grep was truncated with `head -60`, and a follow-up pass still missed one because `process?.env?.X` (optional chaining) does not match a `process\.env` pattern. Only parsing every `test.skip` condition found them all.

Ten more gates are now handled — eight converted, two documented. Four of the ten sit in `@stable` specs driving real completions, including:

- `vector-store-index-query` — **Google**, the same provider whose spend cap wedged shard 2, and a direct sibling of `rag-pipeline.spec.ts` which the first pass *did* convert
- `loop-component-regression` — two sequential completions under an 8-minute budget, a worse blast radius than the case that motivated #1029

Also corrected: seven specs carried a comment claiming the gate had to sit after the in-spec `dotenv.config()` "so a local `.env`-only key is seen". That is false — `playwright.config.ts:4` loads `.env` before any spec module and `dotenv` does not override already-set variables, so those in-spec calls are no-ops. The comment would have sent the next reader hunting for a bug that never existed.

The PR body above has been rewritten; its previous completeness claim was wrong.

Local gates re-run clean: typecheck · lint 0 errors · test:units 61/61 · test:scripts 89/89 · checklist coverage + guard. `vector-store-index-query` verified skipping against a live `providers.json` with `google: inactive` (2 skipped).

Still blocked on the same infra failure as before — #1044 ports the health gate that causes it.

PR #1049, opened 2026-07-29
test(freeze): prove the freeze cache deterministically and drop three inherited specs (#943)

### LINKED ISSUE #943: Validate: Freeze and State — freeze component, freeze path, unfreeze (§15.7)
Promote §15.7 Freeze and State `[-]` items.

**Items:**
- Freeze component
- Freeze path
- Unfreeze component

**Exit:** validated, spec doc, `@stable`, checklist flipped.

Roadmap: Wave 4.

rafaelgiln:
Closes #943. Last item of Wave 4.

Promotes the three `QA-CHECKLIST.md` **§15.7 Freeze and State** items into one consolidated spec, replacing three inherited files.

## What scouting corrected about the premise

**The product has ONE freeze affordance, and its semantics are path freeze.** Labelled `Freeze` (shortcut `F`), it marks the clicked node **and every upstream ancestor**. There is no separate "Freeze Path" entry. Freezing an upstream-most node freezes only itself — its path *is* itself. So the checklist's "Freeze component" and "Freeze path" are the **same control**, told apart by which node is clicked.

**Its location is not fixed.** From `nodeToolbarComponent/index.tsx` the two renders are mutually exclusive:

```tsx
{!hasToolMode && <ToolbarButton  … dataTestId="freeze-all-button-modal" />}
{ hasToolMode && <SelectItem value="freezeAll" data-testid="freeze-all-button-modal">}
```

| Component | tool mode | Freeze lives in |
|---|---|---|
| `Prompt Template` | has | **right-click menu** — its toolbar has 0 controls |
| `Language Model` | none | **selection toolbar** — its context menu has 7 entries, no Freeze |

Verified in both directions on `1.12.0.dev8`. The spec **asserts** the placement rather than hiding it behind a `.first()` that silently resolves to whichever mounted — which is what broke the first implementation attempt here.

## The cache proof, with no provider key

Freeze is an execution-cache control, so a spec that only checks an icon proves nothing. A `Prompt Template` run on its own returns its own template text, which makes the property observable deterministically:

| Step | `frozen` | run output |
|---|---|---|
| run with template **A** | `false` | **A** |
| change template to sentinel **B**, freeze, run | `true` | **A** — cached, B ignored |
| unfreeze, run | `false` | **B** — recomputed |

No `OPENAI_API_KEY`, no model call, ~20 s for the whole file. Every DOM assertion is paired with a persisted-state assertion polled from `GET /api/v1/flows/{id}` reading `data.node.frozen` for **both** nodes — autosave is debounced, and path semantics cannot be proven on the clicked node alone.

## Why the three inherited specs are deleted

| Deleted test | Why |
|---|---|
| `freeze.spec.ts` → *user must be able to freeze a component* | **Asserts the opposite of the product's behaviour and passes anyway.** Reproducing its exact scenario (legacy toggle, `Text Input`, toolbar freeze): the freeze *does* apply — `frozen: true` persisted, one `icon-Snowflake` — and the run returns the **cached** value, not the new one. It cannot notice, because it asserts `icon-FreezeAll`, the control's own icon, instead of the canvas `icon-Snowflake`. With a node frozen and no menu open, Snowflake is 1 and FreezeAll is **0** |
| `freeze-path.spec.ts` → *user must be able to freeze a path* | Logically the **correct** test and the only one asserting the caching property. Dropped on cost, not logic: needs `OPENAI_API_KEY`, ~20 s per run, `waitForTimeout(2000)`, `getByText("Freeze").first()`, no flow cleanup — and it **failed in the baseline batch while passing in isolation**. Its property is preserved deterministically here |
| `freeze-unfreeze-component.spec.ts` → *frozen component shows frozen indicator…* | Asserts only the CSS class `text-blue-500` on the button — no canvas indicator, no persisted state, no behaviour. Red on 1.12 |
| `freeze-unfreeze-component.spec.ts` → *freeze button toggles on each click* | Near-duplicate of the above, same CSS-only observable. Red on 1.12 |
| `freeze-unfreeze-component.spec.ts` → *frozen component can still be selected* | Tracks **no** §15.7 item — node selection is §15.4 |

Baseline on `1.12.0.dev8`: **3 failed, 2 passed**, **5 flows leaked** (none of the three files deleted what it created), no spec doc, and no `@stable` — so the daily never ran them and the redness was invisible.

## Product observation (no assertion depends on it)

The output panel renders `placeholder="common.empty"` — the raw i18n key instead of the resolved string. Cosmetic and outside §15.7, but it is a real localisation defect, and it is *why* the old `getByPlaceholder("Empty")` still matched: Playwright's placeholder matching is substring and case-insensitive, so `"Empty"` hits `"common.empty"` by accident.

## Validation

Langflow **1.12.0.dev8**, `--workers=1 --retries=0`.

```
run 1/3  expected=3 unexpected=0 flaky=0 backendErrors=false
run 2/3  expected=3 unexpected=0 flaky=0 backendErrors=false
run 3/3  expected=3 unexpected=0 flaky=0 backendErrors=false
typecheck=0 lint=0 · QA-CHECKLIST diff ok
```

**Force-fail** — every `test()`, one at a time, each verified red then reverted:

```
FF: frozen output asserted to equal the new sentinel instead of the cached value       → unexpected=1
FF: upstream node asserted as NOT frozen, denying the path semantics                   → unexpected=1
FF: post-unfreeze output asserted as the stale cached value instead of the recomputed  → unexpected=1
FF coverage: complete · no FF-MUTATION markers in diff ✓ · final green run ✓
```

**Cleanup:** the flow is deleted id-scoped in a describe-level `afterEach`; the instance ends at the same flow count it started with, and the 5 orphans from the baseline run were purged. The sidebar legacy toggle enabled during scouting was reverted.

## Reproduce

```bash
npx playwright test \
  tests/tests-automations/regression/flow-functionality/freeze-and-state.spec.ts \
  --workers=1 --retries=0
# expected: 3 passed, ~20s, no provider key required
```


rafaelgiln:
**Rebased onto `main` (`a113831`) — one deliberate conflict resolution.**

While this branch was open, `main` gained the #1029 provider-health sweep (`1437044`, `2e1e2ab`), which retrofitted `freeze-path.spec.ts` to gate on `providerSkipGate("openai")` instead of the bare `OPENAI_API_KEY` presence check. That produced a modify/delete conflict against this PR, which deletes the file.

**Resolved by keeping the deletion.** The sweep was a mechanical gate retrofit across 10 specs, not new intent for this spec, and its motivation — a drained OpenAI key blocking the backend past gunicorn's 300s timeout and killing the shard's Langflow worker — is an argument *for* removing this particular provider-dependent spec, whose caching property is now proven deterministically with no provider at all.

Re-validated after the rebase on `1.12.0.dev8`: **3 passed** (21.5s), typecheck 0 errors, lint 0 errors, QA-CHECKLIST guard and coverage guard both green, no flows leaked.

Issue #1040, opened 2026-07-28
Watch: components are migrating from lfx.components.* to per-vendor distributions (lfx-bundles) — shims are removed at M4


rafaelgiln:
## Summary

Langflow 1.12 is mid-migration: component families are moving out of `lfx.components.*` into **separate per-vendor distributions** (`lfx_openai`, `lfx_anthropic`, `lfx_google`, …) plus an aggregate `lfx-bundles` package. The old import paths still exist, but only as shims that **will be deleted**:

> *"Compatibility shim: lfx.components.groq moved to lfx-bundles. This module re-points to the installed bundle distribution. It contains no component implementations and no third-party dependencies, and **is removed once the deprecation window closes (M4)**."*
> — `lfx/components/groq/__init__.py` on `1.12.0.dev8`

This issue is a watch item: understand the migration's blast radius on this suite before M4 lands, not after.

## Measured on `1.12.0.dev8` (image built 2026-07-28T01:37)

**~75 component directories under `lfx/components/` are already shims:**

```
agentql aiml altk amazon anthropic apify assemblyai azure baidu bing chroma cleanlab
clickhouse cloudflare codeagents cohere cometapi composio confluence couchbase cuga
datastax deepseek elastic exa FAISS git glean google groq homeassistant huggingface
icosacomputing jigsawstack langwatch litellm lmstudio maritalk mem0 milvus mistral
mongodb needle notdiamond Notion novita nvidia olivya ollama openai openrouter
perplexity pgvector pinecone qdrant redis sambanova scrapegraph searchapi serpapi
spider supabase tavily twelvelabs unstructured upstash vectara vectorstores vertexai
vlmrun weaviate wikipedia wolframalpha xai yahoosearch youtube zep
```

**Still in core (non-shim):** `input_output`, `models_and_agents`, `processing`, `logic`, `flow_controls`, `helpers`, `data`, `embeddings`, `knowledge_bases`, `files_and_knowledge`, `files_ingestion`, `langchain_utilities`, `textsplitters`, `toolkits`, `tools`, `utilities`, `prototypes`, `custom_component`, and a few others.

[...]

Issue #1039, opened 2026-07-28
Groq/Mistral specs are @stable but skip on every daily — components are not bundled in the image (and the recorded cause is wrong)


rafaelgiln:
## Summary

`groq-provider.spec.ts` and `mistral-provider.spec.ts` carry `@stable`, run in the daily shard, and **skip on every run** — they have skipped continuously since PR #917 landed the availability gate. The Groq and MistralAI components are not present in `langflowai/langflow-nightly:latest`, and per a product decision **they will stay out of the image** (still supported, but shipped as an installable bundle rather than bundled by default).

Four `QA-CHECKLIST.md` bullets count these two specs as validated `[x]`, so the Coverage Summary and the generated `Phase 0 — Validated` block both credit coverage that never executes a single assertion.

Separately, the cause recorded in our code and issues is **wrong for the current nightly** — it is not a missing `langchain-*` package.

## Measured on `1.12.0.dev8` (image built 2026-07-28T01:37, freshest digest)

1. Installed `langchain-groq` into the container and restarted → **groq still absent** from `GET /api/v1/all`. The langchain package is not the gate.
2. `lfx/components/groq/__init__.py` is now a deprecation shim:
   > *"Compatibility shim: lfx.components.groq moved to lfx-bundles. […] This module re-points to the installed bundle distribution. It contains no component implementations […] and is removed once the deprecation window closes (M4)."*
3. The image installs ~20 **per-vendor distributions** — `lfx_openai`, `lfx_anthropic`, `lfx_google`, `lfx_ollama`, `lfx_amazon`, `lfx_azure`, `lfx_cohere`, `lfx_ibm`, `lfx_datastax`, `lfx_vllm`, `lfx_openai_compatible`, … There is no `lfx_groq` and no `lfx_mistral`.
4. `lfx-groq` / `lfx-mistral` do not exist in the package registry. `lfx-bundles==1.1.5` does.
5. Installing `lfx-bundles` + restart restores them: catalog goes from **3,077,041 bytes / 37 categories** to **4,313,696 bytes / 104 categories**, with `groq` and `mistral` present. (Reverted afterwards — the container is back to the exact baseline byte size with both packages `MISSING`.)

Ollama is present on dev8 because `lfx_ollama` ships in the image, not because `langchain-ollama` returned.

## Why not restore coverage with a bundle lane

[...]

Issue #1011, opened 2026-07-28
Collect models: 3 retries of a failing pre-flight starve the shared Langflow and abort every shard with 0 tests


Victor-w-Madeira:
The `Collect models` step and `Run @stable tests` share the **same Langflow service container** with no health gate between them, and the pre-flight step is allowed to retry 3× (CI default). When both LLM providers are down at once, those retries spend 7–12 minutes hammering the backend until gunicorn kills the hung worker — and the very next step's `globalSetup` preflight then fails on an unresponsive backend, aborting the shard **before a single test runs**.

Surfaced while triaging #1007 (daily run [30351107916](https://github.com/oriontech-me/langflow-e2e/actions/runs/30351107916), 2026-07-28). That run recorded `passed: 0, failed: 0, flaky: 0, skipped: 0` — **zero tests executed across all four shards**.

## The mechanism

1. `Collect models` (`.github/workflows/daily-stable.yml:311`) runs `collect-models.spec.ts` with `continue-on-error: true` — correct by design, a drained provider key must not kill the day.
2. It does **not** set `PLAYWRIGHT_RETRIES`, so it inherits the CI default of 2 (`playwright.config.ts:23`) → **3 attempts**.
3. Each attempt re-imports every provider key and re-walks the Model Providers UI. With Google *and* Anthropic both dead, `validateProviderWithFallback` (`tests/helpers/provider-setup/collect-models.ts:179`) exhausts 36 Google + 13 Anthropic candidates per attempt.
4. Langflow's gunicorn worker stops answering, gets `SIGKILL`ed and re-forked — twice per shard.
5. `Run @stable tests` (`daily-stable.yml:327`) starts and `globalSetup`'s reachability preflight burns its 120 s budget (`tests/globalSetup.ts:51`, throw at `:94`) on an unresponsive backend, exiting the shard with 0 tests.

## Evidence

Langflow Nightly **1.12.0.dev8**. All four shards, same signature:

```
critical  WORKER TIMEOUT (pid:37)                          [gunicorn.error]
error     Worker (pid:37) was sent SIGKILL! Perhaps out of memory?
```

```
Error: [preflight] Langflow backend at http://localhost:7860/ is not reachable after 120000ms
  (last: TimeoutError: apiRequestContext.get: Timeout 8000ms exceeded → GET /api/v1/version)
    at assertBackendHealthy (tests/globalSetup.ts:94:13)
```

[...]

Issue #922, opened 2026-07-24
CI service container omits LANGFLOW_WORKERS=1 → single-backend saturation hangs impacted-specs preflight (blocks #867)


rafaelgiln:
## Symptom
`pr-validation.yml` → **Run impacted E2E specs** fails at `globalSetup` preflight: `[preflight] Langflow backend at http://localhost:7860/ is not reachable after 30000ms` — a single `GET /api/v1/version` hangs for the full 30s. Reproduced **3× consecutively** on PR #921 (fix for #867). The container did NOT crash: its logs show Langflow serving `/api/v1/models` (collect-models: "updated model status: models affected") right up to the preflight window, then `/api/v1/version` stops responding.

Step order in the failing job:
- `Collect models` → **success** (backend alive, providers validated)
- `Run impacted specs` → **failure** (preflight can't reach `/api/v1/version` 30s later)

So the backend is alive but **saturated/blocked** after collect-models' model-status load, not dead.

## Root cause
The GitHub Actions **service container** runs the Langflow backend with the image default of `(2*cpu)+1` workers — each worker holds full in-memory state and, under the collect-models load on the runner, they contend until requests hang (documented in **#773**). The repo's own launch scripts already enforce the fix:

- `scripts/start-langflow-docker.sh` → `-e LANGFLOW_WORKERS="${LANGFLOW_WORKERS:-1}"`
- `scripts/start-langflow-pip.sh` → `--workers "${LANGFLOW_WORKERS:-1}"`

But the `services:` blocks in `.github/workflows/*` set **no** `LANGFLOW_WORKERS` (grep across `.github/workflows/` returns zero), so every service-container job runs multi-worker — contradicting the #773 requirement the scripts encode.

## Evidence
- Local backend with `LANGFLOW_WORKERS=1`: ran collect-models against all 3 providers (openai/anthropic/google), polled `/api/v1/version` throughout — **never blocked** (~10ms baseline held).
- CI multi-worker: `/api/v1/version` hangs 30s after collect-models → preflight throws.
- Classified as **single-backend saturation** (failure-modes.md §1, already lists #867).

## Fix (scope: pr-validation.yml only)
[...]

PR #1051, opened 2026-07-29
feat(provider-setup): build-probe a provider's component, not just its API key (#900)

### LINKED ISSUE #900: Harden collect-models: build-probe a provider's model class, not just the raw API key
Deferred hardening from #898 (upstream: LE-1974).

## Context

On the #879 daily, ~17 `@stable` specs failed only in their **Google** variant because the nightly image shipped without `langchain-google-genai`, so every Google chat/embedding model raised an `ImportError` at build. The failure surfaced downstream as a misleading node-build timeout, indistinguishable from saturation — a full triage cycle to attribute.

`collect-models` did not catch it: its provider probe calls the provider's own API directly (upstream of Langflow), so it validates the **raw API key**, not that Langflow can **build** the model class. A valid key + a missing server-side integration package = a false `active`.

## Proposed hardening

Make `collect-models` (or the pre-flight) verify Langflow can actually build a provider's model class, and downgrade the provider to `inactive` with an attributed error when it cannot — converting the deep per-spec timeout into a single, correctly-attributed failure at collect time (the existing "every env-keyed provider must be `active`" assert then fails loudly with the exact reason).

### Constraint (scouted in #898)

There is **no** standalone endpoint that triggers the model-class import — `/api/v1/models/providers` and `/api/v1/models?provider=` return static metadata only. A faithful probe must **build a Language Model flow** per provider (the `basic-prompting` fixture is reusable) and inspect the build error for the `ImportError: Could not import '...' ... Install the missing package` signature. This is a new mechanism on the critical `collect-models` gate, so it carries some added cost/flake surface — the reason it was deferred rather than shipped inline with #898.

## Done when

- [ ] `collect-models` (or pre-flight) builds a minimal Language Model flow per env-keyed provider and detects the import/build failure deterministically
- [ ] A provider whose model class cannot build is recorded `inactive` with the exact `ImportError` reason (not a false `active`)
- [ ] The check fails loudly (does not silently skip the provider's specs), keeping the daily-failure gate tripping when the environment is genuinely broken
- [ ] Documented in `docs/collect-models.md` (the current "does not cover" note is replaced by "now covered")

## Notes

- Light-touch note already landed (code comment in `tests/helpers/provider-setup/collect-models.ts` + `docs/collect-models.md` "does not cover") as the trap marker until this is built.
- Distinct from #773 (parallel-load saturation) — this is a deterministic package/build gap.

rafaelgiln:
**Closes #900.**

`collect-models` recorded a provider `active` from the raw-key probe alone. That probe calls the provider's own API **upstream of Langflow**, so a valid key on an image that cannot instantiate the component produced a false `active` — and the real failure surfaced tens of layers downstream as a generic node-build timeout. It cost a full triage cycle twice: #898 / LE-1974 (`langchain-google-genai`, ~17 Google `@stable` specs) and #907 / LE-1987 (`langchain-groq` / `langchain-mistralai`).

This adds a second, independent **build axis**.

## Problem

A provider is only usable when two independent things hold, and only one was checked:

| Axis | Question | Failure it catches |
|---|---|---|
| Key | does the provider's cloud API accept this key? | drained credit, revoked key, spend cap |
| **Build** | can *this Langflow image* instantiate the component? | the component's distribution or its `langchain-*` package is missing |

## The measurement that decided the design

The issue originally proposed a cheap catalog scan of `GET /api/v1/all`. **That is not sufficient**, and the reason is an upstream implementation detail that varies *per component*: where the `langchain-*` import sits. Read from the installed sources in the running container, 1.12.0.dev8:

| Component | `langchain-*` import site | Package missing ⇒ |
|---|---|---|
| `OpenAIModelComponent` / `OpenAIEmbeddingsComponent` | module level | absent from the registry |
| `GoogleGenerativeAIComponent` (via `lfx.base`) / `…Embeddings` | module level | absent from the registry |
| **`AnthropicModelComponent`** | **lazy, inside `build_model()`** | **registers normally, fails at BUILD** |

So the failure mode #1039 first hit on Groq is **not a Groq quirk** — among the three providers this spec covers, Anthropic has exactly that shape today.

Proven end to end rather than reasoned about. Hiding `site-packages/langchain_anthropic` and restarting the backend leaves `ext:anthropic:AnthropicModelComponent@official` **present** in `/api/v1/all` (the catalog layer returns *available* — a false positive) while the build layer reports:

```
No module named 'langchain_anthropic'. This flow needs a Python package that is
not installed in this environment.
```

A restart is required to observe it — the import is cached in `sys.modules`, so hiding the package under a warm worker changes nothing.

Also ruled out by measurement: `POST /api/v1/validate/code` does **not** attempt imports (`from langchain_nonexistent_pkg import Foo` returns `{"imports":{"errors":[]}}`), so it cannot serve as a cheap import probe.

## Fix

Two layers in `tests/helpers/provider-setup/probe-component-buildable.ts`:

1. **Catalog** — one `GET /api/v1/all` with an **exact** provider → component-key map. Replaces the substring token match (where `openai` also matched `OpenAI Compatible`) and skips the `component_display_names` pseudo-category.
2. **Build** — every component that survived layer 1, as disconnected vertices of **one** throwaway flow, deleted in a `finally`. A **credentials error is a PASS**: it proves the client class was imported and constructed. Only a packaging signature reproves.

**No paid call is possible.** Every `SecretStrInput` is neutralised (`value: ""`, `load_from_db: false`) before the build. Load-bearing, not hygiene: the registry template ships `api_key` as `{ value: "OPENAI_API_KEY", load_from_db: true }`, so an un-neutralised probe would load the global variable `collect-models` itself just saved, and the default `text_output` output would invoke the model for real. Measured with the neutralisation in place: all five components fail on missing credentials, none reaches the network.

**Its own assert, not a rider on the existing one.** Every build-axis reason is stamped `build axis: `, and a dedicated step asserts no provider carries one. The existing "every env-keyed provider is ACTIVE" check skips providers whose env key is unset — but an unbuildable component is a broken *image* regardless of whether anyone configured a key, so folding it in would let an unbuildable Anthropic pass in silence on a run without `ANTHROPIC_API_KEY`.

**Fails OPEN on its own infrastructure**, with a 60s build timeout against ~9s measured. The daily's `Collect models` pre-flight shares its container with the `@stable` shard and has a history of wedging it (#922/#927, #1011), so a probe timeout degrades to a warning — never to a blocked pre-flight or a false `inactive`.

## Scope note

Only the providers in `providerConfigMap` (openai, anthropic, google) are covered. Groq and Mistral are deliberately not bundled in the image (#1039) and keep their per-spec `isProviderComponentAvailable` gate. Keeping them out is what makes a *declared per-provider expectation map* unnecessary: every provider listed here is one the image is expected to ship, so "component missing" is unambiguously a failure.

## Validation (nightly `1.12.0.dev8`, `--retries=0 --workers=1`)

- 3 clean runs, no flake ✅ — `expected=1 unexpected=0 flaky=0 backendErrors=false` each
- `npm run typecheck` ✅ 0 errors · `npm run lint` ✅ 0 errors
- `npm run test:units` ✅ 16 pass / 0 fail (new `probe-component-buildable.test.ts`)
- QA-CHECKLIST guard ✅ (manual bullets only; generated blocks untouched)
- Zero `🚨 Backend Error` ✅ · 29 flows on the instance, **zero `build-probe` orphans** ✅
- Force-fail ✅ — two mutations, both reverted, final green run after revert:
  - build-axis step also flags `google` → `unexpected=1` (the new assert is not vacuous)
  - `PROVIDER_COMPONENTS.openai` declares a key the registry does not expose → `unexpected=1`, failing with the designed message and **not** downgraded to the billing warning:
    ```
    env-keyed provider(s) inactive for a NON-billing reason: openai — build axis:
    component distribution not installed in this Langflow image —
    ext:openai:ThisComponentDoesNotExist@official absent from GET /api/v1/all.
    Install the provider's lfx distribution (see #1040)
    ```

> Note: `1.12.0.dev9` was published during this work. Validation stayed on `dev8` — the local container has no volume, so upgrading destroys the instance DB (flows + provider global variables). Everything measured and asserted here is internally consistent on `dev8`.

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -46,6 +46,7 @@
 - [-] Google Generative AI Provider Setup → `helpers/provider-setup/setup-google.ts`
 - [-] Provider Map (`providerSetupMap`) — central registration point → `helpers/provider-setup/index.ts`
 - [-] Provider validation via API (credit, valid key) → `helpers/provider-setup/collect-models.ts`
+- [-] Provider build-axis probe (registry key present + component instantiates) → `helpers/provider-setup/probe-component-buildable.ts`
 - [-] Collection of available models via UI (Settings → Model Providers) → `helpers/provider-setup/collect-models.ts`
 - [-] `providers.json` — status of each provider (active/inactive + reason) → `data/providers.json`
 - [-] `models.json` — list of models per provider → `data/models.json`
@@ -384,6 +385,7 @@
 
 #### 7.1 Provider Collection and Validation
 - [x] Validate API keys of all providers via real call → `collect-models.spec.ts`
+- [x] Validate the running build can instantiate each provider's component (registry + build, not just the key) → `collect-models.spec.ts`
 - [x] Collect available models per provider via UI → `collect-models.spec.ts`
 - [x] Inactive providers appear as skipped in tests with reason → `agent-component-regression.spec.ts`
 - [x] Configure provider API key via Save Configuration (first setup) → `collect-models.spec.ts`
--- a/docs/collect-models.md
+++ b/docs/collect-models.md
@@ -9,9 +9,27 @@
 This is a utility spec — not a regression assertion test. It exists to populate two local data files used by LLM agent and model-provider specs as preconditions:
 
 - `tests/helpers/provider-setup/data/models.json` — list of models available per provider (collected from Settings → Model Providers UI)
-- `tests/helpers/provider-setup/data/providers.json` — provider status (`active` / `inactive`) validated via real API calls
+- `tests/helpers/provider-setup/data/providers.json` — provider status (`active` / `inactive`), validated on two independent axes: the raw API key works, **and** the running Langflow build can actually instantiate that provider's component
 
 If this spec is not run before the LLM agent specs, those specs fall back to a hardcoded model and may skip or fail due to missing provider configuration.
+
+### The two axes, and why the second one exists (#900)
+
+A provider is only usable when **both** hold, and they fail independently:
+
+| Axis | Question | Failure it catches |
+|---|---|---|
+| **Key** | does the provider's own cloud API accept this key? | drained credit, revoked key, spend cap |
+| **Build** | can *this Langflow image* instantiate the provider's component? | the component's distribution or its `langchain-*` runtime package is missing from the image |
+
+Before #900 only the first was checked. The key probe calls the provider's API
+**directly, upstream of Langflow**, so a perfectly valid key on an image that
+cannot build the model recorded a false `active` — and the real failure surfaced
+tens of layers downstream as a generic node-build timeout. It cost a full triage
+cycle each time it happened: #898 (`langchain-google-genai`, upstream LE-1974,
+~17 Google `@stable` specs) and #907 (`langchain-groq` / `langchain-mistralai`,
+upstream LE-1987). The build axis converts that into one attributed failure at
+collect time.
 
 ---
 
@@ -44,7 +62,18 @@ surface (the #505 lesson).
    (#570). It stops early on the first model that validates — or, when the SAME
    error repeats 3× in a row, on the conclusion that the error does not depend on
    the model at all (#1011; see Validation criterion).
-5. Write the provider status records to `data/providers.json`
+5. **Probe the build axis** (#900), in two layers, before the records are written:
+   a. **Catalog.** One `GET /api/v1/all` for every provider at once. Each provider
+      declares the exact component keys it needs (chat + embeddings); a key absent
+      from the registry means its distribution is not installed.
+   b. **Buildability.** Every component that passed (a) is placed in **one**
+      throwaway flow and built via `POST /api/v1/build/{flow_id}/flow`. The flow is
+      deleted in a `finally`. A registry hit does **not** prove the component can
+      build — see the contract table below.
+6. Write the provider status records to `data/providers.json`, merging both axes:
+   a build-axis failure records `inactive` with a reason that names the missing
+   **layer** (distribution vs. runtime package) and, when Langflow reports it, the
+   exact module
 
 ---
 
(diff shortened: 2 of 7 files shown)


rafaelgiln:
## CI verdict — the build axis now produces real signal, after two corrections it forced

The first CI run of this PR **failed**, and it exposed defects the local validation could not see. Recording the full chain because each step is a trap that is easy to reintroduce.

### Round 1 — green log, no signal

The job went red on the destructive lane (`backend not reachable after 120000ms`), while the specs themselves passed. Reading the log rather than the status showed the real problem:

```
⚠️ build-axis probe could not run (apiRequestContext.get: Timeout 60000ms exceeded)
```

…and yet the same run printed `build axis: ✅ openai / ✅ anthropic / ✅ google`. **Fail-open and real success produced identical output** — the probe had proven nothing and reported everything fine. That is the exact blind spot this issue exists to remove, reintroduced one layer up.

Three defects, all fixed in `12d9b64`:

| # | Defect | Fix |
|---|---|---|
| 1 | Fail-open looked like a pass | Third state `unknown`, logged `⚠️ … [NOT PROVEN]`, never written to `providers.json`, never overrides the key verdict |
| 2 | Client stopped waiting; server kept building → gunicorn `WORKER TIMEOUT` → backend unreachable for the next step | `POST /api/v1/build/{job_id}/cancel` on timeout |
| 3 | Batching all five in one request discarded the signal for *every* provider when one was slow, and could not name the slow one | One build per component via `stop_component_id`, 20 s each, 90 s ceiling |

### Round 2 — honest signal, still no signal

With the fixes in, CI reported `⚠️ NOT PROVEN` for all three providers, twice. Better — it no longer lied — but still useless. The detail that mattered: several timeouts were on the **`POST` that merely starts the build**. That is not a slow build; it is a backend that cannot accept the request.

### Round 3 — placement was the root cause

`collect-models` saves three provider keys through the Settings UI, and each save makes Langflow validate the provider and fetch its model list. On the single-worker CI backend that is enough load to wedge it — `pr-validation.yml` and `daily-stable.yml` **already carry a dedicated "Wait for the backend to recover from the collect-models load" step** immediately after this spec for exactly that reason (#922/#927/#1044).

The probe was running inside that load, before the recovery step. It consumes nothing the UI collection produces — not `models.json`, not the saved keys, only the component registry — so it now runs **first**, against an idle backend (`660f53a`).

Result, on CI, both the `Collect models` step and the in-suite run:

```
build axis: ✅ openai
build axis: ✅ anthropic
build axis: ✅ google
```

All five checks green, destructive lane included.

### On the two remaining `WORKER TIMEOUT`s — measured, not assumed

They are the ambient collect-models wedge, not this change. Run `30410211167` (PR #1042, the run quoted verbatim in #1048) executed `collect-models` **without** this probe and logged **4**; this run logs **2**. Comparisons against PRs that skip `Collect models` via the `needs_models` gate are void and were discarded.

### Two hypotheses tested and refuted rather than assumed

- **Cold-start imports.** A restarted worker still probes in 9.3 s locally.
- **The CI service container holding provider keys**, which would have made the blank-key build fall back to the env var and issue a real paid call. `pr-validation.yml` passes none to the service container, so no paid call was possible at any point.

Also measured and now recorded as external-dependency contract: `event_delivery` defaults to `polling`, and the events `GET` is a **long-poll** that blocks until the build finishes and returns every event at once. There is no incremental read to lean on — which is why the budget is enforced per component and paired with `cancel`.

Issue #576, opened 2026-07-08
Fix mcp-server.spec.ts: canvas-controls menu left open intercepts mcp-server-dropdown click


Victor-w-Madeira:
## Summary

`tests/tests-automations/regression/mcp/server/mcp-server.spec.ts:622`
(*"mcp server tools should be refreshed when editing a server"*, `@release`)
fails deterministically against `langflowai/langflow-nightly:latest`
(`1.11.0.dev36`) while opening the Add MCP Server modal:

```
TimeoutError: locator.click: Timeout 3000ms exceeded.
  - waiting for getByTestId('mcp-server-dropdown')
  - <html lang="en"> intercepts pointer events
```

## Root cause — test logic, not a product regression

Verified against a running nightly instance:

- The MCP-server testids are **intact and functional**: `mcp-server-dropdown`
  is present, visible and clickable, and clicking it surfaces the **"Add MCP
  Server"** option as expected.
- The click is blocked because the **`canvas_controls_dropdown` Radix menu is
  left open**. The failure snapshot shows `menu "116%" [active]` (Zoom In / Zoom
  Out / Zoom To Fit) still on screen; that menu's full-viewport overlay is what
  intercepts pointer events (`<html> intercepts pointer events`).

The offending sequence (lines ~643–650):

```ts
await page.getByTestId("canvas_controls_dropdown").click();          // 643 opens the zoom menu
await page.getByTestId("fit_view").click();                          // 645 menu item
await zoomOut(page, 3);                                              // 647
await page.getByTestId("canvas_controls_dropdown").click({ force: true }); // 648 toggle -> leaves menu OPEN
await openAddMcpServerModal(page);                                   // 650 mcp-server-dropdown click blocked by the open menu overlay
```

The line-648 toggle ends with the menu **open**, and the test never closes it
before interacting with the MCP node.

## Fix

Close the canvas-controls menu (e.g. `page.keyboard.press("Escape")` or a
click-away, and assert the menu is hidden) before `openAddMcpServerModal(page)`.
Also review the intent of the double `canvas_controls_dropdown` toggle
(lines 643 and 648) — it looks unbalanced.

## Scope / relationship to other work

- Independent of the `/flows` a11y refactor fix (#568). Discovered while
  validating #568; the two `flow-name-div` click sites migrated in that PR
  (lines ~788 and ~898) sit **after** this blocker, so the test cannot reach
  them until this is fixed. The selector migration itself is correct and matches
[...]

Issue #884, opened 2026-07-22
Pre-flight fail-fast gate: backend health, nightly version, required Langflow credentials & flags


rafaelgiln:
## Problem

When the environment is misconfigured (a provider credential not imported into Langflow, a stale/unhealthy backend, wrong nightly), specs fail **mid-run with a misleading signature** instead of an immediate, clear error. Concrete case root-caused in #880: `GOOGLE_API_KEY` present in env but not a Langflow global variable → KB ingest raises "embedding model no longer recognized" → surfaces only as a 90s `node_duration` timeout, **indistinguishable from a saturation timeout**. That misdirection cost a full triage cycle and inflates the apparent `transient-saturation` count.

## Proposal

Add a **pre-flight fail-fast gate** that runs before the suite (a global setup / smoke step):

- **Backend health + version:** assert the instance is up, reports the expected nightly (`/api/v1/version`), and warm it (one trivial build) so the first real spec doesn't eat cold-start latency.
- **Required credentials configured in Langflow** (not just in env): assert each provider key the run depends on is present as a global variable — generalizing the `assertEmbeddingCredentialConfigured` helper added in #880. Fail with an actionable message ("run collect-models") instead of a downstream timeout.
- **Feature flags:** verify `LANGFLOW_ALLOW_CUSTOM_COMPONENTS=true` and other required flags (the nightly ships some off — see #668), so flag drift fails fast, not deep in a spec.

## Why it helps the environment problem

It does not fix saturation, but it **removes the misclassification tax**: a misconfig fails instantly and unambiguously, so the remaining timeouts are *actually* saturation and triage stops chasing phantom product regressions. It also warms the backend, cutting first-spec latency.

## Relationship to other issues

- Generalizes the #880 credential precondition into a suite-wide gate.
- Sharpens the signal for #773 / #818 by keeping misconfig out of the `transient-saturation` bucket.

## Done when

- [ ] Global-setup pre-flight asserts backend health + resolved nightly version and warms it.
- [ ] Pre-flight asserts every run-required provider credential is a configured Langflow global variable, failing with an actionable message.
- [ ] Pre-flight asserts required feature flags; a misconfigured run fails in the gate, not mid-suite.

Issue #773, opened 2026-07-15
Investigate: broad agent/playground execution flakiness under the daily's parallel CI load (gated on clean baseline post-#772)


Victor-w-Madeira:
## Investigate: broad agent/playground execution flakiness under the daily's parallel CI load

Spun out of the #771 daily-failure triage (run [29406577514](https://github.com/oriontech-me/langflow-e2e/actions/runs/29406577514), 2026-07-15). This tracks the **transient** cluster of that run — distinct from the durable OpenAI-quota driver (#772).

### ⛔ Entry gate — do NOT start until there is a clean daily baseline

Right now the OpenAI-quota condition (#772) confounds this signal: the quota churn (collect-models probing 49 models that all 429, plus retries×2 on every OpenAI-bound spec) stretched this run from ~35 → ~47 min, and a saturated/longer run is itself a likely trigger for the transient timeouts below. **Investigation begins only after #772 is resolved and the first clean daily has run**, using that clean daily as the baseline. If the pattern does not survive the quota fix, close this issue as resolved-by-#772.

### Context — 3 consecutive mass-failure guard trips

| Date | Hard | Dominant driver (per triage) |
|---|---|---|
| 2026-07-13 (#704) | 12 | agent run-completion degradation + provider-count drift (#721) |
| 2026-07-14 (#744) | 23 | upstream sidebar/canvas restructure (#746) + traces fixture (#747) + 403 (#748) — fixed & restored |
| 2026-07-15 (#771) | 15 | OpenAI quota (#772) + this transient cluster |

The dominant drivers differ each day and two are already resolved — so the days are **not** the same root cause. What recurs underneath is a small, chronic set of agent/playground executions that fail under load and pass in isolation.

### The transient cluster (2026-07-15) — 10 hard failures, all non-quota

All share one shape: the agent/playground run did not complete in time, or returned empty/truncated output. Observed symptoms (from the run's JSON report):

- **No chat message rendered** (0 bubbles / timeout waiting for the bubble): `memory-history-regression` (0 of 2), `session isolation`, `agent-multi-tool-selection` (Web Search, anthropic), `ollama-provider`.
- **Truncated output** — model responded but the text was cut: `agent-system-prompt` (google) — sentinel `…ODB` instead of `…ODB5EA`.
- **Near-empty output**: `agent-max-tokens` causal control (google) — 6 words vs > 200 expected.
- **Session not persisted in time**: `agent-current-date-tool` toggle ON (anthropic).
[...]

PR #1052, opened 2026-07-29
fix(ui-ux): leave the canvas-controls menu closed, whoever opened it (#997)

### LINKED ISSUE #997: adjustScreenView: trailing dropdown guard is true on both paths — can leave the canvas-controls menu open (86 dependent specs)
`tests/helpers/ui/adjust-screen-view.ts` closes the canvas-controls dropdown behind a guard that is true in **both** branches, so on one path it *opens* a menu that was closed. Harmless on today's Langflow build, but an open canvas-controls menu is a documented click interceptor (#576), and the helper has **86 dependent specs**.

Surfaced by the adversarial review of PR #996 (which adds three of those dependents).

## The bug

```ts
let fitViewButton = await page.getByTestId("fit_view").count();

if (fitViewButton === 0) {
  await page.getByTestId("canvas_controls_dropdown").click();   // open it
  fitViewButton = await page.getByTestId("fit_view").count();   // ← reassigned, now > 0
}

await page.getByTestId("fit_view").click();
await page.waitForTimeout(500);
// ... zoom-out loop ...

if (fitViewButton > 0) {
  await page.getByTestId("canvas_controls_dropdown").click({ force: true, timeout: 1000 });
}
```

The trailing guard is clearly meant to express *"I opened the dropdown, so close it"*. But `fitViewButton` is **reassigned inside the `if`**, so it is `> 0` on both paths:

| `fit_view` initially | Dropdown opened? | Trailing click does | Correct? |
|---|---|---|---|
| hidden (`count === 0`) | yes | **closes** it | ✅ |
| visible (`count > 0`) | no | **opens** it | ❌ leaves an interceptor open |

The condition the code wants is captured *before* the reassignment — e.g. `const openedDropdown = fitViewButton === 0;` and then `if (openedDropdown) { ... }`.

## Why it is dormant, and why that is not reassuring

On Langflow Nightly **1.12.0.dev7** `fit_view` sits behind the dropdown, so every call takes the correct path and the helper's 86 dependents are fine. That is a property of the current UI, not an invariant. The day Langflow surfaces the fit-view control directly on the canvas, **every** `adjustScreenView` call starts leaving the canvas-controls menu open — and per #576 an open canvas-controls menu produces `<html> intercepts pointer events` failures on the next canvas click. That would be a suite-wide break arriving from an unrelated upstream UI change, with a confusing signature.

This is the same shape of latent defect as #988 (`setupPlayground`) and #867 (`expandFocusedNode`): a shared helper whose happy path is load- or version-dependent, with no test covering the branch that goes wrong.

## Secondary: the hard sleep

```ts
await page.getByTestId("fit_view").click();
await page.waitForTimeout(500);
```

A fixed 500 ms sleep in a helper called by 86 specs is both a latency tax and the kind of timing assumption `CONTRIBUTING.md` → *False positive anti-patterns* argues against. If there is an observable "canvas has finished fitting" state, gating on it is strictly better; if there is not, the sleep deserves a comment saying so. Worth addressing in the same pass, but it is not the bug above.

## Deliverables (Done when)

- [ ] The guard uses a boolean captured before the reassignment, so the dropdown is closed only when this helper opened it.
- [ ] The `count > 0` branch is exercised deliberately — e.g. a focused check that the canvas-controls menu is **closed** after `adjustScreenView` returns, so the regression cannot come back silently on a future UI change.
- [ ] The `waitForTimeout(500)` is either replaced by a state-based wait or documented as irreducible.
- [ ] Validated across a representative sample of the 86 dependents (canvas-heavy specs in `flow-functionality/` and `core-components/`) at `--workers=2` — a change to this helper reaches most of the suite.

## Verify the current state

```bash
grep -rl "adjustScreenView" tests/tests-automations/ | wc -l    # → 86
```

Read `tests/helpers/ui/adjust-screen-view.ts` alongside #576 for the interceptor history.

Victor-w-Madeira:
**Closes #997.**

## Problem

`adjustScreenView` closed the canvas-controls dropdown behind a guard that was
true on **both** paths, because `fitViewButton` is reassigned inside the `if`:

```ts
let fitViewButton = await page.getByTestId("fit_view").count();
if (fitViewButton === 0) {
  await page.getByTestId("canvas_controls_dropdown").click();
  fitViewButton = await page.getByTestId("fit_view").count();   // ← now > 0
}
...
if (fitViewButton > 0) { /* toggle the trigger */ }
```

An open canvas-controls menu is a documented click interceptor for the next
canvas action (#576), and this helper reaches **108 spec files** — see below.

### Blast radius

The issue's `grep` reported 86 dependents; that is now 65 after the spec
consolidations in #1046 / #1047 / #1049, of which **61 actually call the
helper**. But a direct `grep` under-counts, because four shared pieces call it
too — so specs that never import it still go through it:

- `tests/helpers/flows/setup-playground.ts`
- `tests/helpers/other/initialGPTsetup.ts`
- `tests/helpers/ui/prompt-template.ts`
- `tests/pages/SimpleAgentTemplatePage.ts`

That adds **47 spec files** invisible to any `adjustScreenView` search:

| Area | direct | indirect only |
|---|---|---|
| `core-functionality/llm-agents/` | 9 | **21** |
| `core-functionality/playground/` | 3 | **17** |
| `core-components/` | 15 | — |
| `flow-functionality/` | 14 | 3 |
| `ui-ux/` | 7 | 2 |
| `core-functionality/model-provider/` | 3 | 3 |
| `mcp/` (client + server) | 4 | 1 |
| `knowledge-ingestion-management/` | 2 | — |
| `templates/`, `project-management/`, `auth/`, `api/flows/` | 1 each | — |
| **Total** | **61** | **47** |

The indirect half is what makes this worth fixing carefully: a break here does
not arrive as 61 obvious canvas failures. It lands mostly in `llm-agents/` and
`playground/`, where a `<html> intercepts pointer events` failure reads as an
agent or provider problem — the misattribution class of #576.

## Fix

**Not the fix the issue proposes**, and that is the substance of this PR.

#997 proposes capturing the boolean before the reassignment and closing only
what the call opened. That is correct for the UI the issue anticipates — the day
fit-view is surfaced directly on the canvas — but it **regresses the build we run
today**. Read against Langflow's `CanvasControlsDropdown.tsx` and verified live
on Nightly 1.12.0.dev7:

| Fact | How it was verified |
|---|---|
| `fit_view`, `zoom_in`, `zoom_out`, `reset_zoom` are all inside `DropdownMenuContent` | source + live count |
| ⇒ `fit_view` is present **iff** the menu is open | count 0 closed / 1 open |
| The trigger carries Radix's `data-state="open"｜"closed"` | attribute dump off the live DOM |
| Clicking `fit_view` does **not** dismiss the menu | `data-state` stays `open` (plain `Button`, not `DropdownMenuItem`) |

So on today's build, "fit_view was already visible" means **the menu was already
open** — #576's leftover interceptor — and the current toggle happens to close
it. Closing only what we opened would hand that menu straight to the next canvas
click.

This PR encodes the **postcondition** instead of a toggle: *the canvas-controls
menu is closed on return, whoever opened it*, guarded on the trigger's real
`data-state`. That holds on both the current layout and the one the issue
anticipates. If `data-state` ever disappears, the helper falls back to intent —
it can then leave a pre-existing menu open, but it can never open one.

**Rejected alternatives:** the issue's boolean (fails the already-open case,
demonstrated below); pressing `Escape` (deselects nodes — too broad a side effect
for 108 dependents).

### Secondary: the 500 ms sleep

Replaced by a viewport-stability poll. Measured, not assumed: `fitView()` is
called with no `duration`, so the `.react-flow__viewport` transform reaches its
final value between the 20 ms and 47 ms samples after the click and never moves
again. Typical cost is now one extra poll (~50 ms); the 500 ms cap keeps the
worst case no slower than the sleep it replaces.

## Scope note

The zoom-out loop, its disabled-button break, the `numberOfZoomOut` option and
the 30 s canvas wait are unchanged. No spec, tag, doc or checklist bullet
changes — this is a helper fix, so there is no `QA-CHECKLIST.md` edit to make.

Three sibling helpers carry the same defect class and are **deliberately left
alone** (shared-file risk, one issue / one PR) — proposed as a follow-up:
`tests/helpers/ui/zoom-out.ts` (identical reassignment, 18 dependents),
`tests/helpers/filesystem/upload-file.ts` (opens and toggles unconditionally, 3
dependents), and `FlowEditorPage.adjustView()` (guards on the trigger's count,
which is always 1 — currently dead code).

### Conclusion

**No failure is attributable to this change**, and the one candidate that looked
like a regression was isolated and refuted with six runs. On the build we run
today the helper's observable behaviour is in fact **unchanged** — both paths
already ended with the menu closed — so what the diff buys is correctness on a
layout that has not shipped yet, ~450 ms per call, and a guard that fails 5/9
against the pre-fix helper so the bug cannot come back silently.

It does **not** claim a green suite: this machine is noisy, for reasons
identified below.

### Static gates

| Check | Result |
|---|---|
| `npm run lint` | 0 errors |
| `npm run test:units` | 73/73 |
| `npm run test:scripts` | 89/89 |
| `node --require ts-node/register --test tests/helpers/ui/adjust-screen-view.test.ts` | 9/9 pass |
| the same tests vs. the **pre-fix** helper | **5/9 fail** — the guards catch the regression |
| the same tests vs. the fix **as proposed in #997** | **3/9 fail** — including the already-open case |

Unit tests live next to the helper because **the branch that was wrong is one the
current UI cannot produce** — no E2E spec can reach it until Langflow moves the
control onto the canvas, at which point all 108 dependents break at once with a
`<html> intercepts pointer events` signature. The tests simulate both layouts.

### E2E

CI could not provide any: `pr-validation.yml` selects no specs for a helper
change (#1054) and `manual.yml` cannot execute a spec at all (#1055, discovered
by dispatching it for this PR). Run locally instead — the same 135-test scope the
dispatch was meant to cover (`@stable` AND `(?:@ui-ux|@workspace)`, 69 files),
twice: once on this branch, once with the helper reverted to `main`.

| 135 tests | passed | failed | skipped |
|---|---|---|---|
| pre-fix helper | 111 | **21** | 3 |
| this PR | 117 | **15** | 3 |

14 failures occur on both sides, 7 only without the fix, and **1 only with it**.

**The regression candidate, resolved.** `execution-error-notification.spec.ts`
failed only in the post-fix run. Isolated and re-run three times each way:
`3 passed / 0 failed` in all six. Deterministically green both ways, so it is a
flake under parallel load — consistent with where it failed, a hidden
`sidebar-search-input` at `setupChatFlow` line 54, which runs *before* both
`zoomOut` (line 68) and `adjustScreenView` (line 81).

The 21→15 delta is directionally positive but sits inside this box's noise; I
would not bank on it.

**Why the noise exists.** 8 of the 15 shared failures carry one signature —
`strict mode violation: getByTestId('add-component-button-chat-input') resolved
to 2 elements`, the second being `saved_componentsChat Input`. A saved component
named "Chat Input" had been left on the local instance by a failing
`saveComponents.spec.ts`; confirmed via the API and deleted. Worth a look beyond
this box: daily shards are file-level against one shared backend, so a leak there
breaks every later spec in the shard that uses the bare testid — the
cross-contamination class of #515.

Separately observed and **not** addressed here: `canvas-zoom-navigation.spec.ts`
fails 2–4 tests per run on this box against unmodified code, including Fit View
leaving the viewport at `scale(2)` (the max-zoom clamp). It is `@stable`, recent,
and does not call this helper — worth watching on the next daily.

### Reproduce

```bash
# regression guards
node --require ts-node/register --test tests/helpers/ui/adjust-screen-view.test.ts   # 9 pass
git stash push tests/helpers/ui/adjust-screen-view.ts
node --require ts-node/register --test tests/helpers/ui/adjust-screen-view.test.ts   # 5 fail
git stash pop

# the E2E scope (needs the nightly on localhost:7860)
PLAYWRIGHT_RETRIES=0 npx playwright test --workers=2 \
  --grep "(?=.*@stable)(?=.*(?:@ui-ux|@workspace))"
```

### DIFF
--- a/tests/helpers/ui/adjust-screen-view.test.ts
+++ b/tests/helpers/ui/adjust-screen-view.test.ts
@@ -1,8 +1,12 @@
 // Unit tests for adjustScreenView (issue #997).
 // Run with: npm run test:units
 //
-// Why this helper is unit-tested at all: it is called by ~65 spec files, and the
-// branch that used to be wrong is one the CURRENT UI cannot produce. On Nightly
+// Why this helper is unit-tested at all: 61 spec files call it directly and
+// another 47 reach it through `setupPlayground`, `initialGPTsetup`,
+// `prompt-template` and `SimpleAgentTemplatePage` — 108 in total, most of the
+// indirect half in `llm-agents/` and `playground/`, where a canvas interceptor
+// reads as an agent or provider failure. And the branch that used to be wrong is
+// one the CURRENT UI cannot produce. On Nightly
 // 1.12.0.dev7 every canvas control (`fit_view`, `zoom_out`, ...) is rendered
 // inside the dropdown's menu content, so `fit_view` is present if and only if
 // the menu is open — which is exactly why the pre-#997 bug was dormant. No E2E
@@ -25,10 +29,12 @@ import type { Page } from "@playwright/test";
 import { adjustScreenView } from "./adjust-screen-view";
 
 type Layout =
-  /** Today's build: the canvas controls live inside the dropdown menu. */
+  /** Today's build: every canvas control lives inside the dropdown menu. */
   | "in-dropdown"
-  /** Hypothetical build: the canvas controls sit directly on the canvas. */
-  | "on-canvas";
+  /** Hypothetical build: every canvas control sits directly on the canvas. */
+  | "on-canvas"
+  /** Hypothetical build, partial migration: only fit-view moved out. */
+  | "fit-view-on-canvas";
 
 interface FakeOptions {
   layout?: Layout;
@@ -36,6 +42,11 @@ interface FakeOptions {
   /** Set false to simulate a build whose trigger has no `data-state`. */
   exposesDataState?: boolean;
   zoomOutDisabled?: boolean;
+  /**
+   * Successive `.react-flow__viewport` style reads. The last entry repeats
+   * forever, so a single-element array models an already-settled viewport.
+   */
+  viewportStyles?: string[];
 }
 
 interface FakeCanvas {
@@ -44,6 +55,7 @@ interface FakeCanvas {
   readonly triggerClicks: number;
   readonly fitViewClicks: number;
   readonly zoomOutClicks: number;
+  readonly viewportReads: number;
 }
 
 function fakeCanvas({
@@ -51,22 +63,40 @@ function fakeCanvas({
   menuOpen = false,
   exposesDataState = true,
   zoomOutDisabled = false,
+  viewportStyles = ["transform: scale(1);"],
 }: FakeOptions = {}): FakeCanvas {
   const state = {
     menuOpen,
     triggerClicks: 0,
     fitViewClicks: 0,
     zoomOutClicks: 0,
+    viewportReads: 0,
   };
 
-  // A control is reachable while the menu is open — or always, once the
-  // controls are rendered on the canvas itself.
+  // A control is reachable while the menu is open, or once its own build has
+  // moved it onto the canvas.
+  const onCanvas = (testId: string) =>
+    layout === "on-canvas" ||
+    (layout === "fit-view-on-canvas" && testId === "fit_view");
+
   const controlCount = (testId: string) =>
     testId === "canvas_controls_dropdown" ||
-    layout === "on-canvas" ||
+    onCanvas(testId) ||
     state.menuOpen
       ? 1
       : 0;
+
+  // Playwright's locator methods do not resolve against a missing element: they
+  // reject once the timeout expires. Modelling that is what keeps the fake from
+  // flattering the helper — an `isDisabled` that quietly answers for an element
+  // that is not there would hide the partial-migration case entirely.
+  const requireRendered = (testId: string, method: string) => {
+    if (controlCount(testId) === 0) {
+      throw new Error(
+        `locator.${method}: Timeout 1000ms exceeded — ${testId} is not rendered`,
+      );
+    }
+  };
 
   const locator = (testId: string) => ({
     count: async () => controlCount(testId),
@@ -78,9 +108,7 @@ function fakeCanvas({
       }
       // Clicking a control must not dismiss the menu — these are plain buttons,
       // not Radix menu items (verified against the running nightly).
-      if (controlCount(testId) === 0) {
-        throw new Error(`clicked ${testId} while it was not rendered`);
-      }
+      requireRendered(testId, "click");
       if (testId === "fit_view") state.fitViewClicks += 1;
       if (testId === "zoom_out") state.zoomOutClicks += 1;
     },
@@ -91,16 +119,26 @@ function fakeCanvas({
       if (!exposesDataState) return null;
       return state.menuOpen ? "open" : "closed";
     },
-    isDisabled: async () => zoomOutDisabled,
+    isDisabled: async () => {
+      requireRendered(testId, "isDisabled");
+      return zoomOutDisabled;
+    },
   });
 
   const page = {
     waitForSelector: async () => undefined,
     waitForTimeout: async () => undefined,
     getByTestId: (testId: string) => locator(testId),
-    // The viewport-settle poll reads a stable value, so it returns on its
-    // second read — the shape the live measurement showed.
-    locator: () => ({ getAttribute: async () => "transform: scale(1);" }),
+    locator: () => ({
+      getAttribute: async () => {
+        const style =
+          viewportStyles[
+            Math.min(state.viewportReads, viewportStyles.length - 1)
+          ];
+        state.viewportReads += 1;
+        return style;
+      },
+    }),
   } as unknown as Page;
 
   return {
@@ -116,6 +154,9 @@ function fakeCanvas({
     },
     get zoomOutClicks() {
       return state.zoomOutClicks;
+    },
+    get viewportReads() {
+      return state.viewportReads;
     },
   };
 }
@@ -148,7 +189,7 @@ test("leaves a menu that was ALREADY open closed, without reopening it", async (
 test("never opens the menu when fit-view is reachable on the canvas", async () => {
   // The #997 bug. Pre-fix, `fitViewButton` was reassigned inside the `if`, so
   // the trailing guard was true on both paths and this call ended with an open
-  // canvas-controls menu — for all ~65 dependent specs at once.
+  // canvas-controls menu — for all 108 dependent specs at once.
   const canvas = fakeCanvas({ layout: "on-canvas", menuOpen: false });
 
   await adjustScreenView(canvas.page);
@@ -180,6 +221,62 @@ test("falls back to intent when the trigger exposes no data-state", async () => 
     "without a state signal it must still never OPEN a menu",
   );
   assert.equal(untouched.menuOpen, false);
+});
+
+test("throws, naming the cause, on a partial layout migration", async () => {
+  // fit-view moved onto the canvas, zoom-out left inside the menu. The helper
+  // has no reason to open a menu, so zoom-out is unreachable. It must NOT guess
+  // at the new layout, and it must not fall back to a blind toggle — it fails
+  // with a message that names the control and points at #997.
+  const canvas = fakeCanvas({ layout: "fit-view-on-canvas", menuOpen: false });
+
+  await assert.rejects(
+    () => adjustScreenView(canvas.page),
+    (err: Error) => {
+      assert.match(err.message, /adjustScreenView/);
+      assert.match(err.message, /zoom_out/);
+      assert.match(err.message, /#997/);
+      return true;
+    },
+  );
+
+  assert.equal(canvas.triggerClicks, 0, "it must not open a menu to recover");
+  assert.equal(canvas.menuOpen, false);
+});
+
+test("skips the zoom-out reachability check when no zoom-out was asked for", async () => {
+  const canvas = fakeCanvas({
+    layout: "fit-view-on-canvas",
+    menuOpen: false,
+  });
+
+  await adjustScreenView(canvas.page, { numberOfZoomOut: 0 });
+
+  assert.equal(canvas.fitViewClicks, 1);
+  assert.equal(canvas.zoomOutClicks, 0);
+  assert.equal(canvas.triggerClicks, 0);
+});
+
+test("keeps polling the viewport until the transform stops changing", async () => {
+  // Guards the settle loop itself: a poll that returned on its first read would
+  // pass every other test here, because they all model an already-stable
+  // viewport. Two moving reads, then a repeat — the earliest it may return is
+  // the read that matches its predecessor.
+  const canvas = fakeCanvas({
+    viewportStyles: [
+      "transform: translate(0px, 0px) scale(1);",
+      "transform: translate(-97px, -120px) scale(1.17);",
+      "transform: translate(-97px, -120px) scale(1.17);",
+    ],
+  });
+
+  await adjustScreenView(canvas.page, { numberOfZoomOut: 0 });
+
+  assert.equal(
+    canvas.viewportReads,
+    3,
+    "must not settle before two consecutive reads agree",
+  );
 });
 
 test("zooms out the requested number of times", async () => {
--- a/tests/helpers/ui/adjust-screen-view.test.ts
+++ b/tests/helpers/ui/adjust-screen-view.test.ts
@@ -1,0 +1,201 @@
+// Unit tests for adjustScreenView (issue #997).
+// Run with: npm run test:units
+//
+// Why this helper is unit-tested at all: it is called by ~65 spec files, and the
+// branch that used to be wrong is one the CURRENT UI cannot produce. On Nightly
+// 1.12.0.dev7 every canvas control (`fit_view`, `zoom_out`, ...) is rendered
+// inside the dropdown's menu content, so `fit_view` is present if and only if
+// the menu is open — which is exactly why the pre-#997 bug was dormant. No E2E
+// spec can reach the failing path until Langflow surfaces the fit-view control
+// directly on the canvas, at which point every dependent spec would start
+// leaving an open canvas-controls menu behind: a documented click interceptor
+// for the next canvas action (#576), arriving as a suite-wide break with a
+// confusing signature.
+//
+// So the widget is simulated here instead: a toggling trigger plus a switchable
+// layout, which lets both today's DOM and the hypothetical one be asserted.
+// The simulated contract was verified live against Nightly 1.12.0.dev7:
+//   - `canvas_controls_dropdown` carries Radix's data-state="open"|"closed";
+//   - `fit_view` count is 0 with the menu closed, 1 with it open;
+//   - clicking `fit_view` leaves the menu open (the zoom-out loop needs that);
+//   - `fitView()` is not animated — the viewport transform settles in a frame.
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import type { Page } from "@playwright/test";
+import { adjustScreenView } from "./adjust-screen-view";
+
+type Layout =
+  /** Today's build: the canvas controls live inside the dropdown menu. */
+  | "in-dropdown"
+  /** Hypothetical build: the canvas controls sit directly on the canvas. */
+  | "on-canvas";
+
+interface FakeOptions {
+  layout?: Layout;
+  menuOpen?: boolean;
+  /** Set false to simulate a build whose trigger has no `data-state`. */
+  exposesDataState?: boolean;
+  zoomOutDisabled?: boolean;
+}
+
+interface FakeCanvas {
+  page: Page;
+  readonly menuOpen: boolean;
+  readonly triggerClicks: number;
+  readonly fitViewClicks: number;
+  readonly zoomOutClicks: number;
+}
+
+function fakeCanvas({
+  layout = "in-dropdown",
+  menuOpen = false,
+  exposesDataState = true,
+  zoomOutDisabled = false,
+}: FakeOptions = {}): FakeCanvas {
+  const state = {
+    menuOpen,
+    triggerClicks: 0,
+    fitViewClicks: 0,
+    zoomOutClicks: 0,
+  };
+
+  // A control is reachable while the menu is open — or always, once the
+  // controls are rendered on the canvas itself.
+  const controlCount = (testId: string) =>
+    testId === "canvas_controls_dropdown" ||
+    layout === "on-canvas" ||
+    state.menuOpen
+      ? 1
+      : 0;
+
+  const locator = (testId: string) => ({
+    count: async () => controlCount(testId),
+    click: async () => {
+      if (testId === "canvas_controls_dropdown") {
+        state.triggerClicks += 1;
+        state.menuOpen = !state.menuOpen;
+        return;
+      }
+      // Clicking a control must not dismiss the menu — these are plain buttons,
+      // not Radix menu items (verified against the running nightly).
+      if (controlCount(testId) === 0) {
+        throw new Error(`clicked ${testId} while it was not rendered`);
+      }
+      if (testId === "fit_view") state.fitViewClicks += 1;
+      if (testId === "zoom_out") state.zoomOutClicks += 1;
+    },
+    getAttribute: async (name: string) => {
+      if (testId !== "canvas_controls_dropdown" || name !== "data-state") {
+        return null;
+      }
+      if (!exposesDataState) return null;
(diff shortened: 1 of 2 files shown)


Victor-w-Madeira:
## Review follow-ups applied — and the E2E gap this PR exposes

Second commit (`79a3983`) closes the findings from an independent review of the
first one.

### The claim that did not hold

The helper advertised that it works on "the layout the issue anticipates". That
was only true of **the menu**. Under a *partial* migration — fit-view moved onto
the canvas, zoom-out left inside the dropdown — the call has no reason to open a
menu, so `zoom_out` is unreachable and the loop died on a bare 1 s `isDisabled`
timeout naming nothing.

It now checks reachability up front and throws with the control, the layout
assumption and the issue number in the message. Guessing at a layout no build
has shipped would grow an untestable branch on a helper 108 specs depend on; a
loud, attributed failure is the point of #997, whose bug was dormant precisely
because it failed quietly.

**The unit test was hiding this**: its `on-canvas` layout moved every control at
once (the optimistic variant), and its fake answered `isDisabled` for elements
it was not rendering — something real Playwright never does. The fake now
rejects like the real locator, and a `fit-view-on-canvas` layout covers the
partial case.

### Also fixed

- **The settle poll's "keep polling while it moves" path was never exercised** —
  every fake returned a constant transform, so a poll that returned on its first
  read would have passed the whole file. The fake now replays a style sequence
  and a test pins the earliest read it may settle on.
- **`fitViewHidden` was doing double duty** as "did this call open the menu".
  They coincide only because the open immediately follows the probe, so the
  second question keeps its own variable now.
- `fit_view` / `zoom_out` extracted as constants alongside `CONTROLS`; comment
  counts corrected to the real reach (108, not the 65 a plain grep reports); and
  the viewport read's `catch` now states what it swallows (absent viewport *and*
  a strict-mode violation), since either silently degrades to the 500 ms sleep
  this replaced.

Unit tests are now **9**, still proven against both alternatives: **5/9 fail**
against the pre-fix helper, **3/9** against the fix as proposed in #997.
`typecheck` clean, `lint` 0 errors, `test:units` 73/73, `test:scripts` 89/89.

### ⚠️ This PR receives no E2E coverage from PR CI

`Run impacted E2E specs` is **skipping**, and that is not a fluke of this PR:
`detect-specs` selects work with

```
git diff --name-only ... -- ':(glob)tests/**/*.spec.ts'
```

— **changed specs only**. `scripts/impacted-tests.ts` maps *Langflow source
paths* to specs via the docs' `External dependencies` sections; it has no
`tests/helpers/**` → dependent-spec edge. So the highest-reach change this repo
can receive — a shared helper on 108 specs — passes PR validation without
executing a single spec.

That is a CI gap rather than a defect of this change, and it deserves its own
`qa-infra` issue. Covered here by hand instead:

**Manual E2E run: [30420649707](https://github.com/oriontech-me/langflow-e2e/actions/runs/30420649707)**
— `manual.yml` on this branch against `langflowai/langflow-nightly:latest`,
filtered to `@stable` AND `(?:@ui-ux|@workspace)`: **135 tests across 69 files**,
the canvas-heavy half of the helper's dependents. `@components` was left out
deliberately — `manual.yml` has no `collect-models` step, so its provider-backed
specs would fail on stale model data for reasons unrelated to this diff.

Victor-w-Madeira:
Follow-ups filed for the two items this PR deliberately leaves out:

- #1053 — the three remaining canvas-controls toggles (`zoom-out.ts` / 18 dependents, `upload-file.ts` / 2, `FlowEditorPage.adjustView()` / dead code), and extracting the postcondition helper this PR keeps private.
- #1054 — the CI gap: `detect-specs` selects nothing when a shared helper or Page Object changes, which is why `Run impacted E2E specs` is skipping on a diff that reaches 108 specs.

Victor-w-Madeira:
## Correction — the manual run did not cover anything

[30420649707](https://github.com/oriontech-me/langflow-e2e/actions/runs/30420649707) came back red having executed **zero tests**, so the coverage I claimed for it in the comment above does not exist. The cause is not this branch:

```
[preflight] backend healthy at http://localhost:7860/ — Langflow Nightly 1.12.0.dev9
Error: [preflight] provider key(s) set in the environment but NOT configured as a
Langflow global variable: OPENAI_API_KEY.
```

`manual.yml` passes `secrets.OPENAI_API_KEY` into `run-e2e` on every dispatch, has **no `collect-models` step** to import it into Langflow, and never sets `PREFLIGHT_SKIP_CREDENTIALS`. So the #884 credential gate fires in `globalSetup` before the first test, on both the main run and the destructive lane.

That gate landed 2026-07-22 (#885). `manual.yml`'s last dispatch was 2026-07-08. **No one has run it since the gate**, which is why this went unnoticed — the workflow is structurally unable to run a single spec whenever the secret is set, and no dispatch input can work around it.

Filed as its own issue rather than fixed here — see the next comment.

Re-running the same 135-test scope locally against nightly 1.12.0.dev7 in the meantime; I will post the result here. Until then the honest status of this PR's E2E evidence is: the 8-file sample from the PR description, plus the before/after baseline, and nothing from CI.

Victor-w-Madeira:
The `manual.yml` breakage is now https://github.com/oriontech-me/langflow-e2e/issues/1055.

Victor-w-Madeira:
## E2E evidence, run locally after #1055 blocked the CI path

Same 135-test scope the manual dispatch was meant to run (`@stable` AND
`(?:@ui-ux|@workspace)`, 69 files), against nightly **1.12.0.dev7**,
`--workers=2`, `--retries=0`, executed twice — once on this branch, once with the
helper reverted to `main`:

| 135 tests | passed | failed | skipped |
|---|---|---|---|
| pre-fix helper | 111 | **21** | 3 |
| this PR | 117 | **15** | 3 |

14 failures occur on both sides, 7 only without the fix, and **1 only with it**.

### The one regression candidate, resolved

`execution-error-notification.spec.ts › executing flow with server error shows
error feedback` failed only in the post-fix run. Isolated and re-run **three
times each way**:

```
WITH the fix:     3 passed / 0 failed   (×3)
WITHOUT the fix:  3 passed / 0 failed   (×3)
```

Deterministically green both ways, so it is a flake under parallel load, not a
regression. Consistent with where it failed: the timeout was
`sidebar-search-input` resolving **hidden**, at `setupChatFlow` line 54 — which
runs before `zoomOut` (line 68) and before `adjustScreenView` (line 81). The
failure is upstream of every line this PR touches.

### Why the rest of the noise exists

8 of the 15 shared failures carry one signature:

```
strict mode violation: getByTestId('add-component-button-chat-input') resolved to 2 elements:
  1) getByTestId('saved_componentsChat Input').getByTestId('add-component-button-chat-input')
  2) getByTestId('input_outputChat Input').getByTestId('add-component-button-chat-input')
```

A **saved component named "Chat Input"** was left on the local instance, so the
sidebar renders two entries with that testid. It came from `saveComponents.spec.ts`,
which failed in the same run with *"exactly one new saved component should be
created by the save"*. Confirmed via the API (`is_component` flows: 1 — "Chat
Input") and deleted.

That spec does have id-scoped `afterEach` cleanup and a `cleanupSavedChatInputs`
helper, and its line 105 comment already notes the testid becomes ambiguous when
a saved copy exists — the leftover survived anyway, which points at the *failing*
save path leaving one behind.

**Worth a look beyond this box:** daily shards are file-level against a single
shared backend, so if `saveComponents.spec.ts` leaks in a shard, every later spec
in that shard using the bare testid hits this violation. Same cross-spec
contamination class as #515. Not filed — flagging it for a decision.

### What this evidence does and does not say

It does **not** say the suite is green on this machine; it plainly is not. The
defensible claims are narrower:

- no failure is attributable to this change — the single candidate was isolated
  and refuted with 6 runs;
- the change is directionally positive (21 → 15), though that delta sits inside
  the noise this box produces and I would not bank on it;
- on the current build the helper's observable behaviour is unchanged anyway
  (both paths already ended with the menu closed) — the diff buys correctness on
  a layout that has not shipped yet, plus ~450 ms per call.

The clean signal still has to come from CI, and today it cannot: #1054 (the PR
gate selects nothing for a helper change) and #1055 (`manual.yml` cannot run a
spec at all) are both open. The next scheduled daily on `main` is the nearest
thing to a control.

rafaelgiln (approved):
## Approve

The substance of this PR is the decision to **not** implement the fix #997 asks
for, and that decision is right. Capturing the boolean before the reassignment
and closing only what the call opened regresses the already-open case — the
leftover interceptor from #576 — which today's toggle happens to handle. Encoding
the postcondition instead holds on both the current layout and the one the issue
anticipates. The reasoning in the description is sound and the rejected
alternatives are the right ones to reject (`Escape` deselects nodes; too broad for
108 dependents).

Scoping is disciplined: the three sibling helpers carrying the same defect class
are deferred to #1053 rather than swept in. I confirmed `zoom-out.ts` has the
identical reassignment and that `FlowEditorPage.adjustView()` guards on the
trigger's `count()`, which is always 1 — dead code, as described.

### Independent validation

Against `langflowai/langflow-nightly:latest` = **1.12.0.dev8** (one dev build
ahead of the description's dev7), `--workers=1 --retries=0`, on a 2-CPU / 4 GB
Colima VM. Scope: the 20 `@stable` tests across the 12 files in
`flow-functionality/` + `ui-ux/` that actually call the helper.

| Run | Result |
|---|---|
| PR branch, clean | **20/20 passed** (3.2 min) |
| PR branch, instrumented | **20/20 passed** (1.8 min) |

- `npm run typecheck` clean, `tsc --noEmit` exit 0 on the branch.
- The 9 unit tests pass; I re-ran them against the pre-fix helper and confirmed
  **5/9 fail**, so the guards do catch the regression.

A third run (with extra instrumentation waits) had 3 failures —
`publish-flow`, `langflowShortcuts`, `sidebar-add-component` — all from
`getAuthToken` / `Flow cleanup failed: 403 — No authentication credentials
provided`. The same three specs passed twice on the same code, so this is login
failing under contention on a small VM, not attributable to this branch.
Consistent with the noise described in the PR body and with #773.

### On the settle poll — measured, and it holds

`if (current !== null && current === previous) return;` accepts *any* two equal
reads, including two reads of the pre-`fitView` transform, so in principle it can
return before the click has applied. I checked whether that actually happens
rather than assuming: instrumented the helper to log, on every return that landed
on the pre-click value, whether the transform moved during the following 400 ms.
Over the 20-test slice:

```
CONFIRMED-EARLY (returned before fitView applied): 0
NOOP  (pre-click value was already final):         2
OK    (settled on a changed transform):           12
```

Both pre-click-value returns were genuine `fitView()` no-ops — the canvas was
already fitted, so returning on that value is correct. Zero early returns.

Worth recording why the obvious hardening is **not** worth doing: excluding the
pre-click value as a settle candidate would force the loop to burn the full
500 ms budget on exactly those no-op calls, which are 2 of 16 here. That trades a
real 500 ms on ~12% of calls for protection against a race with no observed
occurrences. The poll is fine as written.

The one thing I would add is a line in the doc comment noting that the two-equal-
reads rule is load-bearing on `fitView()` being unanimated — so that if a future
build ever animates it, the reader knows this is where it breaks.

### Non-blocking notes

1. **The `data-state` fallback degrades silently into the fix this PR argues
   against.** If the attribute ever moves off the element carrying the testid,
   `getAttribute` returns `null` and the helper falls back to `openedByThisCall`
   — i.e. exactly #997's boolean, re-introducing the already-open leak with no
   signal. The unit tests can't catch it (they simulate the contract). A
   `console.warn` in the `state === null` branch would make that visible.
2. **The `throw` path does not close a pre-existing open menu** — no
   `try`/`finally`. Narrow (needs `zoom_out`'s testid to disappear while the menu
   is open) and the test fails anyway, but it is the one path that contradicts the
   postcondition the helper is named for.
3. **The fake's `page.locator()` ignores its selector** — it answers for any
   argument, so a typo in `.react-flow__viewport` would pass all 9 tests while
   silently degrading to the 500 ms path in production.
4. **Doc nit**: "the 500 ms cap keeps the worst case no slower than the sleep it
   replaces" is not strict — a slow read (200 ms timeout) starting near the
   deadline overshoots to ~750 ms.

### Out of scope, recorded here

- **A fully green run leaked 2 flows.** Run 1 above passed 20/20 and still left
  `New Flow` and `Basic Prompting` behind; the failing run left `New Flow (1)` and
  `New Flow (2)` when cleanup hit the 403. Four orphans total, purged id-scoped.
  Not caused by this PR — the helper creates nothing — but it means green is not
  evidence of cleanup in this slice.
- **Every test logs `500 /api/v1/store/tags` → `[Errno -2] Name or service not
  known`** (the container cannot resolve the Langflow Store). The fixture reports
  `⚠️ 3 HTTP error(s) detected` and does not fail the test. If that is not a
  deliberate allowlist, it is noise that would mask a real backend error.
- Reproducing this PR's evidence required the nightly, which
  `scripts/start-langflow-docker.sh` cannot start — filed as #1076. I started the
  container by hand.

Victor-w-Madeira:
## Review follow-ups applied (`2b71794`)

Thanks @rafaelgiln — the instrumented measurement of the settle poll is the part
I had not done, and the note about the two-equal-reads rule is the one that
changes the file most usefully.

All four notes are in. None changes behaviour on the current build; all four
close a way this fix could erode without a signal.

**1. The `data-state` fallback no longer degrades silently.** You are right that
it decays into exactly #997's boolean, and right that the unit tests cannot catch
it — the test for that branch *asserts* the fallback as acceptable. It now warns,
naming the control, the consequence and the issue. Two additions beyond the
`console.warn` you suggested: the warning is **pinned by a test** (it is now the
only signal, so it should not be able to disappear silently either), and a second
test asserts the normal path stays quiet — a warning that cries wolf on 108 specs
is worth nothing.

**2. The fake now checks its selector.** Same class as the `requireRendered` fix
from the previous round, sitting in the one place I missed. Verified by mutation
rather than by inspection:

```
VIEWPORT mistyped .react-flow__viewprot  → "keeps polling the viewport…" FAILS
console.warn stubbed out                 → "falls back to intent…"      FAILS
```

**3. The settle-poll doc comment was worse than incomplete — it was misleading.**
It presented animation as a cost ("the 500 ms cap keeps the worst case no
slower"), when your point is that the exposure is *correctness*: two 50 ms
samples can coincide mid-movement and the poll returns on a transform still
travelling. The cap bounds how long it waits, not whether what it settled on is
final. Recorded in caps, with what to do instead (stable read over a minimum
span, or React Flow's own transition end).

**4. Corrected** — ~750 ms worst case, as you calculated. Note that the same
inaccurate sentence is in this PR's description under *Secondary: the 500 ms
sleep*; treat the doc comment as the corrected version.

### Note 2 (`try`/`finally`) — deliberately not done

Each test gets a fresh page context, so a menu left open when the helper throws
intercepts nothing: the postcondition only binds actions later in the same test,
and after a throw there are none. Weighed against a cleanup path that can mask
the original error, the value is ~zero. It belongs in #1053, where the
postcondition becomes an extracted helper — added there.

### On your out-of-scope items

- **The `store/tags` 500s are not an allowlist.** Checked `tests/fixtures/fixtures.ts:246`:
  only `flow_error` throws. HTTP errors are `console.log` and nothing else, for
  **any** 4xx/5xx — so there is no allowlist to be deliberate about, and the
  `🚨 Backend Error:` line is purely advisory. That makes checklist item 4
  ("confirm no backend errors logged") a human read of a log the Store noise
  pollutes. Worth its own issue; I will file it.
- **The 2 leaked flows on a green 20/20** is the sharper observation of the two —
  green is not evidence of cleanup. Same class as #515.
- **#1076 is already fixed** (`4e4cc5c`, merged to `main` today), so
  `start-langflow-docker.sh` starts the nightly now — no need to do it by hand
  next time.

### The CI evidence gap has partly closed since the description was written

- **#1055 is closed** and `manual.yml` ran green on another branch today, so the
  135-test scope can finally run in CI on this branch instead of on my noisy box.
- **#1054 is still open** — the PR gate still selects zero specs for a helper
  change, so `Run impacted E2E specs` will keep skipping here.
- The red `E2E via Docker (latest)` at the top of this PR is run
  [30420649707](https://github.com/oriontech-me/langflow-e2e/actions/runs/30420649707)
  from 03:50 — the #1055 zero-test abort already corrected in the thread above,
  not a new failure.

Validation after the follow-ups: unit tests **10** (was 9), `test:units` 74/74,
`test:scripts` 89/89, `typecheck` clean, `lint` 0 errors (3 pre-existing warnings
on this file, unchanged).

Victor-w-Madeira:
Filed as #1084 — the fixture's HTTP error monitor cannot fail a test, and its filter is four exact status codes (400/404/422/500) rather than the `4xx/5xx` both docs claim, so `403`, `502` and `503` are invisible to it. Your `store/tags` observation is the example that exposed it, not the thing being fixed.

Issue #1065, opened 2026-07-29
Triage protocol: drop the "umbrella stays open on a guard-tripped day" clause — history, not the issue, is the standing record


Victor-w-Madeira:
## The rule

`CONTRIBUTING.md` states, in two places, that a triage of a guard-tripped daily must **not** close the umbrella issue:

> Because the collateral has no dedicated issue, the **umbrella stays open** on a guard-tripped run as the standing record; the next clean, non-guarded daily's triage closes it once it confirms recovery, or promotes any cluster that persists into a durable issue. *(line 698)*

> …because that day's today-only collateral is **noted rather than filed**, the triage issue **stays open** as the standing record until a clean, non-guarded daily rechecks it… *(line 663)*

The triage skill mirrors it in `SKILL.md` (Phase 3 and Phase 7 step 5) and in `references/issue-templates.md` (*Guard-Tripped Rule*, bullet 0).

## Why it should go

**Its premise does not hold.** The clause exists so that today-only collateral is not orphaned when it gets no dedicated issue. But the collateral is not carried by the umbrella — it is carried by `reports/daily-history.jsonl`, which is append-only, committed to `main` on every run, and is the *only* input to recurrence detection. `buildTriageDataset` recomputes recurrence on every triage via `rowsWithinDays(..., 30)` + `computeRecurrence()`, reading history and nothing else. If a collateral cluster persists, the next triage sees it as same-signature recurrent and files it — whether or not an umbrella was left open. Nothing is orphaned.

**It costs the next triage.** Phase 2 dedup runs `gh issue list --state open --label daily-failure` and matches every candidate cluster against the result. Every umbrella left open is a permanent extra row in that list, carrying a body full of spec names that no longer correspond to anything actionable. The rule makes each guard-tripped day slightly degrade every triage after it — and guard-tripped days are not rare (of the 12 dailies from 07-13 to 07-29, 7 crossed the threshold).

**It is not needed for discovery.** `matchUmbrella()` keys on the run id inside the body, and `findNewestUmbrella()` on the date in the title — neither depends on prior umbrellas being open or closed.

**It contradicts the paragraph it sits in.** The same passage (line 663) opens by defining the triage issue as *"the analyst's inbox and dispatcher — not the tracking issue for each problem. Its only deliverable is the triage itself … and then close the triage issue."* A standing record of unfiled collateral is exactly the tracking role the sentence before it rules out.

## Proposed change

[...]

Issue #1063, opened 2026-07-29
[Daily #1057] sidebar-search-input never becomes visible after entering a blank flow — setup timeout across 4 specs (recurrent)


Victor-w-Madeira:
Spun out of daily-failure triage #1057 (run [30444299314](https://github.com/oriontech-me/langflow-e2e/actions/runs/30444299314), 2026-07-29).

**Upstream:** _not filed_

> **Corrected 2026-07-30 from the run artifacts.** The original body attributed this
> flake to the error-feedback selector and claimed a single test with no grouping. The
> JSON reports of all three runs show the timeout is in **`setupChatFlow`**, waiting for
> the sidebar search input to become visible — before the error is ever provoked — and
> that three further specs hit the same condition. Symptom, grouping, preliminary read
> and investigation directive were rewritten accordingly; see the correction comment for
> the raw call logs.

## Symptom

Four `@stable` specs flaked (recovered on retry) with the same shape — **`[data-testid="sidebar-search-input"]` is present in the DOM but never becomes visible**, so the sidebar-entry step of the setup times out. In every case the failure is in the setup that enters a blank flow and filters the sidebar, before the test's own subject is exercised.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/ui-ux/execution-error-notification.spec.ts:54` ("executing flow with server error shows error feedback", declared at :139) | `waitForSelector('[data-testid="sidebar-search-input"]')` — state `visible`, 30 s | `TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.` |
| `tests/tests-automations/regression/core-functionality/playground/playground-clear-history.spec.ts:43` ("clear chat on Default session must remove messages but keep the session", declared at :110) | `getByTestId('sidebar-search-input').fill(...)` — 20 s | `TimeoutError: locator.fill: Timeout 20000ms exceeded.` |
| `tests/tests-automations/regression/core-components/api-request-component-regression.spec.ts:100` ("API Request component — non-2xx HTTP response propagates status_code without crashing", declared at :560) | `getByTestId('sidebar-search-input').fill(...)` — 20 s | `TimeoutError: locator.fill: Timeout 20000ms exceeded.` |
[...]

Issue #1059, opened 2026-07-29
[Daily #1057] agent-max-tokens: causal control — unset max_tokens fails toHaveCount (recurrent same-signature, non-wedge)


Victor-w-Madeira:
Spun out of daily-failure triage #1057 (run [30444299314](https://github.com/oriontech-me/langflow-e2e/actions/runs/30444299314), 2026-07-29).

**Upstream:** _not filed_

## Symptom

The `causal control — unset max_tokens generates freely` test hard-failed on `anthropic / claude-sonnet-5` after all 3 attempts, on a `toHaveCount` assertion.

It is the only hard failure of the day whose signature is **not** the `GET /api/v1/auto_login` timeout that accounts for 6 of the other 9, which is why it is filed rather than left as collateral.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests-automations/regression/core-functionality/llm-agents/agent-max-tokens.spec.ts:305` ("causal control — unset max_tokens generates freely") | `the expected number of rendered message elements` | `Error: expect(locator).toHaveCount(expected) failed` |

## Why these failures are one cause

Single test, single signature — no grouping claim is being made. It is filed separately precisely because its signature is distinct from the day's dominant `auto_login` timeout cluster (tracked on #1030), so folding it in there would attribute it to a cause the evidence does not support.

## Preliminary read (descriptive — NOT a verdict)

This surfaced on a **mass-failure day** (10 hard + 12 flaky, guard threshold 5 crossed). The triage verdict on the day itself is that it **was environmental**: 6 of the 10 hard failures are a 20 s timeout on `GET /api/v1/auto_login` in shards where every provider was green and the post-`collect-models` health gate had already passed (tracked on #1030), and two further shards executed zero tests after a credentials-preflight abort (#1058).

Because the day is judged environmental, **no hard failure was manually quarantined** and `@stable` is still in place on this test — the guard had already suppressed the automatic removal. Quarantine it only if it reproduces on a clean, non-guarded daily. Four **recurrent flakes** from the same run were quarantined separately (#1060, #1061, #1062, #1063): the guard suppresses automatic removal for hard failures, it does not exempt a recurrent flake from the quarantine criterion.

[...]

Issue #1058, opened 2026-07-29
[Daily #1057] a shard that executes ZERO tests passes both report guards — the credentials preflight aborts it after a red Collect models


Victor-w-Madeira:
Spun out of daily-failure triage #1057 (run [30444299314](https://github.com/oriontech-me/langflow-e2e/actions/runs/30444299314), 2026-07-29). Signature quoted from the job logs of shards 1 and 2, not from `reports/daily-history.jsonl` — the history row cannot record it, which is precisely the gap this issue is about.

**Upstream:** _not filed_

## Symptom

Shards 1 and 2 executed **zero tests** and the run still reported as an ordinary per-test failure day. Both died ~5 min in (10:36:50 → 10:41:38 and → 10:42:47) while shards 3 and 4 ran 33 and 41 min.

The chain, from the shard-1 log:

```
Models found (google): []
❌ google (no model) — no models collected from the providers panel
Error: env-keyed provider(s) inactive for a NON-billing reason (real key/account/config problem): google
   ↓  Collect models is red, but by design (#980) it does NOT abort the shard
Error: [preflight] provider key(s) set in the environment but NOT configured as a Langflow
  global variable: GOOGLE_API_KEY
   ↓
Run @stable tests (shard 1) → exit 1, zero tests executed
```

A red `Collect models` is deliberately non-fatal (#980), but it also never imports `GOOGLE_API_KEY` as a Langflow global variable — so the credentials preflight in `globalSetup` aborts the shard anyway. The #980 guarantee is defeated by a second gate downstream of it.

**Neither report guard fired.** `shardguard` saw a blob from every shard (all four uploaded — the two empty ones included). `runguard` saw a non-empty merged report, because shards 3 and 4 did produce tests. #1012 closed the case where the *whole run* executes zero tests; the **per-shard** case is uncovered. Recorded totals were `passed: 205` against a ~380 baseline (2026-07-27: 379) — roughly 184 tests silently never ran, and the umbrella rendered as a normal 10-failure day.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/collect-models.spec.ts:35` ("collect providers status and models from UI") | `google models in the providers panel` | `Error: env-keyed provider(s) inactive for a NON-billing reason (real key/account/config problem): google — no models collected from the providers panel` |

## Why these failures are one cause

[...]

PR #1069, opened 2026-07-29
docs(triage-skill): state the guard-day rule as narrowly as CONTRIBUTING does (#1067)

### LINKED ISSUE #1067: Triage skill contradicts CONTRIBUTING: "on a guard-tripped day nothing is quarantined" wrongly exempts recurrent flakes
## The contradiction

`.claude/skills/langflow-e2e-triage/references/issue-templates.md` ends its *Quarantine mechanism* section with:

> **Scope:** quarantine is for **recurrent flakes** removed manually at triage. […] On a **guard-tripped** day nothing is quarantined (tags kept).

`SKILL.md` repeats it in Phase 7 step 3:

> *(Hard failures were already auto-removed by the workflow; on a guard-tripped mass-failure day nothing is quarantined here.)*

`CONTRIBUTING.md` says something narrower. The guard exempts only the **automatic** removal that applies to **hard failures**, and it explicitly adds a deliverable rather than removing one:

> When it trips, the triage issue says so and the tags are still in place — investigate the environment; if it turns out not to be environmental, **manually quarantine** the real hard failures *(line 696)*

> When the mass-failure guard trips, the triage gains one extra deliverable: **decide whether the day was environmental** and, if not, **manually quarantine** the real hard failures *(line 663)*

And the flake criterion carries no guard-day exemption at all:

> Same signature within the window → recurrent → open a **dedicated issue and quarantine the test via PR as prevention** (flake quarantine is always manual; the workflow never auto-removes flakes) *(line 713, and the criteria table at line 796)*

## Why it matters

The skill's blanket wording is strictly wrong in two ways on a guard day:

1. It **skips the quarantine of recurrent flakes**, which the criterion requires regardless of the guard. The guard governs hard failures only.
2. It **drops the environmental-verdict deliverable** — the analyst is supposed to decide whether the day was environmental and manually quarantine the hard failures that were not.

## Observed

This misfired during the triage of #1057 (run 30444299314, 2026-07-29). Following the skill, the triage initially proposed no quarantine at all and stated no environmental verdict. Both were corrected on review: the verdict was recorded on #1058–#1063, and four recurrent flakes were quarantined in PR #1064 — work the skill's wording had ruled out.

## Fix

Replace the blanket claim in both files with the actual rule: on a guard-tripped day the workflow's automatic `@stable` removal is suppressed for hard failures; the triage must state an environmental verdict and manually quarantine any hard failure judged non-environmental; recurrent flakes are quarantined as on any other day. `CONTRIBUTING.md` is authoritative on conflict.

Worth checking the same files for other places where the skill states a guard-day rule more broadly than `CONTRIBUTING.md` does.

## Related

#1065 removes a different guard-day clause (the umbrella staying open) from the same documents. Kept separate: that one is a rule change, this one is the skill misstating a rule that already exists.

Victor-w-Madeira:
## Type

- [x] `docs` — documentation update

---

- [x] **Exception** — off-wave work backed by an issue: a follow-up, a `daily-failure` triage issue, or a `community`-labeled regression. Issue #1067

---

## What this PR does

The triage skill claimed, in two files, that

> On a **guard-tripped** day nothing is quarantined (tags kept).

`CONTRIBUTING.md` says something narrower, and in one direction only: the guard suppresses the **automatic** `@stable` removal, which applies to **hard failures** — and it *adds* a deliverable rather than removing one.

The blanket wording was wrong twice over on a guard day:

1. It **skipped the quarantine of recurrent flakes**, which the flake criterion requires regardless of the guard (`CONTRIBUTING.md` line 713 and the criteria table). The guard governs hard failures; the flake criterion carries no guard-day exemption.
2. It **dropped the environmental-verdict deliverable** — the triage is supposed to decide whether the day was environmental, and manually quarantine any hard failure that verdict does not cover, since the workflow removed nothing.

### It misfired in practice

On the triage of #1057 (run 30444299314, 2026-07-29), following the skill produced a plan with **no quarantine at all and no environmental verdict**. Both were corrected on review: the verdict now appears on #1058–#1063, and four recurrent flakes were quarantined in #1064 — work the skill's wording had explicitly ruled out.

### Where the change lands

The issue named two spots; the sweep it asked for found seven. All are places where the skill stated a guard-day rule more broadly than `CONTRIBUTING.md` does:

| File | Location | Was |
|---|---|---|
| `SKILL.md` | Hard gates | "only **recurrent flakes** are quarantined here" — on a guard day a non-environmental hard failure is quarantined here too |
| `SKILL.md` | Phase 3 | "`@stable` is **kept** on everything" — too broad; and the verdict deliverable was absent |
| `SKILL.md` | Phase 4 | did not say the flake criterion has no guard-day exemption |
| `SKILL.md` | Phase 7 step 3 | the original claim |
| `SKILL.md` | headless `--phase propose` | reported only that `@stable` was left in place, with no verdict |
| `references/issue-templates.md` | *Quarantine mechanism* → Scope | the original claim |
| `references/issue-templates.md` | *Guard-Tripped Rule* + summary bullet | rule 3 treated "kept" as unconditional |

### Two judgement calls worth reviewing

**The environmental verdict is now a numbered rule (4)** in the *Guard-Tripped Rule*, with a worked example. It existed in `CONTRIBUTING.md` but nowhere in the skill — which is why the #1057 triage never produced one.

**Rule 4 is explicitly separated from rule 2**, because they read as contradictory and are not: rule 2 forbids asserting an environmental cause **for a cluster**; rule 4 requires a verdict on **the day**. The text now states that a day judged environmental does not make every failure on it collateral — the exact trap that nearly led to not filing #1059.

- **No behaviour change.** Documentation only — no script, renderer, workflow or config is touched. `renderDedicatedIssueBody()` / `assertDedicatedIssueBody()` are unchanged, so the issue-body contract and the CI guard that enforces it are identical.
- **The guard's own mechanics are unchanged** — threshold, suppression of automatic removal, and the cluster split by durability all stay exactly as they were. Only the skill's *description* of what the triage still owes is corrected.
- **`CONTRIBUTING.md` is not edited.** It was already right; the skill was the one out of step. (The separate guard-day clause removed in #1065 / #1068 was a rule change and is already merged.)

## Known limitations

This is guidance read by an agent, so no test can enforce it — `npm run test:scripts` covers `triage-core.mjs`, not the prose. The real verification is the next red daily that trips the guard: the triage should produce an environmental verdict and quarantine any recurrent flake without a human catching the omission.

---

## Related issue

Closes #1067

---

## Validation

- [x] `npm run test:scripts` — 89 pass, 0 fail
- [x] Grepped the whole skill directory **and** `CONTRIBUTING.md` for the superseded phrasings (`nothing is quarantined`, `kept on everything`, `nothing quarantined on a guard`) — **0 remaining**
- [x] Verified the corrected text against `CONTRIBUTING.md` lines 663, 696, 713 and the criteria table, which stay authoritative on conflict
- [x] Documentation-only change — nothing to run with `--trace=on`, force-fail, or step through in `--debug`

Provenance: raised during the triage of #1057 (daily run 30444299314, 2026-07-29), the run whose triage the defect misdirected.

### DIFF
--- a/.claude/skills/langflow-e2e-triage/SKILL.md
+++ b/.claude/skills/langflow-e2e-triage/SKILL.md
@@ -55,8 +55,11 @@ Talk to the user in **Portuguese (PT-BR)**; produce every GitHub artifact
   only stops the daily, so the test keeps going red on the impacted-specs gate
   (file-diff selected, not tag-filtered — #871); `test.fixme` skips it in every
   context. Full mechanism + why: `references/issue-templates.md` →
-  *Quarantine mechanism*. Hard failures are already auto-removed by the workflow
-  (tag only); only **recurrent flakes** are quarantined here (never automatic).
+  *Quarantine mechanism*. Hard failures are normally auto-removed by the
+  workflow (tag only), so **recurrent flakes** are what gets quarantined here —
+  except on a **guard-tripped** day, where the workflow removed nothing and a
+  hard failure the triage judged non-environmental is quarantined here too. A
+  guard day never suspends the quarantine of a recurrent flake.
   The skill **never restores** either edit and never touches test *logic*, spec
   docs, or `QA-CHECKLIST.md` — un-`fixme` + restore-`@stable` is the dedicated
   issue's deliverable.
@@ -186,8 +189,22 @@ When `guard_tripped` is true the day is a mass-failure day (mostly collateral).
 other non-adjacent dailies — durable) are created/enriched as usual;
 **today-only collateral** is **noted, not filed** (aggregated with counts) and
 recorded in the umbrella's closing comment — the umbrella still **closes** at
-the end of triage (Phase 7). `@stable` is **kept** on everything. Full rule + wording: `references/issue-templates.md` →
+the end of triage (Phase 7). Full rule + wording: `references/issue-templates.md` →
 *Guard-Tripped Rule*.
+
+**The guard adds a deliverable, it does not remove one** (`CONTRIBUTING.md`):
+
+- **Decide whether the day was environmental** and say so, with evidence, on
+  every issue this run produces. This verdict is mandatory on a guard day —
+  it is what the reader needs to weigh any cluster filed from it.
+- **Hard failures:** the workflow's automatic `@stable` removal is suppressed,
+  so the tags are still in place. If the verdict is **not** environmental,
+  **manually quarantine** the real hard failures (Phase 7 step 3). If it **is**,
+  keep `@stable` and say why — quarantine only if one reproduces on a clean,
+  non-guarded daily.
+- **Recurrent flakes: unaffected.** The guard governs hard failures only; it is
+  **not** an exemption from the flake criterion. A flake with `actionable: true`
+  is quarantined exactly as on any other day (Phase 4).
 
 ### Phase 4 — FLAKES
 
@@ -200,7 +217,9 @@ own.
 
 For each actionable flake, the test must be **quarantined as prevention** (so it
 stops running everywhere until it is worked) — part of the triage, not a
-deferred follow-up. Quarantine = **remove `@stable` + add `test.fixme`**
+deferred follow-up, **and not suspended by a guard-tripped day** (the guard
+suppresses the automatic removal that applies to hard failures; the flake
+criterion has no guard-day exemption). Quarantine = **remove `@stable` + add `test.fixme`**
 together (see the hard gate + `references/issue-templates.md` → *Quarantine
 mechanism*). But it is a code change: **record the exact spec path + line here**
 and carry it into the plan (Phase 6) as its own row; it is executed only in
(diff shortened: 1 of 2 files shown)


Victor-w-Madeira:
Review follow-up merged as `0336647`.

Correcting one line of the Validation block above: the grep sweep did **not** return 0 remaining. Two spots survived it because they phrase the claim as "nothing **was** quarantined" — `references/issue-templates.md` Deliverables template (line 199) and its rules list (line 207). Both are fixed in the follow-up commit.

That commit also closes a gap in this PR's own reasoning: the guard-day quarantine was stated in Phase 3 and executed in Phase 7 step 3, but Phase 6 (the plan in between) still filtered hard failures out unconditionally, and step 3 iterated only over recurrent flakes — so the row never reached the confirmation loop. Plus: the worked example's `@stable` note, the stale `rules 1-3` count, and rules 3/4 swapped so the verdict precedes the outcome that depends on it.

Issue #1030, opened 2026-07-28
daily-stable: detect a mid-run backend wedge and abort the shard instead of burning retries on a dead backend


Victor-w-Madeira:
## Problem

The #1011 health gate protects the **start** of the run: it waits for the backend to recover from the `collect-models` load, between that step and `Run @stable tests`. Nothing watches the backend **during** the run.

On run [30374528125](https://github.com/oriontech-me/langflow-e2e/actions/runs/30374528125), shard 2's single Langflow worker was killed by gunicorn six times, ~313s apart (`WORKER TIMEOUT` → `SIGKILL`), from 15:55 to 16:22 — the run had started at 15:46, well past the gate. Consequences:

- ~26 of that shard's 38 minutes were spent retrying against a dead backend (3 attempts per test).
- 14 phantom failures entered the report, identical in shape to real ones — a full triage cycle to attribute. The other three shards, unaffected, produced 5 failures between them.
- The `runguard` (`scripts/check-run-integrity.mjs`) does not catch this: the merged report *does* contain tests. It only covers the zero-test abort (#1012).

## State as of 2026-07-29 — the wedge is steady state, not an outage

Two things landed after this issue was filed, and the second one changes what the right remedy is.

**#1048 (PR #1050, merged 2026-07-28 23:30) capped the cost per wedge.** `LANGFLOW_WORKER_TIMEOUT: 120` is now pinned on `daily-stable.yml`, `pr-validation.yml` and `weekly-stable.yml`, replacing Langflow's 300 s default. Per that issue's analysis the kill now lands 60–120 s after the event loop stops, against 150–300 s before — **bounding** the outage this issue is about, without detecting it. (That band is the configured expectation; it was not re-measured on the run below.) #1029, one of the two trigger classes in the non-goals, is also closed; #976 is still open.

**The only daily since that cap — run [30444299314](https://github.com/oriontech-me/langflow-e2e/actions/runs/30444299314) (2026-07-29) — shows the wedge recurring continuously on both shards that ran tests:**

| Shard | Result | `WORKER TIMEOUT` events |
|---|---|---|
| 1/4 | **0 tests** — aborted in `globalSetup` | 0 |
| 2/4 | **0 tests** — aborted in `globalSetup` | 0 |
| 3/4 | 5 failed | **7** |
| 4/4 | 5 failed, 3 flaky, **101 passed** | **10** |

[...]

PR #1078, opened 2026-07-29
ci(daily-stable): stop a narrow backend wedge from stripping @stable (#1031)

### LINKED ISSUE #1031: Exempt infra-signature failures from @stable auto-removal and label them as wedge collateral
## Problem

When the backend dies mid-run, every test that touches it fails with an infra signature (`apiRequestContext.* Timeout`, `[preflight] ... is not reachable`). Those failures are then attributed to the specs that happened to be running.

On run [30374528125](https://github.com/oriontech-me/langflow-e2e/actions/runs/30374528125) we escaped by luck: 19 hard failures exceeded `MAX_AUTO_REMOVE` (default 5), so the mass-failure guard in `scripts/remove-stable-from-failures.ts:244` tripped and nothing was removed. A wedge that produces **≤5** collateral failures would silently strip `@stable` from innocent specs — auto-removal runs without review by design, and restoration is the path that carries human intervention.

The same misattribution reaches the daily-failure issue body and `reports/daily-history.jsonl`: 14 of the 19 entries described one wedged backend, not 14 broken specs.

## Proposal

Classify a hard failure as **non-attributable** when its last error carries an infra signature and the backend was unhealthy in that window, then:

- exclude it from `@stable` auto-removal, independently of the mass-failure guard, and
- label it as wedge collateral in the daily issue body, so triage does not open per-spec issues for it.

## Done when

- [ ] Infra-signature failures are excluded from auto-removal independently of `MAX_AUTO_REMOVE`
- [ ] The daily issue distinguishes collateral from attributable failures
- [ ] Unit-covered next to the script in `scripts/remove-stable-from-failures.test.ts`, including the ≤5-failure wedge that today would strip tags
- [ ] Triage rules in `CONTRIBUTING.md` updated to match

## Notes

- Needs a signal for "the backend was unhealthy in that window" — the cheapest source is whatever #1030 ends up emitting. It can also stand alone on the error signature plus same-shard clustering.
- Sequencing: the current mass-failure guard already covers the wide-wedge shape seen on 07-28, so this can land after the clean baseline (#818) if capacity is tight. The uncovered case is the narrow wedge.

Victor-w-Madeira:
## Problem

When the backend dies mid-run, every test that touches it fails with a transport-level error, and the report blames the specs that happened to be running. On run [30374528125](https://github.com/oriontech-me/langflow-e2e/actions/runs/30374528125), 14 of 19 hard failures described **one wedged backend**, not 14 broken specs.

We escaped by luck: 19 > `MAX_AUTO_REMOVE` (5), so the mass-failure guard tripped and nothing was removed. The guard only ever covered the **wide** wedge — a wedge costing **≤5** tests would have stripped five innocent tags in an unreviewed commit to `main`, since auto-removal runs without review by design and *restoration* is the human-gated path.

## What this does

Classifies a hard failure as **non-attributable** when its last error is transport-level, and exempts it from `@stable` auto-removal **independently of the threshold**.

| File | Change |
|---|---|
| `scripts/lib/infra-signatures.ts` *(new)* | The signature list + `classifyInfraError`. |
| `scripts/remove-stable-from-failures.ts` | Reads the last failed attempt's error; partitions **before** the guard; emits `attributableFailures`, `exempt[]`, `backendWedged`. |
| `scripts/format-auto-remove-summary.mjs` | Renders collateral in its own block, **ahead of** the removals. |
| `.github/actions/auto-remove-stable/action.yml` | Optional `backend_wedged` input, `exempt_count` output. |
| `.github/workflows/daily-stable.yml` | Forwards `steps.liveness.outputs.wedged` (#1030). |
| `CONTRIBUTING.md` | *Infra-signature exemption (wedge collateral)* note, decision-table row, lifecycle diagram. |
| `.claude/skills/langflow-e2e-triage/SKILL.md` | Phase 3 pointer — don't cluster collateral. |

### Three decisions worth reviewing

**1. The signature list is deliberately narrow.** Only errors that cannot be a product assertion under any reading: `apiRequestContext.*: Timeout`, the globalSetup `[preflight] … is not reachable`, `ECONNREFUSED`/`ECONNRESET`/`socket hang up`, `net::ERR_CONNECTION_*`, DNS failures. The three signatures that *dominate* `reports/daily-history.jsonl` — `locator.click: Timeout`, `page.waitForSelector: Timeout`, `expect(locator).toBeVisible()` — are **excluded on purpose**. A wedge produces them, but so does real breakage; exempting them would switch auto-removal off for most genuine regressions instead of protecting it.

**2. The mass-failure guard keeps counting *every* hard failure**, collateral included. Netting the exempt ones out would make the mechanism *more* aggressive than today: on run 30374528125 the guard trips and removes nothing, while an attributable-only count would remove 5 tags with no review. #1031 asks to protect innocent specs, not to widen auto-removal's reach — so **the removal set is always a subset of what the pre-#1031 script produced**. The partition still happens *before* the guard, so the wide wedge (the run that most needs its cause named) still gets its collateral labelled.

**3. #1030's `wedged` verdict words the exemption, it never decides it.** The exemption has to hold on a run where the liveness recorder produced nothing — that is exactly the run whose backend state is least known. This also keeps faith with `report-backend-outages.mjs`'s own honesty note: at a 33–73% down-share, overlap with an outage window is a *lead, not a verdict*.

## Done when

- [x] Infra-signature failures are excluded from auto-removal independently of `MAX_AUTO_REMOVE`
- [x] The daily issue distinguishes collateral from attributable failures
- [x] Unit-covered next to the script, including the ≤5-failure wedge that today would strip tags
- [x] Triage rules in `CONTRIBUTING.md` updated to match

## Validation

```
npm run typecheck                # clean
npm run test:units               # 93/93  (24 in remove-stable-from-failures, 8 new)
npm run test:scripts             # 119/119 (6 new in format-auto-remove-summary)
npm run lint                     # 0 errors
node scripts/check-checklist-guard.mjs && npm run check:checklist-coverage   # both pass
```

The case the issue names:

```
node --require ts-node/register --test scripts/remove-stable-from-failures.test.ts \
  --test-name-pattern "narrow wedge"
# 1 pass — status "none", 5 exempt, spec file byte-identical
```

New cases: narrow wedge at the threshold · mixed run (collateral exempt, attributable still removed) · ambiguous error must **not** exempt · failure with no error is attributable · guard-tripped run still labels its collateral · exemption independent of the liveness verdict · signature matched on a cause line, not only line one · `lastFailureError` reads the last attempt · pre-#1031 result JSON still renders.

**No local proof for:** the `steps.liveness.outputs.wedged` → `backend_wedged` wiring, which only runs in CI. The YAML parses and the `liveness` step is in the same `merge` job, ahead of `auto_remove`. Final proof is the next red daily — check that the corroboration sentence in the `@stable auto-removal` block agrees with the backend-liveness section directly above it. The change is latent (it only acts on a red day), so it does not warrant a daily re-dispatch.

Closes #1031

### DIFF
--- a/.claude/skills/langflow-e2e-triage/SKILL.md
+++ b/.claude/skills/langflow-e2e-triage/SKILL.md
@@ -177,6 +177,14 @@ occurrences under it.
 than splitting it across per-symptom buckets — this is the #899 fix for the
 misgrouping that hid #898's `langchain-google-genai` cause.
 
+**Set aside the wedge collateral (#1031).** The umbrella's `@stable` section
+renders, ahead of the removals, the hard failures the run could **not** attribute
+to their spec — those whose last error is transport-level. Their tag was kept on
+purpose. Do **not** cluster them and do **not** open a per-spec issue for them:
+they are one backend outage, triaged once for the run against the umbrella's
+backend liveness section. Rule and signature list: `CONTRIBUTING.md` →
+*Infra-signature exemption (wedge collateral)*.
+
 Then cluster the rest of `hard_failures[]` by root cause: **same normalized
 error signature + same area + same failure symptom → one issue.** Don't open one
 issue per failing spec by default — a shared root cause is one problem, not N.
--- a/.github/actions/auto-remove-stable/action.yml
+++ b/.github/actions/auto-remove-stable/action.yml
@@ -27,6 +27,10 @@ inputs:
       backend was not measured. WORDING ONLY — the infra-signature exemption
       never depends on it, so a caller with no liveness step (weekly-stable)
       can leave it unset and still get the exemption.
+      Callers must NOT forward `report-backend-outages.mjs`'s `wedged` output
+      bare: it is also "false" when no shard produced probes at all. Gate it on
+      that step's `measured` output so "unmeasured" arrives here as "", or the
+      summary claims the recorder measured no outage on a run it never measured.
     required: false
     default: ""
 
@@ -74,6 +78,10 @@ runs:
     - name: Commit @stable removals to main
       if: steps.remove.outputs.status == 'removed'
       shell: bash
+      env:
+        REMOVED_COUNT: ${{ steps.remove.outputs.removed_count }}
+        EXEMPT_COUNT: ${{ steps.remove.outputs.exempt_count }}
+        RUN_LABEL: ${{ inputs.run_label }}
       run: |
         set -euo pipefail
         # Same "dubious ownership" workaround as the history commit step: the job
@@ -89,5 +97,12 @@ runs:
         # results.json / payload.json / auto-remove-result.json at repo root are
         # not git-ignored and must not be committed.
         git add tests/tests-automations/regression QA-CHECKLIST.md
-        git commit -m "chore(triage): auto-remove @stable from ${{ steps.remove.outputs.removed_count }} hard-failing test(s) (${{ inputs.run_label }}) [skip ci]"
+        MSG="chore(triage): auto-remove @stable from ${REMOVED_COUNT} hard-failing test(s) (${RUN_LABEL}) [skip ci]"
+        if [ "${EXEMPT_COUNT:-0}" != "0" ]; then
+          # Record in the commit itself why fewer tags were removed than tests
+          # failed — otherwise auditing this commit means re-opening the run
+          # to discover the wedge collateral (#1031).
+          MSG="${MSG}"$'\n\n'"${EXEMPT_COUNT} further hard failure(s) kept @stable as non-attributable wedge collateral (#1031)."
+        fi
+        git commit -m "$MSG"
         git push
--- a/.github/actions/auto-remove-stable/action.yml
+++ b/.github/actions/auto-remove-stable/action.yml
@@ -4,8 +4,10 @@ description: >-
   (failed every retry), regenerate QA-CHECKLIST.md, and commit the change back
   to main — no human review (leadership decision; restoring the tag is the
   human-gated step). A mass-failure guard leaves everything untouched when too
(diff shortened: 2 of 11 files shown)


Victor-w-Madeira:
Review fixes landed in 2aa9a30 (see the commit message for the full rationale):

1. **The classifier was reading a 240-char excerpt.** `lastFailureError` truncated *before* `classifyInfraError` ran, so a transport error under an assertion header + call log — the shape `infra-signatures.ts` names as the reason to match past line one — was not exempted, and the innocent spec lost `@stable` anyway. Classification now runs on the full text; `truncateError` bounds only the issue body.
2. **`backend_wedged` was forwarded bare.** `report-backend-outages.mjs` also emits `wedged=false` when no shard produced probes, so the summary claimed the recorder "measured no outage" over a run it never measured. Now gated on that step`s `measured` output.
3. **Only `errors[0]` was read** — a `timedOut` result puts the wrapper in `error` and the pending call in a later `errors[]` entry. All errors of the attempt are kept now, deduped.
4. **`status` stopped meaning "would the guard have tripped"** — the nothing-attributable exit ran ahead of the guard, so an all-collateral wide wedge reported `none` while the triage skill`s `detectGuard` called the same day a mass-failure day.
5. A backtick in an error broke out of its code span; `exempt_count` had no consumer — the auto-removal commit now records the collateral count.

Validation after the fixes: `test:units` 96/96 (+3), `test:scripts` 121/121 (+2), typecheck / lint / checklist guards clean.

**Follow-up, not in this PR:** `triage-core.mjs`'s `hard_failures[]` carries no exempt flag, so an automated triage pass could still re-cluster collateral. The `SKILL.md` rule covers the human path.

Issue #490, opened 2026-07-02
Create agent-tool-name-validation.spec.ts — invalid tool name blocked with clear message


Victor-w-Madeira:
**Wave 1 deliverable** — see `ROADMAP.md` · Milestone: Wave 1 — Agents & providers (2026-07-14)

**Type:** Create new spec
**Spec file:** `tests/tests-automations/regression/core-functionality/llm-agents/agent-tool-name-validation.spec.ts`
**Spec doc:** `docs/core-functionality/llm-agents/agent-tool-name-validation.md` (to create)

**Checklist bullets (`QA-CHECKLIST.md`):**
- §6.4 — Tool with invalid name — validation prevents execution with clear message

**Depends on:** #480

**Done when:** spec written + `@stable` + spec doc + checklist bullet(s) flipped to `[x]`.

PR #1080, opened 2026-07-29
fix(llm-agents): confirm the context_id write before each agent turn (#1060)

### LINKED ISSUE #1060: [Daily #1057] agent-context-id-isolation: context_id re-tagging assertion is flaky (3× same-signature recurrent)
Spun out of daily-failure triage #1057 (run [30444299314](https://github.com/oriontech-me/langflow-e2e/actions/runs/30444299314), 2026-07-29).

**Upstream:** _not filed_

## Symptom

`switching the agent's context_id re-tags new turns without touching previous ones` (line 570) is same-signature recurrent as a **flake** across three dailies — 2026-07-14, 2026-07-16 and 2026-07-22 — on `openai / gpt-4o-mini`, failing an `Object.is` equality assertion before recovering on retry.

On this run the **same line** also produced a final hard failure on `google / gemini-2.5-flash`, recorded with signature `unknown`. Whether the two are the same defect is open; both rows are listed so the investigation starts from the full picture rather than from the flake alone.

|---|---|---|
| `tests-automations/regression/core-functionality/llm-agents/agent-context-id-isolation.spec.ts:570` ("switching the agent's context_id re-tags new turns without touching previous ones") | `the context_id of newly created messages` | `Error: expect(received).toBe(expected) // Object.is equality` |
| `tests-automations/regression/core-functionality/llm-agents/agent-context-id-isolation.spec.ts:570` ("switching the agent's context_id re-tags new turns without touching previous ones (hard failure, google / gemini-2.5-flash)") | `the context_id of newly created messages` | `unknown` |

Both rows are the same test at the same line. They are listed as two rows because the run recorded them under two different signatures on two different providers — a same-cause claim is **not** being made, and confirming or refuting it is part of the investigation.

This surfaced on a **mass-failure day** (10 hard + 12 flaky, guard threshold 5 crossed). The triage verdict on the day itself is that it **was environmental**: 6 of the 10 hard failures are a 20 s timeout on `GET /api/v1/auto_login` in shards where every provider was green and the post-`collect-models` health gate had already passed (tracked on #1030), and two further shards executed zero tests after a credentials-preflight abort (#1058).

That verdict governs the **hard failures**, none of which was quarantined. It does not cover this test: the guard suppresses the automatic `@stable` removal that applies to hard failures, and does not exempt a recurrent flake from the quarantine criterion. This test was therefore quarantined at triage — see *Flake signal* below.

**Weight of the recurrence signal.** All three flake occurrences (07-14, 07-16, 07-22) fall on dailies that **also tripped the guard** (27, 6 and 22 hard failures). The quarantine follows the criterion as written — same `error_signature` twice or more inside the 30-day window — but the pattern is not distinguishable from the load-induced agent/playground flakiness already under investigation on #773. Rule that out before treating this as a defect specific to `context_id` re-tagging.

An equality assertion on `context_id` failing intermittently is consistent with either a read-before-write race in the test or non-deterministic tagging in the product; neither is asserted here. The `unknown` signature on today's hard failure carries no information by itself and must not be treated as matching anything.

**Cost note.** The `test()` sits inside the `for (const { label, options, skipReason } of targets)` loop, so the `test.fixme` suppresses it for **every** provider target, not only the one where it flakes. That is a limitation of the quarantine primitive, and an argument for lifting this one promptly.

## Investigation directive

1. Check whether it flakes on a **clean, non-guarded** daily. All three known occurrences are mass-failure days, so this determines whether the issue is a real defect or a #773 symptom.
2. Establish a rate: run at `--retries=0` against the current nightly on `openai / gpt-4o-mini` enough times to quantify it, then repeat on `google / gemini-2.5-flash`.
3. Retrieve the actual vs. expected `context_id` values from the 07-14 / 07-16 / 07-22 traces — a stale previous-context value and an empty/undefined one point at different causes.
4. Determine whether the assertion runs before the re-tagging write is observable, i.e. whether a wait-for-postcondition is missing.
5. Recover today's hard-failure error from the run 30444299314 artifacts (it is `unknown` in the history row) and decide whether it is the same defect or a separate one needing its own issue.

## Deliverables (Done when)

## Flake signal

This test is confirmed recurrent (dailies 2026-07-14, 2026-07-16, 2026-07-22). As prevention it was **quarantined** at triage in PR #1064 — `@stable` removed **and** `test.fixme` added — so it stops running in **every** context (daily, PR impacted-specs gate, full suite) until this issue is worked:

- `tests/tests-automations/regression/core-functionality/llm-agents/agent-context-id-isolation.spec.ts` (test at line 570)

Lifting the quarantine after the fix (remove `test.fixme` + restore `@stable`) is a deliverable of this issue.

rafaelgiln:
**Fixes #1060.**

## Problem

`switching the agent's context_id re-tags new turns without touching previous ones` failed the same way on three dailies — 2026-07-14, 2026-07-16 and 2026-07-22 (`openai / gpt-4o-mini`, `Object.is` equality) — and again as a flaky attempt on run [30444299314](https://github.com/oriontech-me/langflow-e2e/actions/runs/30444299314). It was quarantined at triage in #1064 (`test.fixme` + `@stable` removed).

Root cause is a **write race in the test**, not a product regression. The test PATCHes `context_id` on the flow's nodes and runs the turn right after, but the editor keeps firing its own debounced `PATCH /api/v1/flows/{id}` carrying the store's snapshot after a playground turn. That autosave can be *issued* after ours and land last — the endpoint has no version check (same last-writer-wins documented in `wait-for-flow-save-settled.ts`) — silently reverting the switch. Turn 2 then runs under the OLD context and the isolation assert reports cross-tagging that never happened.

Reproduced locally with a request-timeline probe (1 failure in 12 `--retries=0` runs on an idle backend, nightly `1.12.0.dev8`):

```
DIAGPATCH[api-out]  t=…702053 -> ctx-B      our PATCH goes out
DIAGPATCH[fe-out]   t=…702076 ctx=[ctx-A]   editor autosave goes out 23 ms later, stale snapshot
DIAGPATCH[api-done] t=…702123 status=200
DIAGPATCH[fe-done]  t=…702143               the editor's write lands LAST -> flow back on ctx-A
DIAG[turn2-before-send] {"ChatInput":"ctx-A…","ChatOutput":"ctx-A…","Agent":"ctx-A…"}
  1 failed   // exactly the daily's signature
```

The rate explains the recurrence pattern: every recorded occurrence fell on a saturated daily, because parallel shards widen the race window. That correlates with #773 without being #773.

**The `google / gemini-2.5-flash` hard failure on the same line is a different cause** — recovered from the run artifact it is `TimeoutError: apiRequestContext.get: Timeout 20000ms exceeded → GET /api/v1/auto_login` in `getAuthToken`, i.e. the #1030 cluster. It needs no dedicated issue, and it also reproduces locally (3× today, including on the first run after a container restart).

## Fix

Each turn's setup now goes through `prepareTurn`: PATCH → reload → seed the ChatInput → drain the autosave → **read the flow back from the server**, re-patching while the three nodes still disagree (3 attempts). The turn only runs once the server confirms the intended context. If the editor wins every attempt the test fails as an explicit **setup** error naming the reverted write, so a write race can never masquerade as broken isolation.

Both turns use it — turn 1 has the same exposure (the editor also autosaves right after the template load).

Rejected alternatives:
- **Autosave barrier only** (`waitForFlowSaveSettled` before the PATCH): the measured late save came 2.1 s after the previous one, outside the helper's 700 ms quiet window — it would not have caught this.
- **Run both turns through `POST /api/v1/run`**: deterministic, but it removes the Agent from the Playground path the spec exists to cover (`@playground`).

Quarantine lifted: `test.fixme` removed and `@stable` restored.

## Scope note

No assertion changed. The observables are the same: turn-1 messages all carry `CTX-A`, turn-2 messages (id-partitioned) all carry `CTX-B`, turn-1 messages are not re-tagged. Test 1 (model-free retrieval half) is untouched apart from being re-run and force-failed as part of the file. `QA-CHECKLIST.md` needs no edit — the bullets already reference the spec and the counts regenerate from the restored tag.

## Validation (nightly `1.12.0.dev8` / `1.12.0.dev9`, `--retries=0 --workers=1`)

- `npm run typecheck` ✅ · `npm run lint` ✅ (0 errors)
- **9 clean runs of the file, zero occurrences of the defect** ✅ (2 aborts on the #1030 `auto_login` timeout and 1 `built successfully` toast timeout on test 1 with the backend already degraded — none is this failure mode)
- Pre-fix baseline on the same machine: 1 occurrence in 12 runs
- Also green on `google / gemini-3.5-flash` (`gemini-2.5-flash` is gated locally; collect-models settled on 3.5)
- Force-fail — all four executed, each failed as designed, then reverted (`grep FFMUT` → 0 hits) with a green run after:
  - `M1` test 1 expects a `B-*` sentinel in the `CTX-A` retrieval → ✘ `Expected substring: "CTXISO-…-B-1"`
  - `M2` test 1 expects a never-seeded sentinel → ✘ `Expected substring: "NEVER-SEEDED-SENTINEL"`
  - `M3` test 2 expects turn 2 tagged `CTX-A` → ✘ `Received: turn-2 message(s) with wrong context_id: [… "ctx-B-probe-A-…"]` (also proves the switch now reaches the run)
  - `M4` the confirmed-write gate targets a context the flow never receives → ✘ `context_id "ctx-A-…" was reverted by the editor autosave on 3 consecutive attempts`
- Zero `🚨 Backend Error` ✅ · flow cleanup verified: one flow leaked by an aborted run was purged, final count 27 = starter templates only, no duplicates
- `--trace=on` **not run**: it hangs on the Simple Agent template family (known limitation, #490/#518). Step verification came from the `--retries=0` bursts, the force-fails and the request-timeline instrumentation.

### DIFF
--- a/docs/core-functionality/llm-agents/agent-context-id-isolation.md
+++ b/docs/core-functionality/llm-agents/agent-context-id-isolation.md
@@ -1,6 +1,6 @@
 # Agent context_id — switching isolates history between contexts
 
-**Last validated:** Langflow 1.12.x (nightly `1.12.0.dev7`)
+**Last validated:** Langflow 1.12.x (nightly `1.12.0.dev8`)
 
 ---
 
@@ -97,15 +97,32 @@ surface; `@components` — Message History node drives the read-side assert;
 
 1. Load the Simple Agent template (provider/model from `models.json`/`.env`).
 2. Set `context_id = CTX-A` (unique per run) on the **Agent**, **Chat Input**
-   and **Chat Output** nodes (same advanced-field path as #487).
+   and **Chat Output** nodes (same advanced-field path as #487), and **confirm
+   the write survived** before running the turn (see the confirmed-write note
+   below).
 3. Seed the ChatInput with nonce N1; open the Playground, send, wait.
-4. Switch `context_id` to `CTX-B` on the same three nodes; seed nonce N2;
-   send a second turn in the same playground session.
+4. Switch `context_id` to `CTX-B` on the same three nodes — again **confirmed**
+   — seed nonce N2; send a second turn in the same playground session.
 5. **Assert (monitor API):** N1 → its session's turn-1 messages ALL carry
    `context_id === CTX-A` and none carry `CTX-B`; N2 → turn-2 messages ALL
    carry `CTX-B` and none carry `CTX-A`. Message sets are keyed by nonce, so
    the two turns are disjoint by construction.
 6. No `allowFlowErrors`.
+
+> **Confirmed-write setup (#1060).** The `context_id` write is an API PATCH on
+> the flow's nodes, and the editor keeps firing its own debounced
+> `PATCH /api/v1/flows/{id}` carrying the store's snapshot after a playground
+> turn. That autosave can be *issued* after our PATCH and land last — the
+> endpoint has no version check — silently reverting the switch, so the next
+> turn runs under the OLD context and the test reports a cross-tagging failure
+> that never happened. Reproduced locally at ~8% (1/12 `--retries=0` runs);
+> the daily's parallel shards widen the window, which is why every recorded
+> occurrence fell on a saturated day. Each turn's setup is therefore
+> **PATCH → reload → set the ChatInput → drain the autosave → read the flow
+> back**, retried while the server still disagrees, and the turn only runs
+> once the server confirms the intended context on all three nodes. If the
+> editor wins three times in a row the test fails as an explicit **setup**
+> error naming the reverted write — never as a fake isolation defect.
 
 ---
 
@@ -130,10 +147,16 @@ persisted/rendered data — no model judgment anywhere.
   seeding block separate "seed failed" from "isolation broken".
 - **Unique CTX-A/CTX-B/session per run** — monitor rows persist across flow
   deletion; per-run identifiers pin every lookup to THIS run.
+- **Setup failures are distinguishable from contract failures** — a reverted
+  `context_id` write fails in the setup step naming the reversion, so a
+  frontend/backend write race can never masquerade as broken isolation
+  (#1060).
 - **Force-failure checks** (CONTRIBUTING §2): M1 — test 1 asserts a `B-*`
   sentinel present in the `CTX-A` retrieval (inverted negative) ⇒ must fail;
   M2 — test 1 expects a never-seeded sentinel ⇒ must fail; M3 — test 2
-  expects turn-2 messages tagged `CTX-A` (stale context) ⇒ must fail.
+  expects turn-2 messages tagged `CTX-A` (stale context) ⇒ must fail; M4 —
+  test 2's confirmed-write gate targets a context the flow never receives
+  ⇒ must fail in setup, proving the gate is live.
 
 ---
 
--- a/tests/tests-automations/regression/core-functionality/llm-agents/agent-context-id-isolation.spec.ts
+++ b/tests/tests-automations/regression/core-functionality/llm-agents/agent-context-id-isolation.spec.ts
@@ -227,6 +227,28 @@ async function patchContextId(
   expect(patchRes.status()).toBe(200);
 }
 
+// Read the context_id currently stored on the given node types.
+async function readContextIds(
+  request: APIRequestContext,
+  bearer: string,
+  flowId: string,
+  nodeTypes: string[],
+): Promise<Record<string, unknown>> {
+  const res = await request.get(`/api/v1/flows/${flowId}`, {
+    headers: { Authorization: bearer },
+  });
+  expect(res.status()).toBe(200);
+  const flow = await res.json();
+  const values: Record<string, unknown> = {};
+  for (const node of flow.data?.nodes ?? []) {
+    const type = node.data?.type;
+    if (nodeTypes.includes(type)) {
+      values[type] = node.data?.node?.template?.context_id?.value;
+    }
+  }
+  return values;
+}
+
 // Set the task on the ChatInput node (the Playground prompt pre-fills from it;
 // typing into the Playground races an async default re-injection).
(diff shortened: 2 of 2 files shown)


rafaelgiln:
## Why the `Run impacted E2E specs` gate is red

Two consecutive runs of the job produced the same picture, and **neither contains a single `context_id` failure**:

| Target | Result | Cause |
|---|---|---|
| `openai / gpt-4o-mini` (the target this issue is about) | **passed** | — |
| `Context ID isolation — retrieval layer` (test 1) | **flaky** → passed | `TimeoutError: apiRequestContext.get: Timeout 20000ms exceeded → GET /api/v1/auto_login` |
| `google / gemini-2.5-flash` | **failed**, all 3 attempts | same `auto_login` timeout, raised inside `SimpleAgentTemplatePage.load` — the test body never ran |

That is the backend wedge measured in #1074 and closed as **steady state** on #1030 earlier today: gunicorn kills the single worker 7–10 times per shard, and any `getAuthToken` call landing in a restart window dies on its one 20 s attempt. It hits the third provider target because google runs last, after openai and anthropic have already driven two agent turns each through the shared container.

The signature is independent of this PR:

- the same `auto_login` timeout is what produced the `google / gemini-2.5-flash` hard failure on daily run [30444299314](https://github.com/oriontech-me/langflow-e2e/actions/runs/30444299314) — i.e. **before** this change, with the test quarantined;
- it reproduced locally 3× today on nightly `1.12.0.dev8`/`dev9`, including on the first run after a container restart.

Merging with the gate red is a deliberate call: the defect this PR fixes is proven fixed (9 clean local runs at `--retries=0`, 0 occurrences; 1/12 on the same machine before the fix; the `openai` target green in both CI runs), and the red job is an ambient condition tracked and accepted elsewhere. Making `getAuthToken` survive a worker restart is a suite-wide infra change and belongs to its own `qa-infra` issue, not here.

Issue #1077, opened 2026-07-29
daily-stable: give the mid-run worker wedge a baseline that outlives artifact retention


Victor-w-Madeira:
> ### Scope narrowed (2026-09-02) — this issue is now the BASELINE half only
>
> As opened, this issue was *measure, pick and record*. Two months and eight recorded
> occurrences later only the **measure** half turned out to be reachable, and the reason
> is in the mechanism: the wedge is recorded into per-shard `liveness-N` artifacts with
> `retention-days: 7`, so every measurement below is a table typed by hand into a comment
> from data that has since expired — three of the runs named here can no longer be
> re-derived at all. There was no durable baseline to compare a lever against, which is
> the first *Done when* below, and the four axes the second one names appeared on no
> committed record.
>
> PR #1685 closes that: every scheduled daily now writes its outage totals and per-shard
> breakdown onto its `reports/daily-history.jsonl` line, and the five dailies still in
> retention are backfilled — five consecutive runs, every shard executing tests,
> `shards_measured == shard_total` on all four.
>
> **Applying and measuring a lever moved to #1686**, together with the candidate list,
> the rejected levers and the credential-vs-load distinction below — all of which are
> transplanted there verbatim, nothing is lost. That work is blocked on #1679 (the
> spec-side enable write, which may be causing a share of the wedge); holding this issue
> open behind it while its own deliverable was done is what the split fixes.
>
> **New wedge occurrences go on #1686 from now on** — `CONTRIBUTING.md` → *Infra-signature
> exemption (wedge collateral)* still directs run-level outage triage to one issue per run,
> and that issue is now #1686. The eight records below stay here as history.

## Problem

The mid-run wedge is now **measured** but not **relieved**. PR #1074 (issue #1030) added
the in-run liveness recorder and the merge-side reporter, so a wedged daily announces
itself instead of reading as a list of collateral specs. It changes nothing about the
cost: the heavy shards still spend a large share of their span retrying against a
backend that is not answering.

The remaining work is the third *Done when* of #1030, extracted here so it is tracked on
its own (one issue, one PR) instead of holding that issue open behind a blocker with no
date.

## What the data already rules out

On daily run [30444299314](https://github.com/oriontech-me/langflow-e2e/actions/runs/30444299314)
— the first daily after #1048 pinned `LANGFLOW_WORKER_TIMEOUT: 120` on `daily-stable.yml`,
`pr-validation.yml` and `weekly-stable.yml`:

[...]

PR #1088, opened 2026-07-29
fix(auth): retry auto_login across a bounded budget instead of dying on one timeout (#1086)

### LINKED ISSUE #1086: getAuthToken dies on a single auto_login timeout — one worker recycle kills any of 135 specs
Sibling of #1077, from the opposite side. #1077 asks how to **relieve** the mid-run worker wedge (worker count, shard recomposition). This one asks how the suite **survives** one while it still happens — the two are complementary and neither replaces the other.

## Problem

`tests/helpers/auth/get-auth-token.ts` makes exactly one attempt:

```ts
export async function getAuthToken(request: APIRequestContext): Promise<string> {
  const res = await request.get("/api/v1/auto_login");
  if (res.ok()) { /* … */ }
  return "";   // fallback for an environment without auth
}
```

The fallback only covers a **non-ok response**. A request that never answers **throws**, and that is the case the wedge produces: gunicorn recycles the single worker (7–10× per shard, measured on run [30444299314](https://github.com/oriontech-me/langflow-e2e/actions/runs/30444299314) by #1074), the in-flight call hangs, and Playwright kills it at 20 s with

```
TimeoutError: apiRequestContext.get: Timeout 20000ms exceeded.
  - → GET http://localhost:7860/api/v1/auto_login
```

The outage is bounded — the worker comes back — but a call that lands inside it dies outright. **144 files (135 specs) call this helper**, so any of them can be killed by an event that has nothing to do with what it tests.

## Evidence it is costing real runs

| Where | What happened |
|---|---|
| Daily #1057 (run 30444299314) | `agent-context-id-isolation` on `google / gemini-2.5-flash` — 3 attempts, all dead in `getAuthToken` inside `SimpleAgentTemplatePage.load`; the test body never ran. Triage recorded it as signature `unknown` and it took an artifact dive on #1060 to name it |
| PR #1080, both CI runs | Same failure on the same target, plus the model-free test going flaky on it. Zero failures of the behaviour under test — the PR merged with a written justification because the gate could not go green |
| Local, 2026-07-29 | 3 aborts across ~24 runs on nightly `1.12.0.dev8`/`dev9`, including the first run after a container restart (cold start) |

## Proposed fix (deliberately small)

One bounded retry: on a thrown/timed-out `auto_login`, wait a short fixed delay and try **once** more, then give up as it does today. Not a generous loop.

**The trade-off, stated up front:** a client-side retry *masks* the wedge. It is defensible here because the failure is in infrastructure **authentication**, never in the behaviour under test — but if the backend is genuinely dead, the retry only postpones the red. That is exactly why the retry must stay short and capped, and why this issue does not replace #1077: relieving the wedge is still the real fix.

## Done when

- [ ] `getAuthToken` survives a single worker-recycle window, with the retry bounded and the give-up path unchanged
- [ ] Unit test next to the helper (`tests/helpers/auth/get-auth-token.test.ts`, `npm run test:units`) covering: happy path, non-ok response → `""`, one throw then success → token, two throws → the original error surfaces (never a silent `""` that would send callers on with no auth)
- [ ] A spec that exercises the helper stays green; no timeout anywhere else is loosened to compensate
- [ ] Cross-referenced on #1077 so the benchmark there is not read as covering this

rafaelgiln:
**Closes #1086.** Sibling of #1077 from the client side: that issue **relieves** the mid-run wedge, this one **survives** it.

## Problem

`getAuthToken` made exactly one attempt, and its `return ""` fallback only covers a **non-ok response**. A request that never answers **throws** — which is precisely what the shared backend produces while gunicorn recycles its wedged single worker (7–10× per shard, measured by #1074). Playwright kills it at the suite's `actionTimeout` (20 s, `playwright.config.ts`):

```
TimeoutError: apiRequestContext.get: Timeout 20000ms exceeded.
  - → GET http://localhost:7860/api/v1/auto_login
```

**144 files / 135 specs** call this helper, so any of them can be killed by an event that has nothing to do with what it tests:

| Where | What happened |
|---|---|
| Daily #1057 | `agent-context-id-isolation` on `google / gemini-2.5-flash` — 3 attempts, all dead in `getAuthToken` inside `SimpleAgentTemplatePage.load`; the test body never ran, and triage recorded it as signature `unknown` |
| PR #1080, both CI runs | Same target, same death, twice. Zero failures of the behaviour under test — the PR merged with a written justification because the gate could not go green |
| Local, 2026-07-29 | 3 aborts across ~24 runs on nightly `1.12.0.dev8`/`dev9` |

## Fix

A thrown request is retried on a **2 s / 8 s / 20 s** backoff (`DEFAULT_RETRY_DELAYS_MS`, exported so a test can assert the sizing). An answered non-ok response is **not** retried — that is an environment without auth, not an outage, and retrying it would slow every caller down.

**Why this budget and not the one the issue proposed.** The issue asked for a single bounded retry, sized against a worker recycle (~15–20 s: the container log times a full boot at ~18 s, 11 s of it the component registry). Working it showed that shape would not have flipped a single observed failure — in the runs above, *every* test-level attempt died on `auto_login` while a sibling spec recovered on its third, i.e. the backend was returning on a ~1-minute scale. The budget spans what was measured, not what is convenient. Recorded here rather than quietly shipped.

**It does not hide the wedge.** Every retry logs:

```
⚠️  auth: /api/v1/auto_login did not answer (apiRequestContext.get: socket hang up) — retry 1/3 in 2000ms. The backend is wedged or recycling (see #1077).
```

**Budget exhausted rethrows the ORIGINAL error.** It must never degrade into the empty-token fallback — the callers would carry on unauthenticated and fail somewhere far less diagnosable.

## Validation

**Live, against a real outage** — the helper called while the container restarted:

```
⚠️  auth: … socket hang up — retry 1/3 in 2000ms
⚠️  auth: … socket hang up — retry 2/3 in 8000ms
RESULT: token=Bearer eyJhbGciOiJ… after 15s
```

The old code threw on the first hang.

**Unit tests** — `tests/helpers/auth/get-auth-token.test.ts`, 9 cases, `npm run test:units`:

| # | Test | Pins |
|---|---|---|
| 1 | first call succeeds | happy path unchanged, 1 call |
| 2 | non-ok → `""`, no retry | asserted with a NON-empty budget, so it actually proves the answered path never consumes it |
| 3 | ok without token → `""`, no retry | same |
| 4 | one throw, then the token | recycle survived |
| 5 | throws until the third attempt | daily #1057's shape |
| 6 | budget exhausted | `assert.rejects(/Timeout 20000ms/)` — never `""` |
| 7 | bounded | one initial attempt + one per delay = 4, no infinite loop |
| 8 | waits between attempts | backoff really sleeps |
| 9 | default budget ≥ 25 s over 3 attempts | guards the sizing decision itself |

**Force-fail — 4 mutations, all executed, all reverted (`grep FF-MUTATION` → 0), green after:**

- `M1` no retry at all → **6 failed**
- `M2` exhaustion returns `""` instead of rethrowing → **2 failed**
- `M3` default shrunk to `[2000]` → **1 failed** (case 9 earns its place)
- `M4` retry answered non-ok responses too → **passed 9/9 at first.** The hole was in my own case 2, which disabled the budget and so proved nothing; hardened, it now **fails 2**. Worth stating plainly: the force-fail found this, review did not.

**Gates:** `npm run test:units` → 123 passed / 0 failed · `npm run typecheck` ✅ · `npm run lint` exit 0 (0 errors) · 19 API specs that exercise the helper green · `agent-context-id-isolation` (the #1060 spec) 2/2 green in 31 s.

## Still exposed (not in this issue's scope)

Three specs call `/api/v1/auto_login` directly instead of going through the helper, and keep the old single-attempt behaviour: `api-key-expiry-enforcement.spec.ts:26`, `api-custom-component-creation.spec.ts:23`, `api-keys-timezone-display.spec.ts:40`. Worth a follow-up; deliberately untouched here.

### DIFF
--- a/tests/helpers/auth/get-auth-token.test.ts
+++ b/tests/helpers/auth/get-auth-token.test.ts
@@ -1,0 +1,118 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import type { APIRequestContext } from "@playwright/test";
+import { getAuthToken, DEFAULT_RETRY_DELAYS_MS } from "./get-auth-token";
+
+/**
+ * The helper is called by 135 specs, and the daily's shared backend goes
+ * unresponsive mid-run (#1074/#1077) on a ~1-minute scale — long enough that
+ * every one of Playwright's test-level attempts died on `auto_login` for one
+ * target of daily #1057. These tests pin the contract that keeps that from
+ * killing an unrelated spec: a thrown request is retried across a bounded
+ * budget, a non-ok answer is not retried at all, and a budget that runs out
+ * surfaces the ORIGINAL error — never an empty token that would send the
+ * caller on unauthenticated (#1086).
+ */
+
+interface FakeResponse {
+  ok: () => boolean;
+  json: () => Promise<unknown>;
+}
+
+// Each entry is one scripted outcome for the next call: a response or a throw.
+function fakeRequest(outcomes: Array<FakeResponse | Error>): {
+  request: APIRequestContext;
+  calls: () => number;
+} {
+  let calls = 0;
+  const request = {
+    get: async () => {
+      const outcome = outcomes[Math.min(calls, outcomes.length - 1)];
+      calls++;
+      if (outcome instanceof Error) throw outcome;
+      return outcome;
+    },
+  } as unknown as APIRequestContext;
+  return { request, calls: () => calls };
+}
+
+const okResponse = (token: string): FakeResponse => ({
+  ok: () => true,
+  json: async () => ({ access_token: token }),
+});
+
+const timeout = () =>
+  new Error("apiRequestContext.get: Timeout 20000ms exceeded.\n  - → GET /api/v1/auto_login");
+
+test("returns the Bearer header on the first successful call", async () => {
+  const { request, calls } = fakeRequest([okResponse("abc")]);
+  assert.equal(await getAuthToken(request), "Bearer abc");
+  assert.equal(calls(), 1);
+});
+
+test("a non-ok response still yields the empty-token fallback, with no retry", async () => {
+  // An environment without auth answers 401/403 immediately — that is not an
+  // outage, so retrying it would only slow every caller down. The budget is
+  // deliberately NON-empty here: with `[]` the assertion would hold even if
+  // the helper retried answered responses, and the check would prove nothing.
+  const { request, calls } = fakeRequest([{ ok: () => false, json: async () => ({}) }]);
+  assert.equal(await getAuthToken(request, { retryDelaysMs: [0, 0, 0] }), "");
+  assert.equal(calls(), 1, "an answered request must not consume the retry budget");
+});
+
+test("an ok response without a token yields the empty-token fallback, with no retry", async () => {
+  const { request, calls } = fakeRequest([{ ok: () => true, json: async () => ({}) }]);
+  assert.equal(await getAuthToken(request, { retryDelaysMs: [0, 0, 0] }), "");
+  assert.equal(calls(), 1);
+});
+
+test("survives a worker recycle: one throw, then the token", async () => {
+  const { request, calls } = fakeRequest([timeout(), okResponse("xyz")]);
+  assert.equal(await getAuthToken(request, { retryDelaysMs: [0] }), "Bearer xyz");
+  assert.equal(calls(), 2);
+});
+
+test("keeps retrying across the whole budget — the outage outlasts one attempt", async () => {
+  // Daily #1057's shape: the first attempts find nothing, a later one lands
+  // after the backend is back. A single retry would have failed here.
+  const { request, calls } = fakeRequest([timeout(), timeout(), okResponse("late")]);
+  assert.equal(await getAuthToken(request, { retryDelaysMs: [0, 0, 0] }), "Bearer late");
+  assert.equal(calls(), 3);
+});
+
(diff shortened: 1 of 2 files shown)


rafaelgiln:
## What the green checks here do and do not prove

`Run impacted E2E specs` reported **skipping** on this PR. That is not a quirk of the change — [pr-validation.yml#L148](https://github.com/oriontech-me/langflow-e2e/blob/main/.github/workflows/pr-validation.yml#L148) selects the job's targets with

```bash
git diff --name-only … -- ':(glob)tests/**/*.spec.ts'
```

i.e. **only changed `*.spec.ts` files**. A changed helper or POM is never mapped to the specs that consume it, so this PR — which touches a helper with **135 calling specs** — ran zero E2E in CI. The same blind spot covers every change under `tests/helpers/`, `tests/pages/` and the fixtures, which is precisely the highest-blast-radius category in the suite.

So the green checks attest typecheck, lint and the checklist guard. They do **not** attest that any spec exercised this helper. That evidence is local and in the PR body:

- 19 API specs that call `getAuthToken` — green
- `agent-context-id-isolation` (the #1060 spec) — 2/2 green, 31 s
- the live probe against a real restart — two `socket hang up`s, token on the third attempt at 15 s
- 123 unit tests, including the 9 new ones, with 4 executed force-fail mutations

Filed as its own gap so it is not lost in this PR's history.

Issue #1094, opened 2026-07-29
canvas-zoom-navigation: "Fit View centers every node" asserts a scale drop that Fit View is not required to produce


Victor-w-Madeira:
Found while validating #1093 on Nightly 1.12.0.dev7 (`--workers=1 --retries=0`).
Reproduces on clean `main`; absent from the 07-27..29 dailies, so it is environment-
dependent rather than a new product change.

## Symptom

`canvas-zoom-navigation.spec.ts:351` (`@stable`) fails:

```
expect(fitted.scale).toBeLessThan(MAX_ZOOM)   // MAX_ZOOM = 2
```

`fitted.scale` lands exactly on `2` after the `fit_view` click.

## Why the assertion is not sound

The flow under test is `createRunnableChatFlowViaApi` — **two** nodes (Chat Input, Chat
Output). Whether React Flow's `fitView()` produces a scale below `maxZoom` is a property
of *pane size vs node spread*, not of Fit View: on a pane wide enough for two nodes at
`scale(2)`, the fitted scale is clamped **at** `maxZoom` and the assertion fails while
Fit View is working correctly. The daily runner's viewport is apparently narrow enough
that it does not hit the clamp; a local window is not.

## What must not be lost when fixing it

The step's precondition (`overflowing.contained === false`, `:339`) is the teeth that stop
this test passing on a dead Fit View button. Relaxing `:351` to `toBeLessThanOrEqual`
would silently allow the no-op case that precondition exists to exclude. The real
observable is already asserted right below — `geometry.contained === true` plus the
centering deltas. Options worth weighing:

1. Give the flow enough node spread that the fitted scale is provably under `maxZoom`
   (deterministic, but couples the test to node positions).
2. Pin the viewport size in the spec so pane geometry stops varying between environments.
3. Drop `:351` and let containment + centering carry the assertion, keeping the overflow
   precondition as-is.

Option 2 plus 3 looks strongest: the test's stated contract is "every node inside the
pane, centered", not "the scale went down".

Issue #518, opened 2026-07-03
flaky/bug: export-import-flow export produces empty flow (nodes: []) — regression of #384


Victor-w-Madeira:
## Summary

`export-import-flow.spec.ts:66` ("export flow to JSON triggers success toast and produces a valid file") failed on its **first attempt** in the daily-stable run and only recovered on retry. The failure is **not timing noise** — the exported JSON contained `data.nodes: []`, i.e. an empty flow:

```
expect(received).toBeGreaterThan(expected)
Expected: > 0
Received:   0
  at export-import-flow.spec.ts:134
```

This is treated as a **suspected real product bug** (export reads a not-yet-persisted flow), not a pure flake.

## Why this matters

This exact empty-export symptom was already investigated and "fixed" under **#384**, which added a `waitForFlowSaveSettled(page)` guard (see the comment at `export-import-flow.spec.ts:92-96`) to wait for the debounced `PATCH /api/v1/flows/{id}` autosave before navigating away and exporting. **#384 is closed, yet the flake regressed** — so the current mitigation does not fully close the autosave→export race.

## Recurrence

| Date | Run | Result |
|---|---|---|
| 2026-07-02 | 28583683001 | flaky (2 attempts) |
| 2026-07-03 | 28654917894 | flaky — 1st attempt `nodes: []` |

Flaky on consecutive daily runs, plus a suspected real bug → **`@stable` is being removed** from this test in a dedicated PR (this issue stays open until the root cause is fixed and the tag can be restored).

## Deliverables (definition of done)

1. Root cause identified for why the export reads an empty flow despite `waitForFlowSaveSettled`.
2. Fix applied and verified (export reliably contains the added nodes).
3. **`@stable` restored** on `export-import-flow.spec.ts:66` — this is an explicit deliverable, not optional. The tag was removed only as a temporary mitigation; this issue is **not done** until the tag is back on the test.
4. This issue closed.

## Next steps

- Investigate whether the PATCH is genuinely settled, or whether the export path reads a stale/cached flow snapshot.
- Reproduce locally against `langflowai/langflow-nightly:latest`.

Related (previously closed): #384

PR #1100, opened 2026-07-29
fix(ui-ux): return the verified transform from the canvas viewport wait (#1099)

### LINKED ISSUE #1099: Return the verified transform from waitForViewportSettled, and two review nits from #1094
Follow-up of #1094 / PR #1098 (merged as `1e766a0`). Three review items deliberately
left out of that PR: none is a defect, and the fix there is correct without them.
Grouped into one issue because they touch the same two files and one validation run
covers all three.

### 1. `waitForViewportSettled` returns a third read, not the value it verified

`tests/tests-automations/regression/ui-ux/canvas-zoom-navigation.spec.ts`

```ts
return readViewport(page);   // re-reads the DOM after the poll already approved `current`
```

The poll validates `current` (two identical reads plus the `settledWhen`
postcondition) and then discards that string to read the viewport again. The value is
the same in practice — nothing moves the viewport in between — but the returned value
is not the verified value, which is the same class of hazard #1094 removed.

Same shape at the `fit_view` step, where `fitted` and `fittedTransform` come from two
separate reads even though the idempotence assertion treats the latter as canonical:

```ts
const fitted = await waitForViewportSettled(page, movedFrom(displaced));
fittedTransform = await readTransform(page);
```

**Proposal:** extract the regex out of `readViewport` into a pure
`parseViewport(transform: string): Viewport`, have the poll keep the transform it
accepted, and return `parseViewport(settledTransform)`. `readViewport` becomes
`parseViewport(await readTransform(page))`. Behaviour-neutral; it closes the remaining
staleness window and removes two CDP reads per wait.

This is the item that justifies its own validation run: the helper has 14 call sites
in the file.

### 2. The two clamp steps in test 1 lack `expect(clicks).toBeGreaterThan(0)`

Test 3 has it; the `zoom_out` and `zoom_in` clamp steps in test 1 do not. Without it,
a button already disabled on entry gives `clicks === 0`, so `movedFrom` is never
satisfied and the step fails as an opaque 15 s
`the canvas viewport never reached the expected settled state` instead of the
diagnostic value diff (`Expected: 0.25 / Received: 0.88`) — the outcome PR #1098
explicitly set out to avoid for the clamp steps.

Low real risk: the preceding step asserts both buttons enabled, so zero clicks
requires a product change that would already redden that assertion first. One line
each, and it makes the three clamp steps consistent.

### 3. Doc/comment nit: the `0.92` and `0.880331` figures do not reconcile

`docs/ui-ux/canvas-zoom-navigation.md` (the *Fit View scale* row) and the comment at
the `fit_view` step both read:

> the unclamped fit is `min(1000/1090, 672/315) ≈ 0.92` — measured `0.880331` live

`1000/1090 = 0.917` against a measured `0.880`; the ~4 % gap is `fitView()`'s padding.
As written, a future reader spends time hunting a discrepancy that is not one. Adding
*"before `fitView()`'s padding"* resolves it in both places. The argument itself
stands — both values are more than 2x away from the `2` bound.

### Explicitly out of scope

- Shortening `intervals: [150, 150, 150, 200]` to cut the ~150 ms the sentinel now
  costs every wait. The sibling helper measured the `fitView()` commit landing at
  20–47 ms, so a shorter first interval would be safe — but there is no pre-fix
  runtime baseline to size the win against, so it stays a guess.
- Removing the unreachable `current !== UNREAD_TRANSFORM` guard. Dead but harmless,
  and it documents the sentinel's intent.
- Unifying with `tests/helpers/ui/adjust-screen-view.ts`'s own
  `waitForViewportSettled`. That one seeds `previous = null` **and** separates its
  reads by 50 ms, so it does not carry the #1094 bug; it reaches ~108 specs and needs
  its own PR if `settledWhen` is ever worth porting there.

### Validation

Item 1 needs the full spec-validation pass (all four tests, `--trace=on`, force-fail
per test, zero backend errors). Items 2 and 3 ride along on that run. No tag change
expected: all four tests stay `@stable`. No `QA-CHECKLIST.md` change: the eight
bullets are already `[x]` and coverage is unaffected.

Victor-w-Madeira:
**Fixes #1099.** Closes #1099.

Follow-up of #1094 / #1098. Three review items left out of that PR deliberately —
none was a defect, and the fix there is correct without them.

## 1. The wait now returns the transform it verified

`waitForViewportSettled` proved a transform sound (two identical reads plus the
`settledWhen` postcondition) and then went back to the DOM for a fresh read whose
value nothing had checked — the same shape of hazard #1094 removed, one step later:
the guards would prove one transform sound and the caller would assert on another.

- `parseViewport(transform)` extracted as a pure function; `readViewport` becomes
  `parseViewport(await readTransform(page))`.
- The poll keeps the transform it accepted and returns it alongside the parsed
  viewport as `SettledViewport extends Viewport`. The field is **additive**, so all
  14 call sites compile unchanged.
- Fit View idempotence, which compares the raw string across two clicks, now
  asserts on that verified string instead of a third read.

## 2. `expect(clicks).toBeGreaterThan(0)` on test 1's two clamp steps

Test 3 already had it. Without it, a button already disabled on entry gives
`clicks === 0`, which the `movedFrom` wait can only report as an opaque 15 s
timeout instead of the diagnostic value diff — the outcome #1098 explicitly set out
to avoid for the clamp steps.

## 3. The `fitView()` padding clause

`1000/1090 = 0.917` against a measured `0.880331`; the ~4 % gap is `fitView()`'s
padding. Stated in the doc row and the comment so the arithmetic reconciles instead
of reading as a discrepancy to chase. The argument is unchanged — both values are
more than 2x away from the `2` bound.

## A dead assertion this PR introduced and then removed

Worth recording, because the first cut of item 1 made the spec weaker while looking
stricter. Applying `.transform` to the toolbar Fit View step turned

```ts
expect(refitted.transform).not.toBe(displacedTransform);
```

into an assertion that **cannot fail**: that step's wait is
`movedFrom(displacedTransform)`, so "the transform differs from the displaced one"
is the poll's own resolve condition. Asserting it on the returned value restates the
precondition — true by construction.

It also invalidated one of the force-fails: inverting that assert to `.toBe(...)`
fails by definition, so it proved the field carries the post-fit value and nothing
about the assertion having teeth.

That step therefore keeps a **fresh** read, now with the reason stated in the code,
in `waitForViewportSettled`'s docblock (the mirror-image rule: an assertion that
would merely restate its own `settledWhen` must read the DOM itself) and in the spec
doc. The general rule and its one exception are both written down so the next reader
does not "tidy" it back.

## Scope

- No assertion weakened or removed; the `contained === false` precondition and
  `expect(fitted.scale).toBeLessThan(MAX_ZOOM)` are untouched.
- Behaviour-neutral by design: the change removes CDP reads, it does not alter what
  any test asserts.
- No tag change — all four tests stay `@stable`. No `QA-CHECKLIST.md` change: the
  eight bullets are already `[x]` and coverage is unaffected.

## Validation (nightly `1.12.0.dev9`, `--workers=1 --retries=0`)

Full pass on the pre-review-fix diff:

- typecheck ✅ · eslint ✅ 0 errors · `test:scripts` 147/147 ✅ · `test:units` 136/136 ✅
- QA-CHECKLIST guard ✅ · coverage guard ✅
- **5 clean runs, no flake:** 16.1s / 17.6s / 19.7s / 23.4s / 15.5s — same band as
  #1098, so the refactor costs no time
- `--trace=on` ✅ 1 passed in 6.3s (the #518 hang does not affect this family)
- zero `🚨 Backend Error` / `HTTP error(s) detected` ✅
- flow cleanup ✅ — `GET /api/v1/flows/` returns 27 flows, all starter templates,
  zero `Runnable Chat Flow` orphans
- force-fail ✅ — executed per test, reverted, 0 mutation markers left:

| Mutation | Result |
|---|---|
| `zoomUntilClamped` never clicks (test 1) | fails at the new guard — `Expected: > 0 / Received: 0` |
| `ZOOM_STEP` 1.2 → 1.5 (test 1) | fails — `Expected: 1.5 / Received: 1.2000031806218343` |
| `fittedTransform := "MUTANT"` (test 2) | fails — `Received: "translate(-588.052px, -476.054px) scale(0.880331)"` |
| `WHEEL_DELTA` 300 → −300 (test 4) | fails — `Expected: < 0.880331 / Received: 1.33433` |

Re-validation of the review fix (typecheck, lint, clean runs, and a test-3
force-fail that weakens the predicate to a bare settle **and** drops the click — so
the wait resolves on the displaced viewport and the fresh read is what reddens)
is pending and will be posted here before merge.

### DIFF
--- a/docs/ui-ux/canvas-zoom-navigation.md
+++ b/docs/ui-ux/canvas-zoom-navigation.md
@@ -57,6 +57,13 @@ never fail. Reading the DOM again keeps that assertion independent of the wait, 
 it still catches a viewport that snaps back after settling — and it is what fails
 if the predicate is ever weakened.
 
+The same one-read rule applies on the way **in**. The wheel test needs the
+pre-gesture transform twice — as numbers, to place the pointer anchor in flow
+space, and as a string, to pin `movedFrom` — and takes both from a single read
+(`parseViewport(await readTransform(page))`). Two reads could describe two
+different viewports, and the drift between them is precisely what the anchor
+tolerance then measures. There is no `readViewport` helper for that reason.
+
 Four independent tests:
 
 1. **Zoom in / Zoom out** — `zoom_in` multiplies the viewport scale by `1.2` per
@@ -112,9 +119,9 @@ part of test 3's enabled-controls assertion), and the minimap.
 | Zoom-out clamp | `zoom_out` becomes `disabled` with scale exactly `0.25`; `zoom_in` still enabled |
 | Zoom-in clamp | `zoom_in` becomes `disabled` with scale exactly `2`; `zoom_out` still enabled |
 | Fit View precondition | at scale `2` the nodes' union box is NOT contained in the pane rect |
-| Fit View scale | the fitted scale is strictly below `maxZoom` (`2`). Sound for this fixture, not an accident of the pane: the two nodes span `1090 × 315` flow px, so the unclamped fit is `min(1000/1090, 672/315) ≈ 0.92` before `fitView()`'s padding, which brings it to the `0.880331` measured live — either way a factor of ~2.3 away from the bound (#1094) |
+| Fit View scale | the fitted scale is strictly below `maxZoom` (`2`). Sound for this fixture, not an accident of the pane, and the arithmetic closes exactly: the two nodes span `1090 × 315` flow px against a `1000 × 672` pane, and Langflow's toolbar handler calls `fitView({ padding: { left: "20px", right: "20px", top: "80px" } })` (`CanvasControlsDropdown.tsx`), so the fit is width-driven: `(1000 − 40)/1090 ≈ 0.881` against the `0.880331` measured live, agreeing to the rounding of the quoted span (`960 / 0.880331 = 1090.5` flow px). Either figure is a factor of ~2.3 away from the bound (#1094) |
 | Fit View centering | union box inside the pane rect (1 px tolerance) AND \|union center − pane center\| ≤ 4 px on both axes AND both node titles visible |
-| Fit View idempotence | second `fit_view` click leaves the `transform` string byte-identical |
+| Fit View idempotence | second `fit_view` click leaves the `transform` string byte-identical — conditional on no node being **selected** between the clicks, since the handler's right padding jumps to `340px` when the inspection panel is open with a selection; this spec never selects a node |
 | Toolbar collapsed | `main_canvas_controls` visible; `fit_view`/`zoom_in`/`zoom_out`/`reset_zoom` `count() === 0` |
 | Toolbar expanded | all four controls visible after clicking `canvas_controls_dropdown`; `fit_view` and `reset_zoom` enabled |
 | Toolbar Fit View wired | nodes NOT contained before the click; after it the transform differs, every node is inside the pane, and `zoom_in`/`zoom_out` are both enabled |
--- a/docs/ui-ux/canvas-zoom-navigation.md
+++ b/docs/ui-ux/canvas-zoom-navigation.md
@@ -43,6 +43,20 @@ pane) rather than on "something changed". Only the two steps where no change is
 expected — editor hydration and the second, idempotent `fit_view` click — use the
 bare settle.
 
+The wait also **returns the transform it accepted**, parsed, rather than reading
+the DOM again once the poll clears (#1099). A fresh read would hand the caller a
+value none of the guards above had checked — the same shape of hazard, one step
+later: the wait would prove one transform sound and the assertion would run on
+another. Fit View idempotence, which compares the raw string across two clicks,
+therefore asserts on that verified string.
+
+The exception is the toolbar Fit View's "it moved" check, which keeps a **fresh**
+read on purpose. Its wait already demands `movedFrom(<displaced>)`, so asserting
+on the returned transform would restate the predicate's own precondition and could
+never fail. Reading the DOM again keeps that assertion independent of the wait, so
+it still catches a viewport that snaps back after settling — and it is what fails
+if the predicate is ever weakened.
+
 Four independent tests:
 
 1. **Zoom in / Zoom out** — `zoom_in` multiplies the viewport scale by `1.2` per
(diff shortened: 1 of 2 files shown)


Victor-w-Madeira:
## Re-validation of the review fixes (`ab2736e`)

The validation the description left pending, plus the two review items applied on top.

### What changed since the validated diff

1. **The wheel test's anchor and predicate now come from one read.** It read the pre-gesture transform twice — parsed for the pointer anchor, raw to pin `movedFrom` — so the two could describe different viewports, and the drift between them is precisely what the anchor tolerance measures. `readViewport` is gone with it: no wrapper, nothing to invite a second read.
2. **The padding clause states the arithmetic instead of approximating it.** Confirmed at the source: `CanvasControlsDropdown.tsx:85-93` calls `fitView({ padding: { left: "20px", right: "20px", top: "80px" } })`, so the fit is width-driven at `(1000 − 40)/1090 ≈ 0.881` — the `0.880331` measured live, agreeing to the rounding of the quoted span (`960/0.880331 = 1090.5` flow px). Same source records a precondition the byte-identical idempotence assertion depends on: that right padding jumps to `340px` with the inspection panel open on a selected node, so the assertion holds because this spec never selects one.

`normalizeViewport` deliberately keeps its `Promise<Viewport>` return type rather than widening to `SettledViewport`: the narrow type is what stops a future caller from pinning a `movedFrom` on a `baseline.transform` read long before the action — the false-red mirror of the hazard this PR fixes.

### Results (nightly `1.12.0.dev9`, `--workers=1 --retries=0`)

- typecheck ✅ · eslint ✅ 0 errors (2 pre-existing `no-force-option` warnings) · QA-CHECKLIST guard ✅ · coverage guard ✅
- **4 clean runs, no flake:** 16.4s / 17.0s / 15.1s / 15.7s — same band as #1098
- zero `🚨 Backend Error` / `HTTP error(s) detected` ✅
- flow cleanup ✅ — `GET /api/v1/flows/` returns 27 flows, zero `Runnable Chat Flow` orphans

### Force-fails

Executed, reverted, file `shasum` verified identical to the pre-mutation backup, 0 mutation markers left.

| Mutation | Result |
|---|---|
| **(the one this description owed)** test 3: predicate weakened to a bare settle **and** the click dropped | fails at the *fresh read* — `expect(received).not.toBe("translate(-1199.14px, -1401.31px) scale(2)")`. The deliberate exception has teeth: that line, not the wait, is what reddens |
| test 4 step 1: predicate weakened to a bare settle **and** the `wheel` dropped | fails with a value diff — `Expected: < 0.880331 / Received: 0.880331` (which also confirms the `0.880331` behind the padding arithmetic) |
| test 4: anchor fed from a viewport other than the one `movedFrom` is pinned to | fails — `Expected: <= 2 / Received: 64.48`. The two-read divergence is observable, so fix 1 is load-bearing rather than cosmetic |

PR #1118, opened 2026-07-30
refactor(provider-setup): dedupe the inline provider skip maps into provider-health (#1043)

### LINKED ISSUE #1043: Dedupe the 19 inline getProviderSkipReasons() copies into provider-health.ts
## Problem

#1029 added `tests/helpers/provider-setup/provider-health.ts` as the single implementation of the provider-health skip rule, and pointed the provider-**hardcoded** specs at it (PR #1042).

The provider-**parametrized** specs got there first, but each one carries its own inline copy:

```ts
function getProviderSkipReasons(): Map<string, string> {
  const jsonPath = path.resolve(__dirname, ".../data/providers.json");
  if (!fs.existsSync(jsonPath)) { console.warn(...); return new Map(); }
  const records = JSON.parse(fs.readFileSync(jsonPath, "utf-8")) as ProviderRecord[];
  ...
}
```

Nineteen byte-similar copies, plus a fourth hardcoded variant (`getGoogleSkipReason`) in `mcp-client-agent-gemini-tool-regression.spec.ts:66`:

- `mcp/client/mcp-client-agent.spec.ts`
- `mcp/client/mcp-client-agent-gemini-tool-regression.spec.ts` (the `getGoogleSkipReason` variant)
- `core-functionality/llm-agents/agent-*.spec.ts` — 17 files (`system-prompt`, `multi-tool-selection`, `input-sources`, `context-id-isolation`, `context-id-continuity`, `max-iterations`, `tool-name-validation`, `markdown-output`, `max-tokens`, `multimodal-image-input`, `tool-inspection`, `component-regression`, `empty-refusal-response`, `tool-error-handling`, `current-date-tool`, `structured-output`)

## Why this is cleanup, not a fix

**These specs already honor provider health correctly** — nothing is broken and no run is at risk. This is purely about the copies drifting: a change to the skip contract (the fail-open rule, the reason format, the `IGNORE_PROVIDER_HEALTH` escape hatch) currently has to be made in twenty places, and #1029's own audit found the copies had *already* diverged in shape (`Map`-returning vs single-provider, warn-on-missing vs silent).

Deliberately kept out of #1042 to hold that PR to one concern.

## Proposal

Export the parametrized shape from `provider-health.ts` — the `Map<string, string>` of inactive providers the `agent-*` specs consume — and delete the inline copies. Unit-test it alongside the existing matrix in `provider-health.test.ts`.

Mechanical and covered: every touched spec is `@stable`, so the change is visible in the next daily.

## Done when

- [ ] `provider-health.ts` exports the parametrized reason map; `provider-health.test.ts` covers it
- [ ] All 19 inline `getProviderSkipReasons()` copies and the `getGoogleSkipReason` variant are gone
- [ ] The `fs` / `path` / `ProviderRecord` imports they pulled in are dropped where now unused
- [ ] `typecheck`, `lint`, `test:units` clean; the touched specs behave identically on a run with one provider inactive

Victor-w-Madeira:
**Fixes #1043.**

## Problem

#1029 made `provider-health.ts` the single implementation of the provider-health skip rule and pointed the provider-**hardcoded** specs at it (PR #1042). The provider-**parametrized** specs got there first, and each carried its own inline copy:

- 16 `agent-*.spec.ts` files + `mcp/client/mcp-client-agent.spec.ts` → `getProviderSkipReasons(): Map<string, string>`
- `mcp/client/mcp-client-agent-gemini-tool-regression.spec.ts` → a single-provider `getGoogleSkipReason()` variant

**18 call sites**, not the 19+1 the issue's title counts — the issue body enumerates only 16 `agent-*` names, and this PR corrects the arithmetic rather than propagating it.

Nothing was broken: the copies honour provider health correctly. The cost is that the skip contract lived in 18 places — and they had **already drifted**, measurably on `main`:

| copies | behaviour on a missing `providers.json` |
|---|---|
| 15 | `console.warn("… run collect-models.spec.ts first …")` |
| 1 (`agent-component-regression`) | warns about **`collect-providers.spec.ts`**, a spec that does not exist |
| 1 (`mcp-client-agent`) | silent |
| 1 (the gemini variant) | silent, single-provider shape |

None honoured `IGNORE_PROVIDER_HEALTH`, and all rendered a nullable `error` as the literal `inactive — null`.

## Fix

One exported `providerSkipReasons()` in `provider-health.ts` replaces all 18, unit-tested alongside the existing matrix. **Net −199 lines** across the specs.

Three deliberate differences from the copies, each of them the contract this module already owned elsewhere:

1. **`IGNORE_PROVIDER_HEALTH=1` empties the map.** The escape hatch for a stale local `providers.json` now works for the parametrized specs too; before, getting a stale-`inactive` provider's targets back meant deleting the gitignored file, even though `providerSkipGate` honoured the same variable. **Verified safe:** the variable is set in no workflow, no script, `playwright.config.ts`, `globalSetup` or `.env.example` — it appears only in `docs/`, as a local override. CI cannot un-skip a dead-key provider through it.
2. **A nullable `error`** reads `no reason recorded by collect-models` instead of `inactive — null`.
3. **The read goes through `readProviderHealth`**, which fails OPEN on an absent or unparseable file. Named explicitly in the doc comment because it *widens* something: the copies **threw** on malformed JSON (a loud spec-load failure), while these 18 specs now warn and run. That is the module's #980 contract, but it applies to more specs than before.

The reason string is built once (`inactiveReason`) and shared with `unavailableReason`, so the parametrized and hardcoded gates cannot drift apart in what they quote to the report — a unit test pins that they agree for the same record.

Two docs that would have re-created the problem are fixed in the same PR:

- **`CONTRIBUTING.md` → "Model parameterization pattern"** handed the next author the inline copy verbatim, including the `ProviderRecord` import this PR prunes. After the dedupe it was the only remaining occurrence of `getProviderSkipReasons` in the repo. It now shows the shared call.
- **`docs/collect-models.md` → "Who consumes the recorded health"** described the two families as separate mechanisms and scoped the escape hatch to the hardcoded one. Both are false as of this commit; it now says both families share `provider-health.ts`, and carries a warning that a stray `IGNORE_PROVIDER_HEALTH=1` in a developer's `.env` will send the whole agent family at a dead key locally.

## Scope note

No test logic, tags, spec docs or `QA-CHECKLIST` bullets change — every touched spec keeps its behaviour and its tags. Both checklist guards are green.

**Selector reach, and a gap I closed by hand:** `tests/helpers/**` is in the import graph, so this selects **35 specs**; at `IMPACTED_SPEC_CAP=20` the job runs 20 and lists 15 dropped — and **both `mcp/client` files are in the dropped set**. Those are precisely the two non-mechanical rewrites (the silent copy and the single-provider variant), so I ran both locally; results below. Note also that `mcp-client-agent.spec.ts` carries no `@stable`, so the issue's rationale ("every touched spec is `@stable`, so the change is visible in the next daily") does not hold for it — another reason it is covered here by hand.

## Validation (nightly 1.12.0.dev10, `--workers=1 --retries=0`)

- `npm run typecheck` ✅ · `npm run lint` ✅ (0 errors; warnings 1307 → 1306) · `npm run test:units` ✅ **147/147** · `npm run test:scripts` ✅ 147/147 · `npm run check:checklist-coverage` ✅ · checklist guard ✅
- **11 new unit tests** on `providerSkipReasons` (41 in the file): the inactive→reason mapping, active providers absent from the map, fail-open on `null`/`[]`, the escape hatch and its exact-`"1"` contract, the nullable-`error` line, parity with `unavailableReason` for the same record, independence from env keys, that the zero-argument form really reads the file, that an armed hatch skips the read entirely, and that a missing file warns.
- **Mutation-tested**, after the review found three survivors — all three now fail: pointing the default at `null` so the file is never read (#1029's failure mode, restorable by a one-word edit), deleting the `console.warn`, and loosening the hatch to any truthy value. The `records` / `jsonPath` test seam this needed matches the one `readProviderHealth` and `writeProviderHealth` already take.
- **Behavioural parity, proven end to end** by marking `openai` inactive in the local (gitignored) `providers.json` with a sentinel error:
  - the target skips and the report carries `Provider "openai" inactive — SENTINEL-1043: pretend drained key` — identical in shape to what the inline copies produced (`2 skipped`);
  - re-running the same target with `IGNORE_PROVIDER_HEALTH=1` **runs and passes (24.1s)** — the behaviour change, demonstrated rather than asserted.
- Live runs on the restored `providers.json`:
  - `agent-max-tokens` + `mcp-client-agent-gemini-tool-regression` → **2 passed, 1 skipped (41.4s)** (re-run after the review restructure). The skip is the gemini spec, because this machine's `collect-models` records google `inactive` ("no models collected from the providers panel") — the shared gate firing for real, on the file's default path.
  - `mcp-client-agent.spec.ts` → **1 passed (26.4s)** (the file the CI cap drops and the daily never runs).
- Zero `🚨 Backend Error`.

## Done when

- [x] `provider-health.ts` exports the parametrized reason map; `provider-health.test.ts` covers it
- [x] All inline `getProviderSkipReasons()` copies and the `getGoogleSkipReason()` variant are gone (`git grep` → 0, including the one that lived in `CONTRIBUTING.md`)
- [x] The `fs` / `path` / `ProviderRecord` imports they pulled in are dropped where now unused (`ProviderRecord` in all 18; `fs` only in the gemini file, where it became unused; `path`/`fs` retained wherever `getModelsFromJson` still needs them)
- [x] `typecheck`, `lint`, `test:units` clean; the touched specs behave identically on a run with one provider inactive — proven above for both the parametrized family and the hardcoded gemini variant

Closes #1043

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -154,20 +154,10 @@ import fs from "fs";
 import { test, expect } from "../../../../fixtures/fixtures";
 import { SimpleAgentTemplatePage, type LoadSimpleAgentOptions } from "../../../../pages";
 import { hasProviderEnvKeys, type Provider } from "../../../../helpers/provider-setup";
-import type { ProviderRecord } from "../../../../helpers/provider-setup/collect-models";
+import { providerSkipReasons } from "../../../../helpers/provider-setup/provider-health";
 
 if (!process.env.CI) {
   dotenv.config({ path: path.resolve(__dirname, "../../../../.env") });
-}
-
-// Read inactive providers to display as skipped in output
-function getProviderSkipReasons(): Map<string, string> {
-  const jsonPath = path.resolve(__dirname, "../../../../helpers/provider-setup/data/providers.json");
-  if (!fs.existsSync(jsonPath)) return new Map();
-  const records = JSON.parse(fs.readFileSync(jsonPath, "utf-8")) as ProviderRecord[];
-  return new Map(
-    records.filter((r) => r.status === "inactive").map((r) => [r.provider, `Provider "${r.provider}" inactive — ${r.error}`])
-  );
 }
 
 // Read models and apply the .env strategy
@@ -175,7 +165,9 @@ function getTestTargets() {
   const jsonPath = path.resolve(__dirname, "../../../../helpers/provider-setup/data/models.json");
   if (!fs.existsSync(jsonPath)) return [];
   const allModels = JSON.parse(fs.readFileSync(jsonPath, "utf-8"));
-  const skipReasons = getProviderSkipReasons();
+  // Inactive providers, so their targets report as skipped with the collected
+  // reason. Never inline this — one shared implementation (#1043).
+  const skipReasons = providerSkipReasons();
   // apply strategy filter (see agent-component-regression.spec.ts for full implementation)
   return allModels.map((m: any) => ({
     label: `${m.provider} / ${m.model}`,
--- a/docs/collect-models.md
+++ b/docs/collect-models.md
@@ -152,19 +152,31 @@ place: all five components fail on missing credentials, none reaches the network
 ### Who consumes the recorded health
 
 Writing `inactive` is only half the mechanism — a spec has to obey it. Two kinds
-of spec do:
+of spec do, and since #1043 **both go through
+`tests/helpers/provider-setup/provider-health.ts`**, which is the single
+implementation of the rule:
 
 - **Provider-parametrized** specs (the `agent-*` family, `mcp-client-agent`) build
-  their target list from `models.json` and drop a target whose provider is
-  `inactive`, quoting the recorded reason.
-- **Provider-hardcoded** specs gate through
-  `providerSkipGate(...)` in `tests/helpers/provider-setup/provider-health.ts`
-  (#1029). That helper is the single implementation of the rule: missing env key
-  first, then the recorded `inactive` reason. It **fails open** when
-  `providers.json` is absent or unparseable — a fresh clone has no file (it is
-  gitignored) and CI is allowed to run with a failed `Collect models` step (#980),
-  so "no signal" must never skip the suite. `IGNORE_PROVIDER_HEALTH=1` overrides a
-  stale local file.
+  their target list from `models.json` and call `providerSkipReasons()` for the
+  `provider → reason` map, dropping a target whose provider is `inactive` and
+  quoting the recorded reason. Each of them used to carry its own inlined copy of
+  that map (18 of them, already drifted); #1043 deleted the copies.
+- **Provider-hardcoded** specs gate through `providerSkipGate(...)` (#1029), which
+  adds one precedence rule the map does not need: a missing env key is reported
+  before a recorded `inactive`, because without the key the provider cannot be
+  configured at all.
+
+Both **fail open** when `providers.json` is absent or unparseable — a fresh clone
+has no file (it is gitignored) and CI is allowed to run with a failed
+`Collect models` step (#980), so "no signal" must never skip the suite. And both
+honour `IGNORE_PROVIDER_HEALTH=1`, which overrides a stale local file.
+
+> **The escape hatch is local-only, and it is now blunter than it was.** The
+> variable is set in no workflow, script or config — only ever exported by hand.
+> Before #1043 it affected the hardcoded specs alone; it now also un-skips every
+> provider-parametrized target. Since these specs load `.env` whenever `CI` is
+> unset, leaving `IGNORE_PROVIDER_HEALTH=1` in a `.env` file will send the whole
+> agent family at a dead key on your next local run.
 
 Before #1029 the hardcoded specs gated on env-var presence, so a key that existed
 but was drained still made the live call. On run 30374528125 that hung two Google
--- a/tests/helpers/provider-setup/provider-health.test.ts
+++ b/tests/helpers/provider-setup/provider-health.test.ts
@@ -22,6 +22,7 @@ import * as os from "os";
 import * as path from "path";
 import {
   degradeProviders,
+  providerSkipReasons,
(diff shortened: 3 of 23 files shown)


Victor-w-Madeira:
## Attribution of the red `Run impacted E2E specs`

**83 passed, 2 failed, 2 flaky** — the specs ran; nothing mass-skipped, which is the failure mode a broken skip gate would produce. All four failures are **external dependency**, not this change. Every one of them is a spec-level assertion about the URL-fetch tool's output, and the received values name the cause:

| test | received |
|---|---|
| `agent-multi-tool-selection.spec.ts:420` (openai **and** anthropic) | `fetch_content output did not contain /Sample Slide Show/i: … 503 Service Temporarily Unavailable` / `Error loading documents: No documents were successfully loaded from any URL` |
| `agent-tool-inspection.spec.ts:299` (openai) | `fetch_content block found but input/output did not match — input={"urls":["https://httpbin.org/json"]} output={… 503 Service Temporarily Unavailable` |
| `agent-input-sources.spec.ts:252` (anthropic) | `sentinel not yet persisted in any flow` — a persistence-timing assertion, also downstream of the gate |

The fetched pages answered **503** from the runner. Not one of the traces touches `provider-health.ts` or the skip path.

`agent selects the URL tool for a fetch prompt` is a known recurrent offender independently of this branch: it hard-failed in dailies 29087610808 and 29911701167 and flaked in 29489622983. It reproduced locally on `main`-equivalent code during this PR's validation, with the same 503.

The job also logged five `500 Internal Server Error — POST /api/v1/flows/` and a run of teardown 404s, i.e. the shared-backend saturation signature (#1077), not something this PR introduces.

Behavioural parity of the gate itself was proven separately and is in the PR body: marking `openai` inactive in a local `providers.json` produces `Provider "openai" inactive — <collected reason>` and skips the target, and `IGNORE_PROVIDER_HEALTH=1` then runs the same target green.

Issue #1012, opened 2026-07-28
shardguard counts blob files, not tests — a 0-test run reaches triage looking benign


Victor-w-Madeira:
The `shardguard` step measures **blob file presence**, not test count. When every shard dies politely — aborting in `globalSetup` but still uploading a valid, empty blob — the guard reports `complete=true`, the merged `results.json` carries zero tests, and the whole downstream chain treats the run as a normal red day with nothing to triage.

Surfaced while triaging #1007 (daily run [30351107916](https://github.com/oriontech-me/langflow-e2e/actions/runs/30351107916), 2026-07-28), where all four shards aborted before the first test. Root cause of the abort is tracked separately; this issue is about the run being **indistinguishable from a benign one**.

## What the guard actually checks

`.github/workflows/daily-stable.yml:414`:

```bash
EXPECTED="${{ needs.prep.outputs.shard_total }}"
FOUND="$(find all-blobs -maxdepth 1 -name '*.zip' | wc -l | tr -d ' ')"
if [ "$FOUND" -lt "$EXPECTED" ]; then ... complete=false ... else complete=true ... fi
```

Each of the four shards uploaded a **1312-byte** blob containing only the `globalSetup` error and no test results. `FOUND=4`, `EXPECTED=4` → `complete=true`. The guard written to catch "a shard died before producing a blob" missed **four dead shards**, because they all produced one.

## Downstream consequences

| Consumer | Gate | What happened |
|---|---|---|
| `Fail scheduled run if shard blobs were incomplete` (`:665`) | `complete == 'false'` | **Skipped.** The only step whose job is to shout about an under-counted merge stayed silent. |
| `Auto-remove @stable from hard failures` (`:602`) | `complete == 'true'` | **Ran** on an empty `results.json`. Nothing to remove, so no damage — but only by luck of the abort landing in `globalSetup`. The action's internal mass-failure guard was never even reached. |
| `Append daily history` (`:537`) | `always()` | Wrote `passed: 0, failed: 0, flaky: 0, skipped: 0`, `duration_ms: 456176` (7.6 min vs the usual 14–22). Faithful, but semantically ambiguous. |
| `Create issue on failure` (`:616`) | `needs.test.result == 'failure'` | Opened #1007 with the body **"No per-test `@stable` hard failures were auto-removed."** Reads as benign. |
[...]

Issue #1129, opened 2026-07-30
chore: the .gitignore comment on the generated provider-setup files is in Portuguese


Victor-w-Madeira:
## Problem

`CLAUDE.md` → **Language** states that all content in this repository must be
written in English, without exception, and lists inline code comments explicitly.
One comment in `.gitignore` is in Portuguese:

```
.gitignore:27:# Artefatos gerados localmente — cada dev gera via collect-models.spec.ts
```

It documents the two generated provider-setup files on the lines below it
(`tests/helpers/provider-setup/data/models.json` and `providers.json`).

## Scope — this is the only occurrence

A sweep over every tracked file, excluding `docs/` and `.claude/`, returns exactly
two files:

- `.gitignore:27` — the violation above.
- `tests/assets/flows/group_test_iadevs.json` — **not** a violation. The
  Portuguese text is the *payload* of a prompt inside a flow fixture (a
  prompt-engineering guide used as component input), i.e. test data, not repo
  prose. Rewriting it would change what the fixture exercises.

So the fix is one line. Filed separately rather than folded into an unrelated
change so the language rule has a traceable enforcement record.

## Deliverables (Done when)

- [ ] The comment on `.gitignore:27` is in English and still says what those two
      paths are and who generates them.
- [ ] No other tracked file outside `docs/` and `.claude/` carries Portuguese
      prose in a comment or heading (re-verified with the same sweep).
- [ ] `tests/assets/flows/group_test_iadevs.json` is left untouched, and the
      reason is stated in the PR so the next sweep does not re-flag it.

Issue #1128, opened 2026-07-30
qa-infra: go-httpbin is wired into daily-stable only — PR/nightly/manual still hit public httpbin.org and red on its outages


rafaelgiln:
The external-echo decoupling built for #462/#639 exists in **one** workflow. Every other lane still calls the public host, so a third-party outage reds a PR that changed nothing related.

Surfaced on PR #1122 (run [30537081047](https://github.com/oriontech-me/langflow-e2e/actions/runs/30537081047)).

## Evidence

`Run impacted E2E specs` → `2 failed / 4 flaky / 74 passed`. Both hard failures are the same third party answering 503, and **8** occurrences of `503 Service Temporarily Unavailable` appear in the job log:

| Spec (line) | Failure |
|---|---|
| `core-functionality/llm-agents/agent-multi-tool-selection.spec.ts:439` | `Expected: "fetch-tool-returned-expected"` / `Received: "fetch_content output did not contain /Sample Slide Show/i: … 503 Service Temporarily Unavailable …"` |
| `core-functionality/llm-agents/agent-tool-inspection.spec.ts:318` | `Expected: "tool-inspection-captured"` / `Received: "fetch_content block found but input/output did not match — input={\"urls\":[\"https://httpbin.org/json\"]} output={… 503 Service Temporarily Unavailable"` |

The PR under test changed `tests/helpers/flows/load-template-by-name.ts`; neither spec has anything to do with it. It was selected only because the import graph reaches it (#1054), which is correct behaviour — the point is that a correct selection was reded by a third party.

## Root cause: the fix was never extended past `daily-stable`

`daily-stable.yml` self-hosts `ghcr.io/mccutchen/go-httpbin:2.23.1` as a service container and resolves `ECHO_BASE_URL` to its container IP. The specs read that variable and fall back to the public host when it is unset:

```ts
process.env.ECHO_BASE_URL ?? "https://httpbin.org"      // agent-multi-tool-selection, agent-tool-inspection
process.env.ECHO_BASE_URL ?? "https://postman-echo.com" // api-component-regression, api-request-component-regression
```

Service containers per workflow today:

| Workflow | Services | `ECHO_BASE_URL` |
|---|---|---|
| `daily-stable.yml` | langflow, ollama, **go-httpbin** | resolved to the container IP |
| `pr-validation.yml` | langflow | **unset** → public fallback |
| `nightly.yml` | langflow, ollama | **unset** → public fallback |
[...]

Issue #1108, opened 2026-07-30
Extract the flow-cleanup tracker: 50 hand-copied variants, 4 axes of drift


Victor-w-Madeira:
## Problem

The "capture every flow the page creates from its `POST /api/v1/flows` → 201 responses, delete them id-scoped in `afterEach`" block is **hand-copied into 50 spec files**. It is the suite's standard id-scoped cleanup (#490/#681/#515, never the delete-all of #553), but there is no shared implementation — so every copy drifted independently, and a fix to one does not reach the other 49.

Measured on `main` (`fe966d4`), over the 50 files that carry both `createdFlowIds` and a `page.on("response")` listener:

| axis | majority | minority |
|---|---|---|
| accumulator type | **49** use `string[]` | 1 uses `Set` |
| settles pending `resp.json()` before cleaning up | — | **1** |
| navigates off the canvas before deleting | — | **2** (`about:blank`) |
| failed `deleteFlow` is silently swallowed | **45** `.catch(() => {})` | 9 log it |
| URL match | **49** `.includes("/api/v1/flows")` | 4 check the exact `pathname` |

## Why each divergence matters

- **The unsettled-body race (49 files).** The listener calls `resp.json()`, which resolves a tick later, so an id can land in the array *after* `afterEach` has already read it. The flow then leaks permanently — the last test in a worker has no later hook to sweep it. Only `api/flows/api-component-regression.spec.ts` holds those reads and settles them first (`pendingCaptures`, added in #1105).
- **Deleting under a mounted editor (48 files).** #1023 measured this: an editor left on the canvas keeps polling `GET /flows/{id}/events?since=` and 404s once the flow is gone; the fixture logs each as `🚨 Backend Error` and the deterministic pipeline's VALIDATE gate hard-stops on them. #1103 fixed it in the folder specs with an unconditional `about:blank`. Worth stating honestly: this does **not** fire in every spec — a probe on `api/flows/api-component-regression.spec.ts` (failure forced *after* the component run, editor mounted, flow deleted underneath) logged zero backend errors either way, because the build's event stream is already closed. It bites the specs whose editor is still polling.
- **Silent cleanup failure (45 files).** `deleteFlow` throws on purpose — it already absorbs 404-as-done and one transient 5xx, so anything left is a real failure (401/403/422). `.catch(() => {})` turns exactly the signal the helper exists to raise back into a silent leak.
[...]

PR #1133, opened 2026-07-30
fix(fixtures): make the HTTP error monitor cover 4xx/5xx and say what it does (#1084)

### LINKED ISSUE #1084: The fixture's HTTP error monitor cannot fail a test, and covers 4 status codes rather than 4xx/5xx
Raised from @rafaelgiln's review of #1052, where every test in a 20-test slice logged
`500 /api/v1/store/tags` → `[Errno -2] Name or service not known` (the container could not
resolve the Langflow Store) and no test failed. The question asked there was whether that is
a deliberate allowlist. It is not — there is no allowlist to be deliberate about.

## What the code actually does

`tests/fixtures/fixtures.ts` monitors responses and collects two error kinds, then:

```ts
// after use(page)
if (flowErrors.length > 0 && !allowFlowErrors) {
  throw new Error(...);
}
```

Only `flow_error` can fail a test. An `http_error` is pushed into the array, counted in the
`⚠️ N HTTP error(s) detected` line, and then **nothing happens to it** — on any path, for any
endpoint. So the `🚨 Backend Error:` line is purely advisory, and the only thing standing
between a real backend 500 and a green test is a human reading the terminal.

That human step is written down as a validation requirement — `CONTRIBUTING.md` → *Test
validation checklist* step 5 ("Check the terminal logs") and `CLAUDE.md` → step 4 ("Confirm no
backend errors logged"). It is the load-bearing part of the mechanism, and the Store noise
above degrades it: a log that always carries three 500s is a log nobody reads carefully.

## Second, narrower finding

The monitor is not the `4xx/5xx` filter both docs describe:

```ts
status === 400 || status === 404 || status === 422 || status === 500
```

Four exact codes. **401, 403, 405, 409, 502, 503 and 504 are invisible to it** — including the
`403 — No authentication credentials provided` shape that shows up under contention (seen in
the same #1052 review, and in #773), and the `502`/`503` a wedged or restarting backend
produces (#1030/#1048). `CLAUDE.md:86` claims "automatic backend HTTP error monitoring
(4xx/5xx)"; that sentence is wrong today whichever way this issue is resolved.

## The decision to make

Not obvious, which is why this is an issue rather than a PR:

1. **Make `http_error` fail the test** — matches what both docs imply, but needs an allowlist
   first, or the Store 500 alone turns every spec red. Also needs a `page.allowHttpErrors()`
   escape hatch for specs that assert error responses on purpose (the API specs do), mirroring
   `allowFlowErrors()`.
2. **Keep it advisory and make the log trustworthy** — allowlist the known-benign endpoints
   (`/api/v1/store/*` when the Store is unreachable) so a logged error means something, and fix
   the two docs to say "logged, never fails".
3. **Widen the status filter** — orthogonal to 1 vs 2, and cheap, but widening it while errors
   stay advisory only adds noise to a signal nobody is required to act on.

Whichever way it goes, `CLAUDE.md:86` and the two checklist steps have to end up describing
the real behaviour.

## Not in scope

The `store/tags` 500 itself is environmental (no outbound DNS in the container), not a Langflow
defect. It is the example that exposed the gap, not the thing to fix.

## Evidence

- `tests/fixtures/fixtures.ts` — the status filter and the throw that only covers `flow_error`
- #1052 review thread — 20/20 green with 3 logged HTTP errors per test
- `CONTRIBUTING.md` → *Test validation checklist* step 5; `CLAUDE.md:86` and its step 4

rafaelgiln:
Closes #1084

## The problem

`CLAUDE.md` advertised "automatic backend HTTP error monitoring (4xx/5xx)". Neither half of that sentence was true.

**It matched four exact codes**, not a range: `status === 400 || status === 404 || status === 422 || status === 500`. Measured live on Langflow Nightly 1.12.0.dev9 while resolving this — an in-page `405`, and the `403 — No authentication credentials provided` an unauthenticated `GET /api/v1/flows/` returns, were both **invisible**, as were the `502`/`503` a wedged or restarting backend produces (#1030/#1048). `execution-error-notification` had already worked this out and mocked a **503 specifically to slip past the filter**, saying so in a comment.

**And an `http_error` never fails a test** — it is logged, counted, and then nothing happens to it, on any path, for any endpoint. Only `flow_error` throws. So the load-bearing part of the mechanism is a human reading the terminal, which makes noise the real defect: in the #1052 review every test in a 20-test slice logged three `500 /api/v1/store/tags` (no outbound DNS in the container), and a log that always carries the same three 500s is a log nobody reads carefully.

## The decision, and the measurement behind it

**HTTP errors stay advisory.** Making them fail was measured before being rejected, not assumed away:

| Question | Measured |
|---|---|
| Would the 15 API specs that assert 4xx on purpose break? | **No.** `page.request` traffic does not emit `page.on("response")` — verified by probe. Immune either way. |
| Deliberate page-traffic error mocks outside the auth exemption | **2 specs** (503 and 500) |
| Ambient noise over 48 `@stable` tests (ui-ux + project-management) | **2 events**, both `DELETE /api/v1/flows/` → 500 — an ambient upstream bug in *cleanup*, not in any test's subject |
| Extra noise from widening 4 codes → all 4xx/5xx | **0** |

So the blast radius is small — but 48 of **429** `@stable` tests is not a basis for flipping a suite-wide gate. This PR ships the mechanism and honest documentation and leaves enforcement to be decided on a full daily's data, which `PW_HTTP_ERROR_DEBUG=1` now produces.

## What changed

- **`tests/fixtures/http-error-policy.ts` (new)** — `classifyHttpError()` as a pure function: every 4xx/5xx on an `/api/` route, minus exemptions that each must carry a **reason** (auth endpoints; the external Langflow Store, unreachable from CI). Extracted because inside the fixture it only ran in a live browser, so "does the monitor see a 403?" was answerable only by running the suite and reading a terminal. Now covered by `npm run test:units` on every PR.
- **`page.allowHttpErrors()` + a `PageWithErrorHooks` type** — for specs that drive an endpoint into an error on purpose. `llm-invalid-api-key-ui` already carried a comment claiming it allowed exactly this and called `allowFlowErrors()`, which never covered the HTTP side: that spec logged its own mocked 500 as a backend error on every run.
- **The summary line now says what happens next**, because nothing does: `⚠️ N HTTP error(s) detected — ADVISORY: these do NOT fail the test.`
- **Docs corrected** — `CLAUDE.md` (the fixture description and validation step 4) and `CONTRIBUTING.md` (the import note and step 5) now state that `flow_error` fails and `http_error` only logs, and that reading the log is therefore the only gate.

### One extra defect, found while verifying this

The error was pushed into the array **after** `await response.text()`, inside an async handler the fixture teardown does not wait for. An error late in a test raced the summary and disappeared from it — measured: a 405 printed `🚨 Backend Error` with **no** `📋 Found N backend error(s)` line at all. Intermittent, not constant (both errors in the 48-test slice were counted). Since that summary is the entire gate for HTTP errors, an undercount is the one failure the mechanism cannot afford. Fixed by recording before reading the body.

## Validation

Against Langflow Nightly **1.12.0.dev9**, re-run on this branch after #1119 landed.

- `npm run test:units` — **181/181** (9 new) · `npm run test:scripts` — **157/157**
- `npm run typecheck` — 0 errors · `npm run lint` — 0 errors (the new files produce **zero** warnings) · `npm run check:checklist-coverage` — pass
- **50 `@stable` tests** (ui-ux + project-management + `llm-invalid-api-key-ui`) with the widened filter live: **50/50 in 3.3 min**, one single `🚨` — the same ambient `DELETE /api/v1/flows/` 500. Widening added **zero** noise.
- Touched specs: **4 passed, 1 skipped** (the `test.fixme` quarantined under #1063), no `🚨`.
- Wiring verified live: a 405 **is** now reported; `allowHttpErrors()` suppresses it; Store and auth responses are ignored **with the reason printed** under `PW_HTTP_ERROR_DEBUG=1`.
- Flow orphans: 30 before, 30 after — nothing leaked.

**Force-fail (5, executed and reverted):**

| # | Mutation | Result |
|---|---|---|
| FF1 | filter narrowed back to the four codes | *"the codes the old filter missed are now monitored"* fails |
| FF2 | `DELETE /api/v1/flows/` added to the exemption list | 3 tests fail, including the *"stays VISIBLE"* guard — which exists precisely to catch someone silencing that 500 to quieten the log |
| FF3 | an exemption with an empty reason | *"every ignore verdict carries a reason"* fails |
| FF4 | hatch removed from `llm-invalid-api-key-ui` | the spec goes back to logging its **own mock** as a Backend Error on both tests, every run |
| FF5 | revert proven | 0 mutation markers, all green |

## Honest limits

- Measured 50 of 429 `@stable` tests. Sufficient for this change — nothing newly fails, and the worst case is one extra log line — and deliberately insufficient for the enforcement flip, which is why the debug switch ships with it.
- The `DELETE /api/v1/flows/` 500 is **left visible on purpose**. It is a real Langflow 500, and FF2 exists to stop it being allowlisted later just to quieten the log.

## ⚠️ Reviewer: this is a suite-wide change — the PR E2E job does not cover it

`tests/fixtures/fixtures.ts` is one of the paths `pr-validation.yml` treats as suite-wide. The
import-graph selector resolves this diff to **all 236 specs**, and `IMPACTED_SPEC_CAP` selects
20 — so **216 specs are dropped** from the PR run (listed in the job log and run summary, per
#1012's rule). Verified locally with the same script:

```
$ git diff --name-only --diff-filter=d origin/main...HEAD | \
    node scripts/impacted-specs-by-import.mjs --stdin --format=json --cap 20
fullSuite: true   specs resolved: 236   selected: 20   dropped: 216
```

Every spec in the suite imports this fixture, so **dispatch `manual.yml` for a full run before
merging.** The local evidence above covers 50 `@stable` tests; the PR job adds 20 more; neither
is the whole suite.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -84,7 +84,7 @@ Do not restate the current wave here — it changes every cycle; follow the poin
 
 ### Test Infrastructure
 
-- **`tests/fixtures/fixtures.ts`** — Always import `test` from here, never directly from Playwright. It extends the base `test` with automatic backend HTTP error monitoring (4xx/5xx) and flow execution error detection. Provides `page.allowFlowErrors()` for tests that intentionally trigger failures.
+- **`tests/fixtures/fixtures.ts`** — Always import `test` from here, never directly from Playwright. It extends the base `test` with backend HTTP error monitoring and flow execution error detection. **The two are not equally strong, and the difference decides how much a green run is worth (#1084):** a `flow_error` **fails** the test; an `http_error` is **logged and never fails it**, on any path, for any endpoint — so the only thing between a real backend 500 and a green test is a human reading the log (checklist step 4 below, `CONTRIBUTING.md` step 5). Which responses reach that log is decided by `tests/fixtures/http-error-policy.ts` — every 4xx/5xx on an `/api/` route except the documented exemptions (auth endpoints; the external Langflow Store, unreachable in CI). It genuinely covers 4xx/5xx now; before #1084 it matched four exact codes and silently missed 401/403/405/409/502/503, which is why `execution-error-notification` could mock a 503 specifically to slip past it. Escape hatches: `page.allowFlowErrors()` for tests that provoke execution failures, `page.allowHttpErrors()` for tests that drive an endpoint into a 4xx/5xx on purpose (keeps the advisory log trustworthy). `PW_HTTP_ERROR_DEBUG=1` prints what the policy ignored, with reasons.
 
 - **`tests/pages/`** — Page Object Model (POM). `BasePage.ts` provides common navigation; `FlowEditorPage.ts`, `PlaygroundPage.ts`, `MainPage.ts`, `LoginPage.ts`, `SidebarComponent.ts` provide feature-specific selectors and actions.
 
@@ -137,7 +137,7 @@ importable, `testMatch`/discovery gotchas — in `CONTRIBUTING.md` → **Unit te
 1. Run with full trace (`--trace=on`) and verify steps match screenshots
 2. Force a failure to confirm no false positives
 3. Walk through in debug mode (`--debug`)
-4. Confirm no backend errors logged (`🚨 Backend Error:`)
+4. Confirm no backend errors logged (`🚨 Backend Error:`) — **this step is a real gate, not a formality: an HTTP error never fails the test on its own** (#1084)
 5. Update `QA-CHECKLIST.md` coverage symbols
 
 **PR review checklist** — request changes if any of these are missing:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,9 @@ test.describe("Area or feature name", () => {
 });
 ```
 
-> Always import from `fixtures` — never directly from Playwright. The base fixture adds automatic backend error monitoring.
+> Always import from `fixtures` — never directly from Playwright. The base fixture adds
+> backend error monitoring: flow execution errors **fail** the test, HTTP errors are
+> **logged only** (step 5 below explains why that distinction matters).
 
 **4. Use existing helpers and pages**
 
@@ -257,8 +259,22 @@ npx playwright test path/to/test.spec.ts --debug
 
 The base fixture prints backend errors automatically. Look for:
 
-- `🚨 Backend Error:` — unexpected HTTP error
-- `🚨 Flow Error Detected` — silent failure in flow execution
+- `🚨 Backend Error:` — unexpected HTTP error. **Logged, never fails the test** (#1084)
+- `🚨 Flow Error Detected` — silent failure in flow execution. **Fails the test** unless
+  the spec called `page.allowFlowErrors()`
+
+Because an HTTP error cannot fail a test, **this step is the only thing standing between a
+real backend 500 and a green run** — the fixture prints
+`⚠️  N HTTP error(s) detected — ADVISORY` to say so out loud. Treat any line under it as a
+finding to explain, not noise to scroll past.
+
+Which responses reach that log is decided by `tests/fixtures/http-error-policy.ts` (every
+4xx/5xx on an `/api/` route, minus documented exemptions: auth endpoints, and the external
+Langflow Store which is unreachable from CI). If your spec drives an endpoint into a 4xx/5xx
+**on purpose** — including mocking one with `page.route` — call `page.allowHttpErrors()` so
+the deliberate error stays out of the log instead of teaching readers to ignore it. Adding an
+endpoint to the exemption list makes it invisible to all 235 specs, so it needs a reason that
+survives review; run with `PW_HTTP_ERROR_DEBUG=1` to see what is currently being ignored.
 
 **6. Update the checklist**
 
--- a/tests/fixtures/fixtures.ts
+++ b/tests/fixtures/fixtures.ts
@@ -1,5 +1,29 @@
 // tests/fixtures.ts
-import { test as base, expect } from "@playwright/test";
+import { test as base, expect, type Page } from "@playwright/test";
+import { classifyHttpError } from "./http-error-policy";
+
+/**
+ * The escape hatches this fixture bolts onto `page`.
+ *
+ * Specs reach them through `(page as any)` today; this type exists so the two
+ * names are discoverable and spelled consistently.
+ */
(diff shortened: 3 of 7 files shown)


rafaelgiln:
## CI triage: the E2E red is #1128, not this diff

`Run impacted E2E specs` → **2 failed, 1 flaky, 83 passed (8.8m)**. All three are the API Request
component reaching an external echo service:

| Test | URL it uses | Result |
|---|---|---|
| `api/flows/api-component-regression.spec.ts:451` | `https://httpbin.org/get` — **hardcoded** | failed |
| `api/flows/api-component-regression.spec.ts:713` | `https://httpbin.org/post` — **hardcoded** | failed |
| `core-components/api-request-component-regression.spec.ts:552` | `ECHO_BASE` → falls back to `postman-echo.com` | flaky |

The container logged `ComponentBuildError: Error building Component API Request` 36 times.

### Why this diff is not the cause

1. **It cannot fail a test.** `http_error` stays advisory — that is the decision this PR
   implements. None of the three files is in the diff.
2. **`pr-validation.yml` has no go-httpbin service.** `daily-stable.yml` is the only workflow
   that starts one, so `ECHO_BASE_URL` is unset here and the fallback is public.
3. **Two of the three hardcode `httpbin.org` in `fill()`**, bypassing `ECHO_BASE_URL` entirely —
   they hit public httpbin in *every* workflow, including the daily that self-hosts go-httpbin
   for exactly this reason. Three such call sites across two spec files.
4. **httpbin.org is degraded right now**: `/post` → `504`, `/get` → connect timeout. Reproduced
   locally — `:713` fails with the same 30 s `waitForSelector`, while the other two pass.

Both of those are the two findings already filed as **#1128**.

### Why it surfaced on this PR and not the last five

`tests/fixtures/**` is a suite-wide path: the import-graph selector resolves this diff to all 236
specs and the cap picks 20 `@stable`-first, which is how these specs entered the run at all. A
narrow PR never selects them — hence #1108, #1037 and #1129 going green. This is the first
suite-wide diff since go-httpbin was wired in, so it is the first to trip over the gap.

### Suggested order

Land **#1128** first (extend the go-httpbin service to `pr-validation`/`nightly`/`manual`, and
de-hardcode the three call sites); this PR then goes green without a change of its own. Re-running
the job only bets on httpbin's uptime, and merging past the red leaves the same trap armed for the
next suite-wide PR.

Issue #1107, opened 2026-07-30
api/flows: two API Request tests fail — advanced fields are no longer on the node body


Victor-w-Madeira:
## Symptom

Two tests in `tests/tests-automations/regression/api/flows/api-component-regression.spec.ts` fail on every attempt, and are the sole reason the `Run impacted E2E specs` check is red whenever that file (or anything reaching it) is touched:

| test | line | failure |
|---|---|---|
| `include_httpx_metadata=true adds request headers to output` | `:260` | `waitForSelector` on `toggle_bool_include_httpx_metadata` — timeout 10 s (`:268`) |
| `timeout error returns status_code 500 with error field` | `:292` | `locator.fill` on `int_int_timeout` — timeout 20 s (`:299`) |

## Evidence

- **CI**, PR #1105 run [30507197042](https://github.com/oriontech-me/langflow-e2e/actions/runs/30507197042/job/90759453439): 20 tests, **18 passed / 2 failed** — these two, failing on all 3 attempts (`retries=2`).
- **Verified pre-existing, not caused by #1105**: both fail with identical signatures on clean `main` (`3642880`), run locally against `1.12.0.dev9` with `--retries=0`. The #1105 diff touches neither test functionally — they only received the test-title reformat.

## Root cause

The two fields are no longer rendered on the node body. Since the nightly replaced the "Controls" modal with the node inspector side-panel, a parameter that is not added to the body only exists inside the inspector, so a bare `getByTestId("int_int_timeout")` / `toggle_bool_include_httpx_metadata` can never resolve.

The sibling spec already handles exactly this class — `core-components/api-request-component-regression.spec.ts:88-93`:

```ts
async function addTableFieldToBody(page: Page, field: "headers" | "body") {
  await page.getByTestId("title-API Request").click();
  await openAdvancedOptions(page);
  await page.getByTestId(`inspector-add-${field}`).click();
  await closeAdvancedOptions(page);
}
```

`openAdvancedOptions` / `closeAdvancedOptions` live in `tests/helpers/ui/open-advanced-options.ts`.

## Scope

Fix both tests using that pattern — select the node, open the inspector, either interact with the field there or add it to the body via `inspector-add-<field>`, then assert as before. The assertions themselves (status 200 + `"headers"` + `Langflow` in the output; `status_code` 500 with an `error` field) are unchanged and still describe the behaviour worth covering.

## Acceptance

- [ ] Both tests pass against the current nightly with `--retries=0`
[...]

Issue #868, opened 2026-07-21
[Daily #864] Parameters panel — table field edit count mismatch (recurrent)


rafaelgiln:
Spun out of daily-failure triage #864 (run [29822019255](https://github.com/oriontech-me/langflow-e2e/actions/runs/29822019255), 2026-07-21).

## Symptom

The parameters-panel table field edit does not persist as expected — the post-edit row/cell count does not match:

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests-automations/regression/core-components/parameters-panel-field-types.spec.ts:410` ("table field edit persists") | `expect(locator).toHaveCount(expected)` | `Error: expect(locator).toHaveCount(expected) failed` |

## Flake signal

Confirmed recurrent, same signature: failed on dailies **2026-07-20** and **2026-07-21** (`same_signature: true`). As prevention, `@stable` is being removed at triage so it stops running in the daily until this issue is worked:

- `tests-automations/regression/core-components/parameters-panel-field-types.spec.ts` (test at line 411)

Restoring `@stable` after the fix is a deliverable of this issue.

## Preliminary read (descriptive — NOT a verdict)

A `toHaveCount` mismatch after editing a table field means the expected number of rows/cells was not observed. Consistent with (a) the edit not persisting, (b) a changed table DOM structure (different count of matching elements), or (c) the assertion racing the re-render. None confirmed.

## Investigation directive

Investigate all paths independently, **product (Langflow) as prime suspect first**: on the current nightly, edit a table field in the parameters panel and confirm the edit persists and the row/cell count matches. Only after ruling out a product regression, consider: (a) table DOM/testid drift changed the matched-element count → update the locator; (b) the assertion races the re-render → wait on a deterministic post-edit observable.

## Deliverables (Done when)

- [ ] Root cause confirmed (product regression vs. table DOM/testid drift vs. race) with evidence on the current nightly.
- [ ] Spec passes reliably (multiple clean `--retries=0` runs).
- [ ] **`@stable` restored** in the fix PR — the tag was removed at triage as prevention; re-validate per `CONTRIBUTING.md` before restoring.

rafaelgiln:
## Verdict: transient-saturation (contention cluster) — infra-gated, no test fix

Investigated on nightly 1.11.0.dev50. **Not a product regression, not test/DOM drift.**

[...]

Issue #462, opened 2026-07-01
qa-infra: API Request GET test depends on external postman-echo — hard-fails on outages (daily 2026-07-01)


Victor-w-Madeira:
## Summary

`API Request component — GET request returns 200 and output Data contains all required fields` (`api-request-component-regression.spec.ts`, currently line 374) **hard-failed** in the daily `@stable` run of **2026-07-01** (Langflow `1.11.0.dev26`), failing all 3 internal retry attempts.

This is **not a Langflow regression** — it is the external echo dependency being unreachable/slow from the Langflow backend.

## Evidence

- Run: https://github.com/oriontech-me/langflow-e2e/actions/runs/28514231353
- Error: `page.waitForResponse` for `/api/v1/build/**/events` timed out at 45s (`runOnce`, spec line 133). The build event stream never arrived because the component's backend request to the default echo endpoint (`postman-echo.com`) hung.

## History — recurring external-dependency failure

This test (and its sibling verb tests) have a documented history of hard-failing on external-endpoint outages:

- **#383 / #407**: `httpbin.org` returned sustained `503`s → default endpoint moved to `postman-echo.com`, retry logic added to `runAndOpenOutput`.
- **2026-06-22 weekly**: this same GET test hard-failed (`transient upstream 5xx on all 3 attempts`).
- **2026-07-01 daily**: hard fail again (build/events timeout).

The 3-attempt retry only absorbs a blip *within a single test window*; a sustained outage of the public echo service defeats it.

## Decision (daily-cadence triage)

`@stable` **removed** from this test in the triage PR (closes #461). It should not gate the daily suite while it depends on a public third-party service.

## Deliverables

- [ ] Provide a reliable echo endpoint the Langflow backend can reach in CI (self-hosted `mccutchen/go-httpbin`), addressing the blocker documented in the spec's Notes: the component's `validators.url()` rejects single-label service hostnames and needs the service IP in `LANGFLOW_SSRF_ALLOWED_HOSTS`. Wire it via `ECHO_BASE_URL`.
- [ ] Confirm no product regression (the endpoint, once reachable, returns 200 with all required Data fields).
- [ ] **Re-add `@stable`** to the test (and to the spec doc's Tags section) once the endpoint is reliable.

Triage of #461.

PR #1134, opened 2026-07-30
ci: self-host the echo endpoint in every lane, not just the daily (#1128)

### LINKED ISSUE #1128: qa-infra: go-httpbin is wired into daily-stable only — PR/nightly/manual still hit public httpbin.org and red on its outages

rafaelgiln:
Closes #1128

## The problem

The external-echo decoupling built for #462/#639 shipped into **one** workflow. Every other lane kept calling public `httpbin.org` / `postman-echo.com` and reding on their outages.

Most recently **PR #1133**, which lost three API Request specs to an `httpbin.org` 504 while its own diff touched none of them — and whose log read like a product failure rather than a third party being down. Measured while resolving this: `httpbin.org/post` answered `504` and `/get` did not answer at all.

Service containers before this PR:

| Workflow | `go-httpbin` | `ECHO_BASE_URL` |
|---|---|---|
| `daily-stable.yml` | yes | resolved to the container IP |
| `pr-validation.yml` | **no** | unset → public fallback |
| `nightly.yml` | **no** | unset → public fallback |
| `manual.yml` | **no** | unset → public fallback |

## What changed

- **`.github/actions/resolve-echo-endpoint`** (new composite action) — the resolution logic **extracted** from `daily-stable.yml` rather than copy-pasted into three more workflows, and wired into `pr-validation`, `nightly` and `manual` alongside a `go-httpbin` service container. The daily now calls the same action instead of its inline step.
- **`scripts/resolve-echo-endpoint.mjs`** (new) holds the decision, covered by `npm run test:scripts` (12 tests).

  It is a script and not inline bash because the choice is genuinely non-obvious and **both ways to get it wrong fail as an unattributed component error**:
  - Langflow's `validators.url()` **rejects a single-label host**, so `http://go-httpbin:8080` can never be used even though the alias resolves — only a raw IP works (#462).
  - Its SSRF layer **blocks loopback outright**, so `localhost` cannot be handed to Langflow — yet that is exactly the address the **job** must probe on a host-based lane.

  So the address the job probes and the address Langflow calls are **not the same address**, and which is which depends on whether the job runs in a container: `daily-stable` does and resolves the alias with `getent`; the other three run on the runner host, where that lookup fails and the container IP comes from `docker inspect` (matched by image — GitHub names service containers `<hash>_<image-slug>_<hash>`, observed in PR #1133's own run log).

- **The failure mode is now explicit per lane** — the trade-off #1128 asked to be decided and recorded:
  - `mode: fail` on **PR / nightly / manual**. A silent fallback to the public host is precisely what made this look like a product bug, and on a PR a human is waiting on the verdict. On the nightly the stakes are higher still: answering "did the build regress?" from a public host means a third party's outage opens a regression issue against Langflow.
  - `mode: warn` on the **daily**, preserving today's behaviour, on the same reasoning as `Collect models` (#980) — a day of coverage for the dozens of specs that never touch the echo outweighs strictness.

- **`api-component-regression.spec.ts`** — the three `@release`-only tests no longer hardcode `httpbin.org`. The line drawn in #1107 (only the daily lane needs protecting) is what PR #1133 disproved.

## Two things the migration surfaced

**1. The cURL-mode test was proving the payload was sent by accident.** It scraped `textContent` off the output editor, with a comment noting the editor is virtualized and then assuming `evaluate()` returns the whole document. It does not. Measured against the self-hosted echo: **617 characters, cut off mid-token inside `"Host": [`**, with the payload never in it. It had been passing only because `httpbin.org`'s response was short enough for the payload to land inside the rendered window; go-httpbin renders one line per header value and pushed it out. An assertion that depends on how much of a response fits on screen proves nothing, so it now reads the full output via `readOutputJson` and asserts shape-agnostically (the echoed body lands under `json` on httpbin/go-httpbin, `data` on postman-echo).

**2. `nightly.yml` has not run since 2026-03-19** — its `schedule` block is commented out ("disabled until the suite is stable in production"), leaving `workflow_dispatch` only. `CLAUDE.md` claimed it "runs daily at 03:00 BRT"; corrected. Worth knowing that one of the lanes this PR instruments is currently dormant.

## Deliberately NOT migrated

Three `httpbin.org` literals stay, each with the reason recorded at the site — they are **not** external dependencies:

| Site | Why |
|---|---|
| `api-request-component-ui.spec.ts:51-52` | typed into a field and read back; the test never runs the component |
| `parameters-panel-field-types.spec.ts:201` | asserts field **persistence** via the API; no outbound call |
| `parameters-panel-field-types.spec.ts:242` | same |

`manual.yml`'s **external-URL** job is scoped out for a real constraint: it targets a Langflow that is not on this runner's Docker network and cannot reach a service container here. Self-hosting for it would need a publicly reachable echo — a larger decision than #1128 asked for.

## Validation

Against Langflow Nightly **1.12.0.dev9**.

- `npm run test:scripts` — **169/169** (12 new) · `npm run test:units` — **206/206**
- `npm run typecheck` — 0 errors · `npm run lint` — 0 errors · all four workflow YAMLs parse, and the action's step body passes `bash -n`

**The action replayed against real Docker, outside CI** (its bash extracted verbatim), covering all four outcomes:

| Case | Result |
|---|---|
| service up, host topology, `mode: fail` | resolves `172.18.0.2`, probes `localhost:8080`, exports `ECHO_BASE_URL` → exit 0 |
| service absent, `mode: fail` | **blocks** with the reason named → exit 1 |
| service absent, `mode: warn` | continues, naming the exposure → exit 0 |
| resolved but deaf, `mode: fail` | probe loop exhausts, **blocks** → exit 1 |

This is also what caught a real bug before CI could: the step opened `DECISION="$(node …` and closed only the `$(`, leaving the double quote unterminated — the step would have failed to parse on the runner.

**End-to-end proof against a real self-hosted echo** — go-httpbin plus an *isolated* Langflow on port 7861 with the CIDR allowlist (the developer's own instance was left untouched, and its flow count was 30 before and after):

- `api-component-regression.spec.ts` — **5/5**, including `"source": "http://172.18.0.2:8080/post"`, proving the whole chain: service → resolve → SSRF allowlist → `validators.url()` on an `IP:port` URL → assertions.
- `api-request-component-regression.spec.ts` — **14/15**, and the 15th green when run on its own (the file is serial). The one failure was `:736` on `addRowButton.click()` → 0 rows, the known add-row flake class (#868/#919) with no HTTP call involved; **2/2 green re-run in isolation**. Notably `:552` — one of the three specs that failed in PR #1133's CI — is green here.

**Force-fail (executed and reverted):**

| # | Mutation | Result |
|---|---|---|
| FF1 | assert a payload token the echo never returns | cURL-mode test fails |
| FF2 | point the POST test at `/get` | fails on the `"source"` assertion |
| FF3 | pre-fix state | the old `textContent` assertion failed against go-httpbin — the fix's own evidence |
| FF4 | action cases B/C/D above | both failure modes proven |
| FF5 | revert proven | 0 mutation markers, all green |

## Honest limits

- **Workflow YAML cannot be fully run locally.** The action's bash, both failure modes, and the docker discovery are proven above; what only CI can prove is the wiring in each lane. **This PR's own E2E job is that proof** — it touches workflows and API Request specs, so it exercises `pr-validation`'s new service and resolve step directly.
- The **agent** echo specs (`agent-multi-tool-selection`, `agent-tool-inspection`) were not run locally: they need live provider keys the isolated proof instance did not have. They already read `ECHO_BASE_URL`, so no code change was needed for them — only the service wiring, which the same step provides.

## Suggested order

Landing this **before** rebasing PR #1133 should turn that PR green with no change of its own, since its three failures were these same specs on the public host.


rafaelgiln:
## The E2E red here is #1136, not this diff — and the new wiring worked

**Good news first: the resolve step did exactly what it was built to do, in the real host-based topology** — the one thing that could not be proven locally:

```
discovery: getent="none" docker="172.18.0.3" in_container=false mode=fail
echo endpoint via docker inspect (host-based job): Langflow → http://172.18.0.3:8080, job probes http://localhost:8080
echo endpoint is up: probed http://localhost:8080, Langflow will call http://172.18.0.3:8080
```

`getent` found nothing (expected off-container), `docker inspect` resolved the IP, and the probe/Langflow address split held. **20 passed, 1 failed** — and the three API Request specs that reded PR #1133 on public httpbin are **green here**. That is this PR's deliverable-4 proof.

### The one failure

`api/flows/api-request-component-ui.spec.ts` → *"has a headers field accessible in the inspector"*, failing on all three retries.

Pre-existing, not caused by this diff:

1. This PR's change to that file is a **comment only** — six added lines, zero code.
2. It fails locally on 1.12.0.dev9.
3. It fails with **`origin/main`'s exact file** (line 85, before the comment shifted it).
4. No open issue tracked it.

It appeared now only because the comment touched the file, which **selected** it for the impacted-specs run. The spec is `@release` and never `@stable`, so `daily-stable.yml` does not run it — the "invisible red" class.

**Verdict: test defect, not a product regression.** `template.headers.advanced === true` on this build, so the field lives in the node inspector panel; the test looked for it on the node body and clicked a `/advanced/i` control the nightly replaced around dev46.

Filed as **#1136** and fixed in **#1137** (which also gives that file the id-scoped flow cleanup it never had — measured one orphan per test).

### Order

Merging #1137 first should make this PR's E2E job green with no change here.

Issue #995, opened 2026-07-27
edit-flow-name flake: waitForFunction 30s at rename-flow.ts:86 (1 in 6 under load) — classify and fix at the root


Victor-w-Madeira:
`edit-flow-name.spec.ts:7` — "user should be able to edit flow name and see it reflected in the main page listing" — lost `@stable` on 2026-07-13 (`5b78a1b`) as a recurrent flake; its tracker **#727 closed** without the flakiness being classified. Still flaky today, and now with a captured signature that #727 never had.

## Still flaky — reproduced

Langflow Nightly **1.12.0.dev7**, across three runs (`--retries=0`, mixed `--workers=1/2/4` with background load): **5 passed, 1 failed**.

The failure:

```
TimeoutError: page.waitForFunction: Timeout 30000ms exceeded.
  at tests/helpers/flows/rename-flow.ts:86
```

This is new information. #727 recorded recurrence by **test identity only** — it noted explicitly that `reports/daily-history.jsonl` does not store `error_signature` for flaky entries, so the CONTRIBUTING "same signature within 30 days" test could not be applied mechanically. There is now a signature, and it points at a specific line in a shared helper rather than at the test.

#727's own record: flaked on **07-03 and 07-10, both healthy days**, plus one hard failure on 07-08 (a mass-failure day) with a 20s click timeout on the main-page flow list. Same area — the rename/listing round-trip.

## Why this is not a restore candidate

It passes in isolation, which proves nothing: the removal reason was a flake, and isolation is the condition in which a flake passes. Restoring it on that basis is the error corrected in PR #986. The 1-in-6 rate under load is the actual signal.

## Deliverables (Done when)

- [ ] The `waitForFunction` at `helpers/flows/rename-flow.ts:86` is diagnosed: what it waits for, and why that condition intermittently never becomes true (list refresh race, stale query cache, or a genuine product delay in the listing).
- [ ] Classified as test-timing vs. product behaviour, with live evidence — the deliverable #727 left unfinished.
- [ ] Fixed at the root rather than by widening the timeout, then `@stable` restored in the same PR.
- [ ] Validated at `--workers=2 --repeat-each=3` at minimum, since a single pass cannot clear a flake.
- [ ] `Last validated` bumped on the spec doc.

`rename-flow.ts` is a shared helper — check its other callers before changing it.

## Reproduce

```bash
npx playwright test \
[...]

Issue #362, opened 2026-06-08
Weekly triage: Cluster 1 — Text Input/Output are now legacy and hidden from the sidebar (10 @stable tests)


Victor-w-Madeira:
## Summary

Langflow marked the **Text Input** and **Text Output** components as `legacy: true`. The component sidebar **hides legacy components by default**, so searching `"text input"` / `"text output"` now returns *"No components found"*. Every `@stable` test that drags these components from the sidebar fails at the `expect(getByTestId("input_outputText …")).toBeVisible()` step.

This is a **deterministic, permanent product change** — not a transient nightly outage and not a brittle selector. It will recur every week until the suite is updated.

## Evidence (reproduced on the exact CI image)

- Reproduced locally against the **same image digest the weekly ran** (`sha256:55a586ebb4ffbd87b0a07fc5469aa387a23eb60b32671e34fa2ce3cc9d7d93fc`).
- `/api/v1/all` on that image, category `input_output`:
  | name | display_name | legacy |
  |---|---|---|
  | ChatInput | Chat Input | `false` |
  | ChatOutput | Chat Output | `false` |
  | **TextInput** | **Text Input** | **`true`** |
  | **TextOutput** | **Text Output** | **`true`** |
  | Webhook | Webhook | `false` |
- Frontend filter that hides them: `getBooleanFromStorage("showLegacy", false)` and the `if (!showLegacy)` filter in `flowSidebarComponent/index.tsx` (lines 241 / 376). Toggled via `setLocalStorage("showLegacy", …)`.
- Confirmed end-to-end: with `localStorage.showLegacy = "true"` the `input_outputText Output` card reappears and *"No components found"* disappears.
- Weekly failure: run [27133642480](https://github.com/oriontech-me/langflow-e2e/actions/runs/27133642480) (2026-06-08).

## Affected `@stable` tests (10 — `@stable` removed in the PR closing #361)

| Test | Component |
|---|---|
| `core-components/general-bugs-delete-handle-advanced-input.spec.ts:11` | Text Input |
| `core-components/if-else-component-regression.spec.ts` — `129`, `159`, `189`, `223`, `260`, `292`, `323`, `361` | Text Output (×2) |
| `core-functionality/playground/output-modal-copy-button.spec.ts:20` | Text Input |

## Blast radius (not all `@stable` / not all ran this week — watch these too)

14 spec files reference the now-legacy Text Input/Output via the sidebar:
[...]

Issue #684, opened 2026-07-13
Validate & promote lock/unlock flow — §12.5


Victor-w-Madeira:
Wave 2 work item — pointer to a decided `QA-CHECKLIST.md` backlog bullet.

- **Section:** flow-functionality §12.5 Flow Operations
- **Intent:** Validate & promote to @stable (round-trip; note possible dedup between flow-lock.spec.ts and lock-flow.spec.ts)
- **Spec:** flow-functionality/flow-lock.spec.ts + lock-flow.spec.ts
- **Bullet(s):** Lock flow (prevents editing) + Unlock flow

Part of **ROADMAP.md → Wave 2 — Components, RAG, flows & observability** (validation-heavy wave; convergence ~59% → ~74%). Merge with all tests tagged `@stable` and a spec doc under `docs/` mirroring the path.

PR #1152, opened 2026-07-30
fix(project-management): enter edit-flow-name by id, and gate the header on write permission (#1005)

### LINKED ISSUE #1005: renameFlow residual flake: read-only save button and missing flow_name header (~7%, not the #995 clobber)
Surfaced while root-causing #995. Both signatures are **distinct from #995's clobber** — they fire *before* any rename logic runs, so the #995 fix does not address them — and both live in `renameFlow`, which is shared by 7 specs. Measured at ~7% across ~25 local runs on nightly **1.12.0.dev7**, the same rate the pre-#995 helper showed for non-clobber causes.

They are filed together because they were captured in the same validation arms and may share a trigger (flow-list refetch churn); split into two if the investigation shows otherwise.

## Signature A — `save-flow-settings` stays disabled, flow reads as read-only

```
Error: expect(locator).toBeEnabled() failed
Locator:  getByTestId('save-flow-settings')
Expected: enabled
Received: disabled
Timeout:  15000ms
   at helpers/flows/rename-flow.ts   (the first toBeEnabled, before the save click)
```

The call log is the interesting part: most polls resolve to a plain `<button disabled …>`, but **one** carries `title="(Read-Only)"`.

Upstream mechanism to verify: `useIsFlowReadOnly` (`contexts/permissionsContext.tsx`) is

```ts
return Boolean(flowId) && (isLoading || !can(flowId, "write"));
```

i.e. the flow editor **fails closed while the RBAC query is still resolving** — deliberately, per its own docstring ("flow editors fail closed so a denied user cannot briefly mutate the in-memory canvas"), even though `canPerformAction` is fail-open everywhere else. `FlowSettingsComponent` renders `disabled={disableSave || isReadOnly}` with `title={isReadOnly ? t("version.readOnly") : undefined}`, which matches the captured DOM.

Open questions for the investigation:

- Is it `isLoading` (a slow/refetching `POST /api/v1/authz/me/permissions`) or `!can(flowId,"write")` (a loaded map that genuinely omits `write`)? A 15 s window is long for a loading state, so both need ruling in/out.
- The provider (`flow-toolbar-options.tsx`) is keyed on `resourceIds: [currentFlowId]` and `domain: project:{folder_id}`. Does a flow-list refetch change the query key and re-enter `isLoading`?
- Does the suite need a shared "wait until the flow is writable" gate before any editor mutation? That would affect far more than `renameFlow`.

## Signature B — `flow_name` absent for 15 s

```
Error: expect(locator).toBeVisible() failed
Locator: getByTestId('flow_name')
Expected: visible
Error: element(s) not found
   at helpers/flows/rename-flow.ts   (first assertion in the helper)
```

`MenuBar` only renders the `flow_name` span when `onFlowPage` is true, so the editor never mounted. It fires from `edit-flow-name.spec.ts`'s loop, i.e. after either the templates-modal `Basic Prompting` click (iteration 1) or the `list-card-open-button` click (iteration 2) — the click landed but did not navigate.

## Environment caveat — read before chasing this

The local instance used for these measurements carries **~175 accumulated flows**, while CI starts clean. The home list render cost and the `flow-list` refetch volume are therefore not comparable, and both signatures may be local-only artifacts of that accumulation. Reproduce on a clean instance **first**; if they do not reproduce there, the finding is the accumulation itself (several specs — `flow-rename-header`, `general-bugs-save-changes-on-node` — still leak a flow per run) rather than a product defect.

## Done when

- [ ] Each signature reproduced (or shown not to reproduce) on a **clean** nightly instance.
- [ ] Signature A classified: RBAC query latency vs. a genuinely restrictive permission map, with the request/response captured.
- [ ] Signature B classified: click-not-navigating vs. slow editor mount.
- [ ] Decision recorded on whether a shared writable-flow gate belongs in the helpers, since `renameFlow` has 7 callers.
- [ ] If local-only: open the follow-up for the leaking specs instead.

## Reproduce

```bash
npx playwright test \
  tests/tests-automations/regression/core-functionality/project-management/edit-flow-name.spec.ts \
  --workers=2 --repeat-each=8 --retries=0
```

Raise to `--workers=4` with `tests/tests-automations/regression/core-components/api-request-component-regression.spec.ts` as load to make both signatures more frequent.

Victor-w-Madeira:
Closes #1005

## What this fixes

#1005 filed two signatures for the residual ~7% flake `edit-flow-name` carried
after #995, both inside `renameFlow`. Reproducing on a **clean** instance
(1.12.0.dev10, 27 starter projects) surfaced **four** distinct signatures — and
three of them trace to one cause that lives in the spec's ENTRY, not in the
helper.

| Signature | Where it landed | Verdict |
|---|---|---|
| `flow_name.hover()` -> `<html> intercepts pointer events` | `rename-flow.ts:37` | template-entry race |
| `flow_name` not found | `rename-flow.ts:34` | template-entry race |
| `save-flow-settings` stays disabled, dialog then unmounts | `rename-flow.ts:62` | write-permission flip |
| `home-dropdown-menu` not found after `icon-ChevronLeft` | this spec | **product defect — split out** |

### The template-entry race

The spec entered through `awaitBootstrapTest` -> templates modal -> click the
shared **Basic Prompting** card, with **nothing waiting for the navigation that
click starts**. "New Flow" eagerly creates a blank *placeholder* flow and opens
the welcome overlay on it; picking a template then creates a SECOND flow and
navigates to that one. `renameFlow` opens with `waitForFlowSaveSettled` (700 ms
of PATCH silence) and a `flow_name` visibility assertion — both of which are
**already satisfied by the placeholder's header**. So the helper drove the wrong
flow, mid-navigation, behind the welcome overlay.

The hover call log is the proof:

```
locator resolved to <span data-testid="flow_name" …>New Flow</span>
  - <html lang="en">…</html> intercepts pointer events
  - element was detached from the DOM
```

### The write-permission flip

`flow_name` is an `aria-hidden` `<span>` **inside** the `menu_bar_display`
button, which upstream renders `disabled={isReadOnly}` with

```ts
useIsFlowReadOnly = Boolean(flowId) && (isLoading || !can(flowId, "write"))
```

i.e. it fails **closed** for the whole time `POST /api/v1/authz/me/permissions`
is in flight. A span is not a form control, so Playwright's actionability check
never covered the disabled state and a click landed in that window was swallowed
with no error. The same flip landing *after* the dialog opened tears it down
(`<Popover open={openSettings && !isReadOnly}>`) and the remount resets the typed
name — leaving `save-flow-settings` disabled for the rest of the budget.

## Changes

| File | Change |
|---|---|
| `tests/helpers/flows/rename-flow.ts` | Open through the `menu_bar_display` **button** once it reports enabled (same gate #1063/#1147 added for `useAddComponent`); assert `toHaveValue(flowName)` after the fill |
| `.../edit-flow-name.spec.ts` | `createFlowFromStarter` + `page.goto('/flow/{id}')` entry; re-open by id; `afterEach` leaves the canvas before deleting |
| `docs/.../edit-flow-name.md` | Steps rewritten + new section *The template-entry race (issue #1005)* |

`renameFlow` is shared by **7 specs**, which is why the gate is in the helper.
The real Basic Prompting graph is kept deliberately — it is what carries the
mount autosave the #995 clobber rides on, and the validation run still logs the
expected re-apply warnings.

## Validation

Clean local instance, **Langflow Nightly 1.12.0.dev10**.

| Run | Config | Result |
|---|---|---|
| Before | 36x `--workers=4 --retries=0` | **6 failed / 30 passed** (16.7%) |
| After | 36x `--workers=4 --retries=0` | **36 passed**, 0 `Backend Error`, 0 leaked flows |

Other `renameFlow` callers: `flowSettings` PASS, `flow-rename-header` PASS (x2),
`run-flow` PASS. `agent-component-regression`, `general-bugs-save-changes-on-node`
and `auto-login-off` fail — **identically with the changes stashed** (control run
executed): providers inactive locally, the legacy Text Output component
(#362-364), and AUTO_LOGIN on. No failure occurs inside `rename-flow.ts`.

**Force-fail — executed, one mutation at a time, all observed failing, all
reverted** (`grep FF-MUTATION` -> 0 hits):

- `FF:` rename to `${targetName}FFMUTANT` -> read-back assertion failed
  (`Received: "hjppuheyw3sFFMUTANT"`)
- `FF:` `toHaveCount(1)` -> `toHaveCount(2)` -> failed (`Received: 1`)
- `FF:` `menu_bar_display` -> `menu_bar_display_FFMUTANT` -> failed **at the
  gate**, proving it is load-bearing rather than decorative
- `FF:` `fill(flowName + "FFMUTANT")` -> the new `toHaveValue` caught it
  (`Expected "7io5cisk8p4" / Received "7io5cisk8p4FFMUTANT"`)
- `FF:` comment out `createdFlowIds.push(flowId)` -> test still PASSES and leaks
  1 flow (29 -> 30), proving the cleanup is load-bearing

Gates: `tsc --noEmit` clean; eslint 0 errors on the touched files; QA-CHECKLIST
guard + coverage guard pass; `npm run test:units` 215 pass; `npm run test:scripts`
pass. Orphan audit after the 36-run burst: exactly the 27 starter projects, zero
leaks.

`--trace=on` was not used: it is the known local hang (#518/#490). Verification
rests on the `--retries=0` bursts, the executed force-fails, and the failure
screenshots.

**Cost:** median duration at 4 workers goes 22.3s -> 24.7s — three full document
loads per test instead of SPA hops.

## Review follow-up (post-review commits)

Merged `main` in (no force-push) and closed three gaps the review found. Each was
**measured**, not reasoned about, on a clean 1.12.0.dev10 instance.

| Finding | Change |
|---|---|
| The 4th signature is now fixable in-repo: #1156 landed `leaveFlowEditor` after this branch's base and skipped this file to avoid conflicting with this PR, with **nothing tracking the adoption** | the loop exits through `leaveFlowEditor(page, { escapeDeadlock: true })` |
| Dismissing the onboarding tooltip on entry is a **no-op** — upstream arms it on a **10 s** idle timer (`ONBOARDING_TOOLTIP_DELAY_MS`), so a probe at entry looks ~8 s too early and it pops mid-rename, over the settings dialog | `langflow-assistant-discovered` is seeded in localStorage before the first load |
| `getByText(targetName)).toHaveCount(1)` was satisfied by the flow **header** — the assertion never proved this test's own title | scoped to a `list-card` carrying a matching `flow-name-div` (a read, never a click) |

`escapeDeadlock: true` is safe *here specifically*: the rename's persistence is
asserted **after** the exit, against the server, and the next iteration re-enters
by id — so the escape's page load discards nothing this test relies on, and a
rename that never landed still fails the listing assertion.

**Force-fails — executed, one at a time, all reverted** (`grep FF-MUTATION` → 0):

- `FF:` comment out the editor exit → the **unscoped** assertion still **PASSED**
  (this is how the assertion gap was found); with the scoped assertion the same
  mutation fails with `Received: 0`, so exit and assertion are both load-bearing
- `FF:` rename the localStorage key → `assistant-onboarding-tooltip` **appears**
  within 20 s; with the real key it never appears. The exposure is real on a fresh
  Playwright context, and the seed removes it

**Re-validation:** 24 runs of `edit-flow-name` at `--workers=4 --retries=0` on the
final state — **24 passed**, zero `Backend Error`, zero leaked flows (27 starter
projects before and after). `flowSettings` + `flow-rename-header`: 9 passed. The
#995 re-apply warnings still fire, so that exposure is preserved. Gates:
`tsc --noEmit` clean, eslint 0 on the touched files, `test:units` 299 pass,
`test:scripts` 343 pass, checklist guard + coverage guard pass.

Also added, comment-only: `rename-flow.ts` now names the window the gate does
**not** close — `PermissionsProvider` keys on `domain: project:{folder_id}` off
`currentFlow`, which `setCurrentFlow(updatedFlow)` replaces on every save, so a
key change re-enters `isLoading` and can re-disable the button between the
assertion and the click. The fix, if it is ever observed, is to retry
open→dialog-visible — never a wider timeout.

### Note on the red E2E lane

Not attributable to this diff, provable from line ordering:
`agent-component-regression` fails at `:101` (its `renameFlow` is at `:119`, never
reached); `general-bugs-save-changes-on-node` fails at `:64`, i.e. **after**
`renameFlow` at `:57` succeeded; `auto-login-off` fails at `:124`/`:128`, before
its `renameFlow` at `:152`. On top of that `Collect models` was **skipped**
(`needs_models=false`), so no provider was configured — which is what makes
`value-dropdown-model_model` absent at `:101`. Filed as #1216.

## Split out, deliberately not fixed here

- `flow-rename-header.spec.ts` and `general-bugs-save-changes-on-node.spec.ts`
  leak a flow per run (confirmed: the first has no `afterEach`, the second no
  cleanup at all) — tracked in #1154. Re-confirmed locally during this
  re-validation: 3 runs of `flow-rename-header` left exactly 3 flows behind.
- `flowSettings.spec.ts` opens the header by clicking the aria-hidden `flow_name`
  span, carrying the same swallowed-click defect the helper just fixed — #1215.
- `openFlowById` is now a third near-verbatim copy, and the copies have already
  diverged (the two lock specs believe they are protected against an overlay they
  are measurably not protected against) — #1214.


Victor-w-Madeira (commented):
## Review

Checked the diff against the upstream frontend source, issue #1005, and this PR's CI run
(30581596587).

### Verified — what holds up

- **The write-permission mechanism is real, not inferred.** `pages/DashboardWrapperPage/index.tsx`
  mounts `PermissionsProvider` (`resourceIds: [currentFlow.id]`, `domain: project:{folder_id}`)
  **around `<AppHeader/>`**, so `FlowMenu`'s `useIsFlowReadOnly` resolves through it and
  `menu_bar_display` really does render `disabled={isReadOnly}` — added upstream in `887f2a552d`
  (2026-07-15, langflow-ai/langflow#14068). The disabled window exists and the gate is load-bearing.
- **Backward compatible.** On builds before that commit the button carries no `disabled` attribute,
  so `toBeEnabled()` resolves immediately — no risk for a `manual.yml` dispatch at 1.5.1 / 1.10.x.
- **The red E2E lane is not attributable to this PR**, provable from line ordering alone:
  `agent-component-regression` fails at `:101` (its `renameFlow` is at `:119`, never reached);
  `general-bugs-save-changes-on-node` fails at `:64`, i.e. **after** `renameFlow` at `:57` succeeded;
  `auto-login-off` fails at `:124`/`:128`, before its `renameFlow` at `:152`. On top of that,
  `Collect models` was **skipped** in that run (`needs_models=false`), so no provider was configured
  — which is what makes `value-dropdown-model_model` absent at `:101`. 5 tests passed, including
  `edit-flow-name`, `flowSettings`, `flow-rename-header` and `run-flow`.
- **The `strict` cleanup contract survives the migration.** `deleteFlow` throws on a failed delete,
  so the hand-rolled `afterEach` keeps the "a failed cleanup fails the test" behaviour the removed
  comment defended. Worth stating in the code, though (see nits).

### Requested changes

**1. Rebase on `main` and adopt `leaveFlowEditor`.**

This branch is behind `main`, and `main` has since merged #1156 (`tests/helpers/flows/leave-flow-editor.ts`,
closing #1153) — the fix for the **fourth signature** this PR documents as split out. #1156's own body
says it deliberately left `edit-flow-name.spec.ts` alone *because #1152 was in flight and would
conflict*, and no issue tracks that adoption.

As it stands, after merge this spec is the only place left carrying the bare `icon-ChevronLeft` click
plus a `home-dropdown-menu` assertion — keeping a deadlock measured at **2 in 24 runs at
`--workers=4`** (by #1005's own burst) as an unattributed timeout, in the very spec whose
investigation found it, while three other specs are already attributed.

`escapeDeadlock: true` is the right setting here: the rename's persistence is asserted *after* the
exit, against the server (`GET /api/v1/flows/` behind the home listing), and the next iteration
re-enters by id — so discarding editor state cannot mask a rename that did not land.

**2. `openFlowById` does not dismiss the onboarding tooltip.**

`dismissOnboardingIfPresent` exists because the `assistant-onboarding-tooltip` overlay "intercepts
clicks on the canvas **and the Flow Settings modal**" and "appears on flow entry, sometimes a beat
after the canvas mounts" (its own docstring, #684). Both existing `openFlowById` copies (`lock-flow`,
`flow-lock`) call it right after the canvas wait. This one does not — and the first thing the test
does after entry is drive the Flow Settings modal. The local 36/36 does not rule this out: a fresh
CI instance is exactly where that tooltip shows.

### Findings tracked as follow-ups

- **`openFlowById` is now a third near-verbatim copy** (`lock-flow`, `flow-lock`, this spec), and the
  copies have already diverged (`waitForSelector` 100 s + `waitForTimeout(500)` + onboarding dismiss
  vs. `expect().toBeVisible()` 30 s + a permission gate, no dismiss). That is the drift #1108 exists
  to end. Extracting `tests/helpers/flows/open-flow-by-id.ts` is a shared-file change and belongs in
  its own PR.
- **`flowSettings.spec.ts`'s inline `getByTestId("flow_name").click()`** carries the same
  swallowed-click defect the helper just fixed. Noted in the PR body with nothing tracking it.
- **`needs_models` classifier gap (qa-infra).** It returned `false` while the impacted set contained
  `agent-component-regression.spec.ts`, which depends on the model catalog but carries none of the
  classifier's markers (`getTestTargets`, `initialGPTsetup`, …). Until that is closed, every PR
  touching a high-reach helper collects the same unrelated red and the lane's signal is unreadable.

### Nits

- **Residual window between the gate and the click.** The `PermissionsProvider` query key includes
  `domain: project:{folder_id}` derived from `currentFlow`, and `use-save-flow` replaces that object
  via `setCurrentFlow(updatedFlow)` on every save. A response that changes or drops `folder_id`
  changes the key, re-enters `isLoading`, and re-disables the button; landing between `toBeEnabled()`
  and `.click()`, the click is swallowed and the failure surfaces 15 s later as "`input-flow-name`
  not visible" with nothing naming the cause. I verified the key composition and the `setCurrentFlow`
  application, **not** that any PATCH response omits `folder_id` — so this is a named hypothesis, and
  a comment at the call site is enough for now.
- The `menu_bar_display` gate inside `openFlowById` is subsumed by the identical gate now inside
  `renameFlow`, which is the first thing the test does. Harmless, but it deserves one line saying why
  the entry keeps its own contract.
- Worth a line in the new `afterEach` that `deleteFlow` throws by design, so the next migration does
  not "fix" it into a `.catch(() => {})`.

### Checklist

`Closes #1005` ✅ · `@stable` present ✅ · spec doc mirrored under `docs/` with all mandatory sections ✅
· `Last validated: 1.12.x` ✅ · manual Part II bullets present (`QA-CHECKLIST.md` L636, L1164) and no
generated block touched ✅ · typecheck / ESLint / unit lanes green ✅

Direction is right and the root cause is correctly identified. Items 1 and 2 before merge; the rest
as follow-ups.

Issue #1147, opened 2026-07-30
api-request-component-regression: enter the blank flow over the API, not through the welcome overlay


Victor-w-Madeira:
Spun out of #1063, which root-caused this condition and fixed it in the three
specs it could validate. `api-request-component-regression.spec.ts` is the one
spec listed in #1063's body that was **observed failing** with this signature and
is **not** fixed there, so it is tracked here rather than left untracked when
#1063 closes.

## The condition (already root-caused on #1063)

`FlowPage` mounts the whole `FlowSidebarComponent` inside a `display: none`
wrapper while the welcome overlay is open:

```jsx
{!view && (
  <div style={{ display: isWelcomeOpen ? "none" : "contents" }}>
    <FlowSidebarComponent isLoading={isLoading} />
  </div>
)}
```

So `[data-testid="sidebar-search-input"]` is **in the DOM with an empty bounding
box** — exactly what Playwright reports as `hidden`. Confirmed present in the
served `1.12.0.dev10` bundle, so it is not a clone/skew artifact.

Any spec entering a blank flow via home page → "New Flow" → templates modal →
`blank-flow` walks into that window:

1. "New Flow" (`useStartNewFlow`) creates **flow A** and calls `openWelcome(A)`
   **before** navigating — deliberate, so the overlay paints on the first frame.
2. `openNewFlowTemplatesModal` dismisses the overlay via "Browse more templates",
   but `handleBrowseMore` only calls `setIsTemplatesOpen(true)` — it never calls
   `close()`. The welcome stays open behind the modal.
3. Clicking `blank-flow` creates a **second** flow B and navigates to it.
4. The welcome closes only once `FlowBuilderWelcomeMount` observes
   `currentFlowId !== openedForFlowId` — a multi-hop settle (navigate → flow load
   → store update → placeholder delete → re-render) that **nothing in the test
   drives**.

Under parallel CI load that settle can exceed the wait budget, and it always
recovers on retry — which is why it reads as a flake. It is a **test
wait-strategy defect**, not a Langflow regression: the product behaviour is
intentional and documented in its own code comments.

## This spec's exposure

`tests/tests-automations/regression/core-components/api-request-component-regression.spec.ts`
enters through `awaitBootstrapTest` + `blank-flow` and then fills
`sidebar-search-input` with **no wait at all** (it relies on `fill`'s auto-wait),
so it has the same exposure as the three specs fixed on #1063.

[...]

PR #1155, opened 2026-07-30
test(flows): stop flow-rename-header and general-bugs leaking a flow per run (#1154)

### LINKED ISSUE #1154: flow-rename-header and general-bugs-save-changes-on-node leak a flow per run
Spun out of #1005, whose own *Done when* anticipated this: *"if local-only: open
the follow-up for the leaking specs instead"*. The leak turned out **not** to be
the cause of #1005's flake — that was the template-entry race, fixed in #1152 —
but the leak itself is real and confirmed.

## What leaks

Audited against a clean `1.12.0.dev10` instance (27 starter projects, nothing
else) by counting `GET /api/v1/flows/?get_all=true` before and after:

| Spec | Cleanup today | Leaks |
|---|---|---|
| `flow-functionality/flow-rename-header.spec.ts` | no `afterEach` at all; only the **API** test deletes its own flow inline (line 76) | the UI test's `My Renamed Flow {timestamp}`, one per run |
| `core-components/general-bugs-save-changes-on-node.spec.ts` | none whatsoever — no `afterEach`, no `deleteFlow`, no tracker | the blank flow it creates, one per run |

Both create flows through the UI, so `trackCreatedFlows` (`tests/helpers/flows/track-created-flows.ts`)
covers them directly — it captures ids from page-level `POST /api/v1/flows` → 201
responses, which is exactly this creation path.

## Why it matters

Accumulated flows are not cosmetic on the shared home grid. They are what makes
the residual cards of other parallel workers overlap a target card's
absolute-inset `list-card-open-button` and intercept a hit-tested click without
navigating (#580/#588) — the failure mode #1005 spent a burst diagnosing. Every
leaked flow raises the odds for every spec that opens a flow from the listing.

#1005's environment caveat named these two specs by hand; this issue is the
follow-up it pointed at.

## Scope

- [ ] `flow-rename-header.spec.ts` — adopt `trackCreatedFlows` + an `afterEach`
      `cleanup`.
      ~~and fold the API test's inline delete into it so the file has one cleanup
      path rather than two.~~ **Withdrawn.** Two paths is the correct shape here:
      the API test creates through the `request` fixture, a separate
      `APIRequestContext` that emits no page-level response events, so a
      page-attached tracker can never observe that flow (#1147). Folding it in
      would mean a module-scoped `apiFlowId` deleted in a shared hook — more
      shared mutable state than the local `finally` it replaces, for no coverage
      gain.
- [ ] `general-bugs-save-changes-on-node.spec.ts` — same, from nothing.
- [ ] Re-audit: run each spec N times against a clean instance and show the flow
      count returning to its baseline.
- [ ] Decide whether the audit is worth generalising — `scripts/` has no
      "did this spec leak?" check today, and 51 files were hand-migrated onto the
      tracker in #1108, so a drift check may be cheaper than the next audit.

Note `general-bugs-save-changes-on-node.spec.ts` currently fails on the legacy
**Text Output** component (#362-364) before it reaches its assertions, and is not
`@stable`. It still creates — and leaks — its flow before failing, so the cleanup
work stands on its own and does not depend on that spec being green.

## Reproduce

```bash
# baseline
curl -s "$BASE/api/v1/flows/?get_all=true&header_flows=true" -H "Authorization: Bearer $TOKEN" | jq length

npx playwright test tests/tests-automations/regression/flow-functionality/flow-rename-header.spec.ts --workers=1 --retries=0

# same count again -> count grew by one
```

Victor-w-Madeira:
Closes #1154

## What this fixes

Two specs created a flow through the UI and never deleted it:

| Spec | Cleanup before | Leaked |
|---|---|---|
| `flow-functionality/flow-rename-header.spec.ts` | no `afterEach`; only the **API** test deleted its own flow inline | the UI test's `My Renamed Flow {ts}`, one per run |
| `core-components/general-bugs-save-changes-on-node.spec.ts` | none whatsoever | the blank flow, one per run |

That is not cosmetic. Accumulated flows on the shared home grid are what make
another worker's residual card overlap a target's absolute-inset
`list-card-open-button` and swallow a hit-tested click without navigating
(#580/#588) — the failure mode #1005 spent a burst diagnosing. #1005's own
environment caveat named these two files by hand; this is the follow-up it
pointed at.

## Changes

Both adopt the shared `trackCreatedFlows` (#1108) rather than a fourth
hand-copied local variant. The cleanup block is **identical in both files**, and
differs deliberately from the three pre-existing adopters on one point:

```ts
let flows: FlowTracker | undefined;          // not `FlowTracker`
…
const tracker = flows;
flows = undefined;                            // null out BEFORE awaiting
await tracker?.cleanup(request);
```

`flows?.` alone only protects the **first** test in a worker: for a later test
whose `beforeEach` threw, the module-scoped binding still holds the *previous*
test's tracker and the optional chain waves it through. On a non-nullable
declaration TypeScript also treats that `?.` as a dead guard. Latent rather than
live in the three older adopters (they are single-test or their `beforeEach`
cannot fail), but there is no reason to copy it forward.

`flow-rename-header`'s API test **keeps its own `finally` DELETE**. It creates
through the `request` fixture, a separate `APIRequestContext` that emits no
page-level response events, so a page-attached tracker can never observe that
flow (#1147). The issue's scope item asking to fold the two paths into one is
**withdrawn on the issue**, with that reasoning recorded there.

Both use the default log-and-continue `cleanup`, not `strict`: neither file had a
teardown whose contract could be preserved, so failing an otherwise-green test on
a cleanup blip would be a new contract rather than a kept one.

`docs/flow-functionality/flow-rename-header.md` gains the teardown step, the
cleanup criterion and the helper under external dependencies; `Last validated`
moves 1.10.x → 1.12.x. `general-bugs-save-changes-on-node` has no mirrored doc
and no `@stable`, which `npm run check:checklist-coverage` confirms is not a
violation.

## Validation

Local instance, **Langflow Nightly 1.12.0.dev10**. Leak audit = flow count from
`GET /api/v1/flows/?get_all=true` before and after each run.

| Run | Before | After |
|---|---|---|
| `flow-rename-header` ×3 (6 tests) | 35 | **35** |
| `general-bugs-save-changes-on-node` ×3 (all 3 **failing**) | 35 | **35** |
| both files ×2, post-review-fix | 44 | **44** |

The failing-run row is the important one: teardown is proven on the failure path,
which is the path this spec always takes today (see below).

**Force-fail — executed, one mutation at a time, both observed leaking, both
reverted** (`grep FF-MUTATION` → 0 hits):

- `FF:` comment out `flows = trackCreatedFlows(page)` in `flow-rename-header` →
  test still **passes** and the count grows 35 → 36
- `FF:` comment out `await tracker?.cleanup(request)` in `general-bugs` →
  count grows 36 → 37

Gates: `tsc --noEmit` clean; eslint **0 errors and no new warnings** (the 8
warnings on `general-bugs-save-changes-on-node` are pre-existing in untouched
test bodies — verified by re-running eslint with the change stashed);
`check:checklist-coverage` passes.

## Two pre-existing conditions, neither introduced here

- `general-bugs-save-changes-on-node` is **currently red** at
  `getByTestId("input_outputText Output")` — Text Output is hidden from the
  sidebar since Text I/O went legacy upstream (#362-364). It fails identically
  with this change stashed. It carries `@release @components` and no `@stable`,
  so it is not in the daily lane. It still creates — and leaked — its flow before
  failing, so the cleanup stands on its own.
- A `🚨 Backend Error: 500 - POST /api/v1/flows/` fires for this file pair under
  2 workers, on **both** branches (Langflow's non-transactional name suffixing
  under concurrent creation, #588). Naming it so checklist step 4 not coming back
  clean for these two files locally is not mistaken for fallout from the new
  teardown.

## Review

Reviewed by two independent passes before opening. They disagreed on the shape of
`flow-rename-header`'s cleanup; the version here is the synthesis — one shape in
both files, and the safer of the two. Their other accepted findings: the spec doc
update, the `Last validated` bump, and a comment that claimed "the other 50"
adopters when there are in fact 5 (the ~46 remaining files carry hand-copied
local implementations).


Victor-w-Madeira:
## Review — verified, three corrections applied, branch brought current

Read the full diff, checked every claim it makes about the surrounding code, and ran
the gates locally. The change is right: the leak is real, `trackCreatedFlows` is the
correct helper, the force-fail was executed in both directions, and the spec doc
carries the teardown step, the criterion and the `Last validated` bump. Nothing here
blocks it. Three corrections and one root cause below.

### 1. The adopter comparison the cleanup shape rests on was wrong — fixed in 4c6e3b9

The comment justified the null-out with *"the three pre-#1154 adopters … are
single-test or their `beforeEach` cannot fail"*. That does not describe
`api-request-component-ui.spec.ts`:

| Adopter | Tests | Teardown call | Exposure |
|---|---|---|---|
| `edit-flow-name` | 1 | `flows?.cleanup(…, { strict: true })` | dead guard (non-nullable), latent because single-test |
| `flowSettings` | 1 | `flows?.cleanup(…)` | same |
| `api-request-component-ui` | **4** | `flows.cleanup(…)` — **no chain** | a first-test `beforeEach` failure throws `TypeError` from the teardown and **buries the real error** |

So it is the *most* exposed of the three, not the latent one — and the failure it
produces is exactly what `flowSettings`' own comment says its `?.` exists to prevent.
The comment now states this per file instead of collapsing them. Worth a follow-up to
align that file; with 5 adopters it is a small job.

**The counts in the PR body check out.** 29 files reference `trackCreatedFlows`, but
24 of them declare a local function of that name — exactly **5** import the shared
helper. The "~46 hand-copied" framing is accurate.

### 2. Two smaller comment corrections — same commit

- `general-bugs-save-changes-on-node` has **one** test, so the null-out cannot matter
  there. Kept for one shape across both files, but the comment no longer justifies it
  with *"once a file has more than one test"*, which is false of that file.
- The file-scoped `beforeEach` asks for `page`, so the API test — which takes only
  `{ request }` and never touched a page — now has a browser context instantiated for
  it. Real, unstated, and worth about one context launch per run. Accepted for
  uniformity (a test added later is covered without anyone wiring it), now priced in
  the comment with the exit written down.

### 3. The branch was validated against a helper that has since changed — fixed in 46bd940

The branch was **66 commits behind** and its only CI run was 2026-07-30. Since the
merge base (`07d5f19`), #1197 rewrote `tests/helpers/flows/track-created-flows.ts`
(+52/-17) — it added the attribution path and moved the `page.goto("about:blank")`
inside `cleanup()`. The recorded 3× leak audits therefore ran against a different
`cleanup()` than the one this PR now calls. Nothing about the change is incompatible
(the new option is optional and inert), but the evidence predated the code.

`origin/main` is merged in, the merge was clean, and neither spec nor the doc was
touched by it. Gates re-run on the merged tree: `tsc --noEmit` clean · eslint **0
errors** · `test:units` 320 · `test:scripts` 392 · `check:checklist-coverage` passes.
CI is running on the merged head.

Also worth knowing before merge: **PR #1221 modifies `tests/helpers/flows/rename-flow.ts`**,
which both of these specs call. No file overlap and no conflict, but the two interact.

### 4. The recurring `500 /api/v1/flows/` has a root cause, and this PR's attribution of it is inverted

§"Two pre-existing conditions" records it as `POST /api/v1/flows/`, caused by
*"Langflow's non-transactional name suffixing under concurrent creation (#588)"*. The
traceback in this PR's own CI run says otherwise:

```
flows.py line 837, in delete_multiple_flows → _delete_operation → cascade_delete_flow
RuntimeError: ('Unable to cascade delete flow: 8594eced-…',
               OperationalError('(sqlite3.OperationalError) database is locked'))
```

A **DELETE** on the collection endpoint, under SQLite lock contention with 2 workers.
There is exactly one traceback in the whole log and none in `create_flow`.

**The conclusion the PR draws from it still holds, and is now better supported:** this
is not fallout from the new teardown. `cleanup()` deletes per id via
`DELETE /api/v1/flows/{id}` (`deleteFlow`), and nothing in the suite calls the bulk
endpoint at all — it comes from Langflow's own frontend.

It is also not local to this file pair. The same 500 appears **6×** in this branch's
PR lane, **1×** in PR #1212's manual gate run, and **10×** in daily-stable run
30657439522 on `main`. Untracked by any open issue; now that it has a named cause, it
deserves one.

### Not re-verified

`general-bugs-save-changes-on-node`'s red is corroborated but not independently
reproduced: line 90 is `getByTestId("input_outputText Output")`, Text Output went
legacy upstream (#362-364), the spec carries no `@stable`, and the CI failure is a
20 s `locator.waitFor` timeout on that test — all consistent with the PR's account.
It runs on no other lane, so there is no main-branch run to compare against.

Victor-w-Madeira:
## CI on the merged head — [run 30678699548](https://github.com/oriontech-me/langflow-e2e/actions/runs/30678699548)

Same shape as before the merge, which is what re-running was for.

**Everything green except `Run impacted E2E specs`: 2 passed, 1 failed.** The one
failure is now attributed down to the locator:

```
general-bugs-save-changes-on-node.spec.ts:73 › any changes on the node must be
saved on user interaction @release @components
  - waiting for getByTestId('input_outputText Output') to be visible
  93 |       .getByTestId("input_outputText Output")
  95 |     await page.getByTestId("add-component-button-text-output").click();
```

That is Text Output, hidden from the sidebar since Text I/O went legacy upstream
(#362-364): the add-component click lands on nothing, so the node's input never
renders. Identical on both runs of this branch, before and after merging 66 commits
of `main`. The spec carries `@release @components` and no `@stable`, so it is on no
scheduled lane — consistent with it having been broken for a while without anyone
being paged.

Per the repo's own rule for this check, a red from pre-existing breakage is not a
regression and does not block the merge. What the run does prove is that the two
tests this PR is actually about — both of `flow-rename-header` — pass on current
`main`, with the current `track-created-flows.ts`.

### One thing improved, and it revises §"Two pre-existing conditions"

**Zero `🚨 Backend Error` in this run**, and no cleanup warnings. The
`500 /api/v1/flows/` that the previous run logged did not recur.

That makes it intermittent rather than deterministic for this file pair, which fits
the cause the traceback named (SQLite lock contention on `delete_multiple_flows`, see
the review above) better than the PR body's current wording — *"fires for this file
pair under 2 workers, on both branches"* reads as reliable, and it is not. Checklist
step 4 comes back clean for these two files on the merged head.

PR #1157, opened 2026-07-30
ci: extract the post-collect-models health gate into a shared composite action (#1045)

### LINKED ISSUE #1019: ci(pr-validation): the Collect models step still retries 3x against the shared Langflow (#1011 follow-up)
## Context

PR #1015 (issue #1011) cut the chain that cost daily run [30351107916](https://github.com/oriontech-me/langflow-e2e/actions/runs/30351107916) at four independent points. Two of those four are **workflow-local** and were applied to `daily-stable.yml` only:

- `PLAYWRIGHT_RETRIES: "0"` on the `Collect models` step
- the health gate between `Collect models` and the Playwright run

`pr-validation.yml` runs the same `Collect models` spec against the same kind of single, shared Langflow service container — and has neither.

## What is missing

`.github/workflows/pr-validation.yml`, the `Collect models` step in the `e2e` job:

```yaml
      - name: Collect models
        if: needs.detect-specs.outputs.needs_models == 'true'
        env:
          CI: "true"
          PREFLIGHT_SKIP_CREDENTIALS: "1"
          PLAYWRIGHT_BASE_URL: "http://localhost:7860/"
          # ← no PLAYWRIGHT_RETRIES
```

`CI: "true"` selects `retries: 2` in `playwright.config.ts`, so a failing pre-flight runs **three full attempts**, each re-importing every provider key and re-walking the Model Providers UI against the one backend the impacted-specs run is about to use. That is the exact multiplier identified as dominant in #1011: on the daily it turned a ~49 s step into 7–12 min of sustained load and wedged the gunicorn worker.

The worker cap is already there (`LANGFLOW_WORKERS: "1"`, #923), and #927 showed the wedge is process-wide rather than worker contention — so the cap alone is not sufficient, which is why #1011 needed three guards and not one.

## Why this was not done in PR #1015

Scope: that PR is about the daily losing a day of coverage, and `1 issue / 1 PR`. Its "What this PR does not cover" section reasons about the other workflows only for the **worker cap**, not for the retry cap — this issue closes that reasoning gap.

Note that the **fourth** fix in #1015 — the identical-error early exit in `validateProviderWithFallback` — is helper-level and therefore already protects `pr-validation.yml` today: an account-scoped outage now costs 3 probes instead of a full catalog sweep, in every workflow. This issue is about the amplification that remains *on top of* that.

## Impact if left alone

Lower than on the daily: a wedge here costs one PR check (`Run impacted E2E specs`), not a whole day of `@stable` coverage, and the check is already known to be non-blocking when it reddens for pre-existing breakage. But the failure mode is confusing where it lands — the run reports as a `globalSetup` pre-flight timeout on `/api/v1/version` with no mention of what caused it (the #922 signature, which blocked #867), so the next person pays the diagnosis cost that #1011's gate exists to eliminate.

## Proposed change

1. `PLAYWRIGHT_RETRIES: "0"` on `pr-validation.yml`'s `Collect models` step, with the same rationale comment as `daily-stable.yml` (retrying adds no information — the first attempt already produced the diagnosis).
2. Decide explicitly whether the health gate is worth porting. It is ~20 lines of bash and its value is **attribution**, not recovery; the argument for porting is that the `globalSetup` timeout here is precisely the misleading signature from #922. If the answer is no, record the reason in the workflow so the asymmetry with `daily-stable.yml` reads as a decision rather than an oversight.
3. `weekly-stable.yml` stays untouched while disabled — porting a partial mirror would read as fixed without being fixed. Re-enabling it requires porting the whole of #1011 (already recorded in PR #1015).

## Acceptance

- The retry cap is present on `pr-validation.yml`'s `Collect models` step and resolves on the step (a red pre-flight runs one attempt, not three).
- The gate decision is recorded either way — ported, or declined with the reason in the workflow comment.
- No behaviour change for a green pre-flight: on a healthy run nothing about the step's timing changes.

### LINKED ISSUE #1045: ci: extract the post-collect-models health gate into a shared composite action
Follow-up of #1044 (which resolved #1019).

## Context

#1044 ported the post-collect-models health gate from `daily-stable.yml` to
`pr-validation.yml`. It is now the **second verbatim-ish copy** of the same
polling loop, and the two already diverge:

| | `daily-stable.yml` | `pr-validation.yml` |
|---|---|---|
| Deadline | `inputs.recover_timeout_s \|\| 300` | `420` (hardcoded — no dispatch inputs on this workflow) |
| Pre-flight verdict warning | yes (`COLLECT_MODELS_OUTCOME`) | n/a (step is not `continue-on-error`) |
| Failure attribution | `wedged or dead` | `curl` exit code decoded (7 = dead, 28 = wedge) |
| Heartbeat while polling | no | yes (~every 30 s) |
| Trailing log wording | "the @stable run" | "the impacted specs" |

`weekly-stable.yml` has a `Collect models` step and **no gate at all** — harmless
while it is disabled, a latent hole the day it is re-enabled as the documented
fallback.

The divergence is not accidental: #1044 deliberately kept `daily-stable.yml`
comment-only so the daily's behaviour stayed provably unchanged, which parked
two improvements on the PR-validation copy only.

## Scope

Extract the gate into `.github/actions/wait-for-backend` (same pattern as
`setup-playwright`, `run-e2e`, `guard-dedicated-issue`) and adopt it in all
three workflows:

- [ ] Composite action with inputs: `base_url`, `timeout_s`, `next_step_label`
      (the trailing log line), and an optional `collect_models_outcome` for the
      daily's pre-flight warning.
- [ ] Carry the two PR-validation-only improvements into the shared
      implementation, so the daily gets them too:
  - `curl` exit-code decoding in the `::error::` (7 = connection refused → dead;
    28 = timeout → the wedge shape of #922/#927). Today the daily's message
    hedges "wedged or dead" while `curl` already knows which.
  - Heartbeat log every ~30 s, so a step that can sit for 5-7 min does not read
    as hung.
- [ ] Decide one deadline, or justify two. #1044 set PR-validation to 420 s
      because the single measured recovery consumed **223 s** — 74% of the daily's
      300 s budget — and all of it was gunicorn's own timer (WORKER TIMEOUT
      ~3.5 min + SIGKILL + ~12 s of Langflow startup). A slightly slower restart
      blows 300 s. The daily kept 300 s only because changing it was outside
      #1019's scope.
- [ ] Adopt in `weekly-stable.yml` while it is disabled, so re-enabling it does
      not reintroduce the #1011 hole.

## Validation

Structural only — the wedge cannot be reproduced locally (it depends on the
runner's collect-models load). Parse each workflow and assert the gate step
still lands between `Collect models` and the run step carrying the same `if`;
then exercise the extracted script against the four backend states #1044 used
(healthy / dead port / wedged / wedged-then-recovers).

rafaelgiln:
Closes #1045. Follow-up of #1044 (which resolved #1019).

## Problem

The health gate that waits out the wedge `collect-models` can leave behind
(#922/#927) existed as **three** copy-pasted shell loops — `daily-stable.yml`
(#1011), `pr-validation.yml` (#1019/#1044) and `manual.yml` — while
`weekly-stable.yml`, the documented fallback, had **none**. The copies had
already diverged on the deadline (300 vs 420 s), the heartbeat, the failure
attribution and the pre-flight warning.

The issue described three call sites. There were four: `manual.yml` grew its own
copy after the issue was filed, and that copy's comment explicitly asked to be
replaced by this action when it landed. All four are adopted here.

## Change

Extract the gate into `.github/actions/wait-for-backend`, backed by
`scripts/wait-for-backend.mjs`, and adopt it in every lane that shares one
Langflow container between its sweep and its test run.

Carried into the shared implementation:

- **Failure classification**, replacing the daily's `"wedged or dead"` hedge.
  curl's exit codes 7/28 became native Node signals, and the verdict now names
  the state it observed: refused → the backend is dead and waiting cannot help;
  accepted-and-unanswered → the #922/#927 wedge; non-2xx → an application-level
  failure that is **not** the wedge. The inline copies asserted the wedge
  unconditionally, which is false on three of the four states — and a false
  attribution is worse than none, because it sends the next reader after the
  wrong bug.
- **Heartbeat** every ~6 attempts, so a step that can legitimately sit for
  5-7 min does not read as hung.
- **One deadline, 420 s**, for every lane. The two measured recoveries consumed
  223 s (#1019) and 260 s (#1044's validation dispatch) — 74% and 87% of the
  daily's 300 s budget — and every second of both was gunicorn's own timer
  (`WORKER TIMEOUT` → `SIGKILL` → ~12 s of Langflow startup), so a slightly
  slower restart blows 300 s. Over-waiting is nearly free by construction: the
  extra time is only ever spent on a lane already headed for a red, and it is
  bounded well below every caller's `timeout-minutes`. `daily-stable.yml` keeps
  its `recover_timeout_s` dispatch override, now defaulting to 420.
- **`weekly-stable.yml` gains the gate** while it is disabled, so re-enabling it
  as the documented fallback does not reintroduce the #1011 hole.

Unchanged on purpose: a failed `Collect models` is surfaced as a `::warning::`
and never gates the run (#980/#570). `pr-validation.yml` still runs the gate only
under `needs_models == 'true'` — with no sweep there is no wedge to wait out, an
unrelated startup failure is `globalSetup`'s to report, and an LLM-free PR must
not be billed for the wait. `manual.yml` is the only lane passing `base_url`: it
can target a Langflow that is not the job's service container.

Net: **-170 lines of workflow YAML, +106**, four lanes behind one implementation.

## Validation

Structural plus unit, because the wedge cannot be reproduced on demand — it
depends on the runner's `collect-models` load.

`scripts/wait-for-backend.test.mjs` — 28 tests, in `npm run test:scripts` (a PR
gate):

- the loop driven against **real `node:http` servers** in all four backend
  states the issue named: healthy, dead port, wedged (accepts the connection and
  never answers), and wedged-then-recovers;
- the classification, including that only the wedged state is *reported* as the
  wedge, and that a refused connection whose errno arrives only in
  `cause.message` is still classified as dead (observed on Node 25 — without the
  fallback a dead backend was demoted to a generic wiring error);
- `main()`'s exit codes and the exact message each lane reads (next-step label,
  attribution, the `::warning::` for a failed sweep, and the 420 s fallback when
  the deadline input is garbage);
- the loop contract: the deadline is checked *after* a probe, so a budget shorter
  than one interval still gets one real attempt; the heartbeat cadence; and that
  `heartbeat_every: 0` silences it;
- an **adoption guard** over all four workflows: the gate step must sit between
  `Collect models` and the run step, no inline copy of the loop may return, the
  PR lane must keep the gate on the same `if` as its sweep, and the daily must
  keep honouring `recover_timeout_s`.

Both guards were **force-failed**: removing the step from `weekly-stable.yml`
and reordering it above `Collect models` each turn that lane's test red.

Local results:

```
node --test scripts/wait-for-backend.test.mjs   → 28 pass, 0 fail
npm run test:scripts                            → 197 pass, 0 fail
npm run test:units                              → 215 pass, 0 fail
npm run typecheck                               → clean
npm run lint                                    → 0 errors
```

The action YAML and the four workflows parse. The gate binary was also run
directly against a live backend (exit 0) and a genuinely refused port (exit 1,
reported as DEAD, not as the wedge).

**Not provable locally:** the wedge itself, and workflow YAML end-to-end. This
PR touches `.github/workflows/**`, so its own `pr-validation` run exercises the
new action on the PR lane — watch for `Waiting up to 420s for
http://localhost:7860/api/v1/version` in the gate step. The daily and manual
lanes are proven by the next daily and by a `manual.yml` dispatch.


rafaelgiln:
## Runtime proof in CI — the shared gate ran and passed

A correction first: this PR's own `pr-validation` does **not** exercise the
action. `Detect changed specs` reported `Impacted specs: 0 (0 direct, 0
transitive)` and the E2E lane was `skipping` — a diff confined to
`.github/**` + `scripts/` selects no spec, because the import-graph resolver
treats only `tests/fixtures/**`, `playwright.config.ts` and the global hooks as
suite-wide. So the proof had to come from a dispatch.

[Run 30589339315](https://github.com/oriontech-me/langflow-e2e/actions/runs/30589339315) — `manual.yml` on this branch, Docker lane
(`test_tag=@templates`). That lane is the right one: its gate carries no `if:`
and `Collect models` always runs, so the action executes unconditionally.

```
9   Collect models                                                ✅
10  Wait for the backend to recover from the collect-models load   ✅
      Waiting up to 420s for http://localhost:7860/api/v1/version (probe deadline 8000ms, interval 5s).
      Backend answered /api/v1/version on attempt 1 after 0s — proceeding to the run.
```

Everything the happy path should show: the 420 s deadline resolved from the
action's default, `manual.yml`'s own `next_step_label` ("the run"), the probe
deadline and interval as configured, and a first-attempt exit when the backend is
healthy — no wait imposed on a lane that does not need one.

## What this run does NOT prove

- **Recovery from a real wedge.** There was no wedge to wait out. That path stays
  covered by the unit tests (a wedged server that recovers mid-loop) plus the two
  historical measurements the deadline is sized against.
- **The `::warning::Collect models FAILED` branch.** The sweep passed here, so the
  warning never fired. Unit-tested only.

Both were already declared as such in the PR body; naming them again so the green
step is not read as more than it is.

## The run ended red, and not from this diff

```
1 failed, 4 passed (1.8m)
starter-projects.spec.ts:5 › user must be able to interact with starter projects @release @templates
  TimeoutError: page.waitForSelector: locator('[data-testid="template_research-agent"]') — 10000ms
```

`@release` without `@stable`, so it is invisible in the daily — one of the
pre-existing broken `@release` specs on `main` (testid drift: the nightly no
longer exposes `template_research-agent`). It is step 11; the gate is step 10 and
passed, and the spec touches neither `collect-models` nor the gate.

Issue #1162, opened 2026-07-31
fixtures: flow_error cannot fire for a playground/agent run on 1.12.x — the v2 run stream is neither matched nor parsed


Victor-w-Madeira:
## What happens

`tests/fixtures/fixtures.ts` provides the two error monitors every spec inherits.
Per `CLAUDE.md`, they are deliberately asymmetric and that asymmetry is what a
green run is worth (#1084): an `http_error` is **logged and never fails a test**,
while a `flow_error` **fails it**.

On Langflow 1.12.x the `flow_error` half **cannot fire for a playground or agent
run**. Two independent reasons, both measured on 1.12.0.dev10 from a Playwright
trace of `agent-max-tokens.spec.ts`:

1. **The URL filter does not cover the run endpoint.** The detector inspects only

   ```js
   url.includes("/events?event_delivery=") || url.includes("/build/") || url.includes("/run/")
   ```

   The playground run is `POST /api/v2/workflows`. It matches none of the three.

2. **Even if it matched, the body is skipped.** That response is
   `content-type: text/event-stream; charset=utf-8`, and the detector returns
   early for stream-like content types (`Skipping streaming response body parsing`).

So an execution error travels the whole way to the UI with no fixture verdict.
The stream carried, verbatim:

```
data: {"type":"RUN_ERROR","message":"Error code: 400 - {'type': 'error', 'error':
  {'type': 'invalid_request_error', 'message': 'Your credit balance is too low to
  access the Anthropic API...'}}"}
data: {"type":"CUSTOM","name":"langflow.event","value":{"event_type":"error", ...}}
data: {"type":"STATE_DELTA","delta":[{"op":"add","path":"/nodes/Agent-q9bOb",
  "value":{"status":"error","output":{...,"errorMessage":"Error building Component
  Agent: ... Error code: 400 ...","stackTrace":"Traceback..."}}}]}
```

and the run printed **no** `🚨 Flow Error Detected`. The shapes the detector
looks for (`data.build_data.params` starting with `"Error"`, `data.error === true`,
Python-exception regexes) do not match any of these either — note the error
message object even carries `"error": false`.

## Why it matters

- **The strong half of the pair is unreachable where it matters most.** 55 specs
  drive a playground/agent run. For all of them, an execution error is now only as
  visible as whatever assertion happens to trip afterwards — which is exactly how
  #1059 produced a `toHaveCount expected 1, received 0` and a triage discussion
[...]

Issue #976, opened 2026-07-27
Provider API-key fallback: resolve a live key before the suite, fail loud when every candidate is dead


daniellicnerski1:
Follow-up from #967.

## Problem

On the 2026-07-27 daily ([run 30261409427](https://github.com/oriontech-me/langflow-e2e/actions/runs/30261409427)) the Anthropic account was drained — `Your credit balance is too low to access the Anthropic API` — and all 13 candidate Claude models failed the `collect-models` probe. Result: **24 specs skipped silently**. A skip never trips the daily-failure gate, so coverage eroded without a red run.

The chain: `collect-models` records the provider `inactive` in `providers.json`, and the 17 specs that each carry a local `getProviderSkipReasons()` (e.g. `agent-system-prompt.spec.ts:42`) turn that into a `test.skip()` for every target bound to the provider.

The existing fallback in `validateProviderWithFallback` (`collect-models.ts:181`) iterates **models**, not **keys** — each `validate*` reads one fixed env var (`collect-models.ts:29-124`). There is no key-level fallback anywhere in the suite.

## Proposal

A `scripts/resolve-provider-keys.ts` pre-flight step that probes `<PROVIDER>_API_KEY`, `_2`, `_3`… in order and re-exports the first live key to `$GITHUB_ENV` under the canonical name, ahead of `Collect models`. Everything downstream — `collect-models`, the shards, the 17 consumer specs — keeps reading the canonical name and transparently receives a validated key.

A fallback confined to `collect-models` would not work: `Run @stable tests` is a separate step that still receives the dead primary from the workflow `env:` block, and the first spec calling `setup-anthropic.ts:52` would overwrite the working credential with it.

Strict in `daily-stable.yml` (red when a configured provider has zero live keys), warn-only in `pr-validation.yml` — preserving the #952/#955 decision that a billing outage must not redden LLM-free PRs.

Full design: see the first comment on this issue (kept here rather than as a repo file, to avoid provisional docs in the tree).

## Done when

- [ ] `scripts/resolve-provider-keys.ts` resolves per-provider candidate keys and exports the winner to `$GITHUB_ENV`
- [ ] Probe functions extracted from `collect-models.ts` into a shared, key-parameterized module — one implementation, no divergent copy
- [ ] Error classification distinguishes key/account failures (advance to next key) from model and transport failures (do not advance)
- [ ] `daily-stable.yml` wires the step in strict mode, without `continue-on-error`
[...]

Issue #866, opened 2026-07-21
[Daily #864] Agent max_tokens — unset does not generate longer output than capped (recurrent 4×)


rafaelgiln:
Spun out of daily-failure triage #864 (run [29822019255](https://github.com/oriontech-me/langflow-e2e/actions/runs/29822019255), 2026-07-21).

## Symptom

One `@stable` test hard-failed (final status, after retries) — the causal control asserting that an **unset** `max_tokens` generates a longer response than a capped one:

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests-automations/regression/core-functionality/llm-agents/agent-max-tokens.spec.ts:287` ("causal control — unset max_tokens generates freely") | `expect(unsetLength).toBeGreaterThan(cappedLength)` | `Error: expect(received).toBeGreaterThan(expected)` |

## Flake signal / recurrence

**Confirmed recurrent, same signature — 4×:** failed on dailies **2026-07-14, 2026-07-15, 2026-07-16, 2026-07-21** (a 5th occurrence 2026-07-13 was a different signature). This is a durable, cross-day signal, not single-run noise.

`@stable` was **auto-removed** from this test by the daily workflow (hard failure). Restoring it after the fix is a deliverable below.

## Preliminary read (descriptive — NOT a verdict)

`toBeGreaterThan` failing means the unset-`max_tokens` response was **not** longer than the capped one. This is consistent with (a) the model returning a short answer regardless of the cap on this prompt, (b) `max_tokens` not being applied/passed as expected by the agent, or (c) response non-determinism making the length comparison unreliable. None is confirmed — the multi-day recurrence argues against pure non-determinism.

## Investigation directive

Investigate all paths independently, **product (Langflow) as prime suspect first**: on the current nightly, run the agent with `max_tokens` unset vs capped and confirm whether the unset response is actually longer / whether the cap is applied. Only after ruling out a product regression, consider: (a) prompt does not reliably elicit a long enough response → strengthen the prompt so the causal contrast is robust; (b) the length metric is fragile → assert on a more deterministic observable of the cap.

## Deliverables (Done when)

- [ ] Root cause confirmed (product regression in `max_tokens` handling vs. test prompt/assertion fragility) with evidence on the current nightly.
- [ ] Spec passes reliably (multiple clean `--retries=0` runs).
- [ ] **`@stable` restored** in the fix PR — the tag was auto-removed at triage; re-validate per `CONTRIBUTING.md` before restoring.
[...]

PR #1163, opened 2026-07-31
test(llm-agents): make an errored agent run name its own cause (#1188)

### LINKED ISSUE #1188: agent-max-tokens: an errored run fails on a locator timeout, never reaching the message that names the cause
Follow-up of #1166 (merged, closed #1059).

## Symptom — measured on `main`, not inferred

On `main` @ 2b3f1f9, Langflow 1.12.0.dev10, `anthropic / claude-sonnet-5` with a
drained key — a **deterministic error reproducer**, since the provider answers
`400 … Your credit balance is too low` on every run — both tests in
`agent-max-tokens.spec.ts` fail like this:

```
TimeoutError: locator.innerText: Timeout 20000ms exceeded.
Call log:
  - waiting for getByTestId('div-chat-message').last()

  240 |   await expect(page.getByTestId("button-send").last()).toBeVisible({ timeout: 10000 });
  241 |
> 242 |   const reply = (await messages.last().innerText()).trim();
```

2/2 runs, `--retries=0 --workers=1`, on `causal control — unset max_tokens` and on
`max_tokens=50 caps …` (34 s, of which 20 s is this timeout). Both reach it through
the shared `runPrompt()`.

The failure names a **locator**, not a cause — while the cause was available in the
same run, printed by the fixtures:

```
🚨 Flow Error Detected in Event Stream — Error code: 400 - {'type': 'error',
'error': {'message': 'Your credit balance is too low to access the Anthropic API…'}}
```

## Why

Upstream `chat-message.tsx:30` — when `chat.category === "error"` the turn renders
`ErrorView` **instead of** `BotMessage`, so an errored turn carries no
`div-chat-message` at all, only `error-card-stack`. The bubble that satisfies the
turn-started poll (`messages.count() > before`) is gone by the time
`messages.last().innerText()` runs, and that call has no fallback.

So the run **never reaches** the badge assertion where #1166 put the informative
message (*"the finished response must expose a token-usage badge … Rendered reply:
…"*). The one state that message cannot describe is the one it never sees.

## What the doc claims today

`docs/core-functionality/llm-agents/agent-max-tokens.md` states that *"still
generating", "finished without a badge" and "finished with an error"* are three
distinct states, each naming itself. The third is not distinguished today: it
fails as an unattributed locator timeout. Code and doc must agree.

## Done when

- [ ] an errored run fails **naming the provider's message**, in seconds, not on a
      locator timeout
- [ ] the "finished without a badge" and "still generating" paths keep the exact
      behaviour #1166 validated — no assertion weakened, no timeout loosened
- [ ] the doc's three-state claim matches the code
- [ ] verified against the drained-key reproducer above **and** green on a live
      provider key

## Note

PR #1163 — opened on #1059 before #1166 merged, and superseded by it — already
carries a measured implementation of exactly this branch (15 s, provider message
verbatim, plus a `readRunError` that expands the collapsed accordion). That branch
is the intended vehicle for this issue; the rest of it is dropped in favour of what
#1166 already landed.

Victor-w-Madeira:
**Closes #1188.**

> **This PR was repurposed.** It was opened on #1059 and, while it was open, PR #1166 landed a converging fix on `main` and closed that issue. Rather than merge a duplicate or throw the work away, the branch was reset onto `main` and now delivers only the one state #1166 does not resolve. The original investigation — including the 2026-07-29 `blob-4` snapshot, which is the most direct evidence #1059 produced and is written down nowhere else — is preserved verbatim in [this comment](https://github.com/oriontech-me/langflow-e2e/pull/1163#issuecomment-5143284101).

## Problem

#1166 replaced the badge-count completion gate with a model-agnostic one and named three states each failure should distinguish: *still generating*, *finished without a badge*, *finished with an error*. The first two work. The third did not, and the reason is upstream shape rather than timing:

`chat-message.tsx` renders `ErrorView` **instead of** `BotMessage` when `chat.category === "error"`, so an errored turn carries no `div-chat-message` at all — only `error-card-stack`. Every wait keyed on the bubble outlives it.

Measured on `main` (2b3f1f9) at 1.12.0.dev10 against a **drained Anthropic key** — a deterministic reproducer, since the provider answers `400 … credit balance is too low` on every run:

| | Duration | Failure |
|---|---|---|
| before | 34 s | `TimeoutError: locator.innerText: Timeout 20000ms exceeded — waiting for getByTestId('div-chat-message').last()` |
| after | **16 s** | `the agent run errored instead of returning a response — An error occurred Your credit balance is too low to access the Anthropic API…` |

2/2 runs each, on **both** tests in the file (they share `runPrompt()`). The old failure never reached the badge assertion where #1166 put the informative message — so the one state that message was written for was the one it could never see. And the cause was available the whole time: the #1162 fixture printed the provider's 400 as a flow-error advisory *in the same run*.

## The change

- **`failIfRunErrored()` runs at both points an error can appear** — after the turn-start poll (a run can fail before any bubble mounts) and after the completion signal (the measured case: the bubble mounts, then is replaced). Not one check: the two states are reached by different routes.
- **The turn-start poll accepts an error card as a start**, so a run that fails before any bubble no longer waits the full 60 s for an element that is not coming.
- **`readRunError()`** expands the collapsed accordion (upstream `error-message.tsx`) to reach the provider text — and a card carrying *no* provider message says so instead of reading as a cause.
- **The reply read is count-guarded**, so a finished turn that rendered nothing still reaches the badge assertion instead of dying on `locator.innerText`.

**No pass path is added and no timeout is loosened.** An errored run still fails; a badge-less run still fails; every assertion #1166 validated is untouched.

## Validation — 1.12.0.dev10, `--retries=0 --workers=1`

- **Error path** — anthropic/claude-sonnet-5 (drained key): 2/2, provider message verbatim, 16 s
- **Green path** — openai/gpt-4o-mini: both tests passed (17 s, 20 s)
- **Force-fail of the path this PR touches** — badge testid made bogus ⇒ the missing-observable message with the essay quoted, in 34 s, not a locator timeout. `grep FFMARK` on the diff: 0 hits
- `tsc --noEmit` ✅ · eslint ✅ (0 errors, the same 9 pre-existing warnings) · `check:checklist-coverage` ✅ · `check-checklist-guard.mjs` ✅
- `@stable` retained on both tests; no quarantine involved

## Stated limitation

The **anthropic green path** is still unproven: the local key and the CI secret are both out of credit (#976). That is exactly the limitation #1166 recorded and this PR does not change it — the drained key is what makes the *error* path deterministic here.

## Relationship to #1162 / #1165

The v2 flow-error gate is **advisory** today and will fail such runs once step 2 of #1162 flips it. That is a fixture-level verdict for the whole suite; this PR makes the spec's own postcondition self-attributing and removes the 20 s dead wait, independently of when that flip lands.

### DIFF
--- a/docs/core-functionality/llm-agents/agent-max-tokens.md
+++ b/docs/core-functionality/llm-agents/agent-max-tokens.md
@@ -88,7 +88,10 @@ Shared setup per test:
    signal: the turn mounts (`div-chat-message` count rises) and the generating
    indicator clears (`button-stop` hidden, `button-send` back). The
    `chat-message-token-usage` badge is **not** the completion gate — see the
-   note on #1059/#569 below.
+   note on #1059/#569 below. An **error card** at either point ends the run
+   immediately with the provider's message (#1188): upstream renders
+   `error-card-stack` *instead of* the bot bubble, so any wait keyed on the
+   bubble outlives an errored turn.
 6. Assert the badge exists (the `max_tokens` observable), then hover it and read
    the **Output** token count from its tooltip (`Input: 1.0K / Output: 46`
    format; values may be plain integers, `N.NK`, or empty ⇒ 0).
@@ -150,7 +153,9 @@ Shared setup per test:
   failures with three distinct messages instead of one blind
   `toHaveCount … Received: 0`. A false positive is impossible in the other
   direction too — the badge assertion is a hard `toHaveCount(1)`, never skipped
-  when absent.
+  when absent. The third of those states only became true in **#1188**: an
+  errored turn renders no bot bubble at all, so it used to die 20 s later on
+  `locator.innerText` with the provider's error nowhere in the message.
 - **Force-failure check** (CONTRIBUTING §2) run during VERIFY on each hard
   assertion before `@stable`.
 
@@ -240,6 +245,24 @@ Shared setup per test:
   prints the reply that did render. Timeouts were **not** loosened (the total
   budget went from ~248 s to ~205 s) and no assertion was weakened — only the
   attribution changed.
+- **An errored run is resolved explicitly, at both points it can appear**
+  (#1188). Gating on the bot bubble fixed two of the three states but not the
+  third: upstream `chat-message.tsx` renders `ErrorView` **instead of**
+  `BotMessage` when `chat.category === "error"`, so an errored turn carries no
+  `div-chat-message` — only `error-card-stack`. Measured on `main` at
+  1.12.0.dev10 against a drained Anthropic key (a deterministic reproducer: the
+  provider answers `400 … credit balance is too low` every time), both tests
+  failed with `locator.innerText: Timeout 20000ms exceeded — waiting for
+  getByTestId('div-chat-message').last()`, never reaching the badge assertion
+  where the informative message lives. The cause was on screen and in the run
+  stream the whole time — the fixture's flow-error advisory (#1162) printed the
+  provider's 400 in the same run. `runPrompt()` now checks for the error card
+  after the turn-start poll (the run can fail before any bubble mounts) **and**
+  after the completion signal (the measured case: the bubble mounts, then is
+  replaced), failing with the provider's message expanded out of the error
+  accordion. The reply read is also count-guarded, so a finished turn that
+  rendered nothing still reaches the badge assertion instead of a locator
+  timeout. This adds no pass path — an errored run still fails.
 - **Where the #1059 signature came from.** The filed hard failure (daily
   2026-07-29, `anthropic / claude-sonnet-5`) was **environmental collateral**, not
   a product regression: attempt 0 of that very test died on the day's dominant
--- a/tests/tests-automations/regression/core-functionality/llm-agents/agent-max-tokens.spec.ts
+++ b/tests/tests-automations/regression/core-functionality/llm-agents/agent-max-tokens.spec.ts
@@ -210,6 +210,10 @@ async function getSavedMaxTokens(page: Page): Promise<unknown> {
 // same pair that spec uses (the turn mounts, then the generating indicator
 // clears), then assert the badge separately so each failure names its own cause.
 // No timeout was loosened: the worst case went from ~248 s to ~205 s.
+//
+// The third state — "finished with an error" — is resolved explicitly at both
+// points it can appear, because upstream renders the error card INSTEAD of the
+// bot bubble, so every wait keyed on the bubble outlives it (#1188).
 async function runPrompt(page: Page): Promise<string> {
   const node = page.locator(
     '[data-testid^="rf__node-ChatInput"] [data-testid="textarea_str_input_value"]',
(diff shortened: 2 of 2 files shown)


Victor-w-Madeira:
## Superseded — original investigation preserved

PR #1166 landed on `main` (2b3f1f9) while this PR was open and closed #1059 with a converging fix. This branch is being **repurposed** onto `main` to deliver the one state #1166 does not resolve — see #1188.

The original body is kept below, unedited, because the 2026-07-29 `blob-4` snapshot it records is the most direct evidence #1059 produced and is not written down anywhere on `main`. Its verdict framing is superseded by #1166's; its occurrence ledger is corrected by #1166's finding that `daily-history.jsonl` stores only the **last** attempt's signature (so 2026-07-13 was a #866 word-count-floor failure, not this one — 2026-07-22 still stands).

<details><summary>Original PR body</summary>

## Verdict: test-side defect, not environment and not a product regression

`agent-max-tokens.spec.ts` used the token-usage badge for **two jobs at once** — the completion signal *and* the spec's only observable. It is unfit for the first. Upstream renders that badge only when the message carries usage or duration metadata:

```ts
// messageMetadataComponent/index.tsx
if (!hasDuration && !hasTokens) return null;
```

So a response that completes **without** usage metadata renders no badge, and a badge-gated wait can never finish: the test burned the full 120 s on each of 3 attempts and reported `toHaveCount expected 1, received 0`. That signature named a count assertion while the cause was elsewhere — which is why #1059 was left with provider behaviour, load-induced truncation and a wait weakness all still open after triage.

The repo had already recorded this exact trap. `memory-history-regression.spec.ts` names the badge count as the **root cause of #569**: *"not every model/response emits the badge, so the count could sit below the expected value for the full 120s even though the response had already rendered."*

## Evidence

**The 2026-07-29 failure was a completed run.** Extracted the failure snapshot from `blob-4` of run 30444299314:

- `heading "The History of the Ocean" [level=1]` — the essay **rendered**
- the Send button is back, the input is active — generation **finished**
- **no** error card (`grep -c "error-card|An error occurred"` → 0)
- and no token-usage badge

**The occurrence ledger is wider than the issue records.** Checking every `toHaveCount` hit instead of assuming:

| Day | Target | Test | Attempts |
|---|---|---|---|
| 2026-07-13 | google/gemini-2.5-flash | causal control | 144 s |
| 2026-07-22 | google/gemini-2.5-flash | **the capped sibling** | 137 s / 142 s / 139 s — all three |
| 2026-07-29 | anthropic/claude-sonnet-5 | causal control | 247 s / 270 s |

Two providers, three days, always ~120 s+ ending in "no badge". Three consistent attempts pinned at exactly the badge wait is not the shape of wedge collateral, and the signature landing on the **capped** test too rules out the load-truncation reading.

**Two claims in the issue do not hold, and are corrected here:**
- 2026-07-21 is not a same-cause occurrence. It is google/gemini-2.5-flash on `toBeGreaterThan Expected: > 200 Received: 8` — the reply word-count floor that #866 has since **removed**.
- The 2026-07-30 flake on this spec is not a max_tokens problem either: `Expected "GOOGLE_API_KEY" Received "ANTHROPIC_API_KEY"`, a provider-key selection race in the template page object, on the google target. Untouched here.

`@stable` stays on both tests: all three `toHaveCount` occurrences fall on guard days, so no clean-day recurrence is established, and the 2026-07-30 clean daily passed anthropic on both tests. Nothing was quarantined, so nothing to lift.

## The change

`runPrompt()` waits on the **model-agnostic** generating indicator (`button-stop` hidden, `button-send` back — the signal #569 settled on), then resolves three outcomes, each naming itself:

| Outcome | Result |
|---|---|
| error card | fails with the provider's message, e.g. *"the agent run errored instead of returning a response — … Error code: 400 … credit balance is too low …"* |
| badge present | assertions proceed unchanged |
| completed, no badge | fails saying the Output-token observable is absent, so the `max_tokens` contract could not be evaluated — the 07-29 state |

Details that matter:
- **Settle window is conditional.** 15 s after a confirmed completion; the old 120 s budget is kept when the generating indicator was never observed, so a missed probe cannot become a premature failure.
- **The poll's outcome is captured and branched on once.** The error stack lives inside an `AnimatePresence`; re-querying it could drop the verdict into the badge path and restore the uninformative signature.
- **The poll's `message` is the failure's first line** on Playwright 1.58.2 (probed), which is what `append-weekly-history.mjs` records as `error_signature` — so the ledger now stores the state description instead of `toHaveCount expected 1, received 0`.

No new pass path: an errored or observable-less run still fails. What changes is what the failure says and how long it takes to say it — **247 s → 15 s**, measured.

## Verified / not verified

Verified locally on Langflow 1.12.0.dev10: the error path end to end (15 s, provider message verbatim), `tsc --noEmit`, eslint (pre-existing warnings only), the checklist coverage guard, `@stable` retained on both tests, and zero flow leakage on the new throw path (id-scoped `afterEach` still runs — 0 flows before and after).

**Not verified locally, deliberately stated rather than implied:** the green path and the completed-with-no-badge path both need a live provider key producing a real response, and every key reachable here is drained (Anthropic `400 credit balance is too low`, OpenAI quota #772/#775, google inactive). The no-badge branch's logic is small and its message is probe-verified, but it is not exercised end to end. `Last validated` is therefore **not** bumped to 1.12.x — the reason is recorded in the spec doc rather than only here. The next `daily-stable` run is the proof.

**#1059 stays open** for that validation, so this PR does not close it.

## Open product question (recorded, not asserted)

Why did a completed agent response arrive with **no usage metadata** on 07-22 and 07-29, when the same model reports it on 07-30? That is upstream behaviour this spec depends on for its only observable. Characterising it needs a live key; not claimed as a regression here.

## Review

Reviewed independently before opening. The review **rejected the first commit's verdict** ("environment") and it was right — the snapshot above is the evidence that settled it, and the three-way branch replaced a two-way one that would have left the 07-29 class failing at 120 s. Also applied: single-read branching, `readRunError` reading `.last()` and no longer letting a card with no provider message read as a cause, the doc's "Step by step" step 5 (it still described the removed badge-gated wait), and the stale precondition about `load()` wiping all flows (#553). One finding did **not** survive checking — that `expect.poll`'s `message` never reaches the thrown error; on 1.58.2 it is the first line, probed directly.

Refs #1059

---

## Update — the PR's own CI run adds the green-path proof I could not produce locally

`Run impacted E2E specs` on this PR (run 30594964218) executed this spec against a live Langflow with live keys: **3 passed, 2 skipped**, check green. The reworked `runPrompt()` therefore does complete normally with a real provider — the half I could not exercise locally.

Two things that run also shows, neither caused by this PR:

- **The CI Anthropic key is out of credit too.** `Collect models` reported: *provider "anthropic" inactive for a TRANSIENT billing/quota reason — 3 of 13 candidate models failed with the SAME model-independent error … "Your credit balance is too low to access the Anthropic API"*. So the anthropic pair **skipped** (graceful degradation working as designed, #980/#570) and the 2026-07-29 target specifically stays unproven until the key is topped up. This is the class #976 tracks.
- `[google / gemini-3.5-flash] max_tokens=50` failed once at `loadAgent` with `verdict read-failed` and passed on retry — a flake in the template page object's model selection, upstream of anything this PR touches.

So: green path proven on google (and the openai pair in the same run), anthropic still pending a live key. `Last validated` stays at 1.11.x until the anthropic target is exercised, for the reason recorded in the spec doc.


</details>

Issue #985, opened 2026-07-27
8 documented specs are absent from QA-CHECKLIST — nothing enforces the spec↔doc↔checklist triad


Victor-w-Madeira:
Surfaced while auditing #974. The audit needed to know whether the tests missing `@stable` belonged to *validated* specs, which meant cross-checking three sources that are supposed to agree: the spec file, its doc under `docs/`, and its bullet in `QA-CHECKLIST.md`. They do not agree, and nothing detects the disagreement.

## The gap

**8 specs have a spec doc under `docs/` but no entry at all in `QA-CHECKLIST.md`.** Someone invested in documenting them; the checklist — the artifact the Coverage Summary table is derived from — never learned they exist. Their coverage is therefore invisible in every generated count.

| Spec (under `regression/`) | `test()` calls | `@stable` | Doc |
|---|---|---|---|
| `core-components/general-bugs-component-webhook-api-key-display.spec.ts` | 2 | 0 | ✅ |
| `core-functionality/llm-agents/chatInputOutputUser-shard-0.spec.ts` | 2 | 0 | ✅ |
| `core-functionality/llm-agents/chatInputOutputUser-shard-2.spec.ts` | 2 | 0 | ✅ |
| `core-functionality/llm-agents/general-bugs-agent-images-playground.spec.ts` | 2 | 0 | ✅ |
| `flow-functionality/dragAndDrop.spec.ts` | 2 | 0 | ✅ |
| `flow-functionality/general-bugs-hidden-input-edges.spec.ts` | 1 | 0 | ✅ |
| `flow-functionality/general-bugs-shard-3836.spec.ts` | 2 | 0 | ✅ |
| `ui-ux/filterSidebar.spec.ts` | 1 | 0 | ✅ |

Verified by matching each spec's basename against `QA-CHECKLIST.md` — the reference convention for bullets is `- [x] <item> → \`<spec>.spec.ts\``, so a basename miss means genuinely unreferenced (spot-checked against description-based wording too, no false positives).

## Why this is worth fixing rather than back-filling by hand

The invariant that actually matters — **`@stable` ⇒ the spec is referenced in `QA-CHECKLIST.md`** — holds today: **zero** specs currently carrying `@stable` are missing from the checklist. So this is not (yet) a coverage lie in the stable lane.

[...]

Issue #899, opened 2026-07-22
Triage dataset: group failures by per-provider parameterization to surface provider-wide root causes


rafaelgiln:
## Motivation
On the 2026-07-22 daily (#879, run 29911701167), ~17 of the 22 hard failures shared a single deterministic root cause: the nightly shipped **without `langchain-google-genai`**, so every **Google-parameterized** run (agents' google variant, `google-provider`, `language-model-regression`, and the Google-embedding KB specs `rag-pipeline` / `vector-store-index-query`) failed to build the model and timed out (see #898).

The automated `--phase propose` triage ran correctly and produced a solid guard-tripped plan — but it **could not surface that shared cause**. It grouped by error signature and split the Google failures across three buckets: enrich #773 ("broad flakiness under parallel CI load"), reopen #789 ("RAG execution never completes"), and "today-only collateral". It even labeled two of them "google" without connecting them. The load/RAG framing is a **misattribution** of a deterministic package gap.

## Why it was missed
1. **Triage is shallow by design** — it groups symptoms and never root-causes (no systematic-debugging, no container logs). The `ImportError: langchain_google_genai` is only discoverable by investigation, which triage forbids. That is correct and should stay.
2. **The dataset has a blind spot:** it keys each failure by its `test` title, but a model-parameterized spec runs one describe per provider (`[model:gpt-4o-mini]`, `[anthropic / claude-sonnet-5]`, `[google / gemini-3.5-flash]`, …) that all **share the same test title**. So the dataset cannot tell that a failure is specifically the **Google** variant. The "all Google" dimension across ~17 failures is invisible to grouping.

## Proposal
Enhance `build-triage-dataset.mjs` (and the grouping guidance) to:
1. **Parse the parameterization label** out of each failure's full Playwright title / describe path — capture `provider` and `model` (e.g. from `[model:…]` or `[<provider> / <model>]`) as fields on each `hard_failures[]` / `flakes[]` entry.
2. **Add a provider-wide heuristic:** when N failures across ≥2 distinct areas all carry the **same provider variant** on the same run, emit a candidate cluster flagged `provider_wide: true` with that provider — a descriptive signal (not a verdict) that the cause is likely environment/package, not per-test rot or parallel-load flakiness.
3. Surface it in the panorama so the proposal groups the Google variant as one cluster instead of folding it into generic flakiness.

## Done when
- [ ] Dataset entries carry `provider` / `model` when the failing spec is parameterized.
[...]

PR #1181, opened 2026-07-31
docs(analytics): propose the E2E runs/specs analytics data model (#1180)

### LINKED ISSUE #1180: Design review: analytics data model for E2E runs and specs (fact table + spec/checklist dimensions)
## What this asks for

**Review and approve a data model** — no code ships against this issue until the model is approved. The companion PR adds a single design document; nothing executable changes.

The suite already ships every run into Postgres. What it does not have is a shape that answers longitudinal questions. This issue proposes that shape.

## Why now

Four questions get asked repeatedly and none of them can be answered today without reading artifacts by hand:

1. **Is this test actually stable?** `reports/README.md` already admits the limit: without a recorded pass list you can only say "it never appeared as a failure in the captured window", which is necessary but not sufficient.
2. **Which Langflow build introduced this regression?** Done manually every time (see the dev41→dev47 failure-spike investigation in #816/#830).
3. **Which spec inflated the daily from ~14 to ~22 minutes?** `duration_ms` is run-level only — per-test duration is not stored anywhere durable.
4. **Which recurring failure never got an issue?** Nobody measures this. It is the triage hole, and it is invisible by construction.

Five of the nine questions in the design's question→query map are not slow today — they are **impossible**, because the data was never recorded.

## Current state (verified, not remembered)

| Layer | Holds | Analytical limit |
|---|---|---|
| `reports/daily-history.jsonl` | One line per run: totals + `failures[]` + `flaky[]` | Failure-centric by design. No pass list, no per-test duration |
| `e2e_automation_runs` (QA Platform Postgres) | Scalar run columns + **`tests` JSONB** with every test (status, duration, tags, steps, error, screenshot URL) | Per-test data exists **only as a JSONB blob**. No fact table, no spec dimension, no usable index for aggregation |
| `flakiness-report/report.json` | Per-attempt detail, Flakiness.io format | Ephemeral, not persisted |

So the raw data has been landing in Postgres since 2026-06-10 (table born as `nightly_runs`, renamed in migration `20260630183535`). The gap is modelling, not plumbing.

## Proposed model

Four new tables in the QA Platform database, all **100% derived** — no column is editable by a human:

| Table | Grain | Purpose |
|---|---|---|
| `e2e_test_results` | one row per (run × test) | The fact. Every test, not only failures |
| `e2e_spec_catalog` | one row per spec file | Area/module dimension, derived from the repo |
| `e2e_coverage_items` | one row per Part II bullet | `QA-CHECKLIST.md` as queryable data |
| `e2e_test_aliases` | one row per intentional rename | Keeps a test's history intact across renames |

Plus five plpgsql RPCs (`e2e_test_health`, `e2e_spec_duration_trend` / `e2e_duration_movers`, `e2e_version_regression_window`, `e2e_coverage_rollup` / `e2e_stable_set_diff`, `e2e_failure_lifecycle`), matching the RPC pattern the platform already uses (`get_run_failure_treatment`, `get_qa_board_e2e`).

Issue data needs **no new table** — `e2e_dev_issues_cache`, `e2e_issue_spec_refs`, `e2e_product_regressions` and `jira_board_cards` are already populated by cron. What is missing is the left-hand side of the join: the failure as a row, with a timestamp.

## Decisions already taken

| Decision | Value |
|---|---|
| Where the tables live | QA Platform Supabase. `langflow-e2e` only produces payloads; it gains no database and reads none |
| Source of truth | Markdown wins. `QA-CHECKLIST.md` + the spec files stay authoritative; the database is a read-only projection, resynced on merge to `main`. PR guards (#741, #985) are untouched |
| Run sources | `daily-stable` (already POSTs) + `nightly` (new). PR-CI and `manual.yml` excluded — a per-run variable scope would poison trend lines |
| Volume | ~25k rows/month. No partitioning until ~5M |

## Reviewers: three things worth arguing about

1. **Materialising failures as rows reopens a documented decision.** The header comment of `20260726120000_e2e_issue_spec_refs.sql` deliberately chose *not* to: *"Rather than materialize failures as rows — which would duplicate the jsonb and reintroduce a human-declared flag — treatment is DERIVED"*. The design argues the objection was aimed at a **human-declared flag**, which the fact table does not introduce (it is derived and immutable), and that the read pattern is inverted: that decision optimised opening one run, while longitudinal analysis sweeps 90. The `tests` JSONB stays as the raw audit payload. **If this reasoning is wrong, the whole design changes** — say so here.
2. **Stable test identity.** `test_key` is a generated column, `md5(normalized_spec_path || '::' || title_path)`. Renaming a test mints a new key and `e2e_test_aliases` bridges it. Alternative proposals welcome — testid drift has already cost us history once.
3. **Backfill honesty.** Runs from ~2026-07 onward backfill completely from the stored JSONB. Older runs recover only their failures, with `duration_ms = NULL`, and will **never** have a per-test pass rate. This is documented as a known gap rather than back-filled by hand, matching how `reports/README.md` treats its missing weeks.

## Explicitly out of scope

- PR-CI and `manual.yml` ingestion (variable scope per run)
- Any new UI in the QA Platform — this delivers data and RPCs; screens are a separate cycle
- Attempt-level granularity (test-level plus an `attempts` count is enough for the questions asked)
- Partitioning and retention
- Merging with Flakiness.io
- Retiring `reports/*.jsonl` — it stays as the fallback for when the platform is down

## Implementation split (after approval)

- **`quality-platform`** — migrations for the four tables, the `e2e_ingest_run` atomic RPC, the five analytical RPCs, the backfill migration, and the `e2e-automation-runs-create` change from a direct `.insert()` to the RPC. This is the bulk of the work.
- **`langflow-e2e`** — `build-run-payload.mjs` emits `title_path`; new `build-catalog-payload.mjs`; `nightly.yml` gains a JSON reporter plus the payload/POST steps; unit tests in `npm run test:scripts`.

## Roadmap framing

Off the current wave (Wave 4 is *Canvas UI/UX & MCP*). Filed as the **scoping pass** for a pool item, per `ROADMAP.md` → *"Needs scoping first (Rule 1 — no inventing inside a wave)"*. Not scheduled by this issue; to be dated at a review. No milestone attached because the repo currently has none open.

rafaelgiln:
Closes #1180.

**This PR adds one document and nothing else.** No workflow, script, config or spec changes. Merging it approves a data model; it does not ship it.

> **Revision 2** — applies @Victor-w-Madeira's review. Every finding was verified against the repo before being accepted; two of them broke the design as written. Full verdict table in §11 of the document. Corrections to this PR body are at the bottom.

## What is being proposed

The suite has shipped every daily run into Postgres since 2026-06-10, but per-test data lands only as a JSONB blob on `e2e_automation_runs.tests`. That shape answers *"what happened in this run"* and nothing longitudinal. Five of the nine questions we keep asking are not slow — they are **impossible**, because no pass list, no per-test duration and no nominal scope list is ever recorded. `reports/README.md` § *Cannot answer with the current schema* already says so in writing.

Four derived tables in the QA Platform database:

| Table | Grain | Purpose |
|---|---|---|
| `e2e_test_results` | one row per (run × test) | The fact — every test, not only failures |
| `e2e_spec_catalog` | one row per spec file | Area dimension, derived from this repo |
| `e2e_coverage_items` | one row per Part II bullet | `QA-CHECKLIST.md` as queryable data |
| `e2e_test_aliases` | one row per intentional rename | Keeps a test's history intact across renames |

Plus an `is_valid_run` generated column on the existing run table, five analytical RPCs, and `e2e_rebuild_facts(run_id)` — the function that makes "the JSONB stays authoritative" an invariant rather than an assertion.

Issue data needs **no new table** — `e2e_dev_issues_cache`, `e2e_issue_spec_refs`, `e2e_product_regressions` and `jira_board_cards` are already populated by cron. What is missing is the left-hand side of the join: the failure as a row, with a timestamp.

## Constraints the design commits to

- **Markdown stays the source of truth.** `QA-CHECKLIST.md` and the spec files remain authoritative; the database is a read-only projection resynced on merge to `main`. The guards from #741 and #985 are untouched, and PR review does not change.
- **No human-editable column** in any new table. Everything is derived, so nothing can become a competing source of truth.
- **`tests` JSONB stays** as the raw audit payload; the fact table is a materialised index over an immutable source.
- **Correctness is a constraint, not a convention.** Where revision 1 relied on prose — "every RPC must exclude zero-test runs" — revision 2 uses a generated column the RPCs read.

## What changed in revision 2

The four blocking findings, and why two of them were not cosmetic:

1. **The backfill would have aborted, not merely disconnected history.** The historical `tests[]` JSONB carries only the leaf title, so backfilled rows must reconstruct `title_path` — and **17 spec files** use an interpolated `describe` where the leaf title is identical across providers. The daily of **2026-07-29** already holds two `agent-context-id-isolation.spec.ts` rows with the same leaf title. One `INSERT … SELECT` under `UNIQUE (run_id, test_key)` takes the whole migration down. Now: per-run backfill, `ON CONFLICT DO NOTHING`, reported drops, and a `title_path_exact` column so reconstructed identity cannot pass for exact. Ingest gets the same guard — otherwise atomicity turns one bad row into a lost run.
2. **`nightly.yml` is dormant.** Cron commented out (*"Schedule disabled until the suite is stable in production"*), last run **2026-03-19**, failed. Ingesting into it produces zero rows. It moves out of the core deliverable behind an explicit precondition (§2.3), and the volume estimate is recomputed daily-only (~5k rows/month, not ~25k).
3. **`e2e_spec_catalog.module` was not derivable from a path.** `MODULES` has 16 entries over **14 distinct directories** — `core-components/` and `ui-ux/` split by checklist *section*, not by path. `module` now lives only on `e2e_coverage_items`; `e2e_coverage_rollup()` joins the catalog on `spec_path`.
4. **`e2e_normalize_spec_path` is already `IMMUTABLE`** (`LANGUAGE sql IMMUTABLE`, migration `20260726120000`), so the generated column compiles. No work needed — recorded in the document so the next reader does not re-derive it.

Plus: `run_attempt` supersede (`github.run_id` is stable across re-runs, so a re-run of a red daily would otherwise leave the bad attempt permanently); `section` added to the `e2e_coverage_items` key; `p_workflow` defaulting to `'daily-stable'` instead of to the mixed answer the paragraph below it warned against; `attempts` allowing 0 for skipped tests; and extraction of the shared Playwright walk into `scripts/lib/playwright-report-walk.mjs` — `title_path` and `param` already existed in `append-weekly-history.mjs:88,96` while `build-run-payload.mjs:84` had neither, and adding a third consumer to that split without fixing it would have compounded a known capability gap.

## Still open (§11.1)

1. Is `md5(spec_path || '::' || title_path)` the right identity, given a **file move** breaks it exactly like a rename? Alias generation is now wired to the PR tooling, which mitigates but does not remove it.
2. Is Tier-B backfill worth landing, or is a clean cut-over less misleading than rows carrying `title_path_exact = false` and `duration_ms = NULL`?
3. Does `nightly.yml` get re-enabled? Not a #1180 decision, but it decides whether half the implementation split exists.

## Corrections to revision 1 of this PR body

- **Wrong:** *"Off the current wave — Wave 4 is Canvas UI/UX & MCP."* `origin/main:ROADMAP.md:193` has **Wave 5 — 1.11.0 feature coverage (2026-07-31 → 2026-08-14)** as current; Wave 4 closed early on 2026-07-29. The error came from reading a stale working-tree copy instead of `origin/main`.
- **Wrong:** *"No milestone attached because the repo currently has none open."* The **Wave 5 — 1.11.0 feature coverage** milestone is open with 7 issues (#1189–#1195).
- **Withdrawn:** the "scoping pass for a pool item" framing. The pool is Templates plus the auth / project-management tail; this is not in it, and `qa-infra` is not one of the three off-wave exception categories in `CLAUDE.md`. **The off-wave exception is therefore requested at review, not self-declared.** No milestone is attached, because this is not a Wave 5 deliverable.
- `Closes the design pass for #1180` did not auto-close anything; the canonical `Closes #1180` is now used.

## Verification

Nothing executable changed. Standard gates only:

```bash
npm run typecheck && npm run lint
```

Both exit 0 locally (lint reports 1307 pre-existing warnings, 0 errors — none from this branch). The impacted-specs job selects nothing: no spec, helper, fixture or config is touched.

### DIFF
--- a/ISSUE-1180-ANALYTICS-DB-DESIGN.md
+++ b/ISSUE-1180-ANALYTICS-DB-DESIGN.md
@@ -1,0 +1,379 @@
+# ISSUE-1180 — Analytics data model for E2E runs and specs
+
+**Status:** proposed — awaiting review approval. No code ships against this document until the model is approved.
+**Issue:** [#1180](https://github.com/oriontech-me/langflow-e2e/issues/1180)
+**Scope:** the QA Platform database (`quality-platform`) plus the payload producers in this repo.
+**Author:** design pass, 2026-07-31.
+
+---
+
+## 1. Problem
+
+The suite already ships every daily run into Postgres. It has done so since 2026-06-10, when the
+table now called `e2e_automation_runs` was created (born as `nightly_runs`, renamed in migration
+`20260630183535`). What the suite does **not** have is a shape that answers longitudinal questions.
+
+Three layers hold run data today, and each is blocked for analysis in its own way.
+
+| Layer | Holds | Analytical limit |
+|---|---|---|
+| `reports/daily-history.jsonl` | One line per run: `totals` + `failures[]` + `flaky[]` | Failure-centric **by design**. No pass list, no per-test duration. `reports/README.md` § *Cannot answer with the current schema* already enumerates this |
+| `e2e_automation_runs` (QA Platform Postgres) | Scalar run columns, plus a `tests` JSONB carrying every test — status, duration, tags, steps, error, screenshot URL | Per-test data exists **only as a JSONB blob**. No fact table, no spec dimension, no index that serves aggregation. Answering "flake rate per test over 90 days" means `jsonb_array_elements` across the whole history |
+| `flakiness-report/report.json` | Per-attempt detail, Flakiness.io format | Ephemeral, never persisted |
+
+The gap is modelling, not plumbing. The raw data is already arriving.
+
+### 1.1 Questions that are impossible today
+
+Not slow — impossible, because the data was never recorded in a queryable form.
+
+1. **"Is this test actually stable?"** Without a recorded pass list you can only say "it never
+   appeared as a failure in the captured window", which is necessary but not sufficient. The test
+   may have been renamed, silently skipped, or removed from `@stable`.
+2. **"Which Langflow build introduced this regression?"** Answered by hand every time — see the
+   dev41→dev47 failure-spike investigation behind #816/#830.
+3. **"Which spec inflated the daily from ~14 to ~22 minutes?"** `duration_ms` is run-level only.
+4. **"What entered and left the daily between two dates?"** No nominal list of what ran exists.
+5. **"Which recurring failure never got an issue?"** Nobody measures this. It is the triage hole,
+   and it is invisible by construction.
+
+---
+
+## 2. Architecture
+
+Everything new lives in the QA Platform Supabase database. This repo only **produces** payloads; it
+gains no database and reads none.
+
+```
+  langflow-e2e (repo)                    QA Platform (Supabase)
+  ───────────────────                    ──────────────────────
+  daily-stable.yml  ──┐
+  nightly.yml  (new) ─┴──POST──→  e2e-automation-runs-create (edge fn)
+                                        │ one transaction
+                                        ├──→ e2e_automation_runs   (exists, untouched)
+                                        └──→ e2e_test_results      NEW — the fact
+
+  build-catalog-payload.mjs (new)
+  run from update-coverage-summary.yml
+                     ───POST──→  e2e-spec-catalog-sync (edge fn, NEW)
+                                        ├──→ e2e_spec_catalog      NEW
+                                        └──→ e2e_coverage_items    NEW
+
+  (already running on cron, untouched)
+        sync-e2e-issues      ──→ e2e_dev_issues_cache, e2e_issue_spec_refs
+        sync-e2e-regressions ──→ e2e_product_regressions
+        sync-jira-board-hourly ─→ jira_board_cards
+```
+
+### 2.1 Invariants
+
+1. **No human-editable column in any new table.** Every value is derived — from the Playwright
+   payload or from the repository. This is what answers the objection recorded in
+   `20260726120000_e2e_issue_spec_refs.sql`, whose concern was a hand-declared flag becoming a
+   second source of truth.
+2. **The `tests` JSONB stays.** It remains the raw audit payload; the fact table is a materialised
+   index over an immutable source, not a competing copy. If the two ever disagree, the JSONB wins
+   and the fact is rebuildable by replay.
+3. **Every path column is normalised through `public.e2e_normalize_spec_path()`** — the function
+   that already exists. Issue bodies write `core-functionality/foo.spec.ts` while the runner JSONB
+   sometimes carries the `tests/` prefix; without the shared normaliser the join to
+   `e2e_issue_spec_refs` misses on a prefix difference.
+
+### 2.2 Decisions taken before this document
+
+| Decision | Value |
+|---|---|
+| Where the tables live | QA Platform Supabase |
+| Source of truth | Markdown wins. `QA-CHECKLIST.md` and the spec files stay authoritative; the database is a read-only projection, resynced on merge to `main`. The PR guards from #741 and #985 are untouched |
+| Run sources | `daily-stable` (already POSTs) + `nightly` (new). PR-CI and `manual.yml` are excluded — their per-run scope varies (the import-graph subset), which would poison trend lines |
+| Volume | ~25k rows/month. No partitioning until ~5M |
+
+---
+
+## 3. Data model
+
+### 3.1 `e2e_test_results` — the fact
+
+One row per (run × test). Roughly 250 rows per daily run, 800 per nightly.
+
+```sql
+CREATE TABLE public.e2e_test_results (
+  id             BIGSERIAL PRIMARY KEY,
+  run_id         TEXT NOT NULL REFERENCES public.e2e_automation_runs(run_id) ON DELETE CASCADE,
+
+  spec_path      TEXT NOT NULL,          -- always e2e_normalize_spec_path()'d
+  title_path     TEXT NOT NULL,          -- describe > describe > test, full chain
+  test_title     TEXT NOT NULL,          -- leaf title, for display
+  line           INTEGER,                -- display only; drifts on every edit to the spec
+
+  -- Stable identity as a GENERATED column: defined once, in the database, so the
+  -- edge function and the backfill cannot compute it differently.
+  test_key       TEXT GENERATED ALWAYS AS (
+                   md5(public.e2e_normalize_spec_path(spec_path) || '::' || title_path)
+                 ) STORED,
+
+  status         TEXT NOT NULL CHECK (status IN ('passed','failed','flaky','skipped')),
+  duration_ms    INTEGER,
+  attempts       SMALLINT NOT NULL DEFAULT 1,
+  tags           TEXT[] NOT NULL DEFAULT '{}',
+  param          TEXT,                   -- "google / gemini-2.5-flash" when parameterized
+
+  error_signature TEXT,                  -- first line, capped at 240 chars
+  error_full      TEXT,
+  screenshot_url  TEXT,
+  steps           JSONB,
+
+  UNIQUE (run_id, test_key)
+);
+
+CREATE INDEX ON public.e2e_test_results (test_key, run_id DESC);
+CREATE INDEX ON public.e2e_test_results (spec_path);
+CREATE INDEX ON public.e2e_test_results (status) WHERE status <> 'passed';
+CREATE INDEX ON public.e2e_test_results USING GIN (tags);
+```
+
+Two column choices carry the design:
+
+- **`title_path`, not just the leaf title.** The payload currently sends `test: spec.title`, which
+  drops the `describe` chain. A model-parameterized spec repeats the same leaf title under a
+  different `describe` (one per provider — the `param` case from #899), so keying on the leaf alone
+  collides. The producer must be changed to emit the full chain; see §4.1.
+- **`tags` is stored as it was AT THAT RUN**, not as it is today. That single choice makes
+  "evolution of the `@stable` set" and "scope diff between run A and run B" fall out of the fact
+  table, with no slowly-changing dimension to maintain. Both are listed as unanswerable in
+  `reports/README.md` today.
+
+### 3.2 `e2e_spec_catalog` — spec dimension
+
+One row per spec file, current state, derived from the repository.
+
+```sql
+CREATE TABLE public.e2e_spec_catalog (
+  spec_path         TEXT PRIMARY KEY,     -- normalized
+  module            TEXT NOT NULL,        -- one of the 16 in scripts/coverage-summary.ts MODULES
+  area              TEXT NOT NULL,        -- path segment: core-functionality/auth, mcp/server, …
+  functional_tags   TEXT[] NOT NULL DEFAULT '{}',
+  test_count        INTEGER NOT NULL DEFAULT 0,
+  stable_test_count INTEGER NOT NULL DEFAULT 0,
+  is_destructive    BOOLEAN NOT NULL DEFAULT false,
+  has_spec_doc      BOOLEAN NOT NULL DEFAULT false,
+  spec_doc_path     TEXT,
+  repo_commit_sha   TEXT NOT NULL,        -- which commit this row reflects
+  synced_at         TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+```
+
+`module` reuses the vocabulary of the `MODULES` array in `scripts/coverage-summary.ts` — no new
+taxonomy is invented. `repo_commit_sha` on every row is how a stale catalog is detected instead of
+silently trusted.
+
+### 3.3 `e2e_coverage_items` — the checklist as data
+
+One row per Part II bullet of `QA-CHECKLIST.md`.
+
+```sql
+CREATE TABLE public.e2e_coverage_items (
+  id              BIGSERIAL PRIMARY KEY,
+  module          TEXT NOT NULL,          -- same vocabulary as e2e_spec_catalog.module
+  section         TEXT,                   -- "15.10", "3.2" …
+  item_text       TEXT NOT NULL,
+  item_hash       TEXT NOT NULL,          -- md5(item_text)
+  marker          TEXT NOT NULL CHECK (marker IN ('x','-','~','!',' ')),
+  spec_path       TEXT,                   -- NULL when the item is not automated
+  phase           TEXT,                   -- 0 / 1 / 2
+  repo_commit_sha TEXT NOT NULL,
+  synced_at       TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  UNIQUE (module, item_hash)
+);
+```
+
+Keyed on `md5(item_text)` rather than line number: the text can repeat across modules, and a line
+number drifts on every edit. Sync is a **full replace per commit** inside one transaction —
+Markdown is authoritative, so incremental reconciliation would only add ways to drift.
+
+### 3.4 `e2e_test_aliases` — rename bridge
+
+```sql
+CREATE TABLE public.e2e_test_aliases (
+  old_test_key TEXT PRIMARY KEY,
+  new_test_key TEXT NOT NULL,
+  reason       TEXT,
+  created_at   TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+```
+
+Without it, every rename silently resets a test's series and "flake rate over 90 days" starts
+lying. Testid drift has already cost the suite history once.
+
+### 3.5 RLS
+
+Identical to the sibling cache tables: `SELECT` granted to `authenticated`, all writes restricted
+to `service_role`.
+
+---
+
+## 4. Ingestion
+
+### 4.1 Producer changes in this repo
+
+- **`scripts/build-run-payload.mjs`** emits `title_path` — the full `describe > … > test` chain —
+  alongside the existing leaf `test` field. Backwards compatible; the current payload keeps working
+  for any consumer that ignores the new field.
+- **`scripts/build-catalog-payload.mjs`** (new) walks `tests/**/*.spec.ts` for tags, `test()` counts,
+  `@stable` and `@destructive`; resolves docs under `docs/` by content reference; parses the Part II
+  bullets of `QA-CHECKLIST.md` reusing the `MODULES` taxonomy. It **fails loudly** when a module in
+  `MODULES` cannot be located, mirroring what `scripts/coverage-summary.ts` already does.
+- **`nightly.yml`** gains a `json` reporter plus copies of the `Build run payload` and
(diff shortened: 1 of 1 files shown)


Victor-w-Madeira (commented):
## Review

Design-only PR, verified against the repo rather than read on its own terms. The diagnosis holds: `build-run-payload.mjs:97` does emit `{ test: title }` with `title = spec.title` (the leaf), and `reports/README.md:165` already documents the gap in writing. Filename matches the `ISSUE-833-SHARDING-DESIGN.md` convention, no generated block is touched, no guard is exposed. The invariants in §2.1 — JSONB stays authoritative, no human-editable column, one shared normaliser — are the right three to commit to.

What follows is modelling, which is what §10 asks for.

---

### 1. The backfill does not just disconnect history — it aborts (blocking)

`test_key` is `md5(normalize(spec_path) || '::' || title_path)`, and §4.3 backfills from the historical `tests[]` JSONB. That JSONB has **only the leaf title** — `title_path` is precisely what §4.1 is adding now. So the backfill has to synthesise `title_path` from the leaf, and for a model-parameterized spec the leaf is identical across providers.

That is not hypothetical. From `reports/daily-history.jsonl`, run of **2026-07-29**:

```
2 x agent-context-id-isolation.spec.ts
    "switching the agent's context_id re-tags new turns without t…"
    (same file, same test, two different params)
```

Ten-plus specs use the `[${label}]` describe pattern (`agent-system-prompt`, `agent-multi-tool-selection`, `agent-input-sources`, `agent-context-id-*`, `mcp-client-agent`, …), so this recurs across the whole `llm-agents` and `mcp/client` surface.

Two consequences, both from `UNIQUE (run_id, test_key)`:

- **The backfill migration violates the constraint** and, being one `INSERT … SELECT`, takes the whole statement down — not the offending rows.
- Because §4.2 makes ingest **atomic**, any future run whose `title_path` fails to disambiguate loses the **entire run**, not one test.

So §4.3's "Complete" coverage and §8's "roughly 30 runs of history on day one" do not survive contact. And this reframes open question 3: the issue is not whether the *partial* pre-July backfill is worth it, it is that the *complete* one does not run.

This also means §3.1's justification for `title_path` is **more right than it claims** — it presents the collision as a `#899` concern, when it is already happening in production data today. Worth stating that way, since it is the strongest argument in the document.

Worth spelling out in §4.3 how the backfill assigns identity: either seed `e2e_test_aliases` programmatically from it, or accept a documented cut-over and say so, but not `INSERT … SELECT` as written.

### 2. `nightly.yml` is dormant — half the ingest ships zero rows (blocking)

The cron is commented out (`nightly.yml:4-5`), the last run was **2026-03-19**, and its issue step is gated on `github.event_name == 'schedule'`. The design nonetheless makes it a run source (§2.2), sizes the fact table on it ("800 per nightly", §3.1), rests the workflow-filter rationale on it (§5), and allocates work to it in §4.1 and §9.

Either re-enabling the cron becomes an explicit precondition — a team decision, not part of #1180 — or the nightly work leaves the implementation split. As written the PR approves work that produces nothing.

### 3. `e2e_spec_catalog.module` is not derivable from a spec path (blocking)

§3.2 says "one of the 16 in `scripts/coverage-summary.ts` MODULES". There are 16 entries but only ~13 distinct directories: `core-components/` splits into "Component Config" (`### 2.`) and "Core Components" (`### 3.`), and `ui-ux/` into Canvas (`## ui-ux/`) and Settings (`#### 15.10`). Those are **checklist section** splits, not path splits.

For `e2e_coverage_items`, derived from sections, this is fine. For `e2e_spec_catalog`, derived from the file tree, there is no `spec_path → module` function for those two areas — and that is exactly the join `e2e_coverage_rollup()` needs for `items_without_spec` / `specs_without_item`.

### 4. A re-run of a red daily is silently discarded

`run_id` is `github.run_id` (`build-run-payload.mjs:154`, `daily-stable.yml:821`), which is **stable across re-run attempts**; only `run_attempt` changes. With `ON CONFLICT (run_id) DO NOTHING → 200 exists`, re-running a daily that failed on infra leaves the fact table holding the bad attempt permanently, and §4.3 confirms a re-POST will not repair it. This is a real path — the team has re-run a daily over a Chromium-install flake. Suggest `run_attempt` in the key, or an explicit supersede path.

### 5. The zero-test-run exclusion is prose, not a constraint

§7 says "**Every RPC must** exclude runs where `passed = 0 AND failed = 0 AND flaky = 0`". That is an obligation on five RPCs and on every future one, it is NULL-unsafe if any total is nullable (the predicate goes NULL, not TRUE, and the run is kept), and it wrongly drops an all-skipped run, which is not an infra abort. This belongs in a view or a generated `is_valid_run` column the RPCs read from — otherwise #1012 comes back through the query layer.

### 6. `UNIQUE (module, item_hash)` should include `section`

Duplicate bullet text already exists — `QA-CHECKLIST.md:156` and `:470`, both "Outdated component notification → …". They survive only because they land in different modules. Two identical bullets under different sections of the *same* module (two `§15.x` subsections, both "ui-ux Canvas") would abort the full-replace transaction on a legitimate checklist edit.

### 7. The generated column needs an IMMUTABLE function

`GENERATED ALWAYS AS (md5(public.e2e_normalize_spec_path(…) || …)) STORED` only compiles if `e2e_normalize_spec_path` is declared `IMMUTABLE`. If it is `STABLE` — a common default for a helper like this — the migration fails. Worth confirming in `quality-platform` before any DDL, since the whole model rests on that column.

### 8. Nits

- §5: `p_workflow text DEFAULT NULL (NULL means all)` contradicts the sentence right below it ("filtering by workflow is effectively mandatory"). The default hands back the mixed answer.
- §3.1: `attempts SMALLINT NOT NULL DEFAULT 1` — the producer sends `(t.results || []).length`, which is `0` for a skipped test, and DEFAULT does not apply when a value is supplied.

---

## The part I would think hardest about: two histories, already diverging

§6 keeps `reports/*.jsonl` deliberately, as "a second copy, versioned in git, useful exactly when the platform is unreachable". I think that is right as an intent and wrong as currently reasoned, for two reasons the document does not surface.

**The declared safety net cannot restore what the fact table loses.** §7 handles a platform outage with "suite green, run not ingested. The JSONL in the repo remains the safety net." But the JSONL carries no `tests[]` — its keys are `date, duration_ms, failures, flaky, langflow_image, run_id, run_url, totals, version, workflow`. It holds failures and flaky only. So after an outage: the JSONL has the run, Postgres does not, a re-POST will not repair it (§4.3), and the SQL backfill depends on a `tests` JSONB that never arrived. Every outage is a **permanent, unrepairable hole** in the fact table, and the two series drift apart with different gaps. "Safety net" overstates what that file can do — it can restore the failure list, which is exactly the degraded shape §4.3 describes as never having a per-test pass rate.

**The duplication has already cost something, and this design walks into it.** `append-weekly-history.mjs` and `build-run-payload.mjs` are two independent parsers of the same Playwright report, each with its own `visit()`. They have quietly diverged in capability:

- `append-weekly-history.mjs:96` — `visit(node, suitePath = [])`, threads the describe chain
- `append-weekly-history.mjs:88` — `paramFromSuitePath()`, already extracts `[param]` from it
- `build-run-payload.mjs:82` — `visit(node)`, neither

So `title_path`, which §4.1 proposes to add to `build-run-payload.mjs`, is **already computed in the sibling script**, and the `param` column of §3.1 — which §4.1 never assigns a producer — already has a working implementation twenty lines away. Two parsers, one of which solved this problem some time ago, and a design that evolves the other one without noticing.

That is the actual cost of keeping both, and it is not the maintenance cost the document implies. My suggestion is not to retire the JSONL — the offline-copy argument stands. It is to make §4.1 **extract the shared walk into one module both scripts import**, so that the fact table and the JSONL cannot disagree about what ran. Right now the design adds a third consumer to a split that has already produced a silent capability gap, and §6 frames that as a settled trade-off rather than a thing to fix.

---

## On the three questions in §10

**1 — Materialising failures as rows reopens the `20260726120000` decision.** The reasoning holds, in my reading. That comment objected to a *human-declared flag* plus duplication; a derived, immutable, rebuildable fact table with the JSONB as tiebreaker is a materialised index, not a competing source of truth. The read-pattern argument (open one run vs. sweep ninety) is sound too. One gap: §2.1 says "rebuildable by replay", but §9 does not allocate the rebuild. That function is what makes the invariant true rather than asserted — it should be a named deliverable (`e2e_rebuild_facts(run_id)`), especially given finding 4, where a rebuild is the repair path.

**2 — Test identity.** The hash is fine; the weak component is `spec_path`. A **file move** breaks the key exactly as a rename does, and moves are more frequent in this repo than renames. The alias table covers both but manually, which fails precisely when a bulk reorganisation makes it expensive. Worth tying alias creation to the process — the PR tooling already knows which specs changed path — rather than to someone remembering.

**3 — Partial backfill.** Secondary to finding 1; the key story has to land first. Once it does, I would keep the partial backfill: failure-recurrence is the highest-value question and it does not need `duration_ms`. A documented *Known gaps* block is the right treatment, and consistent with how `reports/README.md` handles its missing weeks.

---

## PR body (not the document)

- "Off the current wave — Wave 4 is *Canvas UI/UX & MCP*" — Wave 4 is closed; the current wave is **Wave 5 — 1.11.0 feature coverage** (`ROADMAP.md:133`).
- "No milestone attached because the repo currently has none open" — **Wave 5 — 1.11.0 feature coverage** is open with 7 open issues.
- The "scoping pass for a pool item" framing: the pool in `ROADMAP.md:233` is Templates plus the auth / project-management tail; the analytics model is not in it, and #1180 is labelled `qa-infra`, which is not one of the three off-wave exception categories in `CLAUDE.md`. Not a blocker, but the exception should be granted at review rather than self-declared in the PR.
- `Closes the design pass for #1180` does not auto-close. If merging this closes #1180, use the canonical `Closes #1180`.

---

## Verdict

Strong document, and honest in the right places — it names the decision it reopens instead of quietly overriding it, which is why finding 1 was findable at all. I would not merge as-is, but nothing here needs a rewrite: findings 1, 2 and 3 are paragraph-level corrections, not architectural ones.

Minimum to approve:

1. Fix backfill identity (**1**) — as written the migration aborts on parameterized specs, with production data to prove it.
2. Resolve `nightly.yml`'s dormancy (**2**) — precondition, or drop the scope.
3. Define `spec_path → module` for `core-components/` and `ui-ux/` (**3**).
4. Confirm `e2e_normalize_spec_path` is `IMMUTABLE` (**7**) before any DDL.

And I would fold the shared-walk extraction into §4.1 rather than leaving §6 as a settled trade-off.

rafaelgiln:
Revision 2 is pushed (`b225918`). Every finding was checked against the repo before being accepted; the per-finding verdict table is §11 of the document. What follows is only where verification changed the outcome or diverged from the review.

## Findings 1–3 were load-bearing, and 1 was worse than described

**1 — backfill.** Confirmed, and the magnitude is higher: `describe` interpolation appears in **17** spec files, not ten-plus, concentrated in `llm-agents/` and `mcp/client/`. The 2026-07-29 pair you cited reproduces exactly.

The second consequence you drew — that atomicity turns a future collision into a **lost run** — was the more expensive half, and it changed the ingest design, not just the backfill. §4.2 now uses `ON CONFLICT (run_id, test_key) DO NOTHING` and returns `tests_received` alongside `tests_inserted`, so a collision costs one visible row instead of one silent run. Without that, §4.2's atomicity was actively harmful.

For identity on historical rows I took neither of your two options exactly. Seeding aliases does not help — a backfilled row cannot produce a correct key to alias *from*. A pure cut-over discards recoverable data. Instead the reconstruction is explicit in the schema: `title_path_exact BOOLEAN` (§3.1), false on every backfilled row, and `e2e_test_health` reports `rows_with_reconstructed_identity` as its own count. The degradation is a predicate, not a README paragraph.

Also adopted your framing point: the collision is now presented as production data in §3.1, not as a `#899` concern.

**2 — nightly.** Confirmed: cron commented with *"Schedule disabled until the suite is stable in production"*, last run 2026-03-19, failed. Nightly is out of the core deliverable behind an explicit precondition (§2.3), and the volume estimate is recomputed daily-only — **~5k rows/month, not ~25k**. The workflow filter in §5 stays regardless, since it costs nothing now and is required the moment a second source exists.

Separately: `CLAUDE.md` claims the nightly *"Runs daily at 03:00 BRT"*. Flagged in §2.3 as needing its own issue.

**3 — module.** Confirmed. 16 entries, **14** distinct directories. `module` is removed from `e2e_spec_catalog` entirely rather than made nullable — it is a checklist concept, so it lives only on `e2e_coverage_items`. `e2e_coverage_rollup()` joins the catalog on `spec_path`, which both of its outputs wanted anyway: `items_without_spec` is a coverage row whose path does not resolve, `specs_without_item` is a catalog row with no coverage row at all.

## 7 was already satisfied

`e2e_normalize_spec_path` is declared `LANGUAGE sql IMMUTABLE` in `20260726120000_e2e_issue_spec_refs.sql`. The generated column compiles as written — no DDL risk, no work. Recorded in §2.1 so it does not need re-deriving.

## Two corrections to the review

**6 — the duplicate is at line 459, not 470**, and there is exactly **one** full-line duplicate bullet in the file, not several: *"Outdated component notification → …"* at 156 (`#### 2.3 Component Updates`) and 459 (`#### 8.2 Notifications`). Your mechanism is right and the fix is applied — and note that `section` also had to become `NOT NULL DEFAULT ''`, because a nullable column does not participate in a UNIQUE index and the constraint would have let duplicates through anyway.

**Roadmap — you are right and I was wrong.** `origin/main:ROADMAP.md:193` has Wave 5 as current and the Wave 5 milestone is open with 7 issues. My revision-1 claim came from reading a stale working-tree copy instead of `origin/main`. The PR body is corrected, the "scoping pass for a pool item" framing is withdrawn, and the off-wave exception is now requested at review rather than self-declared.

## On the two-histories section

Adopted, including the reframing. §6 no longer presents the duplication as a settled trade-off: it states that the JSONL cannot restore what an outage costs (no `tests[]`, so recovery lands in the degraded Tier-B shape), and §4.1 extracts `scripts/lib/playwright-report-walk.mjs` so both scripts share one traversal. Confirmed the gap you identified: `append-weekly-history.mjs:96` threads `suitePath` and `:88` has `paramFromSuitePath()`; `build-run-payload.mjs:84` has neither. `param` had no producer assigned in revision 1 — it does now, via the shared module.

Your §10 note on `e2e_rebuild_facts(run_id)` is adopted as a named deliverable in §10 and as the repair path in §8. And the `spec_path` weakness — a file move breaking identity like a rename — is now stated in §3.4, with alias generation wired to the PR tooling that already computes changed paths rather than to someone remembering.

## Remaining disagreement: none blocking

The three items in §11.1 are still open, and question 3 (`nightly.yml`) is a team decision rather than a design one. If you consider the four blocking items resolved, this is ready for a second pass.

Issue #1185, opened 2026-07-31
daily-stable: rotate the tool-calling tier through one provider per weekday, falling back to the next active one


Victor-w-Madeira:
Follow-up of #1169 (PR #1170). That issue pinned `pr-validation.yml`; this one addresses
`daily-stable.yml` — the lane #1170 deliberately left multi-provider.

> **Revised 2026-07-31.** The original proposal was a **fixed pin** to one provider. It is
> now a **weekday rotation with fallback to the next active provider**. Same cost (one
> provider per run), but the blindness window becomes automatic instead of depending on
> someone dispatching a lane, and a drained key costs a deviation rather than a day. The
> reasoning and the superseded design are both kept below.

## Problem

The daily runs **~30 `@stable` agent tests × every active provider**. Each is a multi-turn
agent run, and the Simple Agent tool schemas are re-sent every turn (Langflow sets no
`cache_control`, so nothing is cached on the Anthropic side).

Measured shape of the suite: **25 specs drive a real agent** (`SimpleAgentTemplatePage`),
**17 of them parametrize over providers**. A healthy daily is **466 tests in 16 min** across
4 duration-balanced shards (2026-07-30).

Price of the multiplication:

| Target the spec resolves | $ / MTok in-out | vs `gpt-4o-mini` |
|---|---|---|
| `gpt-4o-mini` | 0.15 / 0.60 | 1x |
| `claude-sonnet-5` (what anthropic settles to today) | 3 / 15 | **20x / 25x** |

The PR lane stopped paying this on 2026-07-31. The daily still does, every weekday.

## Why this does not cost multi-provider coverage

The obvious objection — "the daily is the lane that owes multi-provider coverage" (#1169's
own words) — is answered twice over.

**First, a layer that already exists and is already `@stable`:**
`core-functionality/model-provider/{openai,anthropic,google,ollama,groq,mistral}-provider.spec.ts`,
each carrying a **Settings** test and a **real execution** test. Verified: **none of them
reads `MODEL_TEST_ID` or `MODEL_TEST_PROVIDER`**, so a lane pin does not touch them. Every
provider keeps being exercised end-to-end, every day. #1184 adds a unit test that fails if
one of them starts reading a pin variable, so that independence is guarded rather than
incidental.

**Second, the rotation itself** — agent behaviour still runs on every provider, just not on
the same day.

The distinction the change encodes:

- **Provider contract** — "the key authenticates, the catalog lists, the flow runs on this
[...]

Issue #1184, opened 2026-07-31
provider-setup: extract getTestTargets() into one tier-aware resolver — it is copy-pasted into 17 specs and no spec can declare what it needs from a model


Victor-w-Madeira:
Follow-up of #1169 (PR #1170). Foundation for the rest of the model-selection thread — no
behaviour change, so it can land before the cost baseline exists.

## Problem

**`getTestTargets()` is copy-pasted into 17 spec files.** It is not a helper; each spec
carries its own ~45-line copy, along with its own `getModelsFromJson()`, its own
`ModelRecord`/`TestTarget` interfaces, and its own reading of `MODEL_TEST_ID` /
`MODEL_TEST_PROVIDER` / `ALL_MODELS`:

```
core-functionality/llm-agents/agent-component-regression.spec.ts
core-functionality/llm-agents/agent-context-id-continuity.spec.ts
core-functionality/llm-agents/agent-context-id-isolation.spec.ts
core-functionality/llm-agents/agent-current-date-tool.spec.ts
core-functionality/llm-agents/agent-empty-refusal-response.spec.ts
core-functionality/llm-agents/agent-input-sources.spec.ts
core-functionality/llm-agents/agent-markdown-output.spec.ts
core-functionality/llm-agents/agent-max-iterations.spec.ts
core-functionality/llm-agents/agent-max-tokens.spec.ts
core-functionality/llm-agents/agent-multi-tool-selection.spec.ts
core-functionality/llm-agents/agent-multimodal-image-input.spec.ts
core-functionality/llm-agents/agent-structured-output.spec.ts
core-functionality/llm-agents/agent-system-prompt.spec.ts
core-functionality/llm-agents/agent-tool-error-handling.spec.ts
core-functionality/llm-agents/agent-tool-inspection.spec.ts
core-functionality/llm-agents/agent-tool-name-validation.spec.ts
mcp/client/mcp-client-agent.spec.ts
```

Two consequences:

1. **Any model-selection policy is a 17-file edit**, and a partial migration is invisible —
   a spec left on the old copy silently keeps the old behaviour.
2. **A spec cannot declare what it actually needs from a model.** The resolver treats
   `agent-multi-tool-selection` (whose assertion *is* tool-calling quality) and
   `settings-message-history` (which needs any string back) identically. Every downstream
   decision in this thread — pinning a lane, routing to a cheaper target, routing to local
   inference — has to guess per spec, with no place to record the answer.

## Proposal

[...]

Issue #1183, opened 2026-07-31
Measure the suite's real LLM spend per lane and per provider — every cost decision so far is argued from price ratios, not a bill


Victor-w-Madeira:
Follow-up of #1169 (PR #1170). That issue removed the PR lane's multi-provider volume on a
**price-ratio** argument. Every remaining decision on the cost axis — pinning the daily,
routing specs to local inference, sizing the runner — needs a **measured** baseline, and
none exists.

## Problem

Two independent cost axes are being discussed as if they were one:

| Axis | What spends | Status |
|---|---|---|
| **A** — automation agents | the Claude Code action triaging the daily (`triage-dispatch.yml`) | #1178 → PR #1179, merged: input gate + Sonnet pin, $4.71 → ~$2.80 over the measured window |
| **B** — provider credits in specs | the Langflow Agent calling OpenAI / Anthropic / Google | #1169 (PR #1170, merged) and #1171 (open) — both argued from **$/MTok ratios**, not from a bill |

The only spend figures in the repo are the ones #1179 pulled from a single action log.
They point in a direction that has not been checked: the whole E2E suite's Anthropic spend
on 2026-07-31 read as **~$0.25**, against **$1.66** for one triage-agent run. That comparison
cannot be used as-is — 07-31 is the day the daily **aborted with zero tests** — but it is
enough to make the question real: *we may be optimising the smaller axis.*

Meanwhile the qualitative signal says axis B is not small: the CI Anthropic secret and the
local `.env` key **both drained inside a four-day window** (#1169), which is what started
this whole thread.

Both statements cannot be sized against each other without numbers.

## What to measure

Per day, over a representative window, split by:

- **Provider** — Anthropic / OpenAI / Google.
- **Lane** — `pr-validation.yml`, `daily-stable.yml`, `manual.yml`, the migration workflows,
  and **local `.env` usage** (developer machines).
- **Axis** — automation-agent tokens (the Claude Code action's own `total_cost_usd`, which
  it already prints per run) vs. provider credits burned by the specs.

Cross-reference `reports/daily-history.jsonl` for which days actually executed tests.

## Confounders that will corrupt the reading if not controlled

1. **Anthropic credit is account-scoped.** The CI secret and the local `.env` key share one
   balance (#1169), so console totals do not attribute to a lane on their own. Either
   separate the keys first, or attribute by correlating timestamps against workflow run
   times.
[...]

Issue #1169, opened 2026-07-31
pr-validation: the lane pays for multi-provider LLM coverage it does not owe (141 runs / 3.5 days on claude-sonnet-5)


rafaelgiln:
## Problem

The PR lane pays for multi-provider LLM coverage it does not owe, and the bill is what
drained the Anthropic key twice in four days.

`getTestTargets()` (~17 agent specs, plus 2 more that read `MODEL_TEST_PROVIDER` only)
parametrizes over **one model per ACTIVE provider**. Every provider key in the repo
secrets therefore multiplies the impacted-specs run: an openai variant *and* an anthropic
variant *and* a google variant of every selected LLM spec. That is the right shape for
`daily-stable.yml` — it is the lane that owes multi-provider coverage. It is the wrong
shape for `pr-validation.yml`, which answers "does this PR break the specs it touches",
and one provider settles that.

## What the data says

Measured 2026-07-31, after the CI Anthropic secret went dry mid-morning:

| Lane | Runs since 2026-07-28 (3.5 days) |
|---|---|
| `pr-validation.yml` | **141** (37 / 49 / 49 / 6 per day) |
| `daily-stable.yml` | 7 (466 tests on 07-30) |
| migration + manual | 5 |

Per full `@stable` daily there are ~33 Anthropic-billed agent tests (≈30 `test()`
declarations inside the parametrized loops, plus the three `anthropic-provider.spec.ts`
tests). Order of magnitude across the window: **~10³ paid agent inferences**, the
majority of them driven by the PR lane, not the daily.

Price is the multiplier, not volume:

| Target the spec resolves | $ / MTok in-out | vs `gpt-4o-mini` |
|---|---|---|
| `gpt-4o-mini` (openai settled) | ~0.15 / 0.60 | 1x |
| `claude-sonnet-5` (anthropic settled, confirmed via the `param` field in `reports/daily-history.jsonl`) | 3 / 15 | **20x / 25x** |
| `claude-opus-5` (latent — see below) | 5 / 25 | 33x / 42x |
| `claude-haiku-4-5` | 1 / 5 | 7x / 8x |

Two compounding factors:

1. **No prompt caching on the Anthropic side.** Langflow sets no `cache_control`, so
   every agent turn re-sends the system prompt and the tool schemas at full input price.
   OpenAI discounts a repeated prefix automatically. Effective gap is wider than the
   sticker ratio.
[...]

PR #1209, opened 2026-07-31
perf(collect-models): rank claude-haiku-4-5 ahead of sonnet for anthropic (#1171)

### LINKED ISSUE #1171: collect-models: CANDIDATE_PREFS.anthropic is sonnet-first — now ~87% of the daily's agentic spend after the rotation
Follow-up of #1169 (the PR-lane pin, PR #1170). That issue removed the *volume*; this one is
the *unit price*.

> **Revised 2026-07-31**, after #1185 (PR #1207) landed the weekday rotation. The original
> arithmetic assumed Anthropic ran every weekday; it now runs twice a week. That makes this
> issue **more** valuable, not less — the rotation concentrated the remaining spend into
> Anthropic. The target model, the exact one-line change, and why the other two providers
> stay untouched are all pinned down below.

## Problem

`CANDIDATE_PREFS.anthropic` in `tests/helpers/provider-setup/collect-models.ts` is:

```ts
anthropic: [/^claude-sonnet-5$/, /sonnet/, /haiku/],
```

Sonnet-first. `collect-models` promotes whatever settles to `models.json[0]`, and the shared
resolver (`resolveTestTargets`, extracted in #1184) reads that — so **every Anthropic agent
variant runs on `claude-sonnet-5`**. Confirmed in the current catalog: `claude-sonnet-5` at
index 0, `claude-opus-5` pushed to 1, `claude-haiku-4-5` at index 8 of 13.

## Price, checked against the vendor's current rates

| Model | $ / MTok in-out | vs haiku |
|---|---|---|
| `claude-sonnet-5` | **2 / 10** — introductory, through **2026-08-31** | **2x** |
| `claude-sonnet-5` | 3 / 15 — from 2026-09-01 | **3x** |
| `claude-haiku-4-5` | 1 / 5 | 1x |

The original issue said "3x". That is right from **September 1**; today it is **2x**, because
Sonnet 5 is still on introductory pricing for one more month. Anyone measuring in August and
again in September will see the saving grow on its own — that is the price change, not the
change in this issue.

Haiku 4.5's 200K context (vs Sonnet 5's 1M) is not a constraint here: the agent specs send
short prompts and small tool schemas.

## What the rotation changed — the argument got stronger

#1185 moved the daily from three providers every day to one provider per weekday
(Mon openai · Tue anthropic · Wed google · Thu openai · Fri anthropic). Anthropic went from
5 days to 2 — but because it is ~13x the price of the openai target, it now dominates what
is left.

With `A` = one agent test on `claude-sonnet-5` at today's introductory rates,
`gpt-4o-mini ≈ A/13`, `gemini-2.5-flash ≈ A/7`:

```
weekly, post-#1185:  30 × (A/13 + A + A/7 + A/13 + A)  ≈ 69A
        Anthropic share (Tue + Fri):            60A    ≈ 87% of the total
```

So this is no longer "3x on one of three providers" — it is **the dominant remaining lever on
the daily's agentic spend**:

| | Reduction in the daily's agentic spend |
|---|---|
| Today (2x) | **~43%** |
| From 2026-09-01 (3x) | **~61%** |

Ratio-derived, not measured, and assuming comparable token volume per test across providers
(reasonable — same prompts, same tool schemas; Anthropic additionally pays more because
Langflow sets no `cache_control`). The absolute figure is #1183's job.

## The change

One line, one provider, one model. `claude-haiku-4-5` is the **only** haiku in the 13-model
catalog, so the target is unambiguous:

```ts
// before
anthropic: [/^claude-sonnet-5$/, /sonnet/, /haiku/],

// after
anthropic: [/^claude-haiku-4-5$/, /haiku/, /^claude-sonnet-5$/, /sonnet/],
```

The generic `/haiku/` after the exact id is cheap future-proofing — a future `claude-haiku-5`
matches it. Keeping sonnet at the tail matters: without it, a catalog with no haiku falls
through to raw catalog order, which today leads with `claude-opus-5` — the most expensive
model in the list.

## Why openai and google stay untouched

**openai — deliberately not cheaper.** The catalog holds 41 models, several nominally cheaper
than `gpt-4o-mini`: `gpt-5-nano`, `gpt-5.4-nano`, `gpt-5-mini`, `gpt-5.4-mini`. They are
**reasoning models**, and that has already cost this repo: the suite detects them via
`metadata.reasoning_models` (`setup-language-model-openai.ts:352`) because they hang the
playground for 120 s (#569 / PR #610). The current list —
`gpt-4o-mini` → `gpt-4o` → `gpt-4.1(-mini|-nano)` → `gpt-4*` — is entirely gpt-4 family **by
design**. The one legitimately cheaper in-family candidate, `gpt-4.1-nano`, is already third
in the list; promoting it is a separate experiment with the same 3/3 gate, and a poor
risk/reward while openai is ~1/13 of the bill.

**google — the list is already the cheap tier.** All three entries are *flash*; there is
nothing to optimise.

## Separate finding, worth its own look

**Google currently contributes nothing.** It has **zero entries** in `models.json` and probes
`inactive` ("no models collected from the providers panel"), so its preference list is dead
letter and the rotation's Wednesday slot falls through to openai. If that is local-only it
disappears on the daily; if it is the CI key, the rotation will spend two weekdays on openai
and **Google agent coverage stays at zero** until someone looks. Confirm on the first daily
after PR #1207 merges — the `::warning::` naming `google` will say so explicitly.

## Why this is not a one-line PR

`CANDIDATE_PREFS` does not encode *access*, it encodes **agent compatibility**. It exists
because #570 found `gpt-5.6` passing the raw probe (a ~1-token completion) while the Agent
returned empty replies — the probe cannot tell "the key can reach this model" from "this model
drives our agent specs". A weaker model is exactly where that gap bites:

- tool selection and multi-tool orchestration (`agent-multi-tool-selection`, `agent-tool-inspection`)
- structured output (`agent-structured-output`)
- markdown fence behaviour (`agent-markdown-output` — already flaky under model churn, #826)
- refusal/empty-reply handling (`agent-empty-refusal-response`)
- verbosity-sensitive assertions (#866)

So this needs a **real run**, not a unit test. A unit test can only assert the ordering, and an
ordering assertion that ships a model the Agent cannot drive is worse than no change.

## Validation protocol

Both sides on the same Langflow, `PLAYWRIGHT_RETRIES=0` so a flake is not masked, scoped to the
`@stable` tests under `core-functionality/llm-agents`:

```bash
# candidate
MODEL_TEST_ID=claude-haiku-4-5 MODEL_TEST_PROVIDER=anthropic PLAYWRIGHT_RETRIES=0 \
  npx playwright test tests/tests-automations/regression/core-functionality/llm-agents \
  --grep @stable --workers=2

# baseline, only if the candidate shows failures — needed to separate a haiku weakness
# from a spec that is already red
MODEL_TEST_ID=claude-sonnet-5 MODEL_TEST_PROVIDER=anthropic PLAYWRIGHT_RETRIES=0 \
  npx playwright test tests/tests-automations/regression/core-functionality/llm-agents \
  --grep @stable --workers=2
```

Candidate-first is deliberate: a clean haiku run makes the baseline unnecessary and halves the
spend.

**Confirm the Anthropic key has credit before starting.** The shared balance drained twice in
the week of 2026-07-28 (#1169), and a run that dies on a 400 says nothing about haiku while
still costing the wall-clock.

### Done when

- [ ] The candidate run is recorded here with its `N passed / N failed` summary, per-spec.
- [ ] Any failure is attributed — haiku-specific, or already red on sonnet — with the baseline
      run as evidence when needed.
- [ ] `CANDIDATE_PREFS.anthropic` is reordered **only if** haiku matches sonnet on the whole
      set. A partial result closes this issue as *not adopted*, with the failing specs named,
      rather than shipping a cheaper model that erodes coverage.
- [ ] Orphan flow count checked before the report (every touched spec cleans up id-scoped).

## Out of scope

`resolveTestTargets` reading `models.json[0]` rather than the settled `providers.json` pick.
Post-#570 promotion makes the two agree on a passing sweep, and a failing sweep skips the specs
anyway — tracked as the latent-risk note in #1169, not fixed here.

Victor-w-Madeira:
Closes #1171.

> ### ⛔ Draft on purpose — do not convert until the validation run is recorded
> This ships a **cheaper model**, and `CANDIDATE_PREFS` encodes **agent compatibility**, not price. The evidence that `claude-haiku-4-5` drives the agent suite is a real `@stable` run that **has not happened**: the shared Anthropic balance is dry (`credit balance is too low`, probed 2026-07-31 — both sonnet and haiku fail identically, so it is account-level, not model access). Draft status is the gate; a note in a description is not.
>
> To unblock: top up, then run candidate-first — the command is in #1171 and repeated at the bottom.

## The change

One line, one provider, one model.

```diff
-anthropic: [/^claude-sonnet-5$/, /sonnet/, /haiku/],
+anthropic: [/^claude-haiku-4-5$/, /haiku/, /^claude-sonnet-5$/, /sonnet/],
```

Probe order against the real 13-model catalog:

| | First four probed |
|---|---|
| Before | `claude-sonnet-5` → `claude-sonnet-4-6` → `claude-sonnet-4-5` → `claude-sonnet-4-20250514` |
| After | **`claude-haiku-4-5`** → `claude-sonnet-5` → `claude-sonnet-4-6` → `claude-sonnet-4-5` |

`collect-models` promotes whatever settles to `models.json[0]`, and the shared resolver (`resolveTestTargets`, #1184) reads that — so this decides what every parametrized Anthropic agent variant runs against.

## Why it is worth a PR

| Model | $ / MTok in-out |
|---|---|
| `claude-sonnet-5` | **2 / 10** introductory, through **2026-08-31**; **3 / 15** after |
| `claude-haiku-4-5` | 1 / 5 |

**2x today, 3x from September 1.** #1171 said 3x — right from September, a month early today. Anyone measuring across that boundary will see the saving grow on its own; that is the vendor's price change, not this PR.

**The rotation made this more valuable, not less.** #1185 (PR #1207) moved the daily to one provider per weekday, so Anthropic went from 5 days to 2 — but at ~13x the openai target's price it now carries **~87% of the daily's remaining agentic spend**:

```
weekly, post-#1185:  30 × (A/13 + A + A/7 + A/13 + A)  ≈ 69A
        Anthropic (Tue + Fri):                  60A    ≈ 87%
```

| | Reduction in the daily's agentic spend |
|---|---|
| Today (2x) | **~43%** |
| From 2026-09-01 (3x) | **~61%** |

Ratio-derived, not measured. The absolute figure is #1183's job.

## Two things that look like detail and are not

**The sonnet tail is load-bearing.** Without it, a catalog containing no haiku falls through to **raw api order**, which leads with `claude-opus-5` — the most expensive model Anthropic exposes here. That only bites on a *failing* sweep, where nothing gets promoted, which is exactly #1169's latent-risk note.

**openai and google are deliberately untouched.** openai's cheaper catalog entries — `gpt-5-nano`, `gpt-5.4-nano`, `gpt-5-mini`, `gpt-5.4-mini` — are **reasoning models**, and this repo already paid for that: it detects them via `metadata.reasoning_models` because they hang the playground for 120 s (#569 / PR #610). The `gpt-4o-mini → gpt-4o → gpt-4.1(-mini|-nano) → gpt-4*` list is that constraint, not an oversight. google's three entries are already the flash tier.

## Tests, and what they do not prove

`rankCandidates` is now exported so the ordering can be pinned. Six tests: haiku first; sonnet still reachable; a future `claude-haiku-5` outranks sonnet; openai still leads with `gpt-4o-mini` and no gpt-5.x outranks a gpt-4-family model; google unchanged; an unknown provider falls back to raw order rather than dropping models.

**Ordering is all a unit test can prove here, and the test file says so in its own header.** The probe is a ~1-token completion that cannot tell "the key reaches this model" from "the Agent can drive it" — #570 found `gpt-5.6` passing the probe while the Agent returned empty replies. These guard against a *silent reordering*; they are not evidence that haiku drives the agent suite.

**A force-fail found one of them was vacuous.** Deleting the sonnet tail left the suite green: the fixture it filtered was already in *promoted* order, so removing the tail changed nothing. It now builds a raw unpromoted catalog led by opus — the state the tail actually protects against — and goes red when the tail is removed.

| Check | Result |
|---|---|
| `npm run test:units` | 284 passed, 0 failed (6 new) |
| `npm run test:scripts` | 284 passed, 0 failed |
| `npm run typecheck` / `lint` | clean / 0 errors |
| Force-fail ×2 | sonnet-first restored → 2 red · sonnet tail removed → 1 red (after the fix above) |

## What must happen before this merges

```bash
# 1. confirm credit — this currently returns invalid_request_error
curl -s https://api.anthropic.com/v1/messages \
  -H "x-api-key: $ANTHROPIC_API_KEY" -H "anthropic-version: 2023-06-01" \
  -H "content-type: application/json" \
  -d '{"model":"claude-haiku-4-5","max_tokens":4,"messages":[{"role":"user","content":"Say OK."}]}'

# 2. candidate-first — a clean run makes the sonnet baseline unnecessary and halves the spend
MODEL_TEST_ID=claude-haiku-4-5 MODEL_TEST_PROVIDER=anthropic PLAYWRIGHT_RETRIES=0 \
  npx playwright test tests/tests-automations/regression/core-functionality/llm-agents \
  --grep @stable --workers=2
```

Adopt **only if** haiku matches sonnet across the set. A partial result closes #1171 as *not adopted* with the failing specs named — shipping a cheaper model that erodes coverage is the failure this gate exists to prevent. Record the per-spec summary on #1171 either way.

## Related state worth knowing while this sits

With Anthropic dry **and** google already `inactive` (zero entries in `models.json`), #1185's rotation resolves to openai on every weekday — verified against `scripts/select-daily-model-target.mjs` with the real provider state. The fallback works and no day is lost, but the practical daily is **openai-only**: Anthropic and Google agent coverage are both at zero until the balance is topped up and the Google key is looked at.

### DIFF
--- a/tests/helpers/provider-setup/collect-models.test.ts
+++ b/tests/helpers/provider-setup/collect-models.test.ts
@@ -23,6 +23,7 @@ import assert from "node:assert/strict";
 import * as fs from "fs";
 import * as path from "path";
 import {
+  rankCandidates,
   validateProviderWithFallback,
   type ProviderRecord,
 } from "./collect-models";
@@ -387,4 +388,83 @@ test("first candidate works: exactly one probe, record returned untouched", asyn
   assert.equal(result.status, "active");
   assert.equal(result.model, "gpt-4o-mini");
   assert.equal(result.error, null);
+});
+
+// ─── CANDIDATE_PREFS ordering (#1171) ────────────────────────────────────────
+//
+// What a unit test can and cannot prove here. `rankCandidates` decides which
+// model `collect-models` probes FIRST, and the first one that validates is what
+// it settles on and promotes to models.json[0] — which every parametrized agent
+// spec then runs against. So the ordering is a real cost and coverage lever.
+//
+// But ordering is ALL a unit test can prove. CANDIDATE_PREFS encodes agent
+// COMPATIBILITY, and the probe is a ~1-token completion that cannot tell "the
+// key reaches this model" from "the Agent can drive it" — #570 found `gpt-5.6`
+// passing the probe while the Agent returned empty replies. These assertions
+// therefore guard against a silent reordering; they are NOT evidence that the
+// leading model drives the agent suite. That evidence is the real @stable run
+// recorded on #1171.
+
+test("anthropic ranks claude-haiku-4-5 first — 2x cheaper than sonnet today, 3x from 2026-09-01", () => {
+  const ranked = rankCandidates("anthropic", ANTHROPIC_CATALOG);
+  assert.equal(ranked[0], "claude-haiku-4-5");
+});
+
+test("anthropic keeps sonnet reachable — a haiku-less catalog must not fall through to opus", () => {
+  // The catalog here is in RAW api order, not the promoted order collect-models
+  // writes after a successful sweep. That distinction is the whole test: a
+  // failing sweep promotes nothing, so the raw order stands — and it leads with
+  // claude-opus-5, the most expensive model Anthropic exposes (#1169's latent
+  // risk note). Filtering the promoted ANTHROPIC_CATALOG instead would prove
+  // nothing, because that fixture already starts with sonnet: the assertion
+  // would pass with the sonnet tail deleted.
+  const rawUnpromoted = [
+    "claude-opus-5",
+    "claude-fable-5",
+    "claude-opus-4-8",
+    "claude-sonnet-5",
+    "claude-sonnet-4-6",
+  ];
+  assert.equal(rankCandidates("anthropic", rawUnpromoted)[0], "claude-sonnet-5");
+});
+
+test("a future claude-haiku-5 is preferred over the pinned 4-5 id", () => {
+  // The exact id ranks first, then the generic /haiku/ — so a newer haiku is
+  // picked up without editing this map, as long as it sorts ahead in the
+  // catalog. Both are haiku-tier, so either outcome is correct on price; this
+  // pins that neither falls behind sonnet.
+  const ranked = rankCandidates("anthropic", ["claude-sonnet-5", "claude-haiku-5", "claude-haiku-4-5"]);
+  assert.deepEqual(ranked.slice(0, 2), ["claude-haiku-4-5", "claude-haiku-5"]);
+  assert.ok(ranked.indexOf("claude-sonnet-5") > 1, "sonnet must rank behind every haiku");
+});
+
+test("openai still leads with gpt-4o-mini — the cheaper entries are reasoning models (#569)", () => {
+  // Regression guard for the tempting-but-wrong optimisation: gpt-5-nano and
+  // gpt-5.4-mini are cheaper per token and hang the playground for 120 s.
+  const ranked = rankCandidates("openai", [
+    "gpt-5-nano",
+    "gpt-5.4-mini",
+    "gpt-4o",
+    "gpt-4o-mini",
+    "gpt-4.1-nano",
+  ]);
+  assert.equal(ranked[0], "gpt-4o-mini");
+  assert.ok(
+    ranked.indexOf("gpt-5-nano") > ranked.indexOf("gpt-4.1-nano"),
+    "no gpt-5.x model may outrank a gpt-4 family model",
+  );
+});
+
+test("google is unchanged — its entries are already the flash tier", () => {
+  const ranked = rankCandidates("google", [
+    "gemini-3-pro",
+    "gemini-flash-latest",
+    "gemini-2.5-flash",
+  ]);
+  assert.equal(ranked[0], "gemini-2.5-flash");
+});
+
+test("an unknown provider falls back to raw catalog order rather than dropping models", () => {
+  const catalog = ["b", "a", "c"];
+  assert.deepEqual(rankCandidates("groq", catalog), catalog);
 });
--- a/tests/helpers/provider-setup/collect-models.test.ts
+++ b/tests/helpers/provider-setup/collect-models.test.ts
@@ -18,6 +18,15 @@
 // so is the candidate ORDER of the Google/Anthropic catalogs — read out of that
 // run's `Shard 1/4` log. Two strings are synthesized because that run never
 // produced them; each says so where it is defined.
+//
+// That recorded order is a HISTORICAL snapshot, not the current ranking:
+// #1171 moved `CANDIDATE_PREFS.anthropic` to haiku-first, so
+// `rankCandidates("anthropic", ANTHROPIC_CATALOG)` now leads with
+// `claude-haiku-4-5` rather than `claude-sonnet-5`. That is deliberate — these
+// fixtures are INPUT to the fallback tests (which care about probe count and
+// error classification, not order), and freezing the run's real order is what
+// makes those tests reproduce the incident. The ordering assertions at the
+// bottom of this file call `rankCandidates` directly and are the live contract.
 import { test } from "node:test";
 import assert from "node:assert/strict";
 import * as fs from "fs";
--- a/tests/helpers/provider-setup/collect-models.ts
+++ b/tests/helpers/provider-setup/collect-models.ts
@@ -145,10 +145,33 @@ const CANDIDATE_PREFS: Record<string, RegExp[]> = {
     /^gemini-3\.5-flash$/,
     /^gemini-flash-latest$/,
   ],
-  anthropic: [/^claude-sonnet-5$/, /sonnet/, /haiku/],
+  // Haiku-first (#1171). The Anthropic entry is the one place in this map where
+  // the leading model is chosen for PRICE rather than only for agent
+  // compatibility, so the reasoning is worth stating: `claude-haiku-4-5` is
+  // $1/$5 per MTok against `claude-sonnet-5` at $2/$10 (introductory, through
+  // 2026-08-31) and $3/$15 after — 2x today, 3x from September. After #1185's
+  // weekday rotation, Anthropic runs twice a week but carries ~87% of the
+  // daily's remaining agentic spend, because it is ~13x the openai target's
+  // price. That makes this the dominant cost lever left on the lane.
+  //
+  // Sonnet stays at the tail, and that matters: without it a catalog with no
+  // haiku falls through to raw catalog order, which currently leads with
+  // `claude-opus-5` — the most expensive model Anthropic exposes here. The
+  // generic /haiku/ after the exact id is future-proofing for a later
+  // `claude-haiku-5`.
+  //
+  // NOT extended to the other two providers, deliberately. openai's cheaper
+  // catalog entries (`gpt-5-nano`, `gpt-5.4-mini`, …) are reasoning models,
+  // which hang the playground for 120 s (#569) — the gpt-4-family list below is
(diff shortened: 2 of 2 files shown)


Victor-w-Madeira:
## Review — 107-line diff, one finding, fixed in this PR

Reviewed adversarially against my own change. Three things checked, one real finding.

**Ordering is correct against the real catalog.** `/^claude-haiku-4-5$/` matches at position 0; the generic `/haiku/` then no-ops because the id is already collected. Verified against the live 13-model catalog: `claude-haiku-4-5 → claude-sonnet-5 → claude-sonnet-4-6 → claude-sonnet-4-5`.

**Nothing downstream depends on Anthropic settling to sonnet.** Grepped every `claude-sonnet-5` reference outside this diff. Two looked load-bearing and are not:

| Site | Verdict |
|---|---|
| `agent-tool-name-validation.spec.ts:48` | comment; the `INVALID_NAME_ERROR` regex is provider-agnostic (matches Google / OpenAI / Anthropic wording alike) |
| `agent-structured-output.spec.ts:125` | comment; historical note on #724, no assertion |
| `anthropic-provider.spec.ts:61` | resolves per **family** (`resolveClaudeModel(family)`), so promotion order does not reach it |
| `test-targets.test.ts` | synthetic fixtures in a different module — `CANDIDATE_PREFS` is not in its path |

**Finding (fixed, `c14d6a1`): the test file's header went stale on my own change.** It states the Google/Anthropic catalogs are in `rankCandidates` order from run 30351107916. Still true of the run; no longer true of the ranking, since haiku-first means `rankCandidates("anthropic", ANTHROPIC_CATALOG)` now leads with `claude-haiku-4-5`. Left the fixtures frozen — they are *input* to the fallback tests, which assert probe count and error classification rather than order, and freezing the real order is what makes them reproduce the incident — and marked the header as historical, pointing at the ordering assertions as the live contract. This is the same doc-rot class I flagged on #976 an hour ago; worth catching in my own diff rather than someone else's.

## Converting out of draft — and what that does not settle

The draft gate was mechanical, and it is being lifted deliberately. The reasoning that changes it: **the drained-key problem is independent of this PR and is not amplified by it.** `anthropic-provider.spec.ts` is a provider-contract spec that runs every day regardless — verified locally that it *fails* rather than skipping under a dry key (`1 passed, 1 failed, 1 did not run`), so the daily goes red and `@stable` auto-removal fires with or without this change. Holding this PR in draft protects nothing.

**The residual risk is unchanged and stays on the record:** `CANDIDATE_PREFS` encodes agent compatibility, and no run has shown that `claude-haiku-4-5` drives the agent suite. The unit tests pin ordering only, and say so in their own header.

What that means in practice, given the current state:

- **Merging today is inert.** Anthropic probes `inactive`, so the rotation falls back to openai on both its slots and haiku is never selected. Nothing changes until the balance is topped up.
- **The first daily after top-up is where it lands** — ~30 Anthropic agent tests against an unvalidated model. If haiku cannot drive them, that reads as a product regression and auto-removal strips `@stable` from the failures (#570's trap, at scale).

So the honest sequencing is: merge whenever, but **run the candidate protocol before the first funded daily**, not after. The command is in #1171. If that ordering slips, the cheaper alternative is to revert this one line rather than triage 30 red agent specs.

Issue #1187, opened 2026-07-31
Route the any-completion tier to Ollama — ~30 candidate tests, and resilience is the argument that does not wait on the spend measurement


Victor-w-Madeira:
Follow-up of #1169 (PR #1170).

> **Rewritten 2026-07-31.** The first version scoped this at **13 tests** and framed it as
> coverage-only. Both were wrong, and the correction changes the priority of this issue —
> see *What the first version got wrong* below. It also missed that this is one of **three
> overlapping answers to the same risk** (#976, #1185), which nobody had written down.

## The criterion is not "agentic vs non-agentic"

That was the first version's mistake. The question a spec answers is:

> **Does the deciding assertion read model *quality*, or does it read Langflow's *plumbing*?**

Read against what each spec actually asserts, a large part of the agent suite is the second
kind — the model only has to reply *something*:

| Spec | The assertion that decides it | Why the model's quality is irrelevant |
|---|---|---|
| `agent-context-id-continuity` / `-isolation` | `expect(retrieved).toContain(sentinel-N)` | asserts context-scoped retrieval; the reply's **content is never read** |
| `agent-max-iterations` | `toContainText(/model call limits exceeded/i)` + `/(1\/1)/` | that string is emitted by **Langflow**, not by the model |
| `agent-tool-name-validation` | `slugCell` toContainText, then the control expects `/hello/i` | the first assertion involves no model call at all |
| `agent-system-prompt` | `expect(reply).toContain(sentinel)` | the spec's own doc says it: *"even small models comply with 'always include this word'"* |
| `agent-max-tokens`, `agent-tool-error-handling`, `agent-input-sources`, `agent-component-regression` | caps, error surfaces, input wiring, node behaviour | Langflow contracts |

## The eligible set, counted

Counted by parsing every `test()` declaration and its tag array (not by grepping the file):

| Tier | `@stable` test declarations |
|---|---|
| **Plumbing, agentic** (the table above) | **17** — each parametrized per provider, so ×3 executions today, ×1 after #1185 |
| **Plumbing, non-agentic** (`loop-component-regression` 4, `language-model-regression` 4, `memory-history-regression` 3, `refresh-dropdown-list` 1, `settings-message-history` 1) | **13** — not parametrized |
[...]

PR #1212, opened 2026-07-31
feat(provider-setup): route the any-completion tier to a keyless local model, with one measured pilot (#1187)

Victor-w-Madeira:
Refs #1187 — deliberately NOT `Closes`.

This is the enabling half plus one pilot; #1187's *Done when* still has the per-
candidate 3/3 records, the declared wall-clock ceiling and the layer decision
(#976 / #1185) unmet. Auto-closing it on merge would retire the issue that owns
the adoption gate, so #1187 stays open and the gate is tracked there (see §7).

The enabling half of the issue plus **one** pilot spec, which is the scope agreed
on it: a keyless provider shape, tier routing, `setupOllama`, and the lane input
that can answer the adoption gate. Mass adoption stays out — it needs a measured
3/3 per candidate.

## 1. What this delivers

| Layer | Change |
|---|---|
| **Keyless provider** | `ProviderConfig` is a union on `credential: "api-key" \| "base-url"`; `ollama` added; `KeyedProvider` / `keyedProviders` **derived** from the map |
| **Routing** | `ANY_COMPLETION_PROVIDER=ollama` routes `tier: "any-completion"` to a local model, above the pins, for that tier only |
| **Setup** | `setup-ollama.ts`, dispatched by `providerSetupMap` unchanged; `load()` expects an empty `api_key` for a keyless provider |
| **Pilot** | `agent-system-prompt.spec.ts` declared `any-completion` (2 `@stable` tests) |
| **Lane** | `manual.yml` gains `any_completion_provider` + `retries` |

## 2. Why the routing outranks the model pins

#1185's weekday rotation emits `MODEL_TEST_ID` + `MODEL_TEST_PROVIDER` into
`$GITHUB_ENV`, which is global to the run and reaches every parametrized spec. If
the pin won, a routed spec would run the pinned hosted model and the routing could
never fire on the lane it exists for. The two answer different questions — the pin
says *which hosted provider this run uses*, the tier says *this spec needs none* —
so the narrower statement wins, scoped to that one tier, and an overridden pin is
printed rather than dropped silently.

Every rejection (unknown provider, an API-key provider, missing `OLLAMA_TEST_MODEL`)
returns a **reported** target instead of falling through to the hosted paths.
Falling through is the worse failure: the lane asked for keyless — possibly because
its keys are dead — and quietly spending a paid call is both the cost this avoids
and a green run that proves the wrong thing.

## 3. Two things the compiler could not catch

The union made the compiler name every consumer that reads `keyPlaceholder` /
`invalidKey`. Two call sites read only fields **both** shapes have, so they needed
finding by hand:

- **`collect-models.spec.ts`** compared `providers.json` against
  `Object.keys(providerConfigMap)`. The sweep is keyed-only, so with a keyless
  entry present that assertion **fails on every lane** — and this spec gates the
  daily, the PR run and every manual dispatch.
- **`model-provider-model-toggle`** and **`agent-model-connection-isolation`** pick
  "the first provider with env keys". `hasProviderEnvKeys("ollama")` is true as soon
  as `OLLAMA_BASE_URL` is exported, so on a box with no API key an unscoped `.find()`
  hands them a provider they were never validated against. Both scoped to the keyed
  set — selection identical to pre-#1187 (`--list` confirms).

## 4. Measured, not assumed

Against `llama3.2:1b` — the model the CI Ollama image bakes — on nightly
**1.12.0.dev10**, `PLAYWRIGHT_RETRIES=0`:

- **model alone**, called directly so the verdict is the model's and not the
  harness's: sentinel echoed **3/3**; neutral control emitted no stem.
- **spec end-to-end**: **5 runs × 2 tests, all green**, ~11–13 s per test, label
  `[ollama / llama3.2:1b]` — from an already-configured instance *and* from one
  where the helper configured the provider from scratch.
- **`provider-invalid-auth-error`** (target builder restructured here): **3/3
  passed**. It uses deliberately invalid keys, so no quota is spent.

Provider behaviour recorded while building it, on 1.12.0.dev10: saving the base URL
makes Langflow enumerate the **live** instance and auto-enable exactly what it
serves (one model pulled → `{"llama3.1:latest": true}`; a second pulled → both,
every toggle already `aria-checked="true"`). `OLLAMA_BASE_URL` persists as a
`Global` variable, not a `Credential` (`is_secret: false`).

## 5. Misconfiguration cannot read as a pass

| Misconfiguration | Behaviour |
|---|---|
| No `OLLAMA_TEST_MODEL` | 2 report-skips labelled `provider:ollama (routed)`; does **not** fall back to the 3 hosted providers |
| URL Langflow cannot reach | **fails in 5.9 s naming the rejection** (was: a 100 s dropdown timeout naming nothing) |
| Model the instance does not serve | skips, listing what **is** offered |

The middle row is why `setupOllama` reads the `validate-provider` **body**: that
endpoint answers **HTTP 200** with `{"valid": false, "error": …}` for a URL it could
not reach, so neither the status nor the UI says "rejected". Measured by pointing
Langflow at loopback, which its SSRF layer blocks from inside the container. It
throws `OLLAMA_PROVIDER_UNREACHABLE`, deliberately **not** `MODEL_NOT_AVAILABLE` —
the specs turn that prefix into a skip, and a lane that asked for a local model must
not lose coverage quietly when the wiring is wrong.

## 6. Scope notes

- **No scheduled lane changes behaviour.** The daily sets no routing variable, so
  the pilot still runs hosted there. Routing the daily is the follow-up, after 3/3.
- **Unrouted parametrization is untouched** — `--list` still yields the hosted
  targets for every spec, including the pilot.
- **The external-URL job of `manual.yml` FAILS** when routing is requested rather
  than ignoring it: that target is not on the runner's network and cannot reach a
  service container, so honouring it silently would run the routed specs on a hosted
  provider while the dispatch summary said `ollama`.
- **Before daily adoption**, note that a routed run's title becomes
  `[ollama / llama3.2:1b]`, and a test title is its identity in
  `spec-durations.json` and the `@stable` auto-removal path — history for those
  titles starts fresh.

## 7. The 3/3 gate — recorded, 3/3

`manual.yml` → `any_completion_provider: ollama`, `retries: 0`, grep the pilot.
Three consecutive dispatches, no retries, same runner class:

| # | Run | Result | Advisory HTTP |
|---|---|---|---|
| 1 | [30664476371](https://github.com/oriontech-me/langflow-e2e/actions/runs/30664476371) | **2 passed, 44.9 s** | `500 /api/v1/flows/` — *"An internal error occurred while deleting flows"* |
| 2 | [30672886825](https://github.com/oriontech-me/langflow-e2e/actions/runs/30672886825) | **2 passed, 66.0 s** | none |
| 3 | [30673235886](https://github.com/oriontech-me/langflow-e2e/actions/runs/30673235886) | **2 passed, 56.3 s** | `404 /api/v1/flows/{id}/events?since=` — the known editor-polling teardown 404 |

**The routing fired, and that is checked rather than assumed.** Playwright
reported `Running 2 tests` on all three; `--list` yields **2** targets routed
against **6** hosted (3 providers × 2 tests), so the collapsed matrix is the
routing's signature, not a lane that happened to be green.

**Duration — the input the wall-clock ceiling needs.** ~22 s, ~33 s, ~28 s per
test; mean ~28 s. That corrects the figure #1187 reasons from: it cites
`ollama-provider.spec.ts`'s **>100 s per call on the runner**, and this pilot is
well under it. The spread matters as much as the mean, though — 44.9 s to 66 s
between two identical dispatches is 47 % — so a ceiling derived from this must
carry that margin. The ceiling itself is still **not declared here**; it is a
daily-lane decision and belongs to the first adoption follow-up.

**Neither advisory error is this PR's, and that is measured rather than assumed.**
Run 3's is the teardown 404 that only appears while the editor is still polling
`/events` for a flow the test has deleted. Run 1's `500 /api/v1/flows/` looked
like a one-off until it was counted elsewhere: the **PR lane on this branch logs
it 6 times** across the 20 impacted agent specs, and **daily-stable run
30657439522 on `main` logs it 10 times**. So it is a pre-existing, recurring
advisory on the flow-collection endpoint — untracked by any open issue, and a
candidate for one — not something this change introduces. §8's "zero backend
errors" describes the local runs only.

Still open on #1187 after this PR: the per-candidate adoption of the remaining
~29 declarations, the declared wall-clock ceiling for the daily, and the layer
decision across #976 / #1185. That is why this says `Refs`, not `Closes`.

## 8. Validation

- `npm run typecheck` ✅ · `npm run lint` ✅ (0 errors) · `npm run test:units` ✅ 320
  (10 new for routing, 6 for the endpoint, 6 for the provider map) · `npm run
  test:scripts` ✅ 348 (5 new for the dispatch-input contract, below)
- QA-CHECKLIST guard ✅ · coverage guard ✅ · `regressions:check` ✅
- Zero `🚨 Backend Error` on every run above; flow count 27 before / 27 after — no
  orphans.
- `manual.yml` cannot be proven locally: its proof is the first dispatch (see §7).

## 9. Review fixes (after the first dispatch)

- **`retries` was honoured on `e2e-docker` only.** A dispatch against an external
  URL took the config default with nothing in the log to say the input had been
  dropped — the silent half of the failure the routing rejection already handles
  out loud. Now set on both run jobs, and validated in `detect-target`
  (`Number("O")` is `NaN`, which Playwright resolves to its own default, so a typo
  produced the same silent amplification). `scripts/manual-dispatch-inputs.test.mjs`
  pins the rule: every dispatch input is HONOURED by both run jobs or REFUSED by
  the one that cannot serve it, never ignored.
- **An unreachable Ollama could still skip.** The `validate-provider` guard reads
  the save response, so it only covers a URL saved on that run; an instance
  configured earlier short-circuits the step and reaches the dropdown with an empty
  list, which `MODEL_NOT_AVAILABLE` turned into a `test.skip`. An empty Ollama list
  has no benign reading — Langflow enumerates the live instance on save — so it now
  throws `OLLAMA_PROVIDER_UNREACHABLE` and fails. A non-empty list missing the tag
  still skips. Not reachable on CI (fresh container per job); the persisted local
  Langflow is what hits it.
- **Two comments corrected** where they claimed more than the code delivers: the
  routed target's `skipReason` lookup cannot fire today (kept as a forward seam,
  now labelled), and the settle guard's proof for a keyless provider rests on the
  model axis, not the credential one — `credentialOf()` returns `""` for an absent
  field too.

## 10. Dependencies

- `#1184` (the tier this routes) — merged. `#1183` gates the *spend* argument only;
  the resilience argument does not wait on it.
- A reachable Ollama instance for the routed path; **no provider key at all**. The
  hosted path is unchanged and still needs its keys.


Victor-w-Madeira:
## Review — findings, fixes, and the gate that was missing

Reviewed the full diff, swept every consumer of the changed provider map, reproduced
all four gates locally, and ran the #1187 adoption gate to completion. Six findings,
all resolved below. Three commits added on top of the original four.

### 1. `Closes #1187` would have retired the issue that owns the gate

§7 itself listed the 3/3 gate and the wall-clock ceiling as unmet, and #1187's
*Done when* had three boxes open (per-candidate 3/3 records, the declared ceiling,
the layer decision across #976 / #1185). Merging would have auto-closed it.

→ Changed to **`Refs #1187`**, with the reason stated in the PR body. #1187 stays
open and now carries the recorded result as a
[comment](https://github.com/oriontech-me/langflow-e2e/issues/1187#issuecomment-5148370513).

### 2. `retries` was honoured on `e2e-docker` only — commit 75925d5

`PLAYWRIGHT_RETRIES` was set at the job level of the Docker job and nowhere else, so
a dispatch against an external URL silently took the config default (2 in CI). That
is the invisible half of the failure this PR already handles out loud for the other
new input: `e2e-url` **fails** when `any_completion_provider=ollama` is requested,
because it cannot reach the service container. `retries` has no such obstacle — it
is target-kind agnostic — and a run that retried its way to green is precisely the
evidence a "passes 3/3, no retries" gate exists to reject.

Also validated: `Number("O")` is `NaN`, which Playwright resolves to its own default,
so a typo produced the same silent amplification. The check lives in `detect-target`
(the only job both run jobs depend on) and reads the value from `env:` rather than
interpolating it into the script body.

`scripts/manual-dispatch-inputs.test.mjs` (5 tests) pins the rule that separates the
two cases: **every dispatch input is HONOURED by both run jobs or REFUSED by the one
that cannot serve it, never ignored.** Verified it bites by removing the new `env:`
block and watching it go red.

### 3. An unreachable Ollama could still degrade to a silent skip — commit 91cc326

The `validate-provider` guard reads the save response, so it only covers a base URL
saved on *that* run. An instance configured earlier short-circuits the step
(`alreadyConfigured`) and reaches the model dropdown with an empty list and no
verdict read anywhere — and `MODEL_NOT_AVAILABLE` turns that into a `test.skip`. A
lane that asked for a local model then loses its coverage quietly when its wiring is
wrong, which is #976's "24 specs skipped silently" reproduced on the mechanism built
to prevent it.

An empty Ollama list has no benign reading: Langflow enumerates the live instance on
save, so zero options means it reached nothing. That now throws
`OLLAMA_PROVIDER_UNREACHABLE` and **fails**, attributed. A non-empty list missing the
requested tag is the genuine "this instance does not serve it" case and still skips.
Not reachable on CI (fresh container per job) — the persisted local Langflow is what
hits it.

### 4. Two comments claimed more than the code delivers — commit 4b3c115

- `resolveRoutedTarget`'s `skipReasons.get(provider)` **cannot fire today**:
  `providerSkipReasons()` is built only from `providers.json` inactive records, and
  this PR removes keyless providers from the sweep on purpose. Kept as a forward
  seam and now labelled as one — the unit test exercising it injects a map the
  resolver never emits, which is the fixture trap #1216 just paid for.
- `SimpleAgentTemplatePage.load()` said the settle guard "still proves both axes"
  for a keyless provider. `credentialOf()` returns `""` for an absent field too, and
  on the lane #1187 exists for — every hosted key dead, nothing prefilling `api_key`
  at mount — `""` is also the pre-selection state. The **model** axis carries the
  proof, and the routed path always pins one.

### 5. The 3/3 gate had not run — now recorded

| # | Run | Result |
|---|---|---|
| 1 | [30664476371](https://github.com/oriontech-me/langflow-e2e/actions/runs/30664476371) | 2 passed, 44.9 s |
| 2 | [30672886825](https://github.com/oriontech-me/langflow-e2e/actions/runs/30672886825) | 2 passed, 66.0 s |
| 3 | [30673235886](https://github.com/oriontech-me/langflow-e2e/actions/runs/30673235886) | 2 passed, 56.3 s |

`any_completion_provider=ollama`, `retries=0`, grep the pilot, nightly `latest`.

**The routing is verified, not inferred from green.** All three reported
`Running 2 tests`; `--list` yields 2 targets routed against 6 hosted (3 providers ×
2 tests), so the collapsed matrix is the routing's own signature.

**Duration corrects an input #1187 reasons from.** ~22 s / ~33 s / ~28 s per test
against the **>100 s per call** this issue cites from `ollama-provider.spec.ts`. The
spread is the caveat — 47 % between two identical dispatches — and this is one spec
on an otherwise idle runner, so the ceiling that bounds the daily still has to be
measured *on the daily*. Deliberately not declared here.

### 6. The advisory 500 — I had this wrong first

I initially recorded run 1's `500 /api/v1/flows/` as a one-off worth watching. Counted
elsewhere before calling the gate clean: the **PR lane on this branch logs it 6 times**
across the 20 impacted agent specs, and **daily-stable run 30657439522 on `main` logs
it 10 times**. It is a pre-existing, recurring advisory on the flow-collection
endpoint, untracked by any open issue — a candidate for one, and not this PR's.
Run 3's other advisory is the known editor-polling teardown 404.

---

### Coverage this actually has

- **Routed path:** 3/3 above, retries off.
- **Hosted path:** the PR lane ran 20 of 43 impacted specs (cap announced), the pilot
  among them — **35 passed / 1 skipped**. That is what supports "no scheduled lane
  changes behaviour", rather than the claim alone.
- **Local:** `typecheck` ✅ · `lint` 0 errors ✅ · `test:units` 320 ✅ ·
  `test:scripts` **348** ✅ (343 + 5 new).

### Not proven, and why it is not a blocker

- **The new `e2e-url` `env:` block has never executed.** It only runs on a dispatch
  against an external URL, and there is no external Langflow to point at. Covered
  structurally by the new guard, unproven at runtime — it is a job-level `env` block.
- **The negative path of the retries validation never ran** either; all three
  dispatches passed a valid `0`.

### Context for whoever merges next

- #1216 is still open. Once it lands, a PR touching `provider-setup/` like this one
  will have the provider-dependent specs *pulled in by import* excluded from the PR
  lane — the same 20 that ran green here. It changes what the next such PR can prove,
  not anything in this one.
- `select-pr-model-target` warned on this run that the Anthropic key is **drained
  again** ("credit balance is too low"), which is the scenario #1187 exists for.

Issue #1211, opened 2026-07-31
Harden the token monitor's pricing: dated model ids, dated price bands, and a written blind-spot list


rafaelgiln:
Follow-up of #1197 (PR #1208). Three hardening items for the token monitor's pricing layer, ported from a written specification of an equivalent subsystem in another project of ours (Yoke Agent's `TOKEN_COST_TRACKING.md`), which documents these as known weaknesses of its own implementation. Two of the three are live defects here, with measured exposure.

## 1. Exact-match-only pricing silently zeroes 31 of 90 catalog models

`usdFor` in `scripts/lib/token-cost.mjs` looks the model up by exact key:

```js
const price = prices?.[model];
if (!price) return null;
```

Measured against the catalog `collect-models` actually rotates through (`tests/helpers/provider-setup/data/models.json`): **31 of 90 model ids carry a dated, `-preview` or `-latest` suffix** — `claude-opus-4-20250514`, `claude-sonnet-4-20250514`, `gpt-4o-mini-search-preview`, `gemini-flash-lite-latest`, and so on. Any rotation onto one of those produces `usd_estimated: null`, puts the id in `unpriced_models`, and turns every dollar figure in that run's line into a floor. The token counts stay correct — the cost does not exist.

This already happened once in the first real run: `gemini-3.5-flash` was the largest consumer of run 30647253368 and had no price entry, so the headline read `$0.05 (a FLOOR …)` with `n/a` on the biggest row.

**Fix:** two-phase lookup — exact key first, then a substring match in both directions (a dated id contains its family key; a short alias is contained by a longer key).

**And the trap that comes with it, which the source document flags as a real inaccuracy in its own code:** with substring matching, `gpt-4o-mini-search-preview` matches **both** `gpt-4o` and `gpt-4o-mini`. If iteration order decides the winner, that id gets priced at `gpt-4o`'s $2.50/$10 instead of `gpt-4o-mini`'s $0.15/$0.60 — a **16× overstatement**, silently. The same hazard exists for `gpt-4`/`gpt-4.1`, `claude-sonnet-4`/`claude-sonnet-4-6`, and `gemini-2.5-flash`/`gemini-2.5-flash-lite`.

So the fix is specifically: **candidate keys sorted longest-first**, never declaration order, with a test pinning the `gpt-4o-mini-*` case.

## 2. The price table has no notion of when a price was true

[...]

Issue #1197, opened 2026-07-31
Design review: token consumption + cost monitor for the LLM APIs a run uses (stacked on #1180)


rafaelgiln:
## What this asks for

**Review and approve a design** for measuring the LLM token consumption of a run — per run, per model and per spec — plus its cost estimate. No production code ships against this issue until the design is approved.

It is **stacked on #1180**: the per-spec half of this lands as a sibling fact table in the analytics model that issue proposes. The producer half in this repo does not depend on it.

## Why now

Nothing in the suite measures what it spends. Four questions have no answer today:

1. **What does one daily run cost in dollars?** Nobody knows, per run or per month.
2. **Which specs burn the tokens?** Agent specs run multi-turn loops; none of them is budgeted.
3. **Did consumption regress?** A Langflow change that duplicates context, or an agent that starts looping, is invisible — the suite stays green and the bill moves.
4. **Which model rotation made it worse?** `collect-models` rotates the model per provider; the cost impact of a rotation is never measured.

## Phase 0 spikes — already run, evidence below

Run against a clean `langflowai/langflow-nightly:latest` (`1.12.0.dev10`) container with tracing enabled. Verdict: **go**, and cheaper than the design assumed.

| Spike | Question | Result |
|---|---|---|
| S1 | Does `tokenUsage` populate per provider? | **Yes, all three.** `openai/gpt-4o-mini` 40/48/**88**, `anthropic/claude-sonnet-4-6` 40/7/**47**, `google/gemini-flash-latest` 30/1821/**1851** (prompt/completion/total). `modelName` populated on the provider span in every case. Provider labels accepted by the unified `ModelInput` tweak: `OpenAI`, `Anthropic`, `Google Generative AI` |
| S2 | Does `GET /api/v1/monitor/traces` answer without `flow_id`? | **Yes** — 200, and with traces present it returned 2 traces across 2 distinct flows. The poller needs **one request per tick**, not one per live flow |
| S3 | What does a tick cost? | Flow list 98 ms; per-flow trace query ~13 ms. The naive per-flow loop is 28 requests / 463 ms against 27 live flows. S2 collapses that to 1 request + one detail fetch per new trace |
[...]

Issue #886, opened 2026-07-22
resolveGeminiModel selects the retired gemini-2.5-flash → google-provider + language-model-regression fail durably


rafaelgiln:
Extracted from the #773 investigation as the one **durable, non-load** failure in that cluster.

## Symptom

`google-provider.spec.ts:155` ("configured Google selects a Gemini model in the Agent and executes the flow") and `language-model-regression.spec.ts:130` ("language model must respond with Google provider") fail durably — the agent/model never returns, so the awaited reply times out (`expect(aiMessage).toBeVisible()` 30s). Reproduced **in isolation on nightly 1.11.0.dev50 against a healthy backend** (not saturation).

Backend log at failure:
```
ComponentBuildError: Error calling model 'gemini-2.5-flash' (Not Found): 404
"models/gemini-2.5-flash is no longer available to new users."
```

## Root cause (test-side)

`tests/helpers/provider-setup/resolve-gemini-model.ts` preferred `gemini-2.5-flash` as its **first** preference. Google retired that id for new users, so it 404s at call time. `models.json` lists it alongside the live models (`gemini-flash-latest`, `gemini-3.x-flash`), and the resolver picked the dead one.

Both specs share this resolver, so both fail on the same cause.

## Fix

Exclude the exact retired id from selection and prefer the floating `gemini-flash-latest` alias first, then live dated flash models — so the resolver tracks Google's lifecycle instead of pinning a dated id (same lesson as the RAG answer model in #880).

## Validation (nightly 1.11.0.dev50, `--retries=0 --workers=1`)

- `google-provider.spec.ts` → 2 passed
- `language-model-regression.spec.ts` → 4 passed
- resolver now returns `gemini-flash-latest`
- typecheck ✅ · eslint ✅

## Done when

- [ ] Resolver fix merged; both specs pass reliably.
- [ ] `@stable` retained (never removed — this was diagnosed and fixed same-cycle).

PR #1218, opened 2026-07-31
docs(observability): design the token-usage analytics tables and the dashboard's data path (#1217)

### LINKED ISSUE #1217: Design the token-usage analytics tables and the dashboard's data path — §5 of the #1197 design has gone stale in six ways
Follow-up of #1197. Phase 1 landed the producer: the recorder, the pricing layer, per-spec attribution and `reports/token-history.jsonl`. Phase 2 — the half that makes those numbers queryable next to the failure data, and puts a cost view in the QA Platform's E2E dashboard — is designed only at the sketch level in §5 of `ISSUE-1197-TOKEN-MONITOR-DESIGN.md`, and that sketch has since gone stale in six specific ways. This issue is the design pass that corrects them before any migration is written.

## Why a design pass and not just the migration

Checked against both repositories at `langflow-e2e@ed6eb6e` and `quality-platform@e42cc36`:

| Piece | State |
|---|---|
| `e2e_test_token_usage`, `e2e_model_prices` | No migration exists. Neither does `e2e_test_results` (#1180 is still in design review on PR #1181) |
| FK anchor | **Exists** — `e2e_automation_runs.run_id` is `TEXT NOT NULL UNIQUE` (`20260610210000_nightly_runs.sql`), already read by `TrendView.tsx` and `RegressionsView.tsx` |
| Producer | Carries no token data. `build-run-payload.mjs` has no token block |
| Data | `reports/token-history.jsonl` does not exist on `main` — the history steps are `schedule`-gated, so the series starts with the next scheduled daily |

The last row is the honest constraint on the UI half: five weekdays of scheduled dailies are the first baseline, and the anomaly detector stays silent until then by construction. The database and producer work can proceed in parallel; a screen cannot show what has not been recorded.

## Six corrections to §5

**1. The fact table's grain is not derivable from what the summarizer emits today.** §5 declares the grain `(run × test × model)`. `aggregate()` in `scripts/lib/token-cost.mjs` returns `byModel` (run × model, no spec) and `bySpec` (run × spec, no model split). Neither is the declared grain, and the two cannot be joined after the fact — a spec that called two models contributes one `bySpec` row with a summed total. The per-(spec, model) rollup has to be emitted by `aggregate()`; the inputs are already in hand (`probe.models[]` × the attribution for that `trace_id`), so this is an added accumulator, not new collection.

**2. `provider` is never produced.** The §4.4 example line shows `by_model[].provider: "openai"`, and §5's table declares a `provider TEXT` column. The implementation emits no provider anywhere — not in the probe, not in `byModel`, not on the history line. Either the producer derives it from the model id (an id-prefix rule, which is a guess that rots as ids rotate — #886/#964) or the column goes. Decide it here rather than shipping a column that is always NULL.

**3. `e2e_model_prices` must be banded, not one row per model.** #1211 made prices dated: `claude-sonnet-5` carries two bands (`$2/$10` through 2026-08-31, `$3/$15` from 2026-09-01), and `--summarize` selects the band effective on the **run's own** date, never the newest. A flat price dimension would silently reprice every historical run at today's rate — the exact failure #1211 closed on the JSONL side. Key is `(model, since)`; the join picks the greatest `since <= run_date`.

**4. The model-id resolution rule must not be reimplemented in SQL.** `resolveBands()` is exact-key first, then longest matching substring in either direction, gated by `isAllowedSuffix()` so a tier id (`-lite`, `-mini`, `-nano`) never inherits its non-tier sibling's rate. That gate exists because the naive version priced `gemini-2.5-flash-lite` as Flash — a 3–6× overstatement, found in #1211's second review round. Porting that logic to plpgsql puts two implementations of the repo's most error-prone rule in two languages. **Proposal:** the producer sends the already-resolved `price_key` per row; the database joins on it exactly and only applies band-by-date. Invariant 3 (no dollars stored, no human-editable column) is preserved — the price dimension is still the source of the rate.

**5. In `daily-stable.yml` the token summary runs *after* the payload is POSTed.** `Build run payload` and `POST run to QA Platform` sit at lines ~895/~914; `Summarize token consumption` at ~953. As wired, the payload cannot carry a token block at all. Two things follow:

- The summarize step moves ahead of `Build run payload`. Its inputs allow it — the token artifacts are downloaded at ~733 and `runguard` runs at ~806, both already upstream of the payload build.
- It needs a machine-readable output file that is **not** gated by `TOKENS_SUPPRESS_HISTORY`. That knob deliberately suppresses the history line on a manual dispatch (#1183), but the DB ingest records **every** run, scheduled or manual, by design. Sourcing the payload's token block from the history file would therefore drop it on exactly the runs the platform does record.

**6. The dependency on #1180 is weaker than §5 states.** §5 routes the token ingest through the atomic RPC #1180 §4.2 introduces. That coupling is optional: the FK anchors on `e2e_automation_runs(run_id)`, which exists today. Only the cross-joins — cost of the spec that failed, cost against duration — need `e2e_test_results`. The token fact table can therefore ship, and be queried on its own, before #1180 is approved. Whether it *should* is a scheduling call this design should state explicitly instead of leaving implied.

## Also worth stating plainly in the design

Per-spec coverage is one spec today (`agent-max-tokens.spec.ts` is the only call site passing `attribution` to `cleanup()`; three other specs import the tracker and could each gain one line). So `by_spec` will be one attributed row plus the unattributed bucket for a while, and any "cost per spec" screen has to read as *"the share we can attribute"*, not as a breakdown of the run. That is #1197's invariant 2 applied to a UI: never silently.

## Done when

- [ ] A design document at the repository root, reviewed like #1197 and #1180, that resolves all six corrections and states the #1180 scheduling call.
- [ ] The document specifies the final DDL for `e2e_test_token_usage` and `e2e_model_prices`, the ingest path, the price-join RPC, and the payload block the producer sends.
- [ ] It names what the screen must **not** imply — the attributed share, the unpriced-model floor, and the missing baseline before five scheduled dailies.
- [ ] No migration, no producer change and no UI work ships against this issue; those are the follow-ups it unblocks.

## Explicitly out of scope

The screen itself. The E2E dashboard has three views today (`src/components/e2e/dashboard/{TrendView,RunSummaryView,RegressionsView}.tsx`); a cost view is a fourth, and it is a later cycle for the same reason #1180 §6 defers its own UI — data and RPCs first.

rafaelgiln:
Closes #1217.

Design only — **no migration, no producer change, no UI ships with this PR.** Same posture as #1198 (the #1197 design) and #1181 (the #1180 design): the document is the deliverable, and it exists to be argued with before anyone writes DDL.

## Why

#1197 phase 1 shipped the producer: the recorder, the pricing layer, per-spec attribution, `reports/token-history.jsonl`. Phase 2 — the half that makes those numbers queryable next to the failure data so the QA Platform's E2E dashboard can carry a cost view — was sketched in one short section (§5 of `ISSUE-1197-TOKEN-MONITOR-DESIGN.md`), written before #1211 dated the price table and before anyone read the platform's migrations.

Checked against `langflow-e2e@ed6eb6e` and `quality-platform@e42cc36`, eight of that sketch's assumptions do not hold. Seven are defects that would have shipped.

## The two the issue did not know about

#1217 was opened with six. Two more surfaced while writing the document, both from reading the implementation instead of the design:

**§2.2 — the sum of fact rows can never equal the run total.** `aggregate()` deliberately keeps two totals: per-model sums come from `modelName`-bearing spans, while the run total comes from each trace's own `totalTokens`, because Langflow reports the same usage twice per call (#1197 §2.1). A fact table at the declared grain can only hold the span-derived figure, so §5's claim that *"a query that forgets to filter still sums to the run total"* is false whenever a mismatch exists. Resolved with four nullable columns on the run row, not a third table.

**§2.7 — a live defect in phase 1's own output.** The one attributed call site passes `testInfo.title` (the leaf), and that spec is model-parameterized:

```ts
test.describe(`Agent max_tokens [${label}]`, () => {
  test("max_tokens=50 caps the response's output tokens", …)
```

`aggregate()` keys its spec rollup on `${file}::${test}`, so whenever more than one provider's targets run, **two providers' consumption merges into a single `bySpec` row today**. #1185's weekday rotation masks it by usually resolving one provider per day. The §4.2 claim in the #1197 document that the sidecar already writes #1180's `title_path` chain is simply not what the call site does.

## The other six, in one line each

| # | Correction |
|---|---|
| §2.1 | The declared `(run × test × model)` grain cannot be filled — `aggregate()` emits run × model and run × spec, and the two cannot be rejoined. Needs a `bySpecModel` accumulator |
| §2.3 | `provider` is declared as a column and produced nowhere. It moves to the price dimension, where it is curated in a PR next to the rate, instead of guessed from an id prefix that rots as ids rotate (#886/#964) |
| §2.4 | A flat price dimension would reprice every historical run on the next correction — the failure #1211 closed on the JSONL side, reintroduced in SQL. Key `(price_key, since)`, band chosen by the run's own date |
| §2.5 | Porting `resolveBands()`/`isAllowedSuffix()` to plpgsql would put the rule behind #1211's 3–6× mispricing in two languages, only one of them tested. The producer sends the resolved `price_key`; the database joins on it exactly |
| §2.6 | In `daily-stable.yml` the token summary runs at ~953, **after** the POST at ~914 — as wired the payload cannot carry a token block at all. The step moves ahead of the payload build and writes a summary file not gated by `TOKENS_SUPPRESS_HISTORY`, because the DB records every run while the history file deliberately does not |
| §2.8 | Not a defect — a dependency that does not exist. The FK anchors on `e2e_automation_runs(run_id)`, which already exists |

## Adjusted after #1181 merged mid-review

#1180's document grew from 379 to 565 lines in its review and three sections renumbered, so every cross-reference here was re-verified against the merged version. Two of its additions changed this design rather than just its citations:

- **`is_valid_run`** (#1180 §3.5) — a generated column on `e2e_automation_runs` excluding zero-test runs by construction. §5 here now inherits it instead of restating the per-RPC predicate, which #1180 argues against on two grounds: it is NULL-unsafe, and it wrongly discards an all-skipped run. Noted as a soft ordering constraint, with landing the column here as the recommended trade.
- **The same section is precedent for §3.3.** Adding token columns to `e2e_automation_runs` rather than creating a third table is a move the approved design takes itself, for the same reason: a property every reader needs belongs where every reader already looks.
- **`nightly.yml` is dormant** (#1180 §2.3 — cron commented out, last run 2026-03-19), which is a stronger reason to leave it unwired than the one the #1197 document gave.
- **§2.8 loses its timing argument and keeps its conclusion.** #1180's design is approved, but no migration exists — an approved design is not a schema.

## What the document commits to

Final DDL for `e2e_test_token_usage` (grain run × test × model, `UNIQUE NULLS NOT DISTINCT` so a re-POST cannot duplicate the unattributed bucket) and the banded `e2e_model_prices`; four nullable run-level columns where NULL means *not measured* and `0` means *measured, nothing recorded*; an atomic `e2e_ingest_run_tokens` that never creates a run; four query RPCs including `e2e_token_price_gaps()`; and the producer changes with their tests.

It also names what a screen must **not** imply — #1197's invariant 2 applied to a UI:

- per-spec attribution is one spec today, so `attributed_share` is single-digit percent and ships *with* the rows so a caller cannot present a partial breakdown as a whole;
- an unpriced model makes every dollar figure a floor, surfaced as `is_floor`;
- `reports/token-history.jsonl` has **no committed line yet** (the history steps are `schedule`-gated), so there is no baseline until five scheduled dailies — and zero traces is not zero spend.

## CI expectation

Documentation-only diff: the import graph honestly selects zero specs, and `scripts/ci-change-coverage.mjs` returns `none` because no workflow, action or script changed. `skipping` on the E2E lane is the correct verdict here, not a gap.

## Review focus

The four open questions in §10, in particular: are four token columns on `e2e_automation_runs` the right call versus a run-level table, and is broadening per-spec attribution beyond one call site worth doing before the screen exists.

### DIFF
--- a/ISSUE-1180-ANALYTICS-DB-DESIGN.md
+++ b/ISSUE-1180-ANALYTICS-DB-DESIGN.md
@@ -3,6 +3,11 @@
 **Status:** proposed — revision 2, after review on [#1181](https://github.com/oriontech-me/langflow-e2e/pull/1181). No code ships against this document until the model is approved.
 **Issue:** [#1180](https://github.com/oriontech-me/langflow-e2e/issues/1180)
 **Scope:** the QA Platform database (`quality-platform`) plus the payload producers in this repo.
+**Sibling fact:** `ISSUE-1217-TOKEN-ANALYTICS-DB-DESIGN.md` designs `e2e_test_token_usage` — one row per
+(run × test × **model**) — against this model. It generates the same `test_key` as §3.1 so the two facts
+join on `(run_id, test_key)` and the token series inherits §3.4's rename bridge; it also reuses §3.5's
+`is_valid_run` and adds four nullable token columns to `e2e_automation_runs`. Any change to §3.1's
+identity expression, to §3.4, or to §3.5 has to be made in both places.
 **Revision log:** §11.
 
 ---
--- a/ISSUE-1217-TOKEN-ANALYTICS-DB-DESIGN.md
+++ b/ISSUE-1217-TOKEN-ANALYTICS-DB-DESIGN.md
@@ -1,0 +1,460 @@
+# ISSUE-1217 — Token-usage analytics tables and the dashboard's data path
+
+> Status: **design review**. Nothing executable ships with this document — no migration, no producer
+> change, no screen.
+> Companion issue: #1217. Corrects §5 of `ISSUE-1197-TOKEN-MONITOR-DESIGN.md`, which was written
+> before #1211 and before the platform's own state was checked.
+> Verified against `langflow-e2e@ed6eb6e` and `quality-platform@e42cc36`.
+
+---
+
+## 1. What this is for
+
+#1197 phase 1 answers its four questions from a committed JSONL file. This document is the other
+half: the same numbers as rows, queryable next to the failure data, so the QA Platform's E2E
+dashboard can carry a cost view.
+
+It is a **correction pass**, not a fresh design. §5 of the #1197 document sketched two tables in one
+short section, before #1211 dated the price table and before anyone read the platform's migrations.
+Eight of its assumptions turn out to be wrong or stale. Seven are defects that would have shipped as
+a grain nothing can fill, sums that do not add up, a column that is always NULL, a history repriced
+on every correction, the repo's most error-prone rule written twice in two languages, a payload that
+cannot carry the data at all, and two providers' consumption merged into one row. The eighth is a
+dependency that does not exist.
+
+The issue that opened this pass named six. The two found while writing it — §2.2 and §2.7 — are
+both consequences of reading the implementation rather than the design, and §2.7 is a live defect in
+phase 1's own output, not just in the sketch.
+
+### 1.1 State on the ground
+
+| Piece | State |
+|---|---|
+| `e2e_test_token_usage`, `e2e_model_prices` | No migration exists |
+| `e2e_test_results` (#1180) | Design **approved and merged** (PR #1181). No migration exists yet |
+| `e2e_automation_runs` | **Exists.** `run_id TEXT NOT NULL UNIQUE` (`20260610210000_nightly_runs.sql`), read today by `TrendView.tsx` and `RegressionsView.tsx` |
+| Token data in the run payload | None. `build-run-payload.mjs` has no token block |
+| `reports/token-history.jsonl` | **Does not exist on `main`.** The history steps are `schedule`-gated; the series starts with the next scheduled daily |
+
+The last row bounds the UI half and nothing else: five scheduled dailies are the first anomaly
+baseline, and the detector is silent until then by construction. Database and producer work do not
+wait on it.
+
+---
+
+## 2. The eight corrections
+
+### 2.1 The declared grain cannot be filled from what the summarizer emits
+
+§5 declares `(run × test × model)`. `aggregate()` returns `byModel` (run × model, no spec) and
+`bySpec` (run × spec, no model split). A spec that called two models contributes **one** `bySpec` row
+with a summed total, so the two outputs cannot be joined back into the grain after the fact.
+
+**Decision.** `aggregate()` gains a third accumulator, `bySpecModel`, keyed
+`${file}::${test}::${model}`. The inputs are already in hand — `probe.models[]` crossed with the
+attribution for that `trace_id` — so this is an added rollup, not new collection. The unattributed
+bucket is emitted the same way: one row per model with no spec, not one lump.
+
+### 2.2 Sum of fact rows will not equal the run's authoritative total
+
+`aggregate()` deliberately keeps two totals: per-model sums come from `modelName`-bearing spans,
+while the run total comes from each trace's own `totalTokens`, because §2.1 of the #1197 document
+measured that Langflow reports the same usage twice per call. When the two disagree the summarizer
+reports both and picks no winner (#1197 §7).
+
+A fact table at the declared grain can only hold the **span-derived** figure. §5's claim that "a
+query that forgets to filter still sums to the run total" is therefore false whenever a mismatch
+exists.
+
+**Decision.** The fact table is span-derived, and the authoritative run total lives on the run row
+as four nullable columns (§3.3). A screen can then show the discrepancy instead of a reconciled
+number, and "sum of rows ≠ run total" stays queryable. No third table: the run row is created by the
+same POST, and both `TrendView.tsx` and `RunSummaryView.tsx` already select from it. Widening that
+table is a move the approved #1180 design takes itself — its §3.5 adds a generated `is_valid_run`
+column to the same table, for the same reason: a property every reader needs belongs where every
+reader already looks.
+
+### 2.3 `provider` is never produced
+
+§4.4's example line shows `by_model[].provider: "openai"` and §5 declares a `provider TEXT` column.
+The implementation emits no provider anywhere — not on the probe, not in `byModel`, not on the
+history line. Deriving one from the model id is a prefix guess that rots exactly as ids rotate
+(#886, #964).
+
+**Decision.** No `provider` on the fact table. It moves to the **price dimension**, where it is a
+curated fact reviewed in a PR alongside the rate it belongs to, and provider grouping is recovered
+through the join. A model with no price band therefore has no provider either — which is honest:
+an id we cannot price is an id we have not classified.
+
+### 2.4 The price dimension must be banded
+
+#1211 made prices dated. `claude-sonnet-5` carries two bands ($2/$10 through 2026-08-31, $3/$15 from
+2026-09-01) and `--summarize` selects the band effective on the **run's own** date, never the newest.
+A flat dimension would reprice every historical run at today's rate on the next correction — the
+failure #1211 closed on the JSONL side, reintroduced in SQL.
+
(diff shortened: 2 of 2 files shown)


rafaelgiln:
## Amended — the review question found a ninth correction, and it is the load-bearing one

`23dba3a`. The question was whether the merged #1180 document and this one can be correlated. Two answers; the second was a defect.

**The reference was one-way.** This design cites #1180 fifteen times; the merged #1180 document does not contain the word "token". A reader arriving at the data model had no way to learn a sibling fact was designed against it. Fixed with a `**Sibling fact:**` line in its header naming the three sections both designs share (§3.1's identity, §3.4's alias bridge, §3.5's `is_valid_run`) and stating that a change to any of them has to be made in both places.

**The join did not work.** #1180 §3.1 keys `e2e_test_results` on a generated `test_key`:

```sql
test_key TEXT GENERATED ALWAYS AS (
  md5(public.e2e_normalize_spec_path(spec_path) || '::' || title_path)
) STORED,
UNIQUE (run_id, test_key)
```

and its RPCs resolve renames through `e2e_test_aliases` + `COALESCE(new_test_key, test_key)`. The sketch gave `e2e_test_token_usage` only `spec_path` and `title_path`, so the join needed the digest recomputed on every query — with each RPC free to compute it slightly differently, which is the exact objection #1180 answers by generating the column in the database — and **the token series never reached the rename bridge.**

That second half is the damaging one. #1180 §3.4 records that a **file move breaks the key just as a rename does, and that moves are more frequent in this repository than renames.** One folder reorganisation would have reset the cost series while leaving the failure series intact: two histories disagreeing about when a test began, with nothing reporting the disagreement.

The fact table now generates the same `test_key` from the same expression, and the unique key becomes `(run_id, test_key, model)` — `test_key` already *is* the normalised pair, so keying on both would let a row differing only by a `tests/` prefix insert twice under one identity. The column needs only `e2e_normalize_spec_path()`, which exists and is `IMMUTABLE`, so it does **not** reintroduce the #1180 dependency §2.9 removes.

### Also in this amendment

- **New §3.5 — joining the two facts.** One identity, the inherited rename bridge, and the `title_path_exact` trap: #1180's backfill *reconstructs* `title_path`, so a naive join pairs an exact token row against a guess. Every joining RPC filters that flag and reports its drops.
- **New §5.1** defers the joining RPCs to a follow-up gated on #1180's migrations. No schema change — the shared identity ships here.
- **A gap neither document named: per-attempt billing.** The poller records a trace per attempt, so a flaky test's tokens already arrive summed across attempts with no attempt count anywhere on the token side. Invisible in both facts today; becomes a query only once #1180's `attempts` exists. Open question 5 asks whether it deserves its own issue now.
- **§2.7 is independently confirmed on production rows.** #1180's review aborted its backfill on the same leaf-title collision I found by reading the code: the daily of **2026-07-29** holds two `agent-context-id-isolation.spec.ts` rows with an identical leaf title, and **17 spec files** use an interpolated `describe`, concentrated in `llm-agents/` and `mcp/client/` — which is where this suite's token spend lives. Two independent routes, one bug.
- Two new validation assertions about the shared identity, three new failure-mode rows (rename/move, per-attempt billing, exact-vs-reconstructed join), and two new open questions — including whether the identity expression should become a shared `e2e_test_key()` helper rather than two literal copies, which is the same divergence hazard #1180 generated the column to prevent, one level up.

Diff is still documentation-only: +595 here, +5 in #1180's header.

Issue #1055, opened 2026-07-29
manual.yml aborts in globalSetup with zero tests — the credential gate fires on every dispatch


Victor-w-Madeira:
## Problem

`manual.yml` cannot execute a single spec. Every dispatch dies in `globalSetup`
before the first test, on both the main run and the destructive lane.

```
[preflight] backend healthy at http://localhost:7860/ — Langflow Nightly 1.12.0.dev9
Error: [preflight] provider key(s) set in the environment but NOT configured as a
Langflow global variable: OPENAI_API_KEY. Specs resolve credentials from Langflow,
not the env var, so they would fail with a misleading error ...
##[error]Process completed with exit code 1.
```

Three facts combine:

1. `manual.yml` passes `secrets.OPENAI_API_KEY` into the `run-e2e` action on
   **every** dispatch — it is not conditional on anything.
2. `manual.yml` has **no `collect-models` step**, so nothing ever imports that
   key into Langflow as a global variable. `daily-stable.yml` and
   `pr-validation.yml` both run one.
3. The #884 pre-flight gate hard-fails in CI when a provider key is present in
   the environment but absent from Langflow's variables — which is exactly the
   state above.

No dispatch input reaches this: the workflow exposes `langflow_target`,
`langflow_image`, `test_tag` and `test_grep`, none of which can set
`PREFLIGHT_SKIP_CREDENTIALS` or add a collect-models step.

## Why it went unnoticed

The gate landed on 2026-07-22 (#885, resolving #884). `manual.yml`'s last
dispatch before today was **2026-07-08** (run 28976183193), and the workflow file
itself has not changed since 2026-07-14. So no one has dispatched it since the
gate existed, and its own docs still describe it as the parameterized manual run.

Surfaced on run
[30420649707](https://github.com/oriontech-me/langflow-e2e/actions/runs/30420649707),
dispatched to give PR #1052 the E2E coverage `pr-validation.yml` does not provide
for a shared-helper change (#1054). It reported failure with zero tests executed.

## Fix options

The two workflows that survive the gate do it in different ways, and either
shape works here:

- **Run `collect-models` first**, as `daily-stable.yml` and `pr-validation.yml`
  do — with `PREFLIGHT_SKIP_CREDENTIALS: "1"` on that step and
  `PLAYWRIGHT_RETRIES: "0"`, plus the post-sweep health gate (#1011/#1019/#1045),
  since the sweep can wedge the shared backend. Most faithful to the other
[...]

Issue #967, opened 2026-07-27
[Daily #962] Anthropic provider inactive — all 13 candidate Claude models fail validation (1 hard, 1 flake, 24 skips)


rafaelgiln:
Spun out of daily-failure triage #962 (run [30261409427](https://github.com/oriontech-me/langflow-e2e/actions/runs/30261409427), 2026-07-27, `langflowai/langflow-nightly:latest`).

## Symptom

The `anthropic` provider probed **inactive** in this run: `collect-models` tried **all 13** candidate Claude models and every one failed validation. This is the **dominant driver of the day** — it accounts for 1 hard failure, 1 flake, and **24 of the run's 30 skips**.

The `collect-models` skip reason carried by every affected spec:

```
Provider "anthropic" inactive — all 13 candidate model(s) failed validation
(tried: claude-sonnet-5, claude-sonnet-4-6, claude-sonnet-4-5,
claude-sonnet-4-20250514, claude-haiku-4-5, claude-opus-5, claude-fable-5,
claude-opus-4-8, claude-opus-4-7, claude-opus-4-6, claude-opus-4-5,
claude-opus-4-1, claude-opus-4-20250514);
last error: Your credit balance is too low to access the Anthropic API.
Please go to Plans & Billing to upgrade or purchase credits.
```

**The rejection reason is already in hand** — the run's `results.json` carries it verbatim in every affected spec's skip annotation (quoted in full above): a **credit-balance exhaustion**, not an authentication failure. That narrows directive (a) below from "classify the rejection" to "confirm and top up".

### Tests that ran and failed

| Spec (line) | Assertion / waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/model-provider/anthropic-provider.spec.ts:239` ("configured Anthropic selects a Claude model in the Agent and executes the flow") — **hard failure**, 3 attempts | `getByTestId('div-chat-message').last()` | `expect(locator).toBeVisible() failed` — `element(s) not found`, 30000ms |
| `tests/tests-automations/regression/core-functionality/model-provider/anthropic-provider.spec.ts:181` ("Anthropic API key is configured via Settings → Model Providers") — **flaky** (recovered on retry) | `expect(persistResp.ok()).toBe(true)` (line 234) | `expect(received).toBe(expected)` — `Expected: true / Received: false` |

[...]

PR #1245, opened 2026-08-03
ci(manual): add a provider input so a lane can still ask for multi-provider (#1186)

### LINKED ISSUE #1185: daily-stable: rotate the tool-calling tier through one provider per weekday, falling back to the next active one

### LINKED ISSUE #1186: manual.yml: add a provider input — once the daily pins, no lane runs multi-provider and there is no way to ask for one
Follow-up of #1169 (PR #1170). Counterpart of #1185: once the daily pins to one provider,
multi-provider coverage needs an explicit way to be asked for.

## Problem

Multi-provider is currently the **default**, by omission: with no `MODEL_TEST_ID` /
`MODEL_TEST_PROVIDER` set, `getTestTargets()` returns one model per active provider. Nobody
requests multi-provider — you get it unless a lane opts out, which is exactly what
`pr-validation.yml` did in PR #1170 and what #1185 proposes for the daily.

Once both lanes opt out, **there is no lane left that runs multi-provider**, and no input to
ask for it. `manual.yml` accepts only:

| Input | Purpose |
|---|---|
| `langflow_target` | Docker tag or full URL |
| `langflow_image` | nightly / stable |
| `test_tag` | Playwright tag |
| `test_grep` | extra name filter |

Nothing selects a provider. The full multi-provider run — the one that catches
provider-specific regressions like #643 (Anthropic streaming / `thinking`) and #963
(Gemini "Message empty.") — would become unreachable except by editing a workflow file.

## Fix

Add a `provider` input to `manual.yml`:

| Value | Behaviour |
|---|---|
| `auto` (default) | today's behaviour — one model per active provider (multi-provider) |
| `openai` / `anthropic` / `google` | pin to that provider's **settled** model, via `scripts/select-pr-model-target.mjs --provider <x>` |
| `all-models` | `ALL_MODELS=true` — every model in the catalog, for a deliberate sweep |

Two constraints carried over from #1169, both of which fail as a green run if ignored:

- The variables are a **pair**. `MODEL_TEST_PROVIDER` alone runs the provider's entire
  catalog (41 OpenAI entries on 2026-07-30) rather than narrowing to one model.
- The pinned model must be the one `collect-models` **settled**, never a hardcoded id — an
  id the CI project has lost access to produces a skip and a green run (#570 / #1012).

`manual.yml`'s **external-URL** path needs checking: it targets a Langflow that is not on the
runner's network, and the `Collect models` sweep may not be available there to settle a
model. Either the pin declines with a warning on that path (the #980 shape) or the input is
rejected for it — decide and encode, do not let it silently resolve to nothing.

## Open question this issue must answer

**Who dispatches this, and how far ahead of a release?**

That cadence *is* the detection latency for provider-specific regressions once #1185 lands.
It is the risk half of the same decision and the two issues should agree in writing before
either merges.

## Done when

- [ ] `manual.yml` exposes `provider` with the four values above; `auto` reproduces today's
      behaviour exactly.
- [ ] The provider/model pair is emitted by the shared script, not by inline `env:`.
- [ ] The external-URL path either declines to pin with a `::warning::` or rejects the input
      — never a silent all-skip.
- [ ] The release cadence and its owner are recorded here and cross-referenced from #1185.
- [ ] Covered by `npm run test:scripts`.

## Depends on

#1184 (the resolver) and #1185 (the pin this lane is the escape hatch for). Can be authored
in parallel but should not merge before #1185, so the suite is never left without a
multi-provider lane.

Victor-w-Madeira:
Multi-provider used to be the **default, by omission**: with no `MODEL_TEST_ID` /
`MODEL_TEST_PROVIDER` set, `resolveTestTargets()` returns one model per active provider.
Nobody requested it — you got it unless a lane opted out. Then both scheduled lanes opted
out: `pr-validation.yml` pins openai (#1169) and `daily-stable.yml` rotates one provider
per weekday (#1185). So **no lane ran multi-provider any more**, and there was no input
to ask for one short of editing a workflow file.

`manual.yml` gains a `provider` input:

| Value | Behaviour |
|---|---|
| `auto` (default) | one model per active provider — today's behaviour, emits **nothing** |
| `openai` / `anthropic` / `google` | pin to that provider's **settled** model, as a pair |
| `all-models` | `ALL_MODELS=true` — every model in the catalog, a deliberate sweep |

Resolved by `scripts/select-manual-model-target.mjs`, never by inline `env:`, for the two
reasons #1169 wrote a script for in the first place — both of which fail as a **green
run**: `MODEL_TEST_PROVIDER` alone does not narrow the run, it sweeps that provider's
entire catalog (41 openai entries on 2026-07-30); and a hardcoded id skips silently the
day the CI project loses access, leaving a green dispatch that tested nothing (#570 /
#1012). The decision function and the payload validation are
`select-pr-model-target.mjs`'s, shared with both other lanes, so none of the three can
drift on what `active` means.

## Three things worth reviewing on their own terms

**1. Failure semantics deliberately differ from the two sibling lanes.** Read as "the
same pin, inconsistently configured", this looks like an oversight. It is not.
`select-pr-model-target.mjs` and `select-daily-model-target.mjs` decline to pin and let
their lane fall back to the costlier multi-provider run — #980's trade, correct on a
*scheduled* lane where losing the day costs more than the spend and nobody asked for a
specific provider. Here **the input is the request**: a dispatch that says `anthropic` is
asking "does this break on anthropic", and continuing after a dry anthropic key would run
the other two providers, answer a different question and report success. So a selection
that cannot be honoured **exits 1** with the cause named, and the step carries **no**
`continue-on-error`. This lane is never scheduled, so that costs a re-dispatch, not
coverage. Exit codes: `0` honoured, `1` cannot honour, `2` undecidable payload (#1035).

**2. `auto` emits nothing and no state of the catalog can fail it — but it is NOT
exempt from the guards, and the second commit is about why.** "Emits nothing" is not the
property a dispatcher relies on; "delivers multi-provider" is, and that one is not free.
`collect-models` writes `models.json` only at the *end* of its sweep and is
`continue-on-error` here, so a sweep that throws part-way leaves no catalog — and with no
catalog `resolveTestTargets()` returns **one** `provider:openai (fallback)` target whose
`skipReason` is `undefined` (the health map is empty on an absent `providers.json`), while
openai's key *is* forwarded. ~30 parametrized agent tests would then run one provider,
green, under a log line promising multi-provider: the release-gate dispatch of the
recorded cadence reporting coverage it did not have. So `auto` reads the catalog for
*reporting only* (never failing on it — asserted over a throw, an absent file, an empty
array and a non-array), prints the fan-out **derived from that same file**, and warns when
it is absent, empty, unparseable, or a single provider.

**3. `all-models` declines on an absent or empty catalog.** `ALL_MODELS=true` over an
empty `models.json` does not sweep — `resolveTestTargets()` falls back to a single
`(fallback)` target per spec. Exporting it anyway would log a successful sweep that swept
nothing. When it does sweep, the size is announced **first** (`::warning::` with the
target and provider counts): it is the most expensive shape this lane can run.

## The external-URL path: rejected, not declined

The issue left this open. It is **rejected** (`::error::` + `exit 1`), mirroring the step
that already rejects `any_completion_provider: ollama` on the same job. That job runs no
`Collect models` — deliberately, since the sweep would write our CI keys as global
variables into an instance we do not own (#1055) — and `run-e2e` is called there with no
provider key, so `hasProviderEnvKeys()` already makes every parametrized LLM spec skip
regardless of this input (#967), and `providers.json` is never written, so there is
nothing to pin *to*. Honouring the input would produce a **green run that executed none
of the specs it aimed at** — the silent all-skip the issue's own criteria forbid. A
decline-with-warning would satisfy "never silent" in letter and still hand back that
green run. Empty is accepted alongside `auto`, so a dispatch queued before this input
existed can still be re-run.

## A correction to #1185, recorded while measuring the gap this closes

#1185 recorded the rotation's blindness window as "≤3 days". That is the *intra-week*
gap; the cron is **Mon–Fri**, so the real per-provider detection latency is:

| Provider | Slots | Worst case |
|---|---|---|
| openai | Mon, Thu | ≤4 days |
| anthropic | Tue, Fri | ≤4 days |
| google | Wed | **7 days** |

`google` runs once a week — and google is the provider of #963 (`mcp-client-agent`
returning "Message empty." on gemini while the tool fires). Hence the cadence recorded on
the issue: **one `provider=auto` + `test_tag=@stable` dispatch per Langflow
release-candidate cycle, ≥2 business days before sign-off, owned by whoever signs the
release validation**, plus ad-hoc `provider=<x>` dispatches when triaging a
provider-specific red. Cross-referenced on #1185.

## Validation

- `npm run test:scripts` — **469 passed, 0 failed** (30 new cases).
- `npm run test:units` — 334 passed. `npm run typecheck` clean. `npm run lint` 0 errors.
- **Every guard force-failed by mutation** before being kept — 17 mutations across the
  two commits, all bit, all reverted, plus a control (an unrelated edit that must *not*
  fail). Round 1: `continue-on-error` on the pin step; the external-URL rejection removed;
  `groq` in the choice list; the pin emitting `MODEL_TEST_PROVIDER` alone; `all-models`
  sweeping an absent catalog; a declined pin exiting 0; the step moved after the run.
  Round 2 (the six an independent review got past round 1's guards, see the comment
  below): a hardcoded `--selection auto`; the `env: SELECTION` binding deleted;
  `|| true` appended; `&&` flipped to `||` in the external-URL condition; a run-e2e step
  dropping one provider key; the caller dropping one; and `MODEL_TEST_ID` shadowing the
  pin from run-e2e's step `env:`. Plus the two `auto` guards from that round.
- CLI exercised end-to-end on a realistic payload (openai active, anthropic `inactive`
  with `credit balance is too low`, google active): `auto` → nothing exported, fan-out
  listed; `openai` → both variables; `anthropic` → `::error::` carrying collect-models'
  own message, **exit 1**, nothing exported; `all-models` → `ALL_MODELS=true` plus the
  sweep-size warning.
- **The workflow half cannot be proven locally.** `scripts/ci-change-coverage.mjs` on
  this diff returns `dispatch` naming `manual.yml` — correctly: nothing in the PR lane
  reaches this script. The final proof is a `manual.yml` dispatch; what to watch is the
  `Resolve the run's provider selection` step, and then whether the parametrized agent
  tests in the report carry the pinned provider's label only.

## Structural guards added (`npm run test:scripts`)

- the workflow's `choice` options **equal** the script's vocabulary (a drift makes the
  workflow offer a value that exits 2, or hide one that works);
- the step sits between the post-collect-models health gate and the run;
- the step is **not** `continue-on-error`;
- no `MODEL_TEST_*` / `ALL_MODELS` `env:` key anywhere in the workflow;
- the external-URL job rejects any selection but `auto`, before the run;
- `PIN_PROVIDERS` equals the `*_api_key` inputs `run-e2e` actually forwards — pinning to
  a provider whose key this lane never passes would skip everything and read green
  (#967).

Closes #1186


Victor-w-Madeira:
## Second commit: an independent review round (`f5cc930`)

The first commit was reviewed with clean context, adversarially, against the real
sources. It produced one finding worth the round on its own and six surviving mutations.

### The finding: `auto` was exempt from every guard, on the wrong property

`auto` was argued safe because it "reads nothing it needs". True, and irrelevant — the
property a dispatcher relies on is **"`auto` delivers multi-provider"**, and that one was
unguarded.

`collect-models` writes `models.json` at the very END of its Settings sweep
(`collect-models.ts`), and `Collect models` is `continue-on-error` on this lane. So a
sweep that throws part-way — the #922/#927 wedge, a broken Model Providers page — leaves
**no catalog**. In that state `resolveTestTargets()` takes its empty-catalog branch and
returns **one** target, `provider:openai (fallback)`, whose `skipReason` is `undefined`
(`readProviderHealth()` returns `null` on an absent file, so the skip map is empty). This
lane forwards openai's key, so **nothing skips**: ~30 parametrized agent tests run one
provider and the job is green — under the old log line claiming *"one model per active
provider"* and *"the specs report their own skip reasons"*, both false.

Walk it with the cadence this PR records: nightly regresses the Model Providers page →
`Collect models` red (continue-on-error) → health gate green (the backend is fine) →
`auto` prints its reassuring line → zero anthropic and zero google variants → green →
release signed off on a multi-provider sweep that ran one provider.

Fixed by reading the catalog for **reporting only** — never failing on it, asserted over a
throw, an absent file, an empty array and a non-array — printing the fan-out derived from
`models.json` (the file the resolver actually reads), and warning when that fan-out is
absent, empty, unparseable, or single-provider. `describeAutoFanout` is gone: it read
`providers.json` and could disagree with the run.

### Six mutations that survived all 469 tests

Round 1's YAML guards pinned **spellings**, not wiring — #1226's lesson, recurring in the
half the PR lane cannot exercise (`ci-change-coverage` returns `dispatch` here, so nothing
else catches it either).

| Mutation | What it broke | Now caught by |
|---|---|---|
| `--selection "$SELECTION"` → `--selection auto` | the input becomes **decorative**; every dispatch runs multi-provider and the log agrees with itself | the step must pass `"$SELECTION"` through **and** bind it to `inputs.provider` |
| delete the `env: SELECTION` block | same, via the empty→auto path | same guard |
| append `\|\| true` | removes the exact behaviour the "NOT continue-on-error" test protects, in different characters | `\|\| true`, `; true` and `if:` rejected alongside `continue-on-error` |
| `&&` → `\|\|` in the external-URL condition | **every** external-URL dispatch fails, `auto` and empty included | the condition compared as a normalized whole, not two substring probes |
| a run-e2e step stops forwarding one provider key | `provider=google` pins fine, then every google spec skips — #967 | every step forwarding one key must forward all three |
| the caller stops passing one key | same, from the other side | the Docker job must pass all three into `run-e2e` |

Plus the reviewer's bonus: `MODEL_TEST_ID` in `run-e2e`'s step `env:` **shadows** everything
the pin wrote (step `env:` outranks `$GITHUB_ENV`), and that is where Playwright actually
starts — the no-inline-pin guard now covers `run-e2e/action.yml` too.

All 17 mutations across both commits now fail, and a control mutation (an unrelated edit)
still passes, so the guards are not merely diff-sensitive.

### Two claims corrected rather than defended

- **`all-models` said "for EVERY parametrized spec".** The two specs declaring a `requires`
  capability sweep a capability-filtered subset (`CAPABILITY_EXCLUDES`), so the count is an
  upper bound. The warning says `up to N`, and now also states that an unnarrowed sweep will
  not finish inside the job's 90-minute timeout (~30 parametrized tests × a ~90-model
  catalog, unsharded) — repeated in the input description. **Not gated:** even
  `test_tag=@stable` is far over budget, so any threshold would be arbitrary; the honest
  move is to tell the dispatcher, in the input and in the log, before the run starts.
- **The rotation latencies are best cases, not bounds.** The rotation advances past an
  inactive provider, so a Wednesday with a dry google key costs google its whole week.
  Corrected in the script header, the workflow comment and `CLAUDE.md`.

### Verified and left alone

The review confirmed, from the code rather than the description: nothing leaks on the
`auto` path; the `$GITHUB_ENV` → composite-action → `resolveTestTargets` chain has no
shadowing today; `PIN_PROVIDERS` cannot drift from the catalog's providers (`collect-models`
sweeps keyed providers only, and `providerConfigMap` has no groq/mistral in that set); the
empty-input path is load-bearing, because `github.event.inputs.*` does not apply a declared
default; and the external-URL rejection's premise holds (`models.json` is gitignored, so
that job's fresh checkout has no catalog and `hasProviderEnvKeys` is false with no key).

### Out of scope, recorded

`$GITHUB_ENV` is written as `KEY=value` with no delimiter form and no newline validation in
**all three** lane scripts — a model id containing a newline would define arbitrary
variables for every later step. Pre-existing, identical in `select-pr-model-target.mjs` and
`select-daily-model-target.mjs`, and fixing it belongs in the shared writer rather than in
this lane's copy. Worth its own issue.

Issue #1263, opened 2026-08-04
[Daily #1258] Settings → Model Providers — the provider row never becomes clickable (provider-item-* click timeout, 2 specs)


Victor-w-Madeira:
Spun out of daily-failure triage #1258 (run [30901311395](https://github.com/oriontech-me/langflow-e2e/actions/runs/30901311395), 2026-08-04).

**Upstream:** _not filed_

## Symptom

The two `@stable` tests that configure a provider API key through **Settings → Model Providers** both timed out clicking the provider row on that screen — one as a final hard failure, one as a flake that recovered on retry. The hard failure cascaded a skip onto the dependent test in its own file.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/model-provider/openai-provider.spec.ts:137` ("OpenAI API key is configured via Settings → Model Providers") | `getByTestId('provider-item-OpenAI')` | `TimeoutError: locator.click: Timeout 20000ms exceeded.` |
| `tests/tests-automations/regression/core-functionality/model-provider/google-provider.spec.ts:90` ("Google API key is configured via Settings → Model Providers") | `getByTestId('provider-item-Google Generative AI')` | `TimeoutError: locator.click: Timeout 20000ms exceeded.` |

## Why these failures are one cause

The shared dimension is **one screen and one interaction**, with the **same normalized signature** on both — an unusually strong grouping for this repo.

Both tests open Settings → Model Providers and click the row for their provider to enter a key. Both raise `TimeoutError: locator.click: Timeout 20000ms exceeded.`, and the call logs differ only in which row they name: `getByTestId('provider-item-OpenAI')` and `getByTestId('provider-item-Google Generative AI')`. Same screen, same locator family, same signature, same run.

Two different providers producing the identical failure on the identical control is what argues **against** a per-provider cause and for a cause in the Model Providers screen itself (or in the environment it was rendered in).

Explicitly **not** folded in here: `openai-provider.spec.ts` and `google-provider.spec.ts` each also carry an *Agent model selection* test that failed on this run. Those wait on a different screen and a different observable, and the Google one is filed as its own cluster — this argument does not extend to them.

`openai-provider.spec.ts:201` ("configured OpenAI selects a GPT model in the Agent and executes the flow") was **skipped as a consequence** of the hard failure in the same file — cascading, covered here, not filed separately.

## Preliminary read (descriptive — NOT a verdict)

[...]

Issue #1261, opened 2026-08-04
[Daily #1258] Google — the Agent never settles a Gemini model/credential on the persisted flow (2 specs)


Victor-w-Madeira:
Spun out of daily-failure triage #1258 (run [30901311395](https://github.com/oriontech-me/langflow-e2e/actions/runs/30901311395), 2026-08-04).

**Upstream:** _not filed_

## Symptom

The `@stable` test that drives an Agent with a pinned Gemini model hard-failed (final status, after retries) waiting on a predicate inside `loadAgent()`, so the Agent never reached the state where a Gemini model is selected on the persisted flow. A second spec on the same provider flaked on the explicit guard for that same state.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/model-provider/google-provider.spec.ts:155` ("configured Google selects a Gemini model in the Agent and executes the flow") | `loadAgent() predicate → value-dropdown-model_model containing /gemini\|gemma/i` | `Error: expect(received).toBe(expected) // Object.is equality` |
| `tests/tests-automations/regression/mcp/client/mcp-client-agent-gemini-tool-regression.spec.ts:139` ("Gemini invokes the echo MCP tool (regression for fixed upstream #440)") | `Agent credential settled on the persisted flow (#751 guard, #1072)` | `Error: Agent credential never settled on the persisted flow (#751 guard, #1072).` |

## Why these failures are one cause

The shared dimension is **the provider and the state both specs wait on**, not the signature — the two signatures differ, and they are grouped deliberately.

Both specs load an Agent with a Gemini model pinned from the catalog and then wait for that model/credential to be settled **on the persisted flow** before running it. `google-provider.spec.ts:155` times out on the predicate inside `loadAgent()`; `mcp-client-agent-gemini-tool-regression.spec.ts:139` fails on the explicit `#751`/`#1072` guard for the same condition, which is the named form of that wait.

The cross-day evidence ties the two signatures to one state rather than to two problems: on the **2026-08-03** daily, `google-provider.spec.ts:155` itself flaked with the *other* spec's signature — `Error: Agent credential never settled on the persisted flow (#751 guard, #1072).` So the same test produces either signature depending on where the wait gives up.

The dataset flags `google` as a **provider-wide cluster** (4 tests across 3 files on this run — #899's rule for treating a provider variant as one candidate cluster rather than splitting it per symptom).

[...]

Issue #1126, opened 2026-07-30
[Daily #1121] mcp-client-regression: page.waitForURL times out on the blank-flow entry point (recurrent flake, 3×)


Victor-w-Madeira:
Spun out of daily-failure triage #1121 (run [30534416609](https://github.com/oriontech-me/langflow-e2e/actions/runs/30534416609), 2026-07-30). The run was a clean, non-guarded daily (466 passed, 2 hard failures, 10 flakes — the mass-failure guard did not trip), so this failure is not mass-failure collateral. Backend liveness measured no mid-run outage on 4/4 shards, but 31 single-probe failures were discarded as blips, so a wedge is not fully ruled out for the run (#1030); this test's last error is not transport-level, and the run attributed it to the spec.

**Upstream:** _not filed_

## Symptom

`openBlankFlow()` creates the flow successfully and then never sees the navigation: `page.waitForURL: Timeout 30000ms exceeded` waiting for a URL matching `/flow/` until `load`. Recovered on retry (1 failed attempt of 2). Same-signature recurrent across the dailies of **2026-07-16**, **2026-07-20** and **2026-07-30** (the same test also failed on 07-10 and 07-14 under *different* signatures — context only, not part of the recurrence count).

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests-automations/regression/mcp/client/mcp-client-regression.spec.ts:238` ("unreachable HTTP server results in empty tool dropdown") | `page.waitForURL(/\/flow\//) after the blank-flow click (30 s)` | `TimeoutError: page.waitForURL: Timeout 30000ms exceeded.` |

## Why these failures are one cause

Single test, single signature — no grouping claim. Filed separately from #966, which is `waitForURL`-adjacent in wording only: its signature is an `expect(received).toBe(expected)` inside `open-new-flow-templates-modal.ts` on `run-flow.spec.ts:12`.

## Preliminary read (descriptive — NOT a verdict)

[...]

PR #1279, opened 2026-08-04
fix(specs): attribute the page-entry barrier and stop the provider-panel re-fill (#1262)

### LINKED ISSUE #1262: [Daily #1258] language-model-regression — mainpage_title never renders at the spec's entry point (recurrent 3×, same signature)
Spun out of daily-failure triage #1258 (run [30901311395](https://github.com/oriontech-me/langflow-e2e/actions/runs/30901311395), 2026-08-04). Split out of the run's `google` provider-wide cluster during triage: the failure point is the spec's page-entry bootstrap, not anything provider-specific.

**Upstream:** _not filed_

## Symptom

A `@stable @release` test hard-failed (final status, after retries) **before reaching any provider interaction** — the wait that timed out is the page-entry bootstrap inside `openBasicPrompting()`, which polls for the main page title.

|---|---|---|
| `tests/tests-automations/regression/core-functionality/llm-agents/language-model-regression.spec.ts:135` ("language model must respond with Google provider") | `locator('[data-testid="mainpage_title"]') to be visible` | `TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.` |

One spec, one cause — the argument here is about **why it is not part of the Google cluster** it was first grouped into.

The test is named for the Google provider and the run's dataset flags `google` as a provider-wide cluster, so the obvious grouping is with `google-provider.spec.ts:155`. The call log refutes it: the timeout is `waiting for locator('[data-testid="mainpage_title"]') to be visible`, raised from `openBasicPrompting()` — the **first** statement in the test body, before `setupGoogle()` and before any model is resolved. Nothing Google-specific had run yet.

The shared dimension is the **spec file and its entry point**, and the history shows it is not provider-scoped: on the **2026-07-15** daily the *sibling* test in the same file, `language model must respond with **OpenAI** provider`, hard-failed with the identical signature `TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.` The signature follows the file, not the provider.

`mainpage_title` is the observable this repo's page-entry helpers wait on (`tests/helpers/other/await-bootstrap-test.ts`, `tests/helpers/flows/load-template-by-name.ts`, `tests/pages/MainPage.ts`), which is what makes this an entry-point cluster rather than a model-provider one.

**Recurrent with a stable signature** — this is the durable part of the signal. `recurrence.same_signature` is true: `TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.` on **2026-07-09** (hard), **2026-07-14** (flaky) and **2026-08-04** (hard), plus the same signature on the file's OpenAI sibling on 2026-07-15. Across the full 30-day window the test failed on 6 dailies with mixed signatures (07-07, 07-08, 07-09, 07-14, 07-22, 08-04).

The run carried a measured mid-run backend wedge (#1030/#1048): 13 outages, 12.4 min of unreachable backend across 3 of 4 shards, up to a 27% down-share on the affected shards. This attempt ran on shard 4 and failed **inside** a recorded outage window (`10:47:01→10:48:33` UTC, 92 s) — and a page that never bootstraps is exactly what an unreachable backend looks like from the harness. That is a strong lead and it must be ruled out first, but the umbrella's liveness section is explicit that overlap is a **lead, not a verdict** at a 23.6% down-share, and the same signature recurring on 07-09 and 07-14 is not explained by today's outage.

`page.waitForSelector: Timeout` is deliberately **not** on the infra-signature exemption list (`scripts/lib/infra-signatures.ts`) because a genuine regression produces it too, so `@stable` was auto-removed by the workflow, as designed.

Related entry-point issues, cited for context and **not** claimed as the same cause (different spec, different observable): #1126 (`mcp-client-regression`, `page.waitForURL` on the blank-flow entry point) and #966 (`run-flow`, welcome-panel predicate).

The guard did **not** trip on this run (4 hard failures), so no day-level environmental verdict is claimed.

## Investigation directive

1. On the current nightly, confirm whether the app's main page actually bootstraps and renders `mainpage_title` within 30 s on a cold entry — rule out a regression (or a durable slowdown) in the page-entry path before touching the wait. This is the product-first step and it is mandatory.
2. Then rule out the run environment for **today's** occurrence specifically: check the Langflow service container log for gunicorn `WORKER TIMEOUT` in `10:47:01→10:48:33` UTC on shard 4. Note that this cannot explain 2026-07-09 or 2026-07-14, so a clean environment verdict for today does not close the issue.
3. Only after both, consider wait-strategy fragility: whether a fixed 30 s `waitForSelector` on `mainpage_title` is the right entry-point barrier, and whether the shared page-entry helpers should expose a deterministic readiness observable instead.

Check whether #1126 and #966 turn out to share this cause once reproduced live; if they do, merge per the enrich/merge rule (keep the older number, carry over the claimed spec paths).

## Deliverables (Done when)

- [ ] The recurrence is addressed, not just today's run: the fix accounts for the same signature on 2026-07-09 and 2026-07-14, or explains why those were a different cause.
- [ ] Cross-checked against #1126 and #966 for a shared entry-point cause; merged or explicitly kept separate.

rafaelgiln:
**Fixes #1262.**

## Problem

The issue was filed as an entry-point failure — `mainpage_title` never rendering inside `openBasicPrompting()`, "recurrent 3×, same signature". Neither half of that framing survives the evidence.

**The recorded signature was the LAST attempt's.** In shard 4 of run [30901311395](https://github.com/oriontech-me/langflow-e2e/actions/runs/30901311395) the three attempts failed on three different observables:

| Attempt | Observable | Where |
|---|---|---|
| 1 (initial) | `node_duration_chat output` (60s) + `apiRequestContext.get` timeout on `/api/v1/auto_login` | this file's own build wait (line 187) |
| retry #1 | `[id="new-project-btn"]` | `await-bootstrap-test.ts:61` |
| retry #2 | `[data-testid="mainpage_title"]` | `await-bootstrap-test.ts:16` |

**Environment confirmed for today's occurrence.** The Langflow service logged `WORKER TIMEOUT (pid:37)` at **10:46:42** UTC → `Worker (pid:37) was sent SIGKILL!` → restart finished ~10:47:00 (starter projects at 10:46:57), i.e. inside the triage's recorded outage window `10:47:01→10:48:33`. Both retries ran against a restarting backend, and `openai-provider.spec.ts:137` on the same shard failed on the same `mainpage_title` — collateral, not two broken specs (that sibling was restored separately in #1263 / #1278).

**"Recurrent 3×, same signature" was an artifact of the shared Playwright prefix** `TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.`. On **2026-07-09**, **07-14** and (the OpenAI sibling) **07-15** the locator was `text=built successfully` — the #750 observable, a different failure. The entry point was never the recurring cause.

**Product ruled out first (mandatory step 1):** on 1.12.0.dev15 a cold entry renders `mainpage_title` in **1.4–4.3 s, 5/5**.

**The file's real recurrent cause, reproduced.** Running the whole file, the Google test failed with the build never completing, and the failure snapshot names the reason:

```
Error calling model 'gemini-flash-latest' (INVALID_ARGUMENT): 400 INVALID_ARGUMENT.
{'error': {'code': 400, 'message': 'API key not valid. Please pass a valid API key.',
 'status': 'INVALID_ARGUMENT', … 'reason': 'API_KEY_INVALID' …}}
```

`setupGoogle` decided "already configured" from a **1 s `isVisible`** probe on the Disconnect button, which paints only after the panel's own backend fetch resolves. A slow fetch read as "not configured", so the helper filled the API key field **over a provider that already held a credential**; Langflow then called Google with a key it rejects, the build never completed, and the spec timed out on its build observable — exactly the 07-09/07-14/07-15 and attempt-1 shape. Measured on 1.12.0.dev15: **1 failure in 7 runs**, with the 6 clean runs all observing the field masked as `AIza•••••••••••••••••••••••••••••••••••`.

**Verdict:** attempt 1 + the recurrence = **test defect** (fixed here); the recorded entry-point signature = **environment** (verdict 4, no test change earns anything there — only attribution).

## Fix

**1. The page-entry barrier is attributed** — new `tests/helpers/other/page-entry-barrier.ts`. On timeout it probes `GET /api/v1/version` and says which state it observed:

- backend did not answer, or answered non-2xx → prefixed `[backend-unreachable]`, stating this is not an entry-point regression;
- backend answered 200 → the failure stays attributed to the UI (a genuine entry-point regression must not be dressed up as an outage);
- the probe itself could not run → **UNKNOWN**, never "healthy" (#1012's rule).

The original Playwright error is always appended, so the trace/screenshot stays readable against it. Wired into all six barrier sites: `awaitBootstrapTest` (3), `MainPage.waitForLoad` (2), `loadTemplateByName` (2 — one shared with the retry path), `addFlowToTestOnEmptyLangflow` (1). The probed URL is taken from the page's own origin, not from `PLAYWRIGHT_BASE_URL` — the two can disagree, and naming an origin nobody called is how a reader concludes the wrong backend was healthy.

**2. The re-fill race is fixed at its source** — new `tests/helpers/provider-setup/provider-config-state.ts`. `providerAlreadyConfigured()` polls **both** signals to a 5 s deadline and treats a **non-empty (masked) key field as configured on its own**: Langflow renders a stored credential as `AIza••••…`, and a masked value can never be one this helper typed, so it cannot skip a setup that was genuinely needed. 5 s rather than longer because an unconfigured provider (a fresh CI container before its first setup) pays that whole window on every call.

**3. `@stable` restored** on `language model must respond with Google provider` — its removal was collateral of the wedge, not a verdict on the test.

**Rejected alternatives:** raising the 30 s entry barrier (a 90 s outage is not something 119 specs should wait out silently — retries already cover it, and attribution is what was missing); adding `[backend-unreachable]` to `scripts/lib/infra-signatures.ts` in this PR (the message now meets #1031's bar, but that changes the `@stable` auto-removal contract on `main` — proposed as a separate decision).

## Scope note

Success-path behaviour is unchanged everywhere: same selectors, same budgets, only the failure message differs. `setupGoogle`'s fill/save/toggle flow is untouched apart from the configured-detection. The only spec-file change is the restored tag. `QA-CHECKLIST.md` needs no edit — the manual Part II bullet already references this spec (`npm run check:checklist-coverage` ✅); the generated blocks regenerate on merge.

## Validation (nightly 1.12.0.dev15, `--retries=0`, `--workers=1`)

- `npm run typecheck` ✅ (0 errors) · `npm run lint` ✅ (0 errors) · `npm run test:units` ✅ 413/413 · `npm run test:scripts` ✅ 623/623 · `npm run check:checklist-coverage` ✅
- `language-model-regression.spec.ts` full file, 3 consecutive runs: **4 passed** (56.7 s / 52.2 s / 60 s)
- Google test in isolation before the fix: 3/3 green, 22–24 s — the failure only appears at ~1/7, which is why the whole file was run repeatedly
- Barrier consumers (`project-management/folder-crud.spec.ts`, `flow-functionality/duplicate-flow.spec.ts`): **4 passed**
- Zero `🚨 Backend Error` in the validating runs
- Force-fails **executed**, one per test in the touched file, isolated with `--grep` (serial files mask siblings after a failure), reverted from a pristine copy — post-FF the spec is byte-identical and carries 0 mutation markers:
  - `FF:` OpenAI test — `/4/` → `/FF_MUTANT_999777/` ⇒ failed (`toContainText`, 15000ms)
  - `FF:` Google test — `length > 1` → `> 100000` ⇒ failed (60000ms)
  - `FF:` switch test — `/gemini/i` → `/FF_MUTANT_no_such_model/i` ⇒ failed (15000ms)
  - `FF:` dialog test — `provider-item-OpenAI` → `provider-item-FF_MUTANT_absent` ⇒ failed (10000ms)
  - `FF:` barrier, product branch — healthy backend + a testid that never renders ⇒ message says `IS a product/UI failure`, **no** infra prefix
  - `FF:` barrier, infra branch — dead origin (`127.0.0.1:9`) ⇒ `[backend-unreachable] … did not answer GET /api/v1/version … ECONNREFUSED`
  - `FF:` predicate — `providerAlreadyConfigured` mutated to ignore the masked-key signal ⇒ its unit test fails (pass 2 / fail 1), reverted ⇒ 3/3
- `--trace=on` **not run**: known local hang on this repo beyond the template family (#518/#490) — step verification came from the `--retries=0` runs, the force-fails and the failure snapshots instead.
- Flow cleanup verified: the 4 flows leaked by the validation runs were deleted id-scoped; `GET /api/v1/flows/` is back to the 2 pre-session flows.

## Cross-checks

- **#1261** is this run's real Google cluster — cross-linked, not merged: it fails on the Agent's settled credential/model on the persisted flow, a different state from this file's build.
- **#1126** (`page.waitForURL` on the blank-flow entry point) and **#966** (welcome-panel predicate) stay **separate**: different observables, and the new barrier does not cover `waitForURL`. Both become easier to triage now that a backend outage is named as one.

### DIFF
--- a/docs/core-functionality/llm-agents/language-model-regression.md
+++ b/docs/core-functionality/llm-agents/language-model-regression.md
@@ -37,6 +37,13 @@ PR #775 — the key had no quota, not the test — and **restored in #992** once
 the quota was confirmed back (live HTTP 200 completion on `gpt-4o-mini`) and
 the test ran clean at `--retries=0`.
 
+`@stable` was removed from the **Google** test again by the 2026-08-04 daily's
+auto-removal (run 30901311395) and is **restored by the #1262 fix**. The
+removal was collateral: the attempt that set the recorded signature ran while
+the shard's backend was restarting after a gunicorn `WORKER TIMEOUT`, so the
+recorded error is a page-entry timeout, not this test's own observable — see
+Notes (#1262).
+
 ---
 
 ## Preconditions *(optional)*
@@ -58,7 +65,11 @@ the test ran clean at `--retries=0`.
 ## Step by step *(required)*
 
 Common: open the Basic Prompting template (flow id captured from the page's
-own `POST /api/v1/flows/` response; deleted in `afterEach`).
+own `POST /api/v1/flows/` response; deleted in `afterEach`). The page-entry
+barrier inside `awaitBootstrapTest` now fails **attributed** (#1262): on
+timeout it probes `GET /api/v1/version` and says whether the backend answered,
+so an unreachable backend can no longer be reported as "`mainpage_title` never
+rendered".
 
 **OpenAI test:** `initialGPTsetup` — which pins a deterministic GPT model
 from `models.json` via `resolveGptModel()` (#606; UI preference-ranking is
@@ -97,6 +108,11 @@ visible → Escape.
   `node_duration_` badge within 60s (OpenAI test still uses the `built
   successfully` toast — #750 hardened only the flaking Google test); the
   Playground reply is the end-to-end proof the selected provider executed.
+- **Fix #1262 exit criterion:** the recorded page-entry signature is
+  attributed (backend-reachability probed at the barrier), the Google test's
+  own recurrent observable is addressed at its cause (the provider-panel
+  re-fill race below), and the test proves clean `--retries=0` runs on the
+  current nightly before `@stable` is restored.
 - **Fix #596 exit criterion:** the Google test's root cause is identified
   with evidence (test defect vs product vs environment — triage verdicts),
   the fix does NOT weaken asserts (no timeout inflation to mask latency),
@@ -170,3 +186,44 @@ visible → Escape.
   `setupOpenAI` was already preference-ranked, not blind "first
   available" — the nondeterminism was the last-resort fallback and the
   missing race gate, both addressed.
+- **#1262 root cause (2026-08-04):** the issue was filed as an entry-point
+  failure — `mainpage_title` never rendering inside `openBasicPrompting()` —
+  and that framing does not survive the evidence. Three separate findings:
+  - **The recorded signature was the LAST attempt's.** Attempt 1 failed on
+    `node_duration_chat output` (60s, this file's own build observable), retry
+    #1 on `new-project-btn`, retry #2 on `mainpage_title`. The shard's Langflow
+    logged `WORKER TIMEOUT (pid:37)` at 10:46:42 UTC and SIGKILLed the worker;
+    the restart completed ~10:47:00, inside the triage's recorded outage window
+    (10:47:01→10:48:33). Both retries ran against a restarting backend —
+    **environment**, verdict 4, no test change earns anything there.
+  - **"Recurrent 3×, same signature" was an artifact** of the shared Playwright
+    prefix `TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.`. On
+    2026-07-09, 07-14 and (the OpenAI sibling) 07-15 the locator was
+    `text=built successfully` — the #750 observable, a different failure. The
+    entry point was never the recurring cause.
+  - **The file's real recurrent cause is the provider panel's re-fill race.**
+    `setupGoogle` decided "already configured" from a 1s `isVisible` probe on
+    the Disconnect button, which is painted only after the panel's own backend
+    fetch resolves; a slow fetch read as "not configured" and the helper filled
+    the key field over a provider that already held a credential. Langflow then
+    called Google with an unusable key —
+    `Error calling model 'gemini-flash-latest' (INVALID_ARGUMENT): 400 … 'API
+    key not valid. Please pass a valid API key.' … API_KEY_INVALID` — so the
+    build never completed and the test timed out on its build observable, which
+    is exactly the 07-09/07-14/07-15 and attempt-1 shape. Reproduced on
+    1.12.0.dev15: **1 failure in 7 runs**, the 6 clean runs all observing the
+    provider already configured with the key masked (`AIza•••…`).
+    **Verdict: test defect.** Fix: `providerAlreadyConfigured()`
+    (`helpers/provider-setup/provider-config-state.ts`, unit-tested) polls both
+    signals to a 5s deadline and treats a **non-empty (masked) key field** as
+    configured on its own — a masked value can never be one this helper typed,
+    so it cannot skip a setup that was genuinely needed.
+- **Page-entry attribution (#1262, suite-wide):** the shared barrier now lives
+  in `helpers/other/page-entry-barrier.ts` and is used by `awaitBootstrapTest`,
+  `MainPage.waitForLoad`, `loadTemplateByName` and
+  `addFlowToTestOnEmptyLangflow`. On timeout it probes `/api/v1/version` and
+  prefixes the failure with `[backend-unreachable]` when the backend did not
+  answer (or answered non-2xx), keeps the failure attributed to the UI when it
+  did answer, and reports UNKNOWN when the probe itself could not run (#1012 —
+  an unevaluated probe is not a clean one). Success-path behaviour is unchanged;
+  only the failure message differs.
--- a/tests/helpers/flows/add-flow-to-test-on-empty-langflow.ts
+++ b/tests/helpers/flows/add-flow-to-test-on-empty-langflow.ts
@@ -1,5 +1,6 @@
 import type { Page } from "@playwright/test";
 import { dismissWelcomeOverlayAndWaitForModal } from "./open-new-flow-templates-modal";
+import { waitForPageEntry } from "../other/page-entry-barrier";
 
 export const addFlowToTestOnEmptyLangflow = async (page: Page) => {
   await page.getByTestId("new_project_btn_empty_page").click();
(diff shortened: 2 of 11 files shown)


rafaelgiln:
## The red E2E lane is #1234, not this PR

`core-components/loop-component-regression.spec.ts:152` ("Research Translation Loop … 2 ArXiv papers") failed 3/3 attempts on `[data-testid="template-research-translation-loop"]`. Everything else in the lane passed: **68 passed**, 1 skipped, 1 did not run.

Four pieces of evidence that it is not this diff:

1. **Control run on `origin/main`.** Detached worktree at `b2b3221` (zero of this branch's changes), same command, same nightly (1.12.0.dev15):
   ```
   ✘ loop-component-regression.spec.ts:152 … (14.3s)
     TimeoutError: page.waitForSelector: Timeout 10000ms exceeded.
       - waiting for locator('[data-testid="template-research-translation-loop"]') to be visible
   ```
   Identical failure without the fix present.
2. **The cause is the image, not a wait.** The nightly does not register that starter project at all:
   ```
   [warning] Skipping starter project 'Research Translation Loop'; unavailable components: ArXivComponent
   ```
   The card cannot render, so no wait strategy reaches it.
3. **Already tracked**: #1234 — "the Research Translation Loop template card never appears in the all-templates sidebar (deterministic, 3/3)" — open since the 2026-08-03 daily, before this branch existed. `@stable` was auto-removed there (`32ac9a1`), but the PR lane does not filter by `@stable`, so it still runs.
4. **This PR's barrier passed inside that very spec.** `awaitBootstrapTest(page)` at line 174 returned; the timeout is at line 178. Had the new barrier been involved, the message would carry its attribution (`[backend-unreachable]` or "… IS a product/UI failure at the page entry point").

Why it was selected at all: `loop-component-regression.spec.ts` imports `awaitBootstrapTest`, so the import graph pulls it in with the helper's other consumers — the known cost of a broad helper change meeting an already-red spec.

Also in the log, advisory only: `🚨 Backend Error: 500` on `POST /api/v1/flows/` (the ambient-contention class; it never fails a test by itself).

Follow-up: a separate PR quarantines #1234's test with `test.fixme` so it stops reddening unrelated lanes.

Issue #1189, opened 2026-07-31
Create: Human Input pause/resume in the Playground — decision card, exclusive branch routing (§9.6)


daniellicnerski1:
Wave 5 — 1.11.0 feature coverage.

**Backlog pointer:** QA-CHECKLIST.md §9.6 → `core-functionality/playground/human-input-pause-resume.spec.ts`

**Scope**
- Flow via `POST /api/v1/flows/` (no LLM key needed): Chat Input → Human Input (Approve/Reject) → each branch wired to a distinct Chat Output.
- Run in the Playground: the run suspends and the decision card (`human-input-card`) renders.
- Approve routes only the approved branch and the run completes; a second scenario validates Reject routes only the reject branch.

**Upstream:** langflow-ai/langflow#13633 (durable background execution + HITL suspend/resume), #14090 (HITL polish). Manual validation recipes exist on the QA platform (HITL-01/02/03).

PR #1284, opened 2026-08-04
test(core-components): cover Human Input node configuration (#1190)

### LINKED ISSUE #1190: Create: Human Input node configuration — default/custom action handles, persistence (§3.9)

**Backlog pointer:** QA-CHECKLIST.md §3.9 → `core-components/human-input-node-config.spec.ts`

**Scope**
- Human Input node renders default Approve/Reject branch handles.
- Adding a custom User Action (`action-add-input`) creates a new handle live.
- Configured handles persist after save + page reload.

**Upstream:** langflow-ai/langflow#13633 / LE-1449. Manual recipe: QA platform item HITL-01.

Victor-w-Madeira:
Closes #1190.

Closes the `[ ]` gap in `QA-CHECKLIST.md` §3.9 — first coverage of the **Human Input**
node (HITL, Langflow 1.11.0 — upstream langflow-ai/langflow#13633 / LE-1449). This is the
node's **configuration** surface only: its branch outputs are derived from the
`User Choices` field, and the spec pins that derivation on add, on live change, and after
a reload. Running the flow — pausing, the decision card, exclusive branch routing — is
#1189 (`core-functionality/playground/human-input-pause-resume.spec.ts`) and is
deliberately absent here, which is why this spec needs no provider credentials.

## 1. Covered tests

| 1 | `Human Input renders the default Approve and Reject branch handles when added to the canvas` | A freshly added node already exposes both branch handles (`handle-humaninput-shownode-approve-right`, `…-reject-right`) and the `Approve`/`Reject` chips, with the node's output-handle count **exactly 2**. Catches a regression in the hardcoded `outputs` list a dragged node depends on — `update_outputs()` only fires on a field change, so a node whose handles appeared only after touching a field would fail here. |
| 2 | `adding a custom User Action creates its branch handle without a reload` | Committing `Request Changes` in `action-add-input` makes `handle-humaninput-shownode-request changes-right` appear **on the same page, with no reload or navigation**, the two defaults intact, count 2 → 3. Catches a break in the `real_time_refresh` rebuild (the `POST /api/v1/custom_component/update` round trip) — the failure mode where the handle only shows up after a reload. |
| 3 | `the configured branch handles persist after save and reload` | `GET /api/v1/flows/{id}` reports the node's outputs as exactly `branch_approve, branch_reject, branch_request_changes` — which is also the assertion that the label survived as a slug — and after `page.reload()` all three chips and handles render again, count 3. Catches a break in `update_frontend_node()`, which rebuilds the outputs from the persisted `decisions` on load. |

## 2. How the test was built

- **Setup**: `setupBlankFlow(page)` (flow created via the REST API, opened through the
  dashboard card) and id-scoped cleanup in `afterEach` — `page.goto("/")` to unmount the
  editor first, then `deleteFlow`. Verified flow-neutral: the instance's user-flow count
  was identical before and after the runs (the ~16 flows the burst and the force-fail runs
  created were all deleted, including on the runs that failed on purpose).
- **Live-confirmed selectors** — every testid harvested from the running nightly's DOM
  (`1.12.0.dev10`), none inferred: `add-component-button-human-input`,
  `title-Human Input`, `actionpicker-add-decisions`, `action-add-input`,
  `action-edit-<label>`, `handle-humaninput-shownode-{approve,reject,request changes}-right`.
  Two names in the issue body did not survive the scout and are corrected here: the field
  is **`User Choices`** (`decisions`), not "User Action", and `action-add-input` is the
  inline input the `+` reveals, not the button itself.
- **A two-word custom label is the point.** The label→handle mapping is where the two ends
  of the feature must agree: the backend slugifies it (`_action_id()`: lowercase, spaces →
  underscores) into `branch_request_changes`, while the frontend derives the handle testid
  from the display name and its own `toActionId()` mirror decides whether an existing edge
  still belongs to a renamed choice. A one-word label would pass even if one side dropped
  its normalisation.
- **Counting the handles, not just asserting the expected ones.** `group_outputs: true`
  gives every branch its own handle, so `toHaveCount(2|3)` over
  `[data-testid^="handle-humaninput-shownode-"][data-testid$="-right"]` fails on an extra
  branch as well as a missing one. `-right` excludes the node's single input handle;
  `handle-` excludes the `div-handle-` twin each handle also renders.
- **Persistence is gated on server truth**, not on network silence: `expect.poll` over
  `GET /api/v1/flows/{id}` until the outputs match. `waitForFlowSaveSettled` resolves on
  700 ms of quiet, which also holds while the debounced PATCH has not fired yet (#518).
- **No `page.on("dialog")` handler**, on purpose and commented as such: `FlowPage`
  installs a `beforeunload` that `preventDefault()`s while the store is dirty, and
  Playwright accepts that dialog only while no handler is registered — one here would
  silently cancel the reload the third test depends on (documented in
  `helpers/flows/leave-flow-editor.ts`).
- **Tags**: `@stable @components @ui-ux`, plus `@database` on the persistence test.
  `@regression` is absent deliberately — this is first-time coverage of a new feature, not
  a previously fixed bug. No functional tag maps to HITL today; `@ui-ux` is used because
  every assertion is made against the node's configuration UI.

## 3. Dependencies

- **None — pure UI + REST.** No LLM, no provider key, no `models.json`, no
  `collect-models`. The flow is never run.
- **Parallel-safe**: each test builds its own flow, no named-flow collision, no shared
  state. No `--workers=1` requirement.

- `npm run typecheck` ✅ · `npx eslint <spec>` ✅ (both re-run after rebasing onto
  `origin/main`)
- **3 clean burst runs**, `expected=3 unexpected=0 flaky=0` each, plus a green run after
  the force-fail reverts ✅
- Zero `🚨 Backend Error` on every run ✅
- **Force-fail, executed — one per test, all observed red:**
  - Test 1 — branch-handle count `2 → 3`: proves the count is measured from the live DOM.
  - Test 2 — commit the choice as one word (`RequestChanges`) while the asserts still
    expect `Request Changes`: proves the chip and handle testids derive from the label
    typed, i.e. the label→handle coupling is what is being tested.
  - Test 3 (a) — persistence oracle expecting only `[branch_approve, branch_reject]`:
    proves the poll compares against live API state and is not vacuous.
  - Test 3 (b) — post-reload count `3 → 4`: proves the assertions after `page.reload()`
    are re-read from the rehydrated node, so a reload that dropped a handle fails here.
  - Reverts proven: `grep -c FF-MUTATION` = 0, plus the final green run above.
- `--trace=on` **not** run locally — the known local tracing hang (documented for the
  template family and reproduced on a plain blank-flow spec, #518). Step verification
  rests on the `--retries=0` bursts, the four executed force-fails and the live scout
  instead; CI still captures a trace on first retry.
- **Validation-environment gap, stated rather than papered over**: the local instance is
  `1.12.0.dev10` and the newest nightly tag is `1.12.0.dev15`. `human_input.py` has no
  upstream commit since 2026-07-15 (local Langflow clone at 2026-07-30), and this PR's own
  impacted-specs lane runs the spec against a fresh nightly service container — that lane
  is the cross-check for the 5-build gap.

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -273,7 +273,7 @@
 - [ ] `max_iterations` + `default_route` cycle break (not implementable as a standalone If-Else feedback loop on 1.12.x — Langflow forms graph cycles only via loop-aware target handles (`from_loop_target_handle`, `target_handle.type is None`) that `LoopComponent` ports provide; a feedback edge into the router's regular `match_text` field-input persists in the flow JSON but does not make the graph iterate (`is_cyclic` stays false, router runs once). `conditional_router.py`'s cycle-break only fires when the router already sits inside a Loop-created cycle. Confirmed live on 1.12.0.dev3; product finding filed upstream; #891, follow-up of #822)
 
 #### 3.9 Human Input (HITL, 1.11.0)
-- [ ] Human Input node config: default Approve/Reject branch handles, custom User Action creates a new handle, configured handles persist after save + reload → `core-components/human-input-node-config.spec.ts`
+- [x] Human Input node config: default Approve/Reject branch handles, custom User Action creates a new handle, configured handles persist after save + reload → `core-components/human-input-node-config.spec.ts`
 
 #### 3.10 Data Operations (1.11.0)
 - [ ] Data Operations component: unified JSON/Table/Text operations produce correct outputs per operation mode → `core-components/data-operations-component.spec.ts`
--- a/docs/core-components/human-input-node-config.md
+++ b/docs/core-components/human-input-node-config.md
@@ -13,15 +13,27 @@ Confirms the **configuration surface** of the `Human Input` node shipped with La
 derived from its `User Choices` field**, one handle per choice, and that derivation
 holds on three occasions.
 
-1. **On add** — a freshly dragged node already renders the two default branch handles
-   (`Approve`, `Reject`), before any field is touched. Upstream relies on a hardcoded
-   `outputs` list for exactly this case, because `update_outputs()` only fires on a
-   field change (`lfx/components/flow_controls/human_input.py`).
+1. **On add** — a freshly added node presents **exactly** the two default branch handles
+   (`Approve`, `Reject`), before any field is touched: both present, and no third branch
+   nobody asked for.
+   **Scope caveat, measured rather than assumed:** this does *not* isolate the hardcoded
+   `outputs` list in `lfx/components/flow_controls/human_input.py`. That list carries the
+   comment *"update_outputs only fires on a field change"*, but the frontend fires
+   `POST /api/v1/custom_component/update` **on mount** for any `real_time_refresh` field
+   whose `options` are empty (`hooks/use-fetch-data-on-mount.ts`), and `decisions` is such
+   a field — so the handles would be rebuilt on add even with that list removed. Isolating
+   the fallback needs a different oracle (delay that response with `page.route` and assert
+   the handles are already there); out of scope here.
 2. **On change (live)** — adding a custom choice creates its branch handle **without a
    reload**: the `real_time_refresh` field round-trips through
    `POST /api/v1/custom_component/update` and the node re-renders with the new handle.
-3. **After save + reload** — the configured choices and their handles survive a full
-   page load, rebuilt from the persisted `decisions` by `update_frontend_node()`.
+3. **After save + reload** — the configured choices and their handles survive a full page
+   load. **Two** mechanisms can rebuild them — `update_frontend_node()` on the backend and
+   the frontend's own on-mount round trip — so the spec asserts the **observable** (the
+   handles are there after the reload) without attributing it to either. What it does pin
+   is that the reload actually happened: a `window` sentinel set before it must be gone
+   after, because every DOM assertion in that step is *also* satisfied by the pre-reload
+   page, so a silently-cancelled navigation would otherwise pass it.
 
 The custom choice is deliberately a **two-word label** (`Request Changes`), because the
 label→handle mapping is where the two ends of this feature have to agree: the backend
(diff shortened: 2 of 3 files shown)


Victor-w-Madeira:
## Review response

Two review passes landed on this PR — Copilot's inline comment and an independent adversarial pass against the running nightly (`1.12.0.dev10`). Both are addressed in 6e3db58. Recording the reasoning here, including what I did **not** adopt.

### The persistence test's reload was decorative — fixed

The strongest finding, and it was right: every DOM assertion after `page.reload()` is *also* satisfied by the pre-reload page, so a navigation that silently did not happen passed the whole step. The review proved it two ways — commenting out the reload stayed green, and dismissing the `beforeunload` dialog left a pre-reload `window` marker alive with all eight assertions still passing.

A `window.__reloadSentinel` now bridges the reload and is asserted **gone** after it, which no cancelled navigation can satisfy. Force-failed by commenting out `page.reload()` alone: the test fails on the sentinel (line 236), which is exactly the false-green the review reproduced.

### Not adopted, and why

The same finding proposed PATCHing the persisted `outputs` to a bogus set before the reload, to prove the load-time rebuild. Declined: that tests **rebuild-from-`decisions`**, a separate mechanism — the review itself showed a flow stored with `outputs: [branch_bogus]` renders the three correct handles and gets rewritten server-side. Real behaviour, worth covering, but it is a checklist bullet of its own rather than a widening of §3.9's "handles persist after save + reload".

### Two product claims in the spec doc were wrong — fixed

Both were mine, and the network capture refutes them:

1. The doc said a freshly added node's handles come from the hardcoded `outputs` list *"because `update_outputs()` only fires on a field change"*. That is upstream's own code comment, and it does not hold: the frontend fires `POST /api/v1/custom_component/update` **on mount** for any `real_time_refresh` field with empty `options`, and `decisions` is one — so the handles would be rebuilt on add even with that list removed. Test 1 covers *"exactly the two default branches, no third"*, which is real coverage; it does not isolate the fallback, and the doc now says so explicitly and names the oracle that would (delaying that response via `page.route`).
2. The doc attributed the post-reload rebuild to `update_frontend_node()`. Two mechanisms can do it and the spec never distinguishes them, so it now states the observable instead.

### Copilot's comment — fixed

`getAuthToken` resolves to `""` on an environment without auth (`get-auth-token.ts`, the "no token" fallback), and an empty `Authorization` header is not the same as no header: it overrides the browser context's own auth and would turn the persistence poll into a 401 loop reported as *"the outputs never persisted"*. Guarded the way `setup-blank-flow.ts` does.

### Deliberately unchanged

- **Swallowed `page.goto("/")` in teardown** — `.catch(() => {})` is the house pattern (`singleton-components.spec.ts`); diverging here would be inconsistency without a gain.
- **`CUSTOM_HANDLE` spelled out rather than derived** from `CUSTOM_CHOICE.toLowerCase()` — deriving it the same way the frontend does would make the assertion agree with itself no matter what the app did. The comment now states that.

### Re-validation

2 clean runs at `--retries=0 --workers=1` (3 passed each), typecheck ✅, eslint ✅, zero `🚨 Backend Error`, instance flow count back to its baseline (no orphans). The pre-fix CI run was fully green, including `Run impacted E2E specs` against a fresh nightly service container — which is also what closes the `dev10` → `dev15` gap noted in the PR body.

PR #1286, opened 2026-08-05
test(playground): cover Human Input pause/resume in the Playground (#1189)

### LINKED ISSUE #1189: Create: Human Input pause/resume in the Playground — decision card, exclusive branch routing (§9.6)

Victor-w-Madeira:
Closes #1189.

Closes the `[ ]` gap in `QA-CHECKLIST.md` §9.6 — the **execution** half of Human Input
(HITL, Langflow 1.11.0 — upstream langflow-ai/langflow#13633 durable background execution
+ suspend/resume, #14090 polish; manual recipes HITL-01/02/03). Its sibling #1190
(PR #1284, merged) covers the node's **configuration** surface and explicitly scopes
execution out; this PR never edits the node. No provider credentials are involved:
`route_branch()` returns the prompt text itself.

## 1. Covered tests

| 1 | `approving a Human Input pause routes only the approved branch` | The run **parks**: `human-input-card` renders carrying the prompt, both decisions enabled, and **neither** branch output has emitted — that last part is what separates a pause from a slow completion. Answering **Approve** then produces `chat-message-APPROVED-<prompt>` while `chat-message-REJECTED-<prompt>` stays at count `0` (exclusive routing — the component calls `stop("branch_<action_id>")` on every non-chosen branch), and the card records the answer: the `-reject` button is removed and `-approve` is disabled. |
| 2 | `rejecting a Human Input pause routes only the reject branch` | The mirror. Mirroring is not redundancy: a spec that only ever approves cannot distinguish exclusive routing from *"the approve branch is the only one wired"*. |

## 2. How the test was built

- **Setup**: new fixture `tests/assets/flows/human-input-branching-fixture.json` — Chat Input
  → Human Input (default `Approve`/`Reject`) → one Chat Output per branch — created per test
  via `POST /api/v1/flows/` under a unique name, opened on the canvas, then driven through
  the Playground. Serial describe (named API flows race on the backend's unique-name
  suffixing). Teardown is id-scoped: navigate off the editor, then `deleteFlow` with an
  explicit Bearer. Verified flow-neutral — the instance's user-flow count was identical
  before and after ~14 flows' worth of runs.
- **Wiring both branches is a precondition of the behaviour, not scenery.** With no
  downstream consumer, `_has_downstream_consumer()`
  (`lfx/components/flow_controls/human_input.py`) makes the component **skip the pause
  entirely** — status `"Skipped: no connected outputs"` — because suspending a whole run for
  a decision that routes nowhere would strand it. A "simpler" one-output fixture would never
  pause, so the spec asserts both sinks are on the canvas before running.
- **The two branches carry identical text**, so routing needed a distinguishable sink.
  `route_branch()` returns `Message(text=prompt)` on whichever branch wins; both Chat Outputs
  would emit the same string and "which branch ran" would be invisible in the chat. Each
  Chat Output therefore carries its own `sender_name` (`APPROVED` / `REJECTED`) and display
  name, and the bubble testid is `chat-message-${sender_name}-${text}` — which turns the
  exclusivity claim into an exact locator plus a `toHaveCount(0)` on its counterpart. The
  positive assertion runs **first**, so that zero cannot pass vacuously on a page where
  nothing rendered at all.
- **The fixture is a UI round-trip, not hand-written JSON**, and that was not a preference:
  nodes need `type: "genericNode"` plus their full template to render at all, and an edge
  only attaches when its `sourceHandle` string matches the handle's own `data-handleid`
  **verbatim** (`{œdataTypeœ:œHumanInputœ,œidœ:…,œnameœ:œbranch_approveœ,…}`) — read off the
  live DOM rather than guessed. Upstream's own `src/backend/tests/data/human_input_flow.json`
  is backend-only and renders empty.
- **Live-confirmed selectors**, all harvested on the running nightly (`1.12.0.dev10`):
  `human-input-card`, `human-input-decision-approve` / `-reject` (the `action_id` is the
  slugified label), `playground-btn-flow-io`, `input-chat-playground`, `button-send`,
  `chat-message-${sender_name}-${text}`, `title-Human Input` / `title-Approved Output` /
  `title-Rejected Output`.
- **The prompt is never typed into the Playground input** — it is asserted as *pre-filled*
  from the Chat Input node, because that field re-injects the template default
  asynchronously and races typed text (`authoring-conventions.md`).
- **The transport is deliberately not asserted.** The run is `POST /api/v2/workflows` and
  the answer is `POST /api/v2/workflows/{job_id}/resume`; the user path is the card click,
  and pinning the transport would couple this spec to a mechanism that already moved once
  (v1 build → v2 workflows).
- **Tags**: `@stable @release @playground`. `@release` is claimed here and deliberately not
  on #1190's config spec — if this breaks, a HITL flow cannot be answered at all. No
  `@regression`: first-time coverage of a new feature.

## 3. Dependencies

- **None — pure UI + REST.** No LLM, no provider key, no `models.json`, no `collect-models`.
- **Serial** (`mode: "serial"`) for the unique-name race, not for provider reasons.

- `npm run typecheck` ✅ · `npm run lint` ✅ (0 errors; the 2 `playwright/expect-expect`
  warnings are the standard false positive for helper-held assertions — 35 already exist in
  the repo and the CI gate is errors-only)
- **3 clean burst runs**, `expected=2 unexpected=0 flaky=0` each, plus a green run after the
  force-fail reverts ✅
- Zero `🚨 Backend Error` on every run ✅ · no leaked flows (count back to baseline) ✅
- **Force-fail, executed — one per test, both observed red:**
  - Test 1 — answer by clicking the **counterpart** decision (Reject) while every assertion
    still expects the approved branch: proves the routing assertions track which button was
    pressed, not merely that some output appeared.
  - Test 2 — flip the counterpart-absence assert from `toHaveCount(0)` to `toHaveCount(1)`:
    proves the exclusivity claim is measured against the real DOM (the `APPROVED` bubble
    genuinely never renders) rather than being a vacuous zero.
  - Reverts proven: `grep -c FF-MUTATION` = 0, plus the final green run above.
- `--trace=on` **not** run locally — known local tracing hang (#518/#490). Step verification
  rests on the bursts, the two executed force-fails and the live scout; CI still traces on
  first retry.
- **Validation-environment gap, stated rather than papered over**: the local instance is
  `1.12.0.dev10` and the newest nightly tag is `1.12.0.dev15`. This PR's impacted-specs lane
  runs the spec against a fresh nightly service container — that is the cross-check.

## Timing, measured

Pause appears ~1 s after send; the routed bubble ~1 s after the decision. Per-assertion
budgets are 30 s — generous on purpose, sized for a saturated CI backend rather than for
the local best case.

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -544,7 +544,8 @@
 - [x] DataFrame output renders as Markdown table → `core-functionality/playground/playground-output-data.spec.ts`
 
 #### 9.6 Human-in-the-Loop (1.11.0)
-- [x] Human Input pauses the run durably, decision card renders in the Playground, Approve routes only the approved branch and the run completes; Reject routes only the reject branch → `core-functionality/playground/human-input-pause-resume.spec.ts`
+- [x] Human Input suspends the run server-side, decision card renders in the Playground, Approve routes only the approved branch and the run leaves the suspended state; Reject routes only the reject branch → `core-functionality/playground/human-input-pause-resume.spec.ts`
+- [ ] A suspended Human Input run is recoverable after a page reload (durable execution outliving the tab)
 
 ---
 
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -544,7 +544,7 @@
 - [x] DataFrame output renders as Markdown table → `core-functionality/playground/playground-output-data.spec.ts`
 
 #### 9.6 Human-in-the-Loop (1.11.0)
-- [ ] Human Input pauses the run durably, decision card renders in the Playground, Approve routes only the approved branch and the run completes; Reject routes only the reject branch → `core-functionality/playground/human-input-pause-resume.spec.ts`
+- [x] Human Input pauses the run durably, decision card renders in the Playground, Approve routes only the approved branch and the run completes; Reject routes only the reject branch → `core-functionality/playground/human-input-pause-resume.spec.ts`
 
 ---
 
--- a/docs/core-functionality/playground/human-input-pause-resume.md
+++ b/docs/core-functionality/playground/human-input-pause-resume.md
@@ -43,6 +43,15 @@ outgoing edge, and the component then **skips the pause entirely** — status
 routes nowhere would strand it. So the two Chat Outputs are a **precondition of the
 behaviour under test**, not scenery.
 
+Which is why the setup asserts the **edge count**, not just the node titles: React Flow
+renders nodes independently of edges, so a fixture that lost a branch edge still shows
+`title-Approved Output` and `title-Rejected Output` (measured — both visible with both
+branch edges stripped). Without the edge assertion that defect surfaces 30 s later as
+"the decision card never appeared", i.e. a wiring problem reported as a card-UI failure.
+One nuance worth carrying: on the skip branch the component calls **no** `stop()`, so a
+`successor_map` regression *with* the edges present fires **both** branches rather than
+merely skipping the pause — which is what the exclusivity assertion below catches.
+
 **The two branches carry identical text, so routing needs a distinguishable sink.**
 `route_branch()` returns `Message(text=self._rendered_prompt())` on *whichever* branch
 wins, so both Chat Outputs would emit the same string and "which branch ran" would be
@@ -61,7 +70,9 @@ must have count `0`.
 
 `@release` is claimed here and deliberately **not** on the sibling config spec (#1190):
 this is the execution happy path of 1.11's flagship feature — if it breaks, a HITL flow
-cannot be answered at all — and eight sibling playground specs carry the same set.
+cannot be answered at all. (Measured, since an earlier draft of this doc overstated it:
+exactly **one** other playground spec carries this exact set — `playground-shareable-url` —
+while 11 of 23 carry `@release` alongside further tags.)
 `@regression` is absent: first-time coverage of a new feature, not a previously fixed bug.
 
 ---
@@ -78,9 +89,13 @@ cannot be answered at all — and eight sibling playground specs carry the same 
   renders empty (nodes need `type: "genericNode"` plus their full template) and an edge
   only attaches when its `sourceHandle` string matches the handle's own `data-handleid`
   verbatim (`{œdataTypeœ:œHumanInputœ,œidœ:…,œnameœ:œbranch_approveœ,…}`).
-- The file runs **serial**: each test creates a named flow through the API and those
-  creations race on the backend's unique-name suffixing under parallelism (same reason as
-  the sibling fixture specs).
+- The file is **parallel-safe** and deliberately not serial. Each test owns its flow (unique
+  `${Date.now()}-${random}` name, so the backend's unique-name suffixing has nothing to
+  race) and its own page, and a suspended run is flow-scoped, so the two cannot observe each
+  other. The sibling fixture specs use `mode: "serial"` for that name race; inheriting it
+  here would cost signal, because a flake in the first test **skips** the second and one bad
+  day in the daily would lose the Reject verdict. Measured green at `--workers=2` (and
+  faster: ~7 s against ~10 s serial).
 
 ---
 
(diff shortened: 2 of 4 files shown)


Victor-w-Madeira:
## Review response

Two review passes: Copilot's two inline comments and an independent adversarial pass against the running nightly (`1.12.0.dev10`). Both are addressed in 6236a54. Recording the reasoning, including the one finding I resolved differently from the suggestion.

### The measurement that mattered most — refuted, not fixed

The brief for the independent pass named my own biggest suspicion: that the exclusivity `toHaveCount(0)` is a **timing false-negative** (asserting absence before the counterpart could render). It was **refuted by measurement**: with the fixture rewired so `branch_approve` feeds *both* Chat Outputs — the DOM state of a `stop()` that stopped nothing — the counterpart bubble was present at the instant the routed one became visible in **5/5** runs, including at **6× and 20× CPU throttling**, with a MutationObserver putting the gap at **−21 ms to 0 ms** (counterpart at or *before* the routed one, never after). Both land in one batched render because the resume path invalidates the message query and the list refetches as a unit. The regression is caught 5/5.

### The checklist bullet asserted more than the spec proved — closed by state, not wording

The flipped bullet read *"pauses the run **durably** … and **the run completes**"*. Durability is in this spec doc's own out-of-scope list, and nothing asserted completion. Rather than only softening the text, the spec now asserts the **server side**: `GET /api/v2/workflows/pending?flow_id={id}` must hold exactly **one** suspended request while the card is up, and drop to **zero** after the answer.

That closes a real hole the reviewer identified: the routed bubble arrives through the message-query invalidation, **independent of the run's own event stream**, so a resume that silently never completed passed all six post-decision assertions. `HumanInputCard` also calls `setIsBuilding(true)` before consuming background events, and there is no cheap UI signal for it — `button-stop` never renders at any point in this flow (measured before send, at the pause, at the bubble, +4 s).

Both new assertions were force-failed: expecting `1` pending *after* the answer fails with `Received: 0`; expecting `2` *during* the pause fails with `Received: 1`.

"durably" is gone from the bullet, and the uncovered reload-recovery case is now its own `[ ]` bullet under §9.6 — the checklist should not imply coverage the doc disclaims. What the spec proves is that the run **leaves the suspended state**, not that it reaches a terminal one, and the doc says exactly that.

### The edge guard I advertised did not exist

The two `title-…Output` assertions were commented as protecting against a lost branch edge. Review stripped both edges and proved **both titles stay visible** — React Flow renders nodes independently of edges. The defect surfaced 30 s later as *"the decision card never appeared"*: a wiring problem reported as a card-UI failure. Now an explicit `.react-flow__edge` count of 3, with the comment saying what it actually protects.

### Three factual pointers were wrong

All three corrected in the doc, all three mine:
1. The empty `chat-message-AI-` bubble is **not** the component's `Message(text="")`. It is a host message the frontend synthesizes in `injectHumanInputCard()` (`controllers/API/agui/human-input-card.ts`) purely to carry the card's content block — which means it appears on every pause render, including a replayed one.
2. The resume route lives in `api/v2/workflow.py` (+ `api/v2/hitl.py`), not `workflow_execution.py`. The 409/422 semantics as described were right.
3. "Eight sibling playground specs carry the same tag set" — measured: exactly **one** does (`playground-shareable-url`); 11 of 23 carry `@release` at all.

### Serial mode dropped — measured, not assumed

The reviewer noted the stated rationale (unique-name suffixing race) is weak, since `createFlow` already uses `${Date.now()}-${random}` names, while `mode: "serial"` **skips test 2** when test 1 flakes — losing the Reject verdict on a bad daily. Verified parallel-safe: **4 green runs at `--workers=2`**, and faster (~7 s vs ~10 s).

### Copilot's two

- **Implicit timeouts on the canvas preconditions** — valid: `playwright.config.ts` sets no `expect.timeout`, so those two ran on the **5 s** default while their neighbours had 30 s. Explicit budgets added.
- **`.catch(() => {})` in teardown** — accepted, resolved differently than suggested. Removing the catch (his proposal) would abort the hook **before the deletes** and leak the flow. It now **warns and continues**: the failure is visible, cleanup still runs. He was right that the swallow was silent, and right that my "house pattern" defence was only half true — the spec I mirrored (`split-text-chunking`) has no catch at all; the catch was my addition. I'll carry the same warn-and-continue to #1190's spec as a follow-up rather than smuggling an unrelated file into this PR.

### Also applied

Stale authoring-instance ids (`_frontend_node_flow_id` / `_frontend_node_folder_id`) stripped from the fixture. The post-decision card assertions stay, with their strength now stated honestly in both code and doc: the card sets that state **synchronously from local state** (measured 45 ms after the click, before the request is sent) and keeps it locked on an error, so they pin the UI affordance — never that the backend accepted anything. The `pending` poll is what does that.

### Re-validation

4 green runs at `--workers=2 --retries=0`, typecheck ✅, eslint 0 errors ✅, `check-checklist-guard` ✅, `check:checklist-coverage` ✅, zero `🚨 Backend Error`, instance flow count back to baseline (no orphans). The pre-fix CI run was fully green including `Run impacted E2E specs` against a fresh nightly service container.

PR #1289, opened 2026-08-05
fix(helpers): report a failed teardown unmount instead of discarding it (#1288)

### LINKED ISSUE #1288: The teardown's editor-unmount navigation swallows its own failure in 28 specs and in the shared tracker
Follow-up from the review of #1286 (PR for #1189), where the same objection was raised independently by Copilot and by an adversarial review pass.

## What the pattern is

Every spec that creates a flow leaves the editor before deleting it, because an editor left mounted over a deleted flow keeps polling `GET /flows/{id}/events`, 404s once the flow is gone, and the fixture logs each one as `🚨 Backend Error`. The navigation is written as:

```ts
await page.goto("/").catch(() => {});
```

**Measured on `main`: 28 spec files** carry that exact silent form, and the shared tracker does the same thing with a different target — `track-created-flows.ts` → `cleanup()` does `await page.goto("about:blank").catch(() => {})`.

## Why it is worth a decision rather than a sweep

The swallow is not gratuitous: it protects the **deletes that follow**. If the navigation is allowed to throw, the `afterEach` aborts before `deleteFlow` runs and the flow leaks — strictly worse than the noise the navigation exists to prevent. So "just remove the `.catch()`" (the obvious suggestion, and the one Copilot made) is wrong.

But silent is also wrong, and that is the actual defect: when the unmount fails, the run fills with `🚨 Backend Error` 404 lines from a poll nobody can attribute, and the one line that would explain them was discarded. The advisory error log is only worth reading if what makes it noisy is nameable (the same reasoning as #1084).

## Proposed shape

**Warn and continue** — visible failure, cleanup preserved:

```ts
await page.goto("/").catch((error: unknown) => {
  const message = (error as Error)?.message?.split("\n")[0] ?? String(error);
  console.warn(
    `⚠️  teardown: could not leave the flow editor (${message}) — the deletes ` +
      `below still run, so expect 404 noise from the editor's events poll.`,
  );
});
```

This already ships in `core-functionality/playground/human-input-pause-resume.spec.ts` (#1189), which is where the shape was argued out.

## Scope — deliberately not all 28

1. **`tests/helpers/flows/track-created-flows.ts` → `cleanup()`** — the highest-value single edit. It is the shared cleanup path #1108 built to replace the 51 hand-copied blocks, so every spec migrated onto it inherits the behaviour for free, and its own unit lane can pin it (a fake `page` whose `goto` rejects → the deletes still run **and** the warning is emitted).
2. **`core-components/human-input-node-config.spec.ts`** (#1190) — its sibling #1189 already warns; leaving the pair inconsistent is the kind of divergence that makes a convention unreadable.
3. **A line in `.claude/skills/langflow-e2e/references/authoring-conventions.md`** under *Flow cleanup*, so the next spec is written this way instead of copying whichever neighbour it happened to open.

The remaining 26 copies are **not** in scope: they are being migrated onto the tracker anyway (#1108), a 26-file mechanical sweep is churn with real merge-conflict cost against concurrent spec work, and the pattern is harmless until the navigation actually fails. Migrating one as it is touched is cheaper than a sweep nobody reviews line by line.

## Validation contract

- Unit lane (`npm run test:units`) covers the tracker: `goto` rejects → every captured id still deleted, and the warning emitted. Behavioural force-fail: remove the warn → the assertion on it fails; make the rejection propagate → the delete-count assertion fails.
- One spec run per touched file to confirm nothing else moved.

## Non-goal

Not about *which* URL the teardown navigates to (`/` vs `about:blank` — the tracker deliberately uses `about:blank` so teardown adds no backend traffic of its own). Only about what happens when that navigation fails.

Victor-w-Madeira:
Closes #1288. Follow-up from the review of #1286, where the same objection was raised independently by Copilot and by an adversarial review pass.

**Fixes the diagnosability of the teardown's editor-unmount navigation.**

## Problem

Every spec that creates a flow leaves the editor before deleting it: an editor mounted over
a deleted flow keeps polling `GET /flows/{id}/events`, 404s once the flow is gone, and the
fixture logs each one as `🚨 Backend Error`. That navigation was written:

```ts
await page.goto("/").catch(() => {});
```

Measured on `main`: **28 spec files** carry that silent form, plus the shared tracker
(`track-created-flows.ts` → `cleanup()`, with `about:blank`).

This is a **latent** defect, not an observed failure — and that is stated plainly rather
than dressed up: a rejecting `goto` in the wild is silent *by construction*, which IS the
defect, so there is no failing run to point at. What it costs is attribution: when the
unmount fails, the deletes still run with the editor mounted, the run fills with 404s from a
poll nobody can trace, and the one line that would have named the cause was discarded. The
advisory error log is only worth reading if what makes it noisy is nameable (#1084's rule).

## Fix

**Warn and carry on** — neither of the obvious behaviours is correct:

- **Rethrowing** (the intuitive fix, and the one Copilot proposed) aborts `cleanup()`
  **before** the deletes. `captured` has already been taken out of the tracker by that
  point, so those flows are neither deleted nor tracked: a permanent leak, traded for a
  navigation whose only job was to reduce log noise.
- **Swallowing** discards the only diagnostic.

`cleanup()` also returns it as **`unmountError`**, following `authError` in the same
function — same reason that field exists ("*a 401 that follows is THAT, not a problem with
the flow*"): the consequence arrives later and looks like something else. That is also what
makes the behaviour assertable **structurally**, the way the unit lane already asserts
`authError` without capturing console.

Applied to: the shared tracker, and `core-components/human-input-node-config.spec.ts` —
closing the divergence with its #1189 sibling, which already warned. The convention is now
written down under *Flow cleanup* in `authoring-conventions.md`, so the next spec is not
copied from whichever neighbour it happened to open.

## Scope note

The other **26** spec-local copies are deliberately untouched: they are migrating onto the
tracker anyway (#1108), a 26-file mechanical sweep is churn with real merge-conflict cost
against concurrent spec work, and the pattern is harmless until the navigation actually
fails. The convention says to migrate one as it is touched.

Also unchanged: *which* URL the teardown navigates to (`/` vs `about:blank` — the tracker
uses `about:blank` deliberately, so teardown adds no backend traffic of its own). This PR is
only about what happens when that navigation fails.

- `npm run test:units` — **420 pass / 0 fail** ✅ · `npm run typecheck` ✅ · `npm run lint`
  0 errors ✅ (both re-run after rebasing onto `origin/main`)
- **3 clean runs** of the touched spec, `expected=3 unexpected=0 flaky=0` each, plus a green
  run after the force-fail reverts ✅
- Zero `🚨 Backend Error` ✅ · no leaked flows (instance count back to baseline) ✅
- **Force-fail, executed — five mutations, all observed red:**
  - **Pre-fix, test-first**: the new unit test fails against the unmodified helper at
    *compile* time — `Property 'unmountError' does not exist on type 'FlowCleanupResult'` —
    which is the defect stated precisely: no channel exists to report that failure.
  - Helper A — rethrow instead of warn-and-continue → the delete-count assertion fails,
    proving the deletes are what the warn-and-continue shape protects.
  - Helper B — warn but record nothing → the reporting assertion fails.
  - Touched spec, all three tests re-verified because a cleanup edit can silently defuse a
    sibling's assert (scope = the whole file): branch-handle count `2 → 3`; the custom choice
    committed as one word while the asserts expect two; the persistence oracle expecting 2 of
    the 3 persisted outputs. All red.
  - Reverts proven: `grep -c FF-MUTATION` = 0 across all three files, plus the final green
    run and `420 pass / 0 fail`.
- `--trace=on` not run locally (known local tracing hang, #518/#490); step verification rests
  on the bursts and the executed force-fails.

### DIFF
--- a/.claude/skills/langflow-e2e/references/authoring-conventions.md
+++ b/.claude/skills/langflow-e2e/references/authoring-conventions.md
@@ -339,18 +339,24 @@ failure). Reference: `ollama-provider.spec.ts`, `groq-provider.spec.ts`,
 B** — the bootstrap flow's `POST /flows` competes with the blank-flow click
 and `page.url()` returns the stale bootstrap id (#681).
 
-**The teardown's unmount navigation WARNS and carries on — it neither throws
-nor swallows** (#1288). Both patterns leave the editor before deleting (an
-editor mounted over a deleted flow 404s its `GET /flows/{id}/events` poll and
-the fixture logs each one), and both obvious ways to write that are wrong:
-letting the rejection propagate aborts the hook **before** the delete and leaks
-the flow — strictly worse than the noise the navigation prevents — while
-`.catch(() => {})` discards the one line that would attribute the 404 burst to
-its cause. Log the first line of the error and continue to the deletes.
-`trackCreatedFlows.cleanup()` does this for you and also returns it as
-`unmountError`, alongside `authError`, so a spec on the shared helper needs no
-teardown code of its own. 26 spec-local copies still carry the silent form;
-migrate one as you touch it rather than sweeping them.
+**Use `unmountEditorForCleanup(page, url?)` for the teardown's unmount
+navigation — it neither throws nor swallows** (#1288). Both patterns leave the
+editor before deleting (an editor mounted over a deleted flow 404s its
+`GET /flows/{id}/events` poll and the fixture logs each one), and both obvious
+ways to write that navigation are wrong: letting the failure propagate aborts
+the hook **before** the delete and leaks the flow (measured: 3 leaked flows from
+one spec), while `.catch(() => {})` discards the one line that would attribute
+the 404 burst to its cause. The helper warns with the first line and returns it;
+`trackCreatedFlows.cleanup()` calls it for you and surfaces it as `unmountError`
+alongside `authError`, so a spec on the shared tracker writes no unmount code of
+its own (it still writes the `afterEach` that calls `cleanup`).
+
+**27** spec-local copies still carry the silent form — migrate one as you touch
+it rather than sweeping them. Two of those use `about:blank` rather than `/` and
+hand-roll their own delete loop, so they inherit nothing from the tracker and are
+easy to miss: `api/flows/api-component-regression.spec.ts` and
+`core-functionality/project-management/folder-deletion-integrity.spec.ts` (whose
+local function is *also* named `trackCreatedFlows` and is not this helper).
 
 Validation contract (both patterns): (1) a **behavioral force-fail** of the
 cleanup itself — no-op the delete, run green, count surviving orphans (>0),
--- a/.claude/skills/langflow-e2e/references/authoring-conventions.md
+++ b/.claude/skills/langflow-e2e/references/authoring-conventions.md
@@ -339,6 +339,19 @@ failure). Reference: `ollama-provider.spec.ts`, `groq-provider.spec.ts`,
 B** — the bootstrap flow's `POST /flows` competes with the blank-flow click
 and `page.url()` returns the stale bootstrap id (#681).
 
+**The teardown's unmount navigation WARNS and carries on — it neither throws
+nor swallows** (#1288). Both patterns leave the editor before deleting (an
+editor mounted over a deleted flow 404s its `GET /flows/{id}/events` poll and
+the fixture logs each one), and both obvious ways to write that are wrong:
+letting the rejection propagate aborts the hook **before** the delete and leaks
+the flow — strictly worse than the noise the navigation prevents — while
+`.catch(() => {})` discards the one line that would attribute the 404 burst to
+its cause. Log the first line of the error and continue to the deletes.
+`trackCreatedFlows.cleanup()` does this for you and also returns it as
+`unmountError`, alongside `authError`, so a spec on the shared helper needs no
+teardown code of its own. 26 spec-local copies still carry the silent form;
+migrate one as you touch it rather than sweeping them.
+
 Validation contract (both patterns): (1) a **behavioral force-fail** of the
 cleanup itself — no-op the delete, run green, count surviving orphans (>0),
 revert, count again (0); (2) checking the instance's user-flow count after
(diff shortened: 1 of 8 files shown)


Victor-w-Madeira:
## CI verdict: `ambient-red` — one failure, cause outside this PR, now tracked as #1290

`Run impacted E2E specs` is red: **1 failed, 26 passed**. The failure is
`core-components/general-bugs-save-changes-on-node.spec.ts:73` — *"any changes on the node must be saved on user interaction"* (`@release @components`) — failing all three attempts on:

```
TimeoutError: locator.waitFor: Timeout 20000ms exceeded.
  - waiting for getByTestId('input_outputText Output') to be visible
```

### Why this is not this PR's change

- **Control experiment.** The spec fails **2/2** locally on this branch, and **2/2** with the `origin/main` version of the helper this PR touches — same failure, same locator. The cause is on `main`.
- **Wrong layer.** The failing assertion is in the test **body**; this PR only changes teardown navigation (plus a new helper and its test).
- **The spec was selected because it imports the touched helper.** It picked up `trackCreatedFlows` in #1154, so `impacted-specs-by-import.mjs` correctly pulls it in on any change to that file. That is #1054 working — it surfaced latent rot, it did not cause it.
- **Both specs this PR actually edits were in the selection and passed**: `human-input-node-config.spec.ts` and `human-input-pause-resume.spec.ts`.

### Root cause, verified live

`TextOutput` and `TextInput` are **`legacy: true`** on the running nightly (`1.12.0.dev10`, `GET /api/v1/all` → `input_output`). Legacy components are hidden from the sidebar unless the legacy switch is flipped, so the testid never renders. This is the tail of the Text I/O migration closed by #362 (the 10 `@stable` tests) and #367 (blast radius) — two non-`@stable` files were left behind, and **neither is run by any scheduled lane**, which is why `reports/daily-history.jsonl` has zero occurrences of either. Full triage, both files, and the two candidate fixes: **#1290**.

### On the earlier red

The previous run on this branch also carried the wedged-backend signature (`/api/v1/auto_login` timeout ×3, #1077) plus 2 flaky. This run has **none of that** — no wedge, no flaky, 26 green — so the wedge was incidental noise and the deterministic Text Output failure stands on its own. The `🚨 Backend Error: 500 … /api/v1/flows/` lines are the documented advisory `database is locked` noise on concurrent creation (~9%), which never fails a test.

The red is a pre-existing, product-changed breakage in a spec this PR does not touch and does not worsen; it is filed with a reproduction and two candidate fixes; and everything this PR does change is green — TypeScript Check (which runs the unit lane, 425/425 including the new test that pins the warning a review had deleted while the lane stayed green), ESLint, the QA-CHECKLIST guard, and both edited specs. Blocking on #1290 would hold a fix behind an unrelated defect it exposed.

Issue #1153, opened 2026-07-30
SaveChangesModal can deadlock the editor exit — button-less spinner survives blocker.proceed()


Victor-w-Madeira:
## Title

`SaveChangesModal` can deadlock the editor exit — button-less spinner survives `blocker.proceed()`

## Body

Spun out of #1005, which set out to explain that spec's residual ~7% flake and
found **four** distinct signatures rather than the two it filed. Three of them
traced to the templates-modal entry race and are fixed there. This is the fourth,
and it is **not** a test defect — masking it inside the spec would hide a product
defect, so it is filed on its own.

**Upstream:** _not filed yet_

## Symptom

`edit-flow-name.spec.ts` clicks `icon-ChevronLeft` to return to the home listing
and never gets there:

```
Error: expect(locator).toBeVisible() failed
Locator: getByTestId('home-dropdown-menu').first()
Expected: visible
Timeout: 30000ms
Error: element(s) not found
   at edit-flow-name.spec.ts:54
```

The failure screenshot shows the editor still on the canvas behind a modal:

> **Flow has unsaved changes** · ⟳ *Saving your changes…*

…while the success toast **"Changes saved successfully"** is visible in the
bottom-left corner of the same frame.

Measured 2 in 24 runs at `--workers=4` on a clean `1.12.0.dev10` instance
(0 in a following 36-run burst — intermittent, load-dependent).

## Mechanism (read in the 1.12 frontend source)

`modals/saveChangesModal/index.tsx` — in autosave mode the modal renders with
**no actions at all**:

```tsx
cancelText={autoSave ? undefined : t("flow.exitAnyway")}
confirmationText={autoSave ? undefined : t("flow.saveAndExit")}
onConfirm={autoSave ? undefined : …}
loading={autoSave ? true : saving}
```

so it is a permanent spinner whose only exit is whatever dismisses it from the
outside. That dismissal is `pages/FlowPage/index.tsx` → `handleSave`:

```ts
const handleSave = () => {
  let saving = true;
  let proceed = false;
  setTimeout(() => {
    saving = false;
    if (proceed) { blocker.proceed && blocker.proceed(); setSuccessData(…) }
  }, 1200);
  saveFlow().then(() => {
    if (!autoSaving || saving === false) { blocker.proceed && blocker.proceed(); setSuccessData(…) }
    proceed = true;
  });
};
```

**CORRECTION (2026-07-30).** This issue originally claimed the screenshot's
[...]

Issue #1054, opened 2026-07-29
pr-validation selects no E2E specs when a shared helper or Page Object changes


Victor-w-Madeira:
## Problem

`pr-validation.yml`'s `detect-specs` job selects the E2E work from **changed
spec files only**:

```bash
SPECS=$(git diff --name-only --diff-filter=d "origin/$BASE_REF"...HEAD -- ':(glob)tests/**/*.spec.ts')
```

Nothing maps a changed **helper**, **Page Object** or **fixture** to the specs
that import it. `scripts/impacted-tests.ts` does do path→spec mapping, but its
input is *Langflow source paths* resolved through the `External dependencies`
section of the spec docs — it exists for `file-watcher.yml`, and has no
`tests/helpers/**` → dependent-spec edge.

Consequence: the highest-reach change this repo can receive passes PR validation
**without executing a single spec**.

## Evidence

PR #1052 (fixing #997) changed `tests/helpers/ui/adjust-screen-view.ts`, a
helper that **61 spec files call directly** and another **47 reach indirectly**
through `setupPlayground`, `initialGPTsetup`, `prompt-template` and
`SimpleAgentTemplatePage` — **108 spec files**, roughly half the suite. `Run
impacted E2E specs` reported `skipping`. The PR was covered by hand instead,
with a `manual.yml` dispatch on the branch
([30420649707](https://github.com/oriontech-me/langflow-e2e/actions/runs/30420649707)).

The indirect half is the part a reviewer cannot eyeball: it lands mostly in
`llm-agents/` and `playground/`, where a canvas-interceptor failure reads as an
agent or provider problem — the misattribution class of #576.

## Scope

Teach the PR gate to select specs for changes under the suite's own shared
surfaces, not just for changed specs:

- `tests/helpers/**`
- `tests/pages/**`
- `tests/fixtures/**` — every spec imports the fixture, so this is a full-suite
  signal rather than a per-spec mapping
- `playwright.config.ts` — same, and today equally invisible

Resolution has to be **transitive**: helper A imported by helper B imported by
spec C must select C. A direct-importers-only pass would have selected 61 of
#1052's 108 and called it covered — worse than the current honest zero, because
it reads as coverage.

Bounding matters as much as selection. A fixture or config change resolves to
the whole suite, which is a daily-sized run on a PR. Options to decide between:
cap the selection and say what was dropped (never silently — the silent-cap rule
[...]

PR #1291, opened 2026-08-05
fix(specs): migrate the last two specs off the legacy Text Output (#1290)

### LINKED ISSUE #1290: Two specs still add the legacy Text Output from the sidebar and fail on 1.12 — invisible because neither is @stable
Found while triaging a red `Run impacted E2E specs` lane on PR #1289 (issue #1288). The red was **not** that PR's change — proven by a control experiment — and the real cause is a spec breakage that predates it and that no scheduled lane runs.

## Verdict: product-changed, pre-existing, and unobserved

`TextOutput` and `TextInput` are **`legacy: true`** on the running nightly (verified live on `1.12.0.dev10` via `GET /api/v1/all` → `input_output`). Legacy components are hidden from the sidebar unless the legacy switch is on, so any spec that adds them without flipping that switch waits forever on a component that is not there.

This is the tail of the migration that #362 (the 10 `@stable` tests) and #367 (blast radius) closed: those covered the `@stable` population, and these two files were left behind.

## The two specs, measured

| Spec | Tags | Adds Text I/O | Enables the legacy switch | Result |
|---|---|---|---|---|
| `core-components/general-bugs-save-changes-on-node.spec.ts:73` | `@release @components` | yes | **no** | **fails 2/2** — `locator.waitFor` 20 s on `getByTestId('input_outputText Output')` |
| `ui-ux/general-bugs-minimize-state-error.spec.ts` | not `@stable` | yes | **no** | **fails 1/1** — `locator.hover` 20 s on the same testid |

For contrast, `flow-functionality/general-bugs-reset-flow-run.spec.ts` adds Text I/O too **and calls `addLegacyComponents(page)` first** — it is the correctly-migrated shape, and it passes. So the fix pattern already exists in the repo.

## Why nobody saw it

Neither spec is `@stable`, and `daily-stable.yml` runs only `@stable` — so **no scheduled lane executes either file** (the same class as #1084's "a test without `@stable` runs nowhere"). Consistent with the evidence: `reports/daily-history.jsonl` has **zero** occurrences of either spec in `failures` or `flaky`, across every recorded daily.

What surfaced it was the **import graph**: both import `helpers/flows/track-created-flows.ts` (they got the tracker in #1154), so a PR touching that helper selects them via `impacted-specs-by-import.mjs`. That is #1054 working as designed — it exposed latent rot in specs nothing else runs.

## Evidence that it is not PR #1289's change

- Fails **2/2** locally on that branch.
- Fails **2/2** with the `origin/main` version of the touched helper (control experiment) — same failure, same locator.
- The failing assertion is in the test **body**, while #1289 only changes teardown navigation.

## Proposed fix — decision needed on which of two

1. **Migrate off legacy** (what #367 did elsewhere): replace Text Output with Chat Output. Keeps the specs on the components a user actually sees, and removes the dependency on a switch. Changes what each spec exercises slightly, so each needs its assertions re-read.
2. **Flip the legacy switch** (`addLegacyComponents(page)`, as `general-bugs-reset-flow-run.spec.ts` does). Minimal diff, preserves the exact scenario — but keeps coverage pointed at components upstream has deprecated, and it will break again when they are removed.

Recommend **(1)** for `general-bugs-save-changes-on-node` (its subject is node-level save behaviour, which any component exhibits — Text Output is incidental) and **(2)** only if a spec's scenario genuinely depends on the legacy component.

## Worth deciding alongside

Both files are `@release` / untagged but not `@stable`, which is why they rotted silently. Whether they should be promoted to `@stable` after the fix — or deleted if the behaviour is covered elsewhere — is the more useful question than the fix itself. A spec that no lane runs is a spec that cannot catch a regression.

Related: #362, #367 (the Text I/O legacy migration), #1154 (why these two carry the tracker), #1288 / #1289 (where this was found).

Victor-w-Madeira:
Closes #1290.

## Problem

`TextOutput` and `TextInput` are **`legacy: true`** on 1.12 (verified live via
`GET /api/v1/all` → `input_output`) and hidden from the sidebar, so two specs waited 20 s on a
testid that never renders. `repro-run` measured the baseline at **5/5 failures, 0 voided** —
deterministic, not a flake. This is the tail of the migration #362 (the 10 `@stable` tests) and
#367 closed: they covered the `@stable` population and these two files were left behind.

**Why it was invisible rather than merely broken:** neither file was `@stable`, and
`daily-stable.yml` selects `@stable`, so **no scheduled lane ran either one** —
`reports/daily-history.jsonl` has zero occurrences of both. What surfaced it was the import
graph: both import `track-created-flows.ts`, so a PR touching that helper (#1289) selected
them. That is #1054 working as designed.

## `general-bugs-save-changes-on-node` — migrated, hardened, promoted to `@stable`

**Chat Input replaces Text Output.** Its `input_value` is the same `MultilineInput`
(`multiline: true`), so it renders the *same* `textarea_str_input_value` the assertions already
targeted — the mechanism under test is unchanged, only the host node is one a user can still
add. (The issue suggested Chat Output; that would have failed — `ChatOutput.input_value` is a
`HandleInput` with no textarea.) Chat Input also ships `minimized = True`, so its fields are in
the DOM but **hidden**, which is why the setup now expands the node and seeds the
assistant-onboarding flag `expandFocusedNode` requires.

**The promotion is what pulled the rest in, and reviewing it found the oracle was a race.**
Without a server-side gate, whether the value was persisted by the debounced autosave
(fill + ~1015 ms) or by the exit's own `FlowPage.handleSave` through the #1153 unsaved-changes
blocker depended on who won ~300 ms:

| Measurement | Winner |
|---|---|
| The review's trace of the pre-gate spec | the **exit**, on 4 of 4 exits (blocker rendered each time) |
| Re-measuring the same shape here | the **autosave**, on 4 of 4 exits (verdict `left`, no blocker) |

Both are consistent with a race, and that *is* the defect: either way the spec could pass with
the debounced autosave broken, so the `@stable` promotion would have bought zero
autosave-regression coverage. It now polls `GET /api/v1/flows/{id}` for the persisted
`template.input_value.value` **before** leaving — the value is on the server before the exit is
clicked, so whatever the exit does cannot carry the assertion.

**The exit verdict is asserted too, claiming less than it looks like.** `classifyEditorExit`
polls every 200 ms, so a blocker that *flashes* reads as `left` — proven by mutation: removing
the gate still returned `left` 4/4. What it catches is a blocker that **lingers**, which is the
flake-relevant shape (the one that burns #1153's 15 s grace and then throws). Both the code
comment and the doc say exactly that, rather than implying the stronger guarantee.

Also hardened, all from the promotion audit:
- `leaveFlowEditor(page)` instead of a raw `icon-ChevronLeft` click (`escapeDeadlock` off on
  purpose — its recovery discards the unsaved state this test asserts);
- re-entry anchored on the flow **id** via `list-card-open-button`, not
  `getByText(name).first()` (the home list sorts by `updated_at`, so position 0 belongs to
  whichever worker touched a flow last);
- the flow id resolved by **set difference** against the ids captured before the `blank-flow`
  click — the first captured id is the bootstrap's, the trap #490/#681 paid twice;
- a one-flow guard that prints `flows.failedCreations()`, so a `POST /flows` 500 ("An internal
  error occurred while creating the flow", which this instance does emit under load) is not
  read as a tracker bug;
- both `waitForTimeout(500)` sleeps gone.

## `general-bugs-minimize-state-error` — deleted, its one distinct angle moved

`ui-ux/minimize.spec.ts` (`@stable`) already covers minimize/expand with a non-legacy
component, asserting hidden handles, geometry **and** the persisted `data.showNode`; the deleted
file only counted `hide-node-content`. Its distinct contribution was **repetition** — the
original bug appeared after several cycles, never the first — so that moved into
`minimize.spec.ts` as four further cycles (**five** in total with the one already there, the
deleted file's own loop count), where it also runs for the first time, that file being
`@stable`. Moving rather than migrating follows the no-duplication rule.

Its `Toolset`-absence check did not come along:
`general-bugs-component-as-tool-shortcut.spec.ts` asserts it absent **and then present** after
enabling tool mode, which is strictly stronger than a bare absence check.

Moving the step in raised the ⋮ menu's open count from 2 to 10 per run, so `openNodeMenu` now
**dispatches** the pointer events instead of clicking: a real coordinate click on that trigger
is hit-tested by the ReactFlow pane and DESELECTS the node, unmounting the toolbar as the menu
opens (#867, reproduced at `--workers=1` with no backend load) — which is why
`helpers/ui/expand-focused-node.ts` already drives the same button that way. Each half-cycle is
a retried `toPass` unit, so a lost selection retries instead of timing out on actionability with
no attribution.

The deleted file also **leaked a flow per run** (no tracker, no `afterEach`). The two
`New Flow (n)` orphans my own runs of it left behind were purged; the instance is back at its
baseline.

- Pre-fix baseline: **5/5 failures, 0 voided** (`repro-run`).
- **3 clean bursts per spec** in the pipeline's VALIDATE, plus **2 more per spec** after the
  review fixes, plus the green run after the force-fail reverts. `expected=1 unexpected=0
  flaky=0 backendErrors=false` on every one.
- `npm run typecheck` ✅ · `npm run lint` 0 errors ✅ · `check-checklist-guard` ✅ ·
  `check:checklist-coverage` ✅ (a scratch `coverage:summary` confirms the regenerator picks up
  the new `@stable` test and removes nothing for the deleted non-`@stable` file — not committed,
  per the guard).
- **Zero leaked flows** — instance flow count identical before and after.
- **Force-fails, executed:**
  - read-back expecting `value + "-ff"`, a string never typed → red (the round trip compares
    what persisted against what was filled);
  - the autosave poll reading a value the server never had → red (the gate is real);
  - the expand click dropped from inside the repeated cycle → red (the moved repetition
    re-checks both states each cycle);
  - and the gate-removal mutation, which **passed** — reported here rather than buried, because
    it is what proves the verdict assertion does not catch a flashing blocker. The comment and
    doc were reworded to that narrower claim instead of keeping the stronger one.
- Timing: save-changes 17 s → 24 s (the four autosave waits are the cost of testing the real
  mechanism); minimize 12 s → 14 s.


Victor-w-Madeira:
## Copilot review — accepted, with one correction

**Accepted:** `readPersistedInputValue` resolved a token on every `expect.poll` invocation. The request volume is the smaller half of the problem — with `intervals: [500, 1000, 2000]` across four edits that is dozens of extra `/api/v1/auto_login` calls. The half that matters is **attribution**: `getAuthToken` carries a `[2000, 8000, 20000]` retry budget, so a single hiccup inside the poll could block for up to **30 s — longer than the poll's own 20 s timeout** — and the failure would then read as *"the autosave never persisted"* when the real cause was auth.

That is the same misattribution class this PR already fixes twice over (the creation-500 guard that prints `failedCreations()`, the #1153 exit verdict), so it does not get to stay in the one assertion the `@stable` promotion rests on. Fixed in d3a9a11: the bearer is resolved **once per test** and passed down.

**Force-failed:** passing an empty bearer turns the poll red — so the header is genuinely what authenticates the read, rather than the browser context carrying it silently.

**One detail of the review does not hold:** it reports the same call "on line 94", but there is only one `getAuthToken` in the file — line 94 is the poll's *call site*, not a second resolution. Nothing to change there.

Re-validated: 2 clean runs (25 s, 24 s), typecheck and lint clean.

---

For the record, the CI evidence that matters on this PR: the previous run's `Run impacted E2E specs` selected **exactly the two specs this PR changes** and passed — the same lane that was red on #1289 because of `general-bugs-save-changes-on-node`, now green against a fresh nightly.

Issue #772, opened 2026-07-15
OpenAI API key out of quota — OpenAI-bound @stable specs hard-fail every daily


Victor-w-Madeira:
## OpenAI API key out of quota — OpenAI-bound `@stable` specs hard-fail every daily

Spun out of the #771 daily-failure triage (run [29406577514](https://github.com/oriontech-me/langflow-e2e/actions/runs/29406577514), 2026-07-15). The mass-failure guard tripped (15 hard failures), so `@stable` was left in place; deeper investigation isolated a **durable** driver that will keep failing every daily until resolved.

### Finding (confirmed live)

The `OPENAI_API_KEY` used by CI **and** local `.env` (OpenAI project `proj_vtkyNBz9ytVmv1HpewDvfUaL`) is **out of quota / billing**:

- Direct OpenAI API call returns: `You exceeded your current quota, please check your plan and billing details.` (HTTP 429 on every chat completion; the account can still *list* models — 24 — but cannot *generate*).
- `collect-models.spec.ts` reproduces live on `1.11.0.dev38`: `openai` probes **inactive** (all 49 candidate models fail validation), while `anthropic` (claude-sonnet-5) and `google` (gemini-2.5-flash) probe **active**. The `does not have access to model o3-deep-research` line in the error is a red herring — that is only the error for the last model tried; the real blocker for the accessible models (gpt-4o-mini, gpt-4o, …) is the quota 429.
- The CI run's error carries the **same project id** (`proj_vtkyNBz9ytVmv1HpewDvfUaL`) as the local key → CI and local share the exhausted key.

This is **not** a transient infra blip and **not** test rot — it is an external account-billing condition. It will not self-heal on the next daily.

### Impact

Two distinct effects:

1. **Hard failures** on OpenAI-only `@stable` specs (they run a real completion, so they fail rather than skip):
   - `language-model-regression.spec.ts` — "language model must respond with OpenAI provider"
   - `openai-provider.spec.ts:203` — "configured OpenAI selects a GPT model in the Agent and executes the flow"
   - `general-bugs-agent-images-playground.spec.ts` — "user must be able to send images in the playground with the agent component"
   - (`loop-component-regression.spec.ts` "Research Translation Loop" also fails here, but is tracked separately by #722 — even that fix cannot pass while OpenAI has no quota.)
2. **Silent skips** — every provider-parametrized spec skips its `openai` case (collect-models marks the provider inactive), masking OpenAI coverage across the suite.

[...]

PR #1292, opened 2026-08-05
test(core-components): cover the unified Data Operations component (#1191)

### LINKED ISSUE #1191: Create: Data Operations unified component — per-operation outputs (§3.10)

**Backlog pointer:** QA-CHECKLIST.md §3.10 → `core-components/data-operations-component.spec.ts`

**Scope**
- 1.11.0 unified JSON/Table/Text operations into a single **Data Operations** component.
- Validate correct outputs per operation mode; no LLM key needed.

**Upstream:** langflow-ai/langflow#13743 (unify), #14025 (rename to Data Operations).

Victor-w-Madeira:
Closes #1191.

Closes the `[ ] Data Operations component: unified JSON/Table/Text operations produce correct outputs per operation mode` gap in `QA-CHECKLIST.md` §3.10.

Langflow 1.11.0 merged the three separate operation components — **JSON Operations** (`DataOperations`), **Table Operations** (`DataFrameOperations`) and **Text Operations** (`TextOperations`) — into a single **Data Operations** node (`lfx/components/processing/operations.py`, class `OperationsComponent`, `name="Operations"`); upstream `langflow-ai/langflow#13743` unified and `#14025` named it. The three originals stayed in the sidebar as `legacy: true` with `replacement = ["processing.Operations"]`.

No sibling spec touches this surface: §3.10's *other* bullet (legacy components linking/redirecting to Data Operations, legacy flows still building) is separate coverage and is **not** touched here.

## 1. Covered tests

| 1 | `Data Operations Text mode returns the Case Conversion result as a Message` | Input Type `Text` + `Case Conversion (uppercase)` on `data ops probe abc123` → the node advertises a **Message** output (`output-inspection-message-operations`) and the inspector textarea equals `DATA OPS PROBE ABC123` exactly. Catches a regression in `as_message` / `_case_conversion` or in the Text tab's default output routing. |
| 2 | `Data Operations Word Count switches the Text-mode output to JSON and counts the text` | With the tab still on `Text`, choosing `Word Count` re-advertises the output as **JSON** — `output-inspection-json-operations` exists and `output-inspection-message-operations` has count 0 — and the payload is `word_count: 6`, `unique_words: 4`, `character_count: 30`. This is the most distinctive consequence of the merge: the **operation**, not only the input type, decides the output (`update_outputs`). |
| 3 | `Data Operations JSON mode selects a single key from an upstream JSON output` | Switching to `JSON` re-filters the operation picker (`Select Keys` offered, `Case Conversion` gone — `update_build_config` swapping `OPERATIONS_BY_TYPE`), and `Select Keys ["word_count"]` over an upstream Word Count payload returns `{"word_count": 6}` **without** `unique_words`. The negative half separates "Select Keys ran" from "the payload passed through". |
| 4 | `Data Operations Table mode filters the rows of an upstream Table output` | Switching to `Table` re-filters the picker (`Filter` offered, `Case Conversion` gone), and `Filter: score greater than 15` over an upstream `Text to DataFrame` table returns exactly the cells `beta, 30, gamma, 20`. Also covers the second output override (`Text to DataFrame` → Table). |

## 2. How the test was built

- **Setup:** blank flow created via the REST API (`createFlow`, unique name → parallel-safe), then opened at `/flow/{id}` and populated from the sidebar. Going through the API means the id is known without sniffing the creation response — the canvas URL id is transient on this version.
- **Expected values are derived from the component source, not observed.** `_word_count` (`len(split())` / `len(set(split()))` / `len(text)`) → 6 / 4 / 30; `select_keys` → single-key dict; `filter_rows_by_value` with `greater than` compares numerically after `_convert_numeric_columns` → 10, 30, 20 > 15 keeps beta and gamma. The Word Count sentinel **repeats** `data` and `ops` on purpose so `unique_words` (4) ≠ `word_count` (6), and the table's surviving rows are deliberately **not** the first N, so a `head`-like regression cannot look like a pass.
- **Live-confirmed selectors** (scouted in the running DOM, none invented): `add-component-button-data-operations`, `tab_0_text` / `tab_1_json` / `tab_2_table`, `textarea_str_text_input`, `button_open_text_area_modal_textarea_str_text_input` → `text-area-modal` → `genericModalBtnSave`, `button_open_list_selection_sortablelist_sortablelist_operation`, `list_item_{case_conversion,word_count,text_to_dataframe,select_keys,filter}`, `value-dropdown-dropdown_str_case_type`, `inputlist_str_select_keys_input_0`, `popover-anchor-input-{column_name,filter_value}`, `value-dropdown-dropdown_str_filter_operator`, `handle-operations-shownode-{text,json,table,message}-{left,right}`, `button_run_data operations`, `node_duration_data operations`, `output-inspection-{message,json,table}-operations`, `btn-close-modal`.
- **Chained flows for JSON and Table.** Both inputs are connection-only, so tests 3 and 4 wire a second Data Operations node — which also proves the unified component's own output plugs into its own typed input. The two nodes share every testid (same component type), so once the downstream node is switched they are addressed by an **identity-carrying child** (`textarea_str_text_input` for the Text-mode producer; `handle-operations-shownode-json-left` / `table-left` for the consumer), each asserted to resolve to exactly **one** node — never by DOM order.
- **Two mechanics that took a debug cycle, both fixed rather than worked around:**
  - `adjustScreenView` starts with `fit_view`, which on two **stacked** nodes zooms *in* to fill the canvas and leaves each node ~415 px tall — wider than `separateOverlappingNodes`' 220 px default step, so the separation poll could never converge. Order is now zoom-out → separate (step 300) → fit.
  - The lower node's run control and output-inspector button sit under `main_canvas_controls` (and, right after a build, under the "built successfully" toast), which intercepts hit-tested clicks. Both are dispatched at the event level, each guarded by the assertion that follows (`node_duration_*` / the modal becoming visible), so a no-op dispatch fails loudly — the same fix and safeguard already documented in `if-else-component-regression.spec.ts`.
- **Assertion surface:** the node's own output inspector, whose testid carries the output's display name (`output-inspection-<title>-<type>`). Asking for `output-inspection-json-operations` *is* an assertion that `update_outputs` produced a JSON output — the testid cannot exist otherwise. Message → `textarea` value; JSON → the dialog's text; Table → `role="gridcell"` contents.
- **Cleanup:** each test creates exactly one flow whose id is known; `afterEach` leaves the editor via `unmountEditorForCleanup` (so the events poll stops 404ing against a flow being deleted, #1288) and deletes only that id with an explicit bearer. Verified after the runs: **zero orphans** — the instance holds exactly the flows it held before.

## 3. Dependencies

- **None — pure UI/Python.** No model provider, no API key, no `collect-models`, no `models.json`, no external network. Every operation is string/dict/pandas code.
- **Parallel-safe:** flow names are unique per test and cleanup is id-scoped; no `--workers=1` needed.
- Data Operations is the **non-legacy** component and is in the default sidebar — the Legacy toggle is not used.

## Validation (nightly 1.12.0.dev10, `--retries=0`)

- typecheck ✅ · eslint ✅
- **3/3 clean bursts** at `--retries=0 --workers=1` (`expected=4, unexpected=0, flaky=0` each), plus a 4th green run after the force-fail reverts ✅
- zero `🚨 Backend Error` on every run ✅
- **force-fail executed for all 4 tests**, each isolated by `--grep`, each marked `// FF-MUTATION` and reverted (final marker count in the diff: 0):
  - #1 — expect the untouched text instead of the uppercased result → failed ✅
  - #2 — expect `"unique_words": 6` (equal to `word_count`) → failed ✅
  - #3 — flip the negative half and require `unique_words` to be present → failed ✅
  - #4 — expect the full unfiltered grid with `alpha/10` → failed ✅
- flow cleanup verified via `GET /api/v1/flows/`: zero orphans ✅

> **Nightly note:** validated on `1.12.0.dev10`; the current nightly tag is `1.12.0.dev17`. The local instance was deliberately **not** rebuilt — it runs without a volume, so recreating it would destroy its SQLite state (flows and imported provider global variables). Nothing in the diff depends on behaviour that moved between those builds, and `daily-stable.yml` runs against the newest nightly.
>
> `--trace=on` was not run: it is a known local hang on this stack (see the langflow-e2e-issues skill). Step verification is covered by the clean `--retries=0` bursts, the executed force-fails, and the live DOM scouting the selectors came from.

Generated blocks in `QA-CHECKLIST.md` were **not** regenerated — only the §3.10 bullet was edited (they regenerate on merge, #741).

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -277,7 +277,7 @@
 - [x] Human Input node config: default Approve/Reject branch handles, custom User Action creates a new handle, configured handles persist after save + reload → `core-components/human-input-node-config.spec.ts`
 
 #### 3.10 Data Operations (1.11.0)
-- [ ] Data Operations component: unified JSON/Table/Text operations produce correct outputs per operation mode → `core-components/data-operations-component.spec.ts`
+- [x] Data Operations component: unified JSON/Table/Text operations produce correct outputs per operation mode (Text→Message, Word Count→JSON override, JSON Select Keys, Table Filter) → `core-components/data-operations-component.spec.ts`
 - [ ] Legacy operations components link/redirect to Data Operations (legacy flows keep working) → `core-components/data-operations-legacy-link.spec.ts`
 
 ---
--- a/docs/core-components/data-operations-component.md
+++ b/docs/core-components/data-operations-component.md
@@ -53,7 +53,7 @@ pandas); the spec neither resolves a model nor consumes `models.json`.
 ### Why the assertion surface is the output inspector
 
 Each test reads the value the component actually returned, from the node's output inspector
-(`output-inspection-<output>-operations`), not from a downstream sink:
+(`output-inspection-<output display name>-operations`), not from a downstream sink:
 
 - `Message` output → the inspector renders a `textarea`; assert `toHaveValue(...)`.
 - `JSON` output → the inspector renders a JSON viewer; assert the dialog's text contains the
@@ -61,11 +61,17 @@ Each test reads the value the component actually returned, from the node's outpu
   is absent).
 - `Table` output → the inspector renders an ag-grid; assert the `role="gridcell"` contents.
 
-The inspector testid itself carries the output's display name
-(`output-inspection-${title.toLowerCase()}-${type.toLowerCase()}`, from
-`CustomNodes/GenericNode/components/NodeOutputfield/index.tsx`), so asserting on
-`output-inspection-json-operations` *is* an assertion that the node advertises a JSON output
-— the testid cannot exist unless `update_outputs` produced that output.
+The inspector testid itself carries the output's display name **first**, the component type
+second — `output-inspection-${title.toLowerCase()}-${classNameFromType(id).toLowerCase()}`,
+from `CustomNodes/GenericNode/components/NodeOutputfield/index.tsx`. So in
+`output-inspection-json-operations`, `json` is the **output's display name** (what
+`update_outputs` advertised) and `operations` is the **component type** (`name="Operations"`,
+identical on every node of this component). Reading it the other way round — component first
+— is the easy mistake here, precisely because both halves are lowercase nouns and the
+component is *called* Data Operations: it would suggest the testid is stable per node, when
+in fact its first half is exactly the thing under test. Asserting on
+`output-inspection-json-operations` *is* therefore an assertion that the node advertises a
+JSON output — the testid cannot exist unless `update_outputs` produced that output.
 
 ---
 
@@ -242,7 +248,9 @@ by an empty output, a passthrough, or a node that failed to build (the run gate 
   emits `button_open_list_selection_sortablelist_sortablelist_<field>` and the
   `list_item_<snake_case_name>` options.
 - `src/frontend/src/CustomNodes/GenericNode/components/NodeOutputfield/index.tsx` — emits
-  `output-inspection-<output>-<type>`, the spec's per-operation output-type assertion.
+  `output-inspection-<output display name>-<component type>` — output first, component
+  second (e.g. `output-inspection-json-operations`). This is the spec's per-operation
+  output-type assertion.
 - `tests/helpers/flows/create-flow.ts`, `delete-flow.ts`,
   `add-component-from-sidebar.ts`, `unmount-editor-for-cleanup.ts`.
 - `tests/helpers/ui/zoom-out.ts`, `adjust-screen-view.ts`, `separate-overlapping-nodes.ts`,
--- a/docs/core-components/data-operations-component.md
+++ b/docs/core-components/data-operations-component.md
@@ -1,0 +1,307 @@
+# Spec: Data Operations — unified component, per-operation outputs
+
+**Test file:** `tests/tests-automations/regression/core-components/data-operations-component.spec.ts`
+
+**Last validated:** Langflow 1.12.x
+
+---
+
+## What this test validates
+
+Langflow 1.11.0 merged the three separate operation components — **JSON Operations**
+(`DataOperations`), **Table Operations** (`DataFrameOperations`) and **Text Operations**
+(`TextOperations`) — into a single **Data Operations** component
+(`lfx/components/processing/operations.py`, class `OperationsComponent`, `name="Operations"`).
+Upstream: `langflow-ai/langflow#13743` (unify) and `#14025` (naming). The three originals
+remain in the sidebar as `legacy: true` with `replacement = ["processing.Operations"]`.
+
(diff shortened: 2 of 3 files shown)


Victor-w-Madeira:
CI closes the nightly gap noted in the Validation block: the `Run impacted E2E specs` job ran this spec against **Langflow Nightly 1.12.0.dev17** (the newest tag) — `Running 4 tests using 2 workers` → **4 passed (19.4s)**.

So the spec is now proven on both builds: 1.12.0.dev10 locally (3 clean `--retries=0 --workers=1` bursts + the four force-fails) and dev17 in CI, at 2 workers — which also exercises the parallel-safety claim (unique flow names, id-scoped cleanup).

The `Collect models` sweep in that job is red for the known drained-OpenAI/Google-key reason (#772/#1029), not for anything in this PR: it is `continue-on-error`, and this spec needs no provider at all.

Victor-w-Madeira:
Copilot review addressed in bf309c4 — one taken, one corrected.

**1. `closeOutputInspector` not waiting for the dialog — taken.** Real gap, and it had a sharper edge than "animation race": `openOutputInspector` resolves the modal with `page.locator('[role="dialog"]').last()`, so a dialog still on its way out stays a live candidate for the *next* open. It now waits for `btn-close-modal` to reach count 0 — deliberately that testid rather than `[role="dialog"]`, which the assistant onboarding tooltip also carries (the repo's documented strict-mode trap).

Proven to gate, not just to exist: removing the click and keeping the wait fails the test at `closeOutputInspector` (line 221), and the marker is reverted (`grep -c FF-MUTATION` = 0) with a green 4/4 run after.

**2. Testid order — the comment inverts it, so I clarified the doc instead of applying the change.** The suggestion was that the convention is `output-inspection-<type>-<title>`. Upstream emits:

```tsx
data-testid={`output-inspection-${title.toLowerCase()}-${classNameFromType(id).toLowerCase()}`}
```
(`CustomNodes/GenericNode/components/NodeOutputfield/index.tsx`)

Title comes **first**. In `output-inspection-json-operations`, `json` is the **output's display name** and `operations` is the **component type** (`name="Operations"`) — not the reverse. Applying the suggestion would have documented it backwards, and that matters here beyond wording: the first half is exactly what this spec asserts on (it changes to `message` / `json` / `table` with the selected operation), while the second half is constant on every node of this component. Reading it the other way makes the testid look node-stable and hides the assertion.

The placeholders **were** ambiguous, which is the fair half of the comment — `<output>`/`<type>` are now `<output display name>`/`<component type>` in both the doc and the spec comment, with the source expression and an explicit note about the tempting reverse reading.

Re-validated after both edits: typecheck 0, eslint 0, **4 passed (28.9s)** at `--retries=0 --workers=1`, zero `🚨 Backend Error`, zero orphan flows.

PR #1293, opened 2026-08-05
test(core-components): cover the legacy operations link to Data Operations (#1192)

### LINKED ISSUE #1192: Create: legacy operations components link to Data Operations (§3.10)

**Backlog pointer:** QA-CHECKLIST.md §3.10 → `core-components/data-operations-legacy-link.spec.ts`

**Scope**
- Legacy JSON/Table/Text operations components are linked/redirected to Data Operations; legacy flows keep working.

**Upstream:** langflow-ai/langflow#14118 (link legacy operations to Data Operations).

Victor-w-Madeira:
Closes #1192.

Closes the `[ ] Legacy operations components link/redirect to Data Operations (legacy flows keep working)` gap in `QA-CHECKLIST.md` §3.10 — the second and last bullet of that section, whose first bullet was closed by #1291 (`data-operations-component.spec.ts`, #1191). §3.10 is now fully covered.

When 1.11.0 merged JSON Operations (`DataOperations`), Table Operations (`DataFrameOperations`) and Text Operations (`TextOperations`) into the unified **Data Operations** node, the three originals were not deleted — they were marked `legacy = True` with `replacement = ["processing.Operations"]`, and upstream `langflow-ai/langflow#14118` wired that pointer into the UI. This PR covers the promise that pairing makes to a user who already has one of the old components: it tells you what replaced it **by name**, gives you a one-click route to it, is still reachable when you search its old name, and **keeps building** in the meantime.

**Distinct from the two specs that look adjacent.** `flow-functionality/general-bugs-frontend-crashing-on-invalid-replace.spec.ts` also mentions `No direct replacement` — but it is the opposite journey and a different subject: it builds a *custom* component with a bogus `replacement` and proves the frontend degrades to that fallback instead of crashing. It asserts the fallback path **works**; this PR asserts the three real legacy operations components are **not on it**. And `legacy-components-toggle-regression.spec.ts` covers the sidebar toggle's visibility behaviour with an unrelated component pair, not the replacement-pointer contract.

## 1. Covered tests

| 1 | `All three legacy operations components name Data Operations as their replacement` | Each of JSON / Table / Text Operations renders the Legacy banner (`dismiss-warning-bar`) reading **`Use Data Operations.`**, and **not** `No direct replacement.`. Catches a `replacement` pointer that stopped resolving — a rename or a move of the unified component. |
| 2 | `The legacy banner link filters the sidebar to Data Operations` | Clicking `Data Operations` inside the banner activates a component filter (`sidebar-filter-reset` appears), the sidebar offers `add-component-button-data-operations`, and `add-component-button-json-operations` drops to count 0. The link is a route to the replacement, not decoration. |
| 3 | `Searching a legacy operations name surfaces Data Operations with legacy components hidden` | With the Legacy toggle in its **default OFF** state, each of the three old names typed into the sidebar search offers `Data Operations` and no legacy button. The unified component carries the old names in `metadata.keywords`, so the search that used to land on the legacy node now lands on its replacement. |
| 4 | `A legacy operations component still builds and returns its result` | A legacy Text Operations node with `Case Conversion = uppercase` builds (`node_duration_text operations`) and its Message output equals `LEGACY STILL WORKS ABC123`. The "legacy flows keep working" half of the checklist bullet. |

## 2. How the test was built

- **The banner text is a far sharper observable than a UI string usually is,** and that is the core of the design. `NodeLegacyComponent` does not print `replacement` raw — it resolves it through `useGetReplacementComponents`, which splits `"processing.Operations"` into category + name and looks up `data.processing.Operations.display_name` in the types store. Only a successful lookup renders `Use <display name>.`; anything else falls back to the literal **`No direct replacement.`**. So asserting the positive string proves the whole chain (the component still declares a replacement, the pointer still names something that exists, the frontend resolved it to an actionable name), and asserting the fallback's **absence** is what catches the silent degradation a rename would cause — the node still renders and nothing errors.
- **Setup:** blank flow via the REST API (`createFlow`, unique name → parallel-safe), opened at `/flow/{id}`; the existing `addLegacyComponents` helper flips the sidebar Legacy toggle where the test needs it.
- **Live-confirmed selectors** (scouted in the running DOM, none invented): `add-component-button-{json,table,text}-operations`, `title-{JSON,Table,Text} Operations`, `dismiss-warning-bar`, `sidebar-filter-reset`, `sidebar-options-trigger` + `sidebar-legacy-switch`, `textarea_str_text_input`, `button_open_list_selection_sortablelist_sortablelist_operation`, `list_item_case_conversion`, `value-dropdown-dropdown_str_case_type`, `button_run_text operations`, `node_duration_text operations`, `output-inspection-message-textoperations`, `btn-close-modal`.
- **Legacy nodes keep their ORIGINAL class names in every testid** — the node type is `DataOperations` / `DataFrameOperations` / `TextOperations`, never `Operations`. Hence `output-inspection-message-textoperations`, and hence no testid collision with the unified component if both ever share a canvas.
- **The banner link has no testid**, so it is matched by accessible name (`getByRole("button", { name: "Data Operations", exact: true })`), scoped to the node — otherwise it could resolve to the sidebar entry of the same name.
- **Test 1 deliberately leaves the three nodes stacked.** It only reads text and testids *inside* each node (located by its own `title-…`), so the pointer-interception problem that forces `separateOverlappingNodes` in the chained #1191 tests simply does not arise. Adding the separation would be cost with no assertion behind it.
- **`showLegacy` was verified to be client-side before relying on its default.** It is localStorage-backed (`getBooleanFromStorage("showLegacy", false)`, `flowSidebarComponent`), not a server-side user setting — so enabling it in tests 1, 2 and 4 cannot leak into a parallel worker, and test 3's OFF precondition holds without any reset. Had it been server state, test 3 would have been a cross-worker hazard of exactly the class #553 documents.
- **Expected value in test 4 comes from the legacy source** (`text_operations.py` → `_case_conversion` → `str.upper`), not from observation.
- **Cleanup:** one flow per test, id known from creation; `afterEach` leaves the editor via `unmountEditorForCleanup` (so the events poll stops 404ing against a flow being deleted, #1288) and deletes only that id with an explicit bearer. Verified after the runs: **zero orphans**.

## 3. Dependencies

- **None — pure UI/Python.** No model provider, no API key, no `collect-models`, no `models.json`, no external network.
- **Parallel-safe:** unique flow names, id-scoped cleanup, and the one piece of shared-looking state (`showLegacy`) is per browser context. No `--workers=1` needed.

## Validation (nightly 1.12.0.dev10, `--retries=0`)

- typecheck ✅ · eslint ✅ · QA-CHECKLIST guards ✅
- **3/3 clean bursts** at `--retries=0 --workers=1` (`expected=4, unexpected=0, flaky=0` each), plus a green run after the force-fail reverts ✅
- zero `🚨 Backend Error` on every run ✅ · suite runs in ~16 s
- **force-fail executed for all 4 tests**, each isolated by `--grep`, each marked `// FF-MUTATION` and reverted (final marker count: 0):
  - #1 — expect the banner to read `Use JSON Operations.` (the legacy component's own name) → failed ✅
  - #2 — expect the legacy entry to survive the sidebar filter → failed ✅
  - #3 — search a term matching nothing, proving the assertion is driven by the **search term** and not by a sidebar that always lists Data Operations → failed ✅
  - #4 — expect the untouched lowercase text instead of the uppercased result → failed ✅
- flow cleanup verified via `GET /api/v1/flows/`: zero orphans ✅

> **Nightly note:** validated on `1.12.0.dev10` (the same local instance as #1291). It was deliberately not rebuilt — it runs without a volume, so recreating it would destroy its SQLite state. The `Run impacted E2E specs` job runs this spec against the newest nightly.
>
> `--trace=on` was not run: known local hang on this stack (see the langflow-e2e-issues skill). Step verification is covered by the clean `--retries=0` bursts, the executed force-fails, and the live DOM scouting the selectors came from.

Generated blocks in `QA-CHECKLIST.md` were **not** regenerated — only the §3.10 bullet was edited (they regenerate on merge, #741).

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -278,7 +278,7 @@
 
 #### 3.10 Data Operations (1.11.0)
 - [x] Data Operations component: unified JSON/Table/Text operations produce correct outputs per operation mode (Text→Message, Word Count→JSON override, JSON Select Keys, Table Filter) → `core-components/data-operations-component.spec.ts`
-- [ ] Legacy operations components link/redirect to Data Operations (legacy flows keep working) → `core-components/data-operations-legacy-link.spec.ts`
+- [x] Legacy operations components link/redirect to Data Operations (banner resolves the replacement, link filters the sidebar, old names still find the new component, legacy flows keep working) → `core-components/data-operations-legacy-link.spec.ts`
 
 ---
 
--- a/docs/core-components/data-operations-legacy-link.md
+++ b/docs/core-components/data-operations-legacy-link.md
@@ -1,0 +1,248 @@
+# Spec: legacy operations components link to Data Operations
+
+**Test file:** `tests/tests-automations/regression/core-components/data-operations-legacy-link.spec.ts`
+
+**Last validated:** Langflow 1.12.x
+
+---
+
+## What this test validates
+
+The other half of `QA-CHECKLIST.md` §3.10. When 1.11.0 merged JSON / Table / Text
+Operations into the unified **Data Operations** node (covered by
+`data-operations-component.spec.ts`, #1191), the three originals were not deleted — they
+were marked `legacy = True` with `replacement = ["processing.Operations"]`, and upstream
+`langflow-ai/langflow#14118` wired that pointer into the UI. This spec covers the promise
+that pairing makes to a user who already has one of the old components:
+
+> A legacy operations node **tells you what replaced it, by name**, gives you a one-click
+> route to it, is still reachable when you search for its old name — and **keeps building**
+> in the meantime.
+
+Four tests, one claim each:
+
+1. **All three legacy nodes resolve their replacement.** Each of JSON / Table / Text
+   Operations renders the Legacy banner reading `Use Data Operations.`
+2. **The banner link is a route, not decoration.** Clicking `Data Operations` in the banner
+   filters the component sidebar down to that component.
+3. **Searching the old name finds the new component**, with the Legacy toggle in its
+   default OFF state.
+4. **A legacy operations component still runs**, returning the value its operation defines.
+
+### Why the banner text is the assertion, and why `No direct replacement.` is the trap
+
+`NodeLegacyComponent/index.tsx` does not print the raw `replacement` string. It resolves it
+through `useGetReplacementComponents`, which splits `"processing.Operations"` into category
+and name and looks up `data["processing"]["Operations"].display_name` in the types store.
+Only when that lookup succeeds does it render `Use <display name>.`; otherwise it falls back
+to the literal string **`No direct replacement.`**
+
+That makes the observable unusually sharp for a UI string. Asserting `Use Data Operations.`
+proves the *whole chain*: the legacy component still declares a `replacement`, the pointer
+still names a category and component that exist in this build, and the frontend resolved it
+to the display name a user can act on. Rename or move the unified component and the banner
+silently degrades to `No direct replacement.` — the node still renders, nothing errors, and
+only this assertion notices. So every banner check asserts the positive string **and** the
+absence of the fallback.
+
+### Distinct from the sibling spec that looks similar
+
+`flow-functionality/general-bugs-frontend-crashing-on-invalid-replace.spec.ts` also mentions
+`No direct replacement` — but it is the **opposite** journey and a different subject: it
+builds a *custom* component with a deliberately bogus `replacement`
+(`THISISNOTEXISTING.COMPONENT`) and proves the frontend degrades to that fallback instead of
+crashing. It asserts the fallback path **works**; this spec asserts the three real legacy
+operations components are **not on it**. Neither covers the other.
+
+`core-components/legacy-components-toggle-regression.spec.ts` covers the sidebar Legacy
+toggle itself (visibility on/off) with an unrelated component pair, not the
+replacement-pointer contract. Test 3 below depends on that toggle's default only as a
+precondition.
+
+---
+
+## Tags
+
+`@stable` `@components` `@ui-ux`
+
+`@components` is the cross-cutting tag (canvas/sidebar component behaviour); `@ui-ux` the
+functional one — the banner, the sidebar filter and the sidebar search are interface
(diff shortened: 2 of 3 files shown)


Victor-w-Madeira:
CI closes the nightly gap noted in the Validation block: the `Run impacted E2E specs` job ran this spec against **Langflow Nightly 1.12.0.dev17** (the newest tag) — `Running 4 tests using 2 workers` → **4 passed (16.1s)**.

Proven on both builds, then: 1.12.0.dev10 locally (3 clean `--retries=0 --workers=1` bursts + the four force-fails) and dev17 in CI at 2 workers, which also exercises the parallel-safety claim — including that `showLegacy` is per browser context and not shared state.

With this merged, `QA-CHECKLIST.md` §3.10 is fully covered: the unified component's per-operation outputs (#1291) and the legacy link/redirect contract (this PR).

Victor-w-Madeira:
Copilot review applied in 64f6cbd. Both comments were the same underlying point and it was a fair one: `LEGACY_COMPONENTS[].title` held a **testid** (`title-JSON Operations`), not a label — and it was interpolated into the `test.step()` name, so a Playwright report/trace read `title-JSON Operations carries the Legacy banner …`. The display name and its testid differ only by a prefix here, which is exactly what makes the two easy to conflate.

What changed:

- `title` → split into **`label`** (human name — step titles and the new failure message) and **`titleTestId`** (selector).
- `nodeByTitle` → `nodeByTitleTestId`, so the helper name says what it takes.
- `LEGACY_COMPONENTS[0]` / `[2]` at the call sites → named `JSON_OPERATIONS` / `TEXT_OPERATIONS`. Which legacy component a test drives is part of what the test means, and an index hides it. (Not in the review — it is the same readability problem one line over.)
- The node-count assertion gained a message naming the component, so a missing node fails with the component's name rather than a bare count mismatch.

**No assertion changed** — this is a rename plus one added failure message. The force-fail evidence in the Validation block was produced against these same assertions and still stands; re-running the four mutations would exercise byte-identical expectations.

Re-validated after the edit: typecheck 0, eslint 0, **4 passed (16.1s)** at `--retries=0 --workers=1`, zero `🚨 Backend Error`, zero orphan flows.

Issue #1023, opened 2026-07-28
project-management: residual 404s after the destructive lane — an editor polling a deleted flow, and a folder refetched after its own deletion


rafaelgiln:
Split out from #1010 (PR #1021), which fixed the account-wide folder **wiper** by moving it to a destructive lane. These `404`s survive that fix and are a **different collider** — filed rather than folded in, so the wiper fix stayed scoped.

## What survives

The four folder-touching specs run together at `--workers=2 --retries=0` against `1.12.0.dev8`, six times after the lane landed:

| | Failures | `404` on a project id | `404` on a flow id |
|---|---|---|---|
| Before the lane (3 runs) | 1 | 3, across 2 of 3 runs | 1 |
| After the lane (6 runs) | **0** | 1, in 1 of 6 runs | 5, in that same run |

No test fails. Both residuals are `🚨 Backend Error` lines only — but `tests/fixtures/fixtures.ts` flags every `/api/` 4xx, and the deterministic pipeline's VALIDATE gate hard-stops on `backendErrors=true`, so while they stand they can block a PR that touches these specs (the same mechanism #1008 already demonstrated).

## Residual A — an editor polling a flow that was deleted

Five errors on one flow id, all inside a single run:

```
🚨 Backend Error: 404 - /api/v1/flows/247410f0-…                      (x3)
🚨 Backend Error: 404 - /api/v1/flows/247410f0-…/events?since=1785247509.431
🚨 Backend Error: 404 - /api/v1/flows/247410f0-…/note_translations
```

`/events?since=` + `/note_translations` is the flow-editor page polling. So a page that had a flow open kept polling it after it was deleted.

**Not the wiper:** it is excluded from these runs by the lane (7 tests scheduled, not 8) and the `422` the wiper emits (#1008) disappeared along with it.

**Attribution is not settled.** The run's `line` reporter printed the `404`s under two different headers (`folder-deletion-integrity.spec.ts:77` and `:193`) because stdout interleaves at `--workers=2`, while the terminating `📋 Found 5 backend error(s)` belongs to a single test. Both of those tests open the "Basic Prompting" starter template (which lands in the editor) before returning to the folder view, so either is a plausible owner.

**Not reproducible in pairs** — each of these is 0 `404`s:

| Combination | `404`s |
|---|---|
| `folder-deletion-integrity` alone | 0 |
| `+ bulk-actions` | 0 |
| `+ folder-crud` | 0 |

[...]

PR #1294, opened 2026-08-05
fix(fixtures): declare the projects/undefined 422 instead of logging it as a backend error (#1008)

Victor-w-Madeira:
Resolves the repo-side half of #1008 — deliberately **without** a `Closes` keyword; see *Why this does not close the issue* below.

Refs #1008

## Verdict — upstream frontend defect

No test produces the request. It is a `GET` the app issues, and any user who deletes their last project reaches it. Reproduced on `1.12.0.dev7`, `dev8` and `dev10`; the code path is identical on `origin/release-1.12.0`, the branch the nightly image is built from (not `main` — the two lines diverge). The chain:

1. `use-get-folders.ts` — `myCollectionId = data?.find(default)?.id ?? data?.[0]?.id`. With no project left, `data` is `[]` and the value stored is `undefined`.
2. `homePage/index.tsx` — passes `id: folderId ?? myCollectionId!` into `useGetFolderQuery`. `folderId` exists only on the nested `folder/:folderId` route, never on `/all`. `IGetFolder.id` is a **required** `string`, and `!` is a compile-time assertion with no runtime effect.
3. `use-get-folder.ts` — nests its existence guard inside `if (params.id)`, so it is skipped for exactly the `undefined` case it should block, and `` `${getURL("PROJECTS")}/${params.id}` `` interpolates the string.

Nothing downstream stops the fetch: no `enabled` is passed, so react-query's default applies. The backend's `422 uuid_parsing` is correct — the defect is that the request is sent at all.

**The query string is the attribution evidence**, and it looks like a contradiction until you see the mechanism: the id is interpolated into the *path* yet absent from the *query*, because `buildQueryStringUrl` skips `undefined` values while the path template had already stringified it. What remains is `page, size, is_component, is_flow, search` — byte-for-byte the observed URL. And the chain is **exclusive**: `useGetFolderQuery` has one call site, only `use-get-folder.ts` builds a `GET …/{id}?…` of the seven `getURL("PROJECTS")` references, and `is_flow` appears as a query param nowhere else.

**It is user-facing, not a test artifact:** `delete_project` refuses only `ASSISTANT_FOLDER_NAME`, so the default Starter Project is deletable and a user can empty their account through the UI.

## The fix, and why not `allowHttpErrors()`

New narrow hatch `page.expectKnownHttpError({ pathname, status, reason })`. It quietens **only** that exact status-plus-pathname, printing `📌 Known backend defect` rather than `🚨 Backend Error` — the string the deterministic pipeline's VALIDATE gate greps for, which is what made any PR touching this spec unable to pass that gate.

`allowHttpErrors()` was rejected because test 4's body deletes N projects through the UI, and `DELETE /api/v1/projects/{id}` → 500 with the toast still reading "deleted successfully" is a *separate* filed defect (#965/LE-2020) this loop is unusually well placed to catch. Blanket silence would have traded one known error for blindness to the other. An `IGNORED` entry was rejected for the opposite reason — it would hide the response from all 247 specs, permanently.

Two properties make it safe to add:

- **Consulted last**, after the `/api/` test and after `IGNORED`, so a declaration can only quieten what the policy would have reported — never widen monitoring onto an asset or un-ignore a global exemption.
- **Verified in both directions.** A declared defect that stops firing **fails the test** and names the call to delete. An exemption whose justification expired silently is what #1084 was raised about, and a printed warning would be one more line nobody reads.

**Declared on test 4 only, deliberately.** The `422` is a property of the zero-project state, so tests 1–3 emit it too when the account is *already* empty — but on a normal account they do not, so an unconditional declaration there would be stale on every healthy run and would fail them. Left undeclared, that `422` in tests 1–3 still reports what it means: this run started with no projects (#1010's fingerprint). The rule the hatch implies is written down once: declare a known defect only in a test whose own body guarantees the state that fires it.

## `@stable` restored on tests 1–3

Which the spec doc had gated on this issue. Three back-to-back `--workers=1 --retries=0` runs on `1.12.0.dev10`: 3/3 green. **Test 4 stays untagged** — `@destructive` must never carry `@stable` (#1010), since `daily-stable.yml` has no destructive lane and the pair would mean "runs nowhere".

## Evidence

- `422` declared **6/6** destructive-lane runs on `1.12.0.dev10`; never a `🚨` line for it.
- Force-failure: mutating the declared `pathname` puts the `422` back as `🚨` **and** fails with `1 declared known backend defect(s) did NOT occur`. Both directions.
- **Mutation battery: 11 applied one at a time, 10 killed.** Two survived the first pass and both were acted on. One was a real gap — the code claimed the `📌` line is "printed on the FIRST occurrence only" and nothing pinned it, because the gate spec fired the defect once, which cannot tell the two designs apart; it now fires three times. The other is accepted and recorded in the gate spec header rather than left unknown (#1012): the debug-only `ignoredByPolicy` double-count changes no verdict, count or gate string.

## Findings from the independent review, both fixed

- **The stale check raced the response.** `page.on("response")` is an event and the check runs in teardown, so a response issued in the last moments of the body could still be in flight — reporting "the defect is gone" about a defect that is still there. Fixed with a 1 s grace period entered **only** when something has not fired, so a healthy run never pays for it and the cost only lands on a run already headed for a red. Pinned by a test that kills it: the probe server withholds the response 300 ms, the body fires it without awaiting and returns; removing the grace period turns that test red.
- **Per-page scope.** A defect firing only on a `context.newPage()` page or a popup is never counted, so the declaration reports stale every run — unfixable, not flaky. No current caller does this; now documented on `KnownHttpDefect`.

## One honest caveat, corrected during review

An earlier draft of the spec doc claimed "no `🚨 Backend Error` line, in either lane". True for the three runs it cited, not in general: over 6 destructive-lane runs, **2** also logged `404 … GET /api/v1/flows/{id}`. It correlates with the **entry state** — both occurrences were the first destructive run after a non-destructive one, with flows already on the account — and is the #1023 teardown-order class surfacing inside the test rather than in teardown. **Out of scope for #1008** and worth its own issue. The claim in the doc is now the accurate one, and the residual `404` is a reason the `422` was declared narrowly instead of silencing the test.

## Why this does not close the issue

#1008's own *Done when* list keeps it open past this PR: the verdict is recorded here, and the spec runs without the `422` reading as a backend error, but it also asks for the defect to be **filed upstream** (Jira `LE-####` or a `langflow-ai/langflow` issue) and for the issue to stay open **until the fix lands on `langflowai/langflow-nightly:latest`**. Filing in another org's public tracker was left for a maintainer to authorise. A `Closes` here would auto-close an issue that still has tracking work.

One consequence worth knowing: test 4 is `@destructive` and no scheduled lane runs `@destructive` (`daily-stable.yml` has none, `nightly.yml` is dormant), so the "the defect is gone" alarm arrives on a PR touching this file or a `manual.yml` dispatch — not the day after the upstream fix merges.

## Roadmap

#1008 carries no roadmap milestone and none of the exception labels (`follow-up` / `daily-failure` / `community`); the only milestone in the repo, *Wave 5*, is fully closed. Worked at the maintainer's direction rather than off a wave — flagging rather than implying wave coverage.

## Gates

`npm run typecheck`, `npm run lint` (0 errors), `npm run test:units` (430/430), `npm run check:checklist-coverage`, `scripts/check-checklist-guard.mjs` — all green. Only the manual Part II bullets were edited; no generated block touched.

> **Reviewer note:** this changes `tests/fixtures/**`, which the impacted-specs resolver treats as suite-wide — expect the PR lane to warn and cap, and dispatch `manual.yml` for a full run before merging.


Victor-w-Madeira:
### Correction to my own reviewer note — the pre-merge run advice

The note at the end of the description says to "dispatch `manual.yml` on this branch for the full suite". That is the workflow's standing wording, and it is **not actionable as written**: #1174 records that `manual.yml` cannot finish the full suite inside its own 90-minute timeout. Repeating it would send a reviewer at a dispatch that times out.

**Measured coverage of this PR's green:** the lane resolved **250** impacted specs (the whole suite, as expected for a `tests/fixtures/**` change), capped at 20, so **230 were not run** — the cap warning and the dropped list are both in the `Detect changed specs` job.

The actionable version, in order of what it buys:

1. `manual.yml` with `test_tag=@stable` — ~170 specs, the recurring lane's own selection, and the set whose green is the release signal. This is the one worth spending.
2. Accept the 20-spec cap. Defensible here on the specific shape of this change: the fixture's behaviour is **unchanged for every spec that declares nothing** — `declaredDefects` is empty, `classifyHttpError`'s second parameter defaults to `[]`, and a unit test asserts the one-argument and two-argument spellings classify identically. Exactly one spec declares anything.

I have not dispatched either — the run cost is the maintainer's call, and neither is needed to review the diff.

Issue #1303, opened 2026-08-05
[Daily #1296] global-variables-crud — a newly created variable never appears in the Global Variables grid (recurrent 3×)


Victor-w-Madeira:
Spun out of daily-failure triage #1296 (run [30997773754](https://github.com/oriontech-me/langflow-e2e/actions/runs/30997773754), 2026-08-05).

**Upstream:** [LE-2599](https://datastax.jira.com/browse/LE-2599) — _Late or failed `/api/v1/session` breaks global-variables refetch under auto-login_ (filed 2026-09-11)

## Symptom

A `@stable` test flaked (recovered on retry) on Settings → Global Variables. Despite the test's name, the failure is **upstream of anything about hiding a credential value**: it fails on the test's own sanity assertion that the variable it just created is present in the grid — `getByRole('treegrid').getByText(varName, { exact: true })` — with `element(s) not found` after 15 s. The variable was created and then never appeared in the list, so the secret-leak assertion the test exists for was never reached.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/ui-ux/global-variables-crud.spec.ts:144` ("Credential variable value is hidden from the variable list") | `getByRole('treegrid').getByText('<varName>', { exact: true }) to be visible (timeout 15000ms)` | `Error: expect(locator).toBeVisible() failed` |

## Why these failures are one cause

One spec, one cause, filed because it meets the **recurrence criterion**: the same test failed with the same `error_signature` on **2026-07-15**, **2026-07-17** and **2026-08-05** — three same-signature occurrences inside the 30-day window.

Worth recording because the test name misdirects: anyone picking this up from the title alone would investigate whether a credential value leaked into the DOM, which is not where it fails. The failing wait is the sanity check that precedes that assertion.

**Deliberately not folded into #1235**, which tracks the same surface (the Global Variables ag-grid). #1235 is about a **row interaction not producing the state it should** — clicking a row does not open the Update Variable modal, ticking a row checkbox does not enable the delete button. This is a different step: no row interaction has happened yet, the row itself never renders after a create. The signatures also cannot carry a match — `expect(locator).toBeVisible() failed` is the most generic string Playwright emits, and #1235's own grouping argument rests on a shared run-pair (07-27 and 08-03) that this test does not share. If the investigation finds one grid-refresh mechanism behind both, merge into #1235 (keep the older number) and carry the spec path over.

## Preliminary read (descriptive — NOT a verdict)

[...]

Issue #1302, opened 2026-08-05
[Daily #1296] ollama-provider — the flow produces no chat message within 180 s (recurrent 2× same-signature, 6 hits in the 30-day window)


Victor-w-Madeira:
Spun out of daily-failure triage #1296 (run [30997773754](https://github.com/oriontech-me/langflow-e2e/actions/runs/30997773754), 2026-08-05).

**Upstream:** _not filed_

## Symptom

A `@stable` test flaked (recovered on retry) waiting for the playground to show a chat message after running a flow through the local Ollama model. The wait is `expect(getByTestId('div-chat-message')).toHaveCount(1)` with a **180 s** budget; Playwright resolved the locator to **0 elements 183 times** and the message never arrived.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/model-provider/ollama-provider.spec.ts:224` ("the Ollama component lists the local model live and executes the flow") | `getByTestId('div-chat-message') to have count 1 (timeout 180000ms)` | `Error: expect(locator).toHaveCount(expected) failed` |

## Why these failures are one cause

One spec, one cause, filed because it meets the **recurrence criterion**: the same test failed with the same `error_signature` on **2026-07-30** and **2026-08-05**. `CONTRIBUTING.md` requires the cause and not just the count, and the signature is identical on both, so it is recurrent rather than two unrelated flakes.

Worth recording alongside it: this test has **6** occurrences in the 30-day window (2026-07-15, 07-22, 07-23, 07-24, 07-30, 08-05), of which only the last two carry this signature. Read that as a test that has been unsettled for a month with varying observables, and treat the earlier four as context rather than as part of this cause.

## Preliminary read (descriptive — NOT a verdict)

A **flake**, not a hard failure — it passed on retry.

The observable that failed is downstream of everything else in the test: the model list *had* already been read live (the test got past that step), so what did not happen is the **run producing output**. A timing lead worth stating descriptively, not as a verdict: this repo has measured Ollama on a GitHub runner at **>100 s per call** (`ollama-provider.spec.ts`'s own measurement, recorded in `CLAUDE.md`), so a 180 s budget for a full flow run is closer to the observed cost than the number suggests, and a slower-than-usual runner would exhaust it without anything being broken. That is a hypothesis to test, not a conclusion — and it does not explain the other four occurrences, which carry different signatures.

[...]

Issue #1301, opened 2026-08-05
[Daily #1296] Canvas — a generic node never becomes clickable, so the click times out at 20 s (2 specs, recurrent 2× each)


Victor-w-Madeira:
Spun out of daily-failure triage #1296 (run [30997773754](https://github.com/oriontech-me/langflow-e2e/actions/runs/30997773754), 2026-08-05).

**Upstream:** _not filed_

## Symptom

Two `@stable` tests flaked (each recovered on retry) on the **same action**: clicking a node on the flow canvas. In both, Playwright waited the full 20 s on `getByTestId('div-generic-node')` and the click never landed — the canvas node never reached a clickable state.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-components/edit-name-description-node.spec.ts:42` ("user should be able to edit name and description of a node") | `getByTestId('div-generic-node') — click` | `TimeoutError: locator.click: Timeout 20000ms exceeded.` |
| `tests/tests-automations/regression/flow-functionality/stop-building.spec.ts:24` ("user must be able to stop a building from the canvas") | `getByTestId('div-generic-node').nth(1) — click` | `TimeoutError: locator.click: Timeout 20000ms exceeded.` |

## Why these failures are one cause

The two specs live in different directories and test different features (editing a node's name/description vs. stopping a build from the canvas), so the grouping is argued rather than assumed. Three dimensions are shared:

1. **Same observable, not just the same signature.** Both wait on `div-generic-node` — the same testid — and both fail as `locator.click: Timeout 20000ms exceeded.`. `locator.click: Timeout` on its own is a generic string and would be too weak to group on; what carries the argument is that the *same locator* is the one that never becomes clickable.
2. **Same class of action, same point in the test.** Both reach for a node the flow has just rendered and act on it immediately — one clicks the node to open its header controls, the other clicks the second node to start and then stop a build.
3. **Neither failing attempt has any outage cover.** They ran on different shards (1 and 4), each with its own Langflow container, and both shards measured **0 outages / 0 s of unreachable backend** for the whole run (see the preliminary read). That rules out a single per-container hiccup as the shared dimension, so the shared dimension is the canvas, not the backend instance.

If the investigation shows the two clicks fail for different reasons, split this issue and cross-link.

## Preliminary read (descriptive — NOT a verdict)

[...]

Issue #1274, opened 2026-08-04
The #751 credential-settle guard asserts a binding upstream removed on purpose (#14311) — 19 @stable specs fail by design on 1.11.2


Victor-w-Madeira:
The `#751` credential-settle guard (`tests/helpers/flows/agent-credential-settle.ts`, hardened in #1072) asserts that the Agent's credential field carries the provider variable name in the **persisted flow**:

```
observed  api_key="" models=[claude-haiku-4-5]
verdict   credential-pending
```

Upstream removed the behaviour it asserts. `646bdd6b` — *"fix(models): stop automatic provider field binding"* (langflow-ai/langflow#14311, merged into `release-1.11.2` on 2026-07-29) deletes the block in `lfx/base/models/unified_models/build_config.py` that installed the variable name on the field:

```diff
-        # Pre-populate with the variable name (never the raw secret) when a
-        # credential is available in the database or environment.
-                    field_config["value"] = var_key
-                    field_config["load_from_db"] = True
```

The PR body states the new contract: *"Provider build-config refresh now changes only field visibility and metadata … It never changes `value` or `load_from_db`. Empty fields use the existing provider-specific DB/environment fallback at runtime."* So an empty `api_key` on the persisted flow is now **correct**, and the guard's premise is stale.

## Measured cost, on 1.11.2 today

Release-validation dispatches against `langflowai/langflow:1.11.2rc3`:

| Run | Selection | Retries | Result |
|---|---|---|---|
| `1.11.1` | 96 tests (`@stable` ∧ `@agents`) | 0 | 94 passed, 2 skipped, **0 failed** |
| `1.11.2rc3` | the same 96 | 2 | **20 failed**, 16 passed, 59 did not run |
| `1.11.2rc3` (`@stable`, 4 shards) | 427 tests | 2 | **19 of the 27 failures are this guard**, plus **14 skips** cascading out of the serial blocks it aborts |

It fires on all three providers (`openai / gpt-4o-mini`, anthropic, `google / gemini-flash-latest`), so it is not provider-specific.

## Why this is worth doing before it breaks

The change is on the 1.11.2 line only:

| ref | auto-bind present |
|---|---|
| `v1.11.1` | yes |
| `v1.11.2` | **no** |
| `release-1.12.0` | yes |
| `main` | yes |

[...]

PR #1312, opened 2026-08-05
fix(triage): extend the infra-signature exemption to flakes (#1310)

### LINKED ISSUE #1310: The infra-signature exemption covers hard failures only — a wedge-collateral flake is still filed and quarantined
Follow-up from the triage of #1296 (daily run [30997773754](https://github.com/oriontech-me/langflow-e2e/actions/runs/30997773754), 2026-08-05).

## The gap

The **infra-signature exemption** (`CONTRIBUTING.md` → *Infra-signature exemption (wedge collateral)*, `scripts/lib/infra-signatures.ts`, #1031) is scoped to hard failures, and says so literally:

> A **hard failure** whose last error is **transport-level** — the harness could not reach or talk to the backend — is not attributable to the spec that reported it.

That scope was correct for what #1031 was protecting: `@stable` **auto-removal** (`scripts/remove-stable-from-failures.ts`), which only ever touches hard failures. It is the *only* consumer — the two `.mjs` files that appear to import the module (`report-backend-outages.mjs`, `watch-backend.mjs`) reference it in **comments only**.

The **flake** criterion has no equivalent filter. `.claude/skills/langflow-e2e-triage/scripts/lib/triage-core.mjs:536` decides actionability with:

```js
const flakes = dedupeEntries(run.flaky).map(withRecurrence).map((f) => ({
  ...f,
  actionable: f.recurrence.same_signature,
}));
```

`triage-core.mjs` contains **zero** mentions of infra signatures. So a flake whose error is transport-level — the harness failing to reach the backend, an error that cannot be a product assertion under any reading — still satisfies "same signature twice in 30 days", and `CONTRIBUTING.md` → *Triage protocol* then requires a dedicated issue **and a quarantine PR** for it.

The result is the exact misattribution #1031 exists to prevent, arrived at through the flake door instead of the hard-failure door: a spec loses `@stable`, gains `test.fixme`, and gets an investigation directive, because the backend stopped answering.

## The measured instance

`tests/tests-automations/regression/core-functionality/llm-agents/agent-context-id-isolation.spec.ts:512` on #1296:

| | |
|---|---|
| Signature | `TimeoutError: apiRequestContext.get: Timeout 20000ms exceeded.` — on `GET http://localhost:7860/api/v1/auto_login` |
| Infra list | **matches** `api-request-timeout` (`\bapiRequestContext\.\w+:\s*Timeout\b`) |
| Recurrence | 2× same-signature (2026-07-29, 2026-08-05) ⇒ `actionable: true` |
| Measured backend state | ran on shard 2; **retry 1 lasted 119 s with 108 s of measured backend downtime inside it (91 %)**, crossing 2 outage windows; retry 2 crossed a third |

The triage of #1296 declined to file or quarantine it and said so in the closing comment, but that was a **judgement call against the written criterion** — not something the rules support. The next person to triage a wedge day, or the automated `--phase propose` job, will file and quarantine it.

## Why this is not a one-line change

Two constraints found while reading the code, both of which shape the fix:

**1. The classifier is TypeScript; both places that need it are `.mjs`.** `classifyInfraError` lives in `scripts/lib/infra-signatures.ts` under `module: "commonjs"`, `moduleResolution: "node"`, no `allowJs`. `triage-core.mjs` is dependency-free ESM, and so is `append-weekly-history.mjs`. Copying the patterns is not acceptable — this repo has twice removed exactly that kind of duplication after it drifted (#1043 `providerSkipReasons`, #1184 `resolveTestTargets`, which had drifted into five variants). One source of truth is required.

**2. The stored signature is one line, so classifying from it is strictly weaker than classifying from the error.** `append-weekly-history.mjs` (`errorSignature()`) stores only the **first non-empty line**, capped at 240 chars. `classifyInfraError` deliberately matches *anywhere* in the message, because — in its own words — "a wedge often surfaces as an assertion whose *cause* line is the transport error, and the first line alone would miss it".

That difference is not hypothetical, and #1296 contains both halves of it:

- `agent-context-id-isolation:512` — the transport error **is** line 1. Classifiable from the stored signature.
- `agent-context-id-continuity:405` — line 1 is `Error: Agent credential never settled on the persisted flow (#751 guard, #1072).` and the transport error sits deeper, on a `last read err` line. **Not** classifiable from the stored signature.

So a fix that only pattern-matches the history's `error_signature` would silently keep missing the guard-wrapped shape — which is the shape the #751/#1072 guard produces, i.e. the most common wrapper in this suite.

## Proposed fix

1. **One source of truth reachable from both languages.** Move the pattern list to data (`scripts/lib/infra-signatures.json`, pattern as `source` + `flags`) consumed by both `infra-signatures.ts` (`resolveJsonModule` is already on; `scripts/lib/model-prices.json` is the precedent) and a new dependency-free `.mjs` accessor. The patterns themselves do not change, so `scripts/remove-stable-from-failures.test.ts` must keep passing untouched — that is the regression check on this move.
2. **Classify at write time, where the full error still exists.** `append-weekly-history.mjs` records `infra_signature: "<id>" | null` on both `failures[]` and `flaky[]`, classified from the complete error text rather than from the truncated signature. Backwards-compatible schema addition — no `version` bump (`reports/README.md`).
3. **Demote, do not drop.** In `triage-core.mjs`, a flake classified as infra is **not** `actionable`; it carries the id and reason so the triage notes it against the run's wedge instead of filing and quarantining it. Never silently: an excluded flake must be visible in the dataset and in the proposal (#1012's rule).
4. **Fallback for the existing window.** Every row currently in `reports/daily-history.jsonl` predates the new field, so recurrence over the 30-day window must still work: when `infra_signature` is absent, fall back to classifying the stored `error_signature`. State plainly that the fallback is the weaker check from constraint 2 — it catches `agent-context-id-isolation:512`, not `agent-context-id-continuity:405`.
5. **`CONTRIBUTING.md`**: extend the exemption's wording to cover flakes, and remove the sentence in the *Triage protocol* that currently makes the flake criterion unconditional. The skill's `SKILL.md` and `references/issue-templates.md` both currently assert the opposite ("that exemption governs hard-failure auto-removal, not flakes") and must be corrected in the same change, or the next triage will follow the stale text.

## Non-goals

- **Not** widening `INFRA_SIGNATURES`. The list stays narrow for the reasons its own header gives; this issue changes *who consults it*, not what is on it.
- **Not** touching hard-failure auto-removal. `remove-stable-from-failures.ts` behaviour is unchanged.

## Deliverables (Done when)

- [ ] The pattern list has exactly one source of truth, reachable from TypeScript and from `.mjs`, with `scripts/remove-stable-from-failures.test.ts` passing unchanged.
- [ ] `reports/daily-history.jsonl` entries carry `infra_signature` for both failures and flakes, classified from the full error text; `reports/README.md` documents the field.
- [ ] A flake classified as infra is not `actionable`, and the exclusion is **visible** in the dataset and the proposal rather than silent.
- [ ] Rows without the field still classify via the documented weaker fallback, and the weakness is stated where a reader will meet it.
- [ ] `CONTRIBUTING.md`, the triage `SKILL.md` and `references/issue-templates.md` agree with the new rule — no remaining text asserting the exemption is hard-failure-only.
- [ ] Unit tests in both lanes (`npm run test:scripts`, `npm run test:units`), including the two shapes from #1296: the transport-error-on-line-1 flake, and the guard-wrapped one that only the full-error path catches.
- [ ] `agent-context-id-isolation.spec.ts:512` is verified to come out **non-actionable** when the #1296 run is re-triaged (`--run 30997773754`), which is the behavioural check that the gap is closed.

Victor-w-Madeira:
Closes #1310

Follow-up from the triage of #1296.

## The gap

`#1031` exempted wedge collateral from `@stable` auto-removal — but that path (`scripts/remove-stable-from-failures.ts`) only ever sees **hard failures**, and it is the exemption's only consumer. The flake criterion had no equivalent filter: `triage-core.mjs` decided actionability with `actionable: f.recurrence.same_signature` and contained **zero** mentions of infra signatures.

So a flake whose error is transport-level — the harness failing to reach the backend, which cannot be a product assertion under any reading — still satisfied "same signature twice in 30 days", and `CONTRIBUTING.md` then required a dedicated issue **and a quarantine PR** for it. A spec quarantined because the backend stopped answering: the exact misattribution `#1031` exists to prevent, reached through the flake door instead of the hard-failure door.

Measured instance, from #1296: `agent-context-id-isolation.spec.ts:512` timed out at 20 s on `GET /api/v1/auto_login`, recurrent 2× (07-29, 08-05); its retry lasted 119 s with **108 s of measured backend downtime inside it**. That triage declined to file it — but only as a judgement call against the written criterion, which the next triage (or the automated `--phase propose` job) would not repeat.

## Three findings that shaped the fix

**1. The list had to become data — JSON is the only format both consumers reach.** On Node 20 a CommonJS module cannot `require()` an ESM one, and a `.ts` compiled to CommonJS cannot import a `.mjs`. The auto-removal consumer is TypeScript; both new triage consumers (`append-weekly-history.mjs`, `build-triage-dataset.mjs`) are ESM. No single *code* module can serve both, so `scripts/lib/infra-signature-patterns.json` is now the single source of truth, with two thin accessors over it. Copying the patterns was rejected: this repo has twice removed exactly that after it drifted (#1043 `providerSkipReasons`, #1184 `resolveTestTargets`, by then five variants).

The patterns were **generated from the existing module** rather than retyped, and `scripts/remove-stable-from-failures.test.ts` passes **unchanged** (27/27) — that is the regression check on the move, and the reviewer's cheapest one.

**2. A resolution trap that `tsc` cannot see.** Naming the data file `infra-signatures.json` made the extensionless `require("./lib/infra-signatures")` in `remove-stable-from-failures.ts` resolve to the **JSON** — CommonJS tries `.js`, `.json`, `.node` before ts-node's hook offers `.ts`. The script that edits spec files on `main` died at runtime with `classifyInfraError is not a function`, while `tsc --noEmit` stayed green throughout, because TypeScript resolves `.ts` first. Found the only way it can be: by running the auto-removal test. The file is renamed, the trap is documented where someone would re-introduce it, and a test pins the extensionless require.

**3. A pre-existing defect this work exposed** (scope added beyond the issue — flagged rather than folded in silently). `lastFailed` selected the last **non-passed** result. For a test that failed and whose retries were then *skipped* — what a `test.describe.serial` block does when it aborts — that is a skipped attempt carrying no error at all, so the row recorded `error_signature: "unknown"` while the real message sat on attempt 0. Two of the 24 hard failures on run 30997773754 landed that way, and `unknown` is not a signature but the absence of one, which triage's recurrence rule matches on — clustering unrelated message-less failures. `build-run-payload.mjs` has always excluded `skipped` here and its header claims to mirror this script's parsing; the two had drifted. I could not classify an error that was being discarded, so this had to be fixed here.

## The fix

- **Classification happens at write time**, in `append-weekly-history.mjs`, because it needs the **full** error text. `error_signature` is line 1 only, and a wedge routinely surfaces as an assertion whose *cause* line carries the transport error (the `#751` credential guard being the usual wrapper). Run 30997773754 contains both shapes — `agent-context-id-isolation:512` (line 1) and `agent-context-id-continuity:405` (three lines down). New field `infra_signature` on `failures[]` and `flaky[]`; additive, no schema version bump.
- **`triage-core.mjs` stays pure** — the classifier is *injected* by `build-triage-dataset.mjs` (the accessor reads its JSON at load), pinned by a structural test, because every behavioural test can pass while production omits it.
- **Demoted, never dropped.** An excluded flake stays in `flakes[]` carrying an `infra_excluded` block with the signature and *how it was decided*, so the exclusion is visible in the proposal rather than the flake quietly vanishing from the list (#1012).
- **A row that already carries `infra_signature` is trusted.** Only an *absent* field triggers the weaker signature-only fallback — re-deriving a recorded verdict could only make it worse. Absence vs. falsiness is asserted in a test.

## Verification

- **Independent reproduction of the workflow's own verdict**: re-running the appender over #1296's real `results.json` produces exactly the **same 7** collateral hard failures the umbrella listed, by a different code path. `agent-context-id-continuity:405` → `null` (its last error is the #1274 premise break), `agent-context-id-isolation:557` → `api-request-timeout`.
- **The issue's behavioural criterion**: re-triaging run 30997773754 returns `agent-context-id-isolation:512` as **non-actionable** with an `infra_excluded` block, while the four flakes legitimately filed from that run (#1301 ×2, #1302, #1303) stay actionable. Nothing else changed.
- **+23 tests** across both lanes (`test:scripts` 676 → 699, `test:units` 425 unchanged), including a **cross-language guard** that the two accessors never diverge — which exists because the first draft of the `.mjs` dropped an ESC prefix and widened `+` to `*`, a divergence that survives any by-eye reading of the diff.
- **Mutation-checked**, per #1226's lesson that a guard which cannot fail its own mutation is worthless: reverting the `lastFailed` fix fails 2 tests; dropping the injected classifier fails 1; both restore to green.
- Gates: `typecheck`, `lint` (0 errors), `test:scripts` 699/699, `test:units` 425/425, both QA-CHECKLIST guards, `regressions:check`.

## Two limitations, stated rather than buried

- **The fallback is weaker by construction.** Every row currently in `reports/daily-history.jsonl` predates the field, so those are classified from the one-line signature: that catches `:512`, and **cannot** catch the guard-wrapped shape. Asserted in a test and documented where a reader meets it, not left to be discovered.
- **One deliverable from the issue was void.** It required correcting text in `SKILL.md` and `references/issue-templates.md` asserting the exemption is hard-failure-only. Those docs never mention the exemption at all — the wrong claims I had seen were in *generated issue bodies* (#1265 and #1296's automated proposal), not in documentation. `SKILL.md` Phase 4 is updated to state the new two-condition rule; there was nothing else to correct.

## Not touched

`INFRA_SIGNATURES` is not widened — this changes *who consults the list*, not what is on it. Hard-failure auto-removal behaviour is unchanged.


Victor-w-Madeira:
## Response to the Copilot review (`2acb12d`)

**1. `triage-core.mjs` JSDoc vs. actual behaviour - valid, fixed.**

Correct, and it was the more important of the two. `actionable` is `recurrent && !infra_signature`, and an `unclassified` entry carries `infra_signature: null` - so a recurrent transport-level flake in that state stays actionable and would be filed and quarantined exactly as before this PR. The doc's "never to attributable" claimed a safety the code did not provide. The label named the gap; it did not close it.

Rewording the comment would have made it honest, but the claim is worth *having*, so it is made true instead. There is no safe default here - assuming collateral silently drops real flakes, assuming attributable is the bug - so the gap is now **visible**: `buildDataset` returns `infra_classification_gap`, `null` when every entry reached a verdict (recorded or fallback) and otherwise naming how many did not, and why an unclassified entry is not a cleared one. Non-null presence is itself the signal, the same convention `run_errors` already uses in the history schema.

Both JSDoc blocks now describe the real behaviour and point at the two things that actually prevent the regression - the structural test on the injection, and this flag. `SKILL.md` gains the reporting rule. +3 tests, including one asserting the honest outcome (unclassified means still actionable) alongside the announcement.

**2. `no-control-regex` on `infra-signatures.mjs` - not applied; the premise does not hold in this repo.**

Checked rather than assumed, three ways:

- `npx eslint --print-config scripts/lib/infra-signatures.mjs` gives **23 rules active, with `no-control-regex` reported as `NOT SET`**.
- Injecting a real violation into that file produced **no ESLint output at all**.
- `npm run lint` is at **0 errors** (1313 pre-existing warnings, unchanged).

`eslint.config.js` scopes its rule blocks to `tests/**/*.ts`, so nothing under `scripts/**` is subject to that rule. Adding an `eslint-disable-next-line no-control-regex` would be a directive for a rule that is not configured - a comment asserting something untrue, which is worse than the absence it would paper over. (The TypeScript sibling's disable comment predates this PR and is vestigial for the same reason; churning it is out of scope here.)

Worth noting that the regex is written with a `\u001b` escape rather than a literal control byte - and the accompanying test compares both accessors' `stripAnsi` on ESC-bearing input precisely because an earlier draft of the ESM half lost that prefix and widened `+` to `*`.

Gates after the fix: `typecheck` clean, `lint` 0 errors, `test:scripts` **702/702**, `test:units` **425/425**, both QA-CHECKLIST guards, `regressions:check`. Re-running the builder over run `30997773754` still yields `gap: null`, 4 actionable flakes, and `agent-context-id-isolation.spec.ts:512` infra-excluded.

Issue #1072, opened 2026-07-29
Agent api_key rebind does not settle in 20s — waitForAgentCredentialSettled flakes across the agent family (#751 class)


rafaelgiln:
Split out of #967, where it was recorded as *"related context, not folded in"* with the instruction to confirm whether it was downstream of the Anthropic outage or independent. **It is independent** — the analysis is below, so #967 can close on its own deliverables.

## Symptom

`tests/tests-automations/regression/core-functionality/llm-agents/agent-tool-name-validation.spec.ts:321` ("causal control — a valid custom tool name executes normally", **google** variant) flaked in the 2026-07-27 daily ([run 30261409427](https://github.com/oriontech-me/langflow-e2e/actions/runs/30261409427)) with:

```
Expected: "GOOGLE_API_KEY"
Received: "ANTHROPIC_API_KEY"
```

A google-parameterized test observing an Anthropic credential.

## Cause — the #751 guard timing out, not a wiring bug

The only place in the suite that compares credential **names** is the guard added by #751, `tests/pages/SimpleAgentTemplatePage.ts:73-92`:

```ts
expect(agent?.data?.node?.template?.api_key?.value).toBe(expectedCredential);
// inside .toPass({ timeout: 20000, intervals: [500, 1000, 2000] })
```

Every `SimpleAgentTemplatePage.load()` ends there. On the 1.11+ unified model selector, opening the Agent model dropdown auto-binds `api_key` to the **default** credential (`ANTHROPIC_API_KEY`); picking the target provider's model rebinds it **asynchronously**, persisted through a debounced autosave PATCH. The guard exists to block until that lands, so a caller cannot build a flow with the wrong provider's key.

So the failure is the guard doing its job: the rebind did not persist within 20 s and the node was still on the default credential. Not a false positive, and **not** downstream of the drained Anthropic key of #967 — a zero-credit key does not affect a UI rebind, and no inference happens before this point. `ANTHROPIC_API_KEY` appears only because it is the default binding.

## Why it needs its own tracker

#751 (the daily-#744 cluster that introduced the guard) is **closed**, as is #894. Nothing open owns this. The guard is on the shared `load()` path, so any spec in the agent family inherits the flake.

Recurrence in `reports/daily-history.jsonl` — the `expect(received).toBe(expected)` signature shows up repeatedly across the family, with the 2026-07-27 `:321` entry the one confirmed as this guard (its full expected/received strings are recorded in #967):

| Date | Test | Param |
[...]

PR #1315, opened 2026-08-05
fix(agents): settle the #751 guard on the provider, not on api_key (#1274)

### LINKED ISSUE #1274: The #751 credential-settle guard asserts a binding upstream removed on purpose (#14311) — 19 @stable specs fail by design on 1.11.2

Victor-w-Madeira:
Closes #1274

## What broke

Upstream #14311 ("stop automatic provider field binding", `646bdd6b`) deleted the block that wrote the credential variable name into the Agent node's `api_key`. It reached the 1.12 line on **2026-08-04** (`b65e195eb1`, merge of `release-1.11.2` into `release-1.12.0`), and the **2026-08-05** daily was the first nightly carrying it: **13 `@stable` specs** failed the `#751` credential-settle guard, each after waiting its full 20 s for a transition that can no longer happen.

Measured on `1.12.0.dev16` (local nightly; auto-bind confirmed absent from the image itself):

| read | `api_key` | `model.value[0]` |
|---|---|---|
| 0 | `""` (`load_from_db: false`) | `{name: "claude-opus-5", provider: "Anthropic"}` ← mount default |
| 1–7 | `""` | `{name: "gemini-2.5-flash", provider: "Google Generative AI"}` |

## The fix, and why it is not a capitulation

The credential is no longer **stored**, but it is still **determined** — by the provider of the selected model. With `api_key` empty, `get_api_key_for_provider` (`lfx/base/models/unified_models/credentials.py`) resolves `get_provider_secret_variable_key(provider)` from the user's global variables. So the guard settles on `template.model.value[0].provider`: not a weaker proxy for the credential, but the input the runtime derives it from.

**The race #751 exists for is unchanged and still real**, which is why the guard is re-pointed rather than deleted — read 0 above is the mount-time default, and a caller that opens the Playground in between runs the *default* provider's model (the original #744 signature).

`api_key` is asserted in **neither** direction. Requiring it empty would swap one dated premise for another and break `manual.yml` lanes that can dispatch a pre-#14311 build.

The provider string is Langflow's own spelling, **derived** from `providerTestId` rather than listed again (the panel testid is `provider-item-<the same name>`; both come from Langflow's provider metadata). A second hardcoded table is what #1043 and #1184 had to remove after it drifted.

This also fixes what the credential axis could not express for a **keyless** provider (#1187): `""` was both the settled state and the pre-selection state, so the guard leaned on the caller pinning a model. `provider: "Ollama"` is a positive assertion — which is also why the three callers that pin no model keep a real check instead of settling on the first read.

## Two commits, on purpose

`95ad238` is the change; `48aeebd` applies an **independent adversarial review** run with clean context. Two of its eight findings were real errors in the first commit's own claims:

- **The #1072 vacuity was not closed, and the docstring said it was.** For `anthropic` + no pinned model the mount state already satisfies both checks, so the first read settles and nothing is proved. The pre-#1274 code had the identical hole by the identical route, so it is inherited rather than introduced — but the claim was wrong. Closing it needs an observable this module does not have (default vs. deliberate pick), so the vacuum is made **visible**: the Page Object warns when it settles on the first read with no pinned model. Matters on the anthropic rotation day (#1185).
- **Provider and model were compared on two independent arrays**, so a payload whose entries each carry half of the wanted pair classified as `settled` for a combination no entry contains. The probe now carries **paired** entries. The test fixture takes `[provider, model]` tuples, because the parallel-array fixture could not express the bug it was meant to catch.

Also from the review: `provider-pending` asserted a provider nobody observed when `providers=[]` (new `provider-unobservable` verdict); **`langflowProviderName` had zero coverage** — replacing its body with `return providerTestId` passed all 436 unit tests while making the guard time out for all 29 dependent specs, and that mutation now fails three tests; "14 specs" was 13; a dead import; four comments contradicted by the code; and the `api_key is EMPTY` note printed even where no api_key was observed.

The review also **closed two gaps the first commit admitted**, by measuring what it could not: the unverified `"OpenAI"` and `"Ollama"` strings are correct (checked against `GET /api/v1/models/providers` on the live nightly, the upstream `agentics/constants.py`, and the frontend's testid template — the derivation is structural, not a convention), and `model.value[0]` already carries `provider` at v1.10.0 through v1.11.1, so the backward-compatibility claim holds.

## Validation

Against the local nightly `1.12.0.dev16`. **Only google has credit here** — openai and anthropic both return `insufficient_quota` from the providers themselves — so live coverage is single-provider; the other three rest on the measured/derived evidence above.

- `model-provider-model-toggle.spec.ts` → **1 passed** (a no-pinned-model caller, and one of the 13)
- `google-provider.spec.ts` → **2 passed**, including the end-to-end Agent execution — which is what proves the runtime credential fallback works with `api_key: ""`
- **Forced failure**: expected provider replaced with a bogus string → the guard fails with `provider-pending` and prints `observed providers=[Google Generative AI] models=[gemini-3.5-flash-lite] api_key=""`. Not a false positive.
- Re-validated after the classifier rework; the new vacuity warning does not fire for google (correct — google is not the selector's default).
- Gates: `typecheck` clean, `lint` 0 errors, `test:units` **442**, `test:scripts` **702**, both QA-CHECKLIST guards.

## Deliberately NOT in this PR

Two specs assert the old contract **themselves**, outside the shared guard, so this change does not reach them. Both need their spec doc updated first (the repo's SDD gate), which is why they are not bundled here:

- `mcp/client/mcp-client-agent.spec.ts:86` — carries a **private copy** of the guard with `expect(...api_key?.value).toBe(expectedCredential)`. Found by the review; my own blast-radius claim had missed it because a private copy matches no import of the shared module. It is `@mcp @agents @regression` with **no `@stable`**, so the daily gives no signal on it.
- `openai-compatible-provider-setup.spec.ts:620` — `@stable`, still red on this root cause; its spec doc explicitly declares `template.api_key.value`.

**So the daily stays red on one `@stable` spec after this merges**, and that is stated rather than discovered later.

One unrelated failure seen while validating, not caused by and not fixed here: `agent-system-prompt.spec.ts` got past the guard and the model answered, then failed on a **truncated sentinel** (`…PINEAPPLE-1785939935193-2` vs `…-2XPDFN`) on `gemini-2.5-flash`.

### DIFF
--- a/tests/helpers/flows/agent-credential-settle.test.ts
+++ b/tests/helpers/flows/agent-credential-settle.test.ts
@@ -526,26 +526,33 @@ test("#1274 no verdict claims a check that did not run, for any caller", () => {
   }
 });
 
-test("#1274 the message says the empty api_key is not the finding", () => {
-  // Every reader of this message arrives knowing the pre-#14311 behaviour, and the
-  // observed line still prints `api_key=""`. Without this note the empty field
-  // reads as the defect — which is how the 2026-08-05 daily's 14 failures were
-  // first triaged as a product regression.
+test("#1274 the note says api_key is not asserted — without claiming it is empty", () => {
+  // Every reader arrives knowing the pre-#14311 behaviour, and the observed line
+  // still prints an api_key. Without a note the field reads as the defect, which is
+  // how the 2026-08-05 daily's failures were first triaged as a product regression.
+  //
+  // Asserted on the INTENT, not on a phrase about the value: `manual.yml` can
+  // dispatch a pre-#14311 build where api_key IS written, and a note claiming it is
+  // empty on "every build" would contradict the observed line above it. Pinning the
+  // phrase is what let that false claim ship (caught in review of this PR).
   const message = failure();
-  assert.ok(message.includes("api_key is EMPTY"), message);
-  assert.ok(message.includes("#14311"), message);
+  assert.ok(message.includes("NOT asserted in either direction"), message);
   assert.ok(message.includes("#1274"), message);
+  assert.ok(
+    !/api_key is EMPTY on every build/.test(message),
+    "the note must not claim the field is empty on builds where it is written:\n" + message,
+  );
 });
 
 test("#1274 the api_key note is printed only when an api_key was observed", () => {
   // Review caught it printing unconditionally, including for read-failed, where no
   // api_key was seen — the same unobserved-claim defect this module exists to avoid.
   const observed = failure();
-  assert.ok(observed.includes("api_key is EMPTY"), observed);
+  assert.ok(observed.includes("api_key is NOT asserted"), observed);
   for (const verdict of ["read-failed", "no-agent-node"] as const) {
     const message = failure({ probe: null, verdict, reads: 0 });
     assert.ok(
-      !message.includes("api_key is EMPTY"),
+      !message.includes("api_key is NOT asserted"),
       `${verdict} observed no api_key, so it must not comment on one:\n${message}`,
     );
   }
--- a/tests/helpers/flows/agent-credential-settle.test.ts
+++ b/tests/helpers/flows/agent-credential-settle.test.ts
@@ -17,12 +17,18 @@
 //     distinction became unreachable — while the tests passed, because the fixture
 //     used a string too. Every fixture below therefore uses the real shape, and
 //     `MODEL_VALUE_SHAPE` documents where it is verified in the product.
-//  2. **The verdict distinctions.** Several states produce the same "wrong
-//     credential" observation and they send a reader to opposite places: a lagging
-//     autosave, a model click that never registered, a credential that matches only
-//     because it is the selector's default, and a read that never succeeded. A
+//  2. **The verdict distinctions.** Several states produce the same "not the
+//     provider I asked for" observation and they send a reader to opposite places: a
+//     lagging autosave, a model click that never registered, a provider that matches
+//     only because it is the selector's default, and a read that never succeeded. A
 //     classifier that collapses any of those silently re-creates the mis-triage,
 //     and no Playwright spec would fail for it.
+//  3. **The axis itself (#1274).** The guard settles on the PROVIDER of the
+//     persisted model, not on `api_key` — upstream #14311 stopped writing that
+//     field, so a classifier that consults it can only ever refuse to settle, which
+//     is how 14 `@stable` specs failed on the 2026-08-05 daily. Tests below pin both
+//     directions: an empty api_key must not block settling, and a stale one must not
+//     either.
 import { test } from "node:test";
 import assert from "node:assert/strict";
 import {
@@ -82,12 +88,15 @@ test("the fixture itself uses the real array-of-objects shape", () => {
   assert.equal(value[0]?.name, "gpt-4o-mini");
 });
 
-test("reads the credential and the SELECTED model names off the real shape", () => {
+test("reads the provider, the SELECTED model names and the credential off the real shape", () => {
   const probe = readAgentCredentialProbe(
     flowWithAgent("OPENAI_API_KEY", MODEL_VALUE_SHAPE("gpt-4o-mini")),
   );
   assert.deepEqual(probe, {
     credential: "OPENAI_API_KEY",
+    // The axis #1274 moved the guard onto — it comes off the SAME entry as the
+    // model name, so a payload that carries one carries the other.
+    selectedProviders: ["OpenAI"],
     selectedModels: ["gpt-4o-mini"],
   });
 });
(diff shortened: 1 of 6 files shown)


Victor-w-Madeira:
## Response to the Copilot review (`ce29710`)

All four findings are real and applied. One of them exposed a worse problem than itself.

**1. The note asserted something false on builds this helper supports — valid, fixed.**

Correct, and it is the sharpest of the four. The note read *"api_key is EMPTY on every build since upstream #14311"*, while this very change argues in its own commit message that `manual.yml` can dispatch a **pre-#14311** Langflow where the field IS written. On such a run the note would have contradicted the `observed` line printed directly above it — the exact mis-triage this module exists to prevent.

It now states what the guard *does* rather than what the field contains, with #14311 given as the reason an empty value is expected:

```
  observed       providers=[Anthropic] models=[claude-opus-5] api_key="ANTHROPIC_API_KEY"
  note           api_key is NOT asserted in either direction — the provider
                 above is the axis (#1274). Upstream #14311 stopped writing
                 this field on the 1.12 line, so an empty value here is
                 expected and is not the finding.
```

Verified against a probe carrying a non-empty `api_key`, which is the case the old wording broke on.

**2. The tests pinned the exact phrase — valid, and it is how the false claim shipped.**

Both assertions (and the suppressed third) matched the literal `"api_key is EMPTY"`, so they locked in the wrong claim instead of catching it. They now assert the intent, plus a **negative** assertion that the note must never claim the field is empty on every build.

**3. The stale `credential-only` bullet — valid, and verifying it found a worse bug.**

Applied. But checking it revealed that the fix for this exact bullet in `48aeebd` **never applied**: that commit rewrote three comment blocks with `str.replace()` and no assertion, so one silent miss went unnoticed while the other two landed. I audited all three, re-applied the missing one, and the substitution now asserts its anchor.

Worth recording because it is the same failure mode this repo keeps finding in its own guards — an edit that could not fail loudly did not fail loudly. Copilot found it by reading the file rather than the diff, which is precisely where that class of error hides.

**4. Portuguese comment — valid, fixed, and it was not alone.**

`index.ts:22` plus three lines in `provider-config.ts`. Both pre-existing, both in files this PR modifies, and `CLAUDE.md` allows no exception — so they are translated here rather than left sitting next to the diff. Portuguese elsewhere in the suite is out of this PR's scope and untouched.

---

Re-validated after the change: `google-provider.spec.ts` **2 passed** against the local nightly `1.12.0.dev16`; `typecheck` clean, `lint` 0 errors, `test:units` **442**, `test:scripts` **702**.

Issue #1265, opened 2026-08-04
[Daily #1258] modelInputComponent — the sidebar search input never renders (recurrent flake 2×)


Victor-w-Madeira:
Spun out of daily-failure triage #1258 (run [30901311395](https://github.com/oriontech-me/langflow-e2e/actions/runs/30901311395), 2026-08-04).

**Upstream:** _not filed_

## Symptom

A `@stable` test flaked (recovered on retry) waiting for the **component sidebar's search input** to become visible. Despite the test's name, the failure is upstream of anything about the Language Model node's model selector — the sidebar it needs to search in never appeared.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/llm-agents/modelInputComponent.spec.ts:63` ("the Language Model node renders its model selector") | `getByTestId('sidebar-search-input') to be visible` | `TimeoutError: locator.waitFor: Timeout 30000ms exceeded.` |

## Why these failures are one cause

One spec, one cause, filed because it meets the **recurrence criterion**: the same test failed with the **same `error_signature`** on two separate dailies in the 30-day window — `TimeoutError: locator.waitFor: Timeout 30000ms exceeded.` on **2026-07-15** and **2026-08-04**. `CONTRIBUTING.md` requires the cause, not just the count: the signature is identical on both, so it is recurrent rather than two unrelated flakes.

Worth recording because the test name misdirects: the wait that times out is `getByTestId('sidebar-search-input')`, i.e. the component sidebar's search field, reached before the Language Model node is ever configured. Anyone picking this up from the title alone would investigate the model selector, which is not where it fails.

## Preliminary read (descriptive — NOT a verdict)

The run carried a measured mid-run backend wedge (#1030/#1048): 13 outages, 12.4 min of unreachable backend across 3 of 4 shards, up to a 27% down-share on the affected shards. This test ran on shard 2, and the umbrella names `modelInputComponent.spec.ts` among the specs failing inside one of that shard's seven outage windows — a **lead, not a verdict**, at a 26.8% down-share.

The second data point deserves a caveat rather than being quoted flat: **2026-07-15** was itself a very broad failure day (15 hard failures, 29 flakes), so both same-signature occurrences land on days with heavy unrelated breakage. The recurrence criterion is met as written, and the quarantine follows from it, but the investigation should not treat "recurrent 2×" as proof of a test-local cause.

[...]

Issue #1002, opened 2026-07-28
loadTemplateByName is not concurrency-safe — random template spec fails in setup and leaks a flow


Victor-w-Madeira:
`tests/helpers/flows/load-template-by-name.ts` is the template-loading twin of the `setupPlayground` defect fixed in #988: under concurrency it fails **in setup**, before any assertion runs, taking down a *random* member of whichever template-loading specs happen to be running. #988 hardened the blank-flow/playground path; the template path was never touched.

Surfaced while verifying the five restorations on #992 — it is the reason that issue's load re-check could not come back clean.

## The failure

Always the same observable: the flow **is created** (`POST /api/v1/flows/` → 201, the id is captured), but the canvas never renders, so the gate two lines later times out:

```
tests/helpers/flows/load-template-by-name.ts:55
  page.waitForSelector('[data-testid="canvas_controls_dropdown"]', {timeout: 30000})
  → TimeoutError: Timeout 30000ms exceeded
```

The same wait in `tests/helpers/ui/adjust-screen-view.ts:11` fails identically when the caller reached it first. The page snapshot at failure shows the app sitting on the **flows list**, not the editor — the SPA never routed into the canvas it had just created a flow for.

Every run also logs `500 Internal Server Error` on `POST /api/v1/flows/` — `{"detail":"An internal error occurred while creating the flow."}` — the same backend signal #988 recorded.

## Evidence — it is the helper, not the tests

Langflow Nightly **1.12.0.dev7**, `MODEL_TEST_ID=gpt-4o-mini MODEL_TEST_PROVIDER=openai`, `--retries=0`.

| Run | Set | Result | Victim |
|---|---|---|---|
| `--workers=1` | the five #992 tests | **5 passed** | — |
| `--workers=2 --repeat-each=2` | same five | 9 passed / **1 failed** | `agent-component-regression:274` |
| `--workers=2 --repeat-each=2` | same five, repeated | 8 passed / **2 failed** | `agent-tool-error-handling:296`, `language-model-regression:68` |
| `--workers=2 --repeat-each=2` | **control**: 3 specs already `@stable` | 11 passed / **1 failed** | `agent-current-date-tool:383` |

Three different victims across two identical runs, and a **fourth victim that already carries `@stable`** — the random-victim signature of a shared-setup defect, not four independent flakes. Serial runs are always clean.

## It is already hitting the daily

[...]

PR #1320, opened 2026-08-05
fix(flows): verify the sidebar add landed and repair a swallowed click (#1304)

### LINKED ISSUE #1304: [Daily #1296] modelInputComponent — the canvas renders no node at all, so the model-name trigger is never reachable (hard failure, no outage cover)
Spun out of daily-failure triage #1296 (run [30997773754](https://github.com/oriontech-me/langflow-e2e/actions/runs/30997773754), 2026-08-05). Filed after the triage decided to **quarantine** this test (see *Preliminary read*), so that restoring it has an owner — the guard-day rule would otherwise leave a today-only hard failure unfiled.

**Upstream:** _not filed_

## Symptom

A `@stable` test **hard-failed** (all 3 attempts, retries exhausted) waiting for *any* node to exist on the canvas. Despite the test's name, the failure is upstream of anything about the model-name trigger: the assertion that fails is `expect(locator('[data-testid^="rf__node-"]').first()).toBeVisible()` with `element(s) not found` after 15 s, i.e. `addLanguageModelNode(page)` returned but the Language Model node is not on the canvas. The trigger this test exists to check was never reached.

|---|---|---|
| `tests/tests-automations/regression/core-functionality/llm-agents/modelInputComponent.spec.ts:106` ("the trigger shows the selected model name") | `locator('[data-testid^="rf__node-"]').first() to be visible (timeout 15000ms)` | `Error: expect(locator).toBeVisible() failed` |

One spec, one test, one cause. Filed as its own issue rather than folded anywhere, for two reasons worth recording:

1. **Not #1265, despite the same file.** #1265 tracks `modelInputComponent.spec.ts:63` ("the Language Model node renders its model selector"), whose wait is `getByTestId('sidebar-search-input')` — the *component sidebar* never opening — with signature `TimeoutError: locator.waitFor: Timeout 30000ms exceeded.`. This is a different test, a different observable (the node on the canvas, not the sidebar) and a different signature. The two are plausibly related — both are "the editor never reached a usable state in this spec" — and if the investigation finds one mechanism, merge into #1265 (keep the older number) and carry this spec path over.
2. **Not the run's credential cluster.** The dominant cluster of this daily (14 failures) is the `#751` credential guard against upstream #14311 (#1274). This failure never reaches a credential check — it fails before any node exists — and its signature is unrelated.

A **hard failure**, not a flake: it failed on all 3 attempts, so the retry budget did not absorb it.

**It has no outage cover, and that was measured rather than assumed.** The run carried a mid-run backend wedge (#1030/#1077: 35 outages, 15.8 min of unreachable backend, up to a 32.8% down-share on 2 of 4 shards). This test ran on **shard 3**, whose two measured outage windows were `10:36:37→10:38:25` and `10:42:39→10:44:11` UTC. Its three attempts ran `10:47:57→10:48:18`, `10:48:19→10:48:49` and `10:48:51→10:49:14` — **zero overlap** with either window. So the wedge is not an available explanation here without new evidence.

**`@stable` was removed and `test.fixme` added at triage, and the reason is procedural rather than a verdict.** The mass-failure guard **tripped** on this run (24 hard failures against a threshold of 5), which suppresses the workflow's automatic `@stable` removal for every hard failure — so this tag survived only because of the guard. The triage's day-level verdict is that 2026-08-05 was **not** environmental (its dominant cluster is the #1274 premise break, not the environment), and on a normal daily this tag would have been removed automatically. The quarantine restores that outcome; it asserts nothing about the cause.

This is a **first occurrence** — no same-signature recurrence anywhere in the 30-day window. Read the single data point as such: it is a genuine final hard failure, but one run is not a pattern, and `expect(locator).toBeVisible() failed` is the most generic string Playwright emits.

## Investigation directive

1. On the current nightly, add a Language Model node to a blank flow repeatedly and confirm whether it always lands on the canvas. Rule out a product regression in adding that component before touching the spec — note that the 1.12 line has been moving components into per-vendor distributions (#1040), so a node that silently fails to instantiate is a live possibility, not a hypothetical.
2. Check whether the failure reproduces at all. It is a first occurrence with retries exhausted on one run; if it does not reproduce, the deliverable is to establish *why it failed that day* well enough to justify restoring `@stable`, not simply to observe a green run.
3. Only then, inspect `addLanguageModelNode()`: whether it can return before the node is committed to the canvas, in which case the helper needs a deterministic post-condition (this helper is shared, so a change there reaches other specs — check its importers).
4. Decide the relationship to **#1265**: if one mechanism explains both this test and `:63`, merge into #1265 and carry this spec path over; if not, record why they are separate.

## Deliverables (Done when)

- [ ] The relationship to #1265 is decided and recorded: merged into it (carrying this spec path over), or kept separate with the reason stated.
- [ ] Because this is a first occurrence, restoring `@stable` is justified by evidence about *why it failed on 2026-08-05* — not by a green run alone.

rafaelgiln:
Closes #1304.

## Problem

`modelInputComponent.spec.ts:106` ("the trigger shows the selected model name") hard-failed on the 2026-08-05 daily ([run 30997773754](https://github.com/oriontech-me/langflow-e2e/actions/runs/30997773754)) on **all 3 attempts**, waiting for `[data-testid^="rf__node-"]`: `addLanguageModelNode` returned and the canvas had no node.

The defect is not in that test, and not in this file. **`addComponentFromSidebar` was a fire-and-forget `fill` + `click` with no post-condition, and Langflow drops that click outright a measurable fraction of the time** — the DOM click is accepted, the app never registers the add, and no node reaches the canvas. Root-caused with an instrumented scout on nightly `1.12.0.dev17`:

| Measurement | Result |
|---|---|
| Scout, 20 iterations (add → poll 4 s → on a miss, capture state and re-issue) | **4 of 20** produced no node; in **all 4** an identical second `fill`+`click` produced it (`nodes=1`), with the `+` button still visible, the search input still holding `"language model"`, and **zero** `POST/PATCH /api/v1/flows` after the first click in 3 of the 4 |
| Pre-fix solo baseline, `--retries=0 --workers=1` | **2 of 11** runs, exact issue signature — and one of them failed on a **different** test of the file (`:93`), not the quarantined one |
| Contended shape post-fix, 3 rounds `PW_SHARD_FILE_LEVEL=1 --workers=2` vs `openai-compatible-provider-setup.spec.ts` | **12/12** victim executions passed (pre-fix measurement in that shape: 3/12 failed) |

That is the swallowed-click class (#420/#966) one layer later, on the very surface #537 recorded as re-rendering while its component catalog streams in. Whichever test of a file is running when the drop happens is the one that reports it, which is why this looked like several problems: on that daily it cost **three specs on one shard inside 100 s under three unrelated messages** — `flow-functionality/stop-building.spec.ts:24` on `div-generic-node`, `ui-ux/langflowShortcuts.spec.ts:47` on "the Chat Output component should be on the canvas", and this test on `[data-testid^="rf__node-"]` — none of them naming the add.

**Why it hard-failed that day rather than flaking** (the issue's own bar for restoring `@stable`): the window WAS degraded. The triage's "no outage cover" rests on the liveness recorder's probe-measured windows (`10:36:37→10:38:25`, `10:42:39→10:44:11`), which by construction miss degradation that does not fail the probe — a limitation `tests/helpers/other/page-entry-barrier.ts` states about itself. In the same 100 s the run also lost the two specs above, `openai-compatible-provider-setup.spec.ts:620` after 48 s and `mcp-client-agent-gemini-tool-regression.spec.ts:139` three times at 32–40 s, while 152–157 s page-entry barrier failures straddled the window. This file's own siblings passing on the **same worker** (w17) at `10:47:39` and `10:47:50` rule out instance-wide breakage and the spec's own logic.

**Ruled out:** a product regression in instantiating the component (it lands solo, and every drop was repaired by an identical second click — #1040's per-vendor distribution move is not implicated); cross-worker destructive cleanup; and **provider-credential churn**, the standing hypothesis carried in this spec's comment — the drop reproduces solo with no neighbour at all, and the same class hit `core-components/edit-name-description-node.spec.ts:42` on that daily with **no** provider-mutating spec running.

**#1265 stays a separate issue**, as this issue's item 4 asked us to decide: its observable is the component sidebar never opening, upstream of any add.

## Fix

`tests/helpers/flows/add-component-from-sidebar.ts` snapshots the node count, re-issues the interaction **once** when it did not move, and otherwise fails naming the swallowed click plus the observed sidebar state.

Rejected alternatives, explicitly:

- **A longer wait / raised timeout.** The first attempt's failure is not slowness — nothing is in flight to wait for. This test already waited 15 s and lost 3/3.
- **A `try/catch` or a retry in the spec.** It would fix one of 23 call sites and leave the mechanism unnamed in the other 22.
- **Claiming infra in the message.** Deliberately not done, per the #1262 rule: an `INFRA_PREFIX` here would exempt a genuine regression in adding components from `@stable` auto-removal. A unit test pins the message as unclassifiable by `scripts/lib/infra-signatures.ts`.
- **An `expectNewNode` opt-out knob.** All 23 importers were audited — every one expects the add to produce a node (`toHaveCount(1/2/before+1)` or a title visible); `singleton-components.spec.ts` checks the refusal path via the `+` button disappearing, never by a second add through this primitive. So the post-condition is default-on.

The decision and message builders are pure and unit-tested next to the helper (`add-component-from-sidebar.test.ts`, 9 assertions, `npm run test:units`), in the idiom of `page-entry-barrier.test.ts`. `classifyAddOutcome` reports `duplicated` and `lost` as distinct outcomes rather than folding them into "swallowed", because a canvas with 2 nodes and a canvas that lost one are different problems and the re-issue is what can produce the former.

## Scope note

Quarantine from the triage of #1296 (PR #1308) is **lifted**: `test.fixme` removed and `@stable` restored on "the trigger shows the selected model name". **No assertion was weakened, no selector loosened and no timeout raised** — anywhere. The four tests' observables are byte-identical to before the quarantine. The spec doc's step 4 also drops a claim that was already known false ("key-independent"): with no provider credential the field renders a **Setup Provider** CTA and `model_model` is absent (#1265).

## Product observation (not claimed as a regression)

A user clicking the sidebar `+` gets no node ~20 % of the time under load on `1.12.0.dev17`, silently. There is **no evidence this is new in 1.12** — the class is recorded in this repo since #420/#966 and, for this exact surface, #537 — so it is reported here with its measurement (4/20, second click repairs 4/4) for the team to decide whether it becomes an upstream question. The suite-side change makes it **louder**, not quieter: the failure now points at the helper with the mechanism named instead of surfacing as some caller's generic 15 s visibility timeout.

## Validation (nightly `1.12.0.dev17`, `--retries=0 --workers=1`)

- typecheck ✅ · eslint ✅ (0 errors) · `npm run test:units` **451/451** ✅ — all re-run after rebasing onto current `main`
- **3 clean burst runs** of the touched spec, `expected=4 unexpected=0 flaky=0` ✅ · final green run after FF reverts ✅
- **Force-fail 5/5**, each reverted (`grep FF-MUTATION` empty):
  - `model_model` → `model_model_ff` → failed as expected
  - catalog floor `toBeGreaterThan(1)` → `toBeGreaterThan(9999)` → failed as expected
  - `manage-model-providers` → `manage-model-providers-ff` → failed as expected
  - placeholder assertion inverted (`not.toMatch` → `toMatch`) → failed as expected
  - **the new failure path**: `click()` → `hover()` in the primitive (button resolved, no click delivered) → failed **at the helper** with `the sidebar add was swallowed: no node reached the canvas after 2 attempt(s) of 12000ms each on getByTestId("add-component-button-language-model") … node count: 0 before, 0 after; "+" button still visible: yes; sidebar search input: "language model"`
- Contended shape (the reproducing neighbour): **12/12** of this file's executions passed
- Flow cleanup: orphan count checked via `GET /api/v1/flows/` after every batch — **0 leaked by this file** across ~15 runs. The contended rounds leaked 2 flows, both the **neighbour's** (`Basic Prompting (1)` from its `createFlowFromStarter`, plus the untracked `New Flow` its `awaitBootstrapTest` creates per #1002); purged id-scoped, instance back to 26 starter projects. Worth a follow-up on that spec — out of scope here.
- **`🚨 Backend Error`: not zero, and pre-existing.** Runs intermittently log `500 - /api/v1/flows/` (no id in the path, so not this suite's `deleteFlow`, which uses `/api/v1/flows/{id}` and already retries). The backend log names it: `RuntimeError: Unable to cascade delete flow: <id>` from `sqlite3.OperationalError: database is locked` — the **frontend** bulk-deleting the transient flow that clicking "New Flow" creates (#1002) while another write holds the SQLite lock. Measured **2 of 6 PRE-fix runs** on this instance with the original helper and the test still quarantined. It became more frequent after the lift for a structural reason: the restored 4th test takes the file from 6 to 8 flows created and cascade-deleted per run. Every affected run still had all 4 tests green.

### DIFF
--- a/docs/core-functionality/model-provider/provider-management.md
+++ b/docs/core-functionality/model-provider/provider-management.md
@@ -227,10 +227,15 @@ sidebar entry to the canonical **Language Model** component:
 > page-entry barrier failures straddled it. This file's own siblings passing on the
 > **same worker** 7 s and 17 s earlier rule out instance-wide breakage.
 > **The fix is in `tests/helpers/flows/add-component-from-sidebar.ts`, not here:**
-> it verifies the node count moved, re-issues the fill+click **once** when it did
-> not, and otherwise fails naming the swallowed click and the observed sidebar
-> state (unit-pinned in `add-component-from-sidebar.test.ts`, including that the
-> message is deliberately **not** infra-classifiable, per the #1262 rule). No
+> it requires a node **id that was not on the canvas before** to appear, re-issues
+> the fill+click **once** when none did, and otherwise fails naming the swallowed
+> click and the observed sidebar state (unit-pinned in
+> `add-component-from-sidebar.test.ts`, including that the message is deliberately
+> **not** infra-classifiable, per the #1262 rule). The post-condition is a set
+> difference rather than a `before + 1` count delta because a canvas can still be
+> mounting a loaded flow's own nodes when the baseline is taken — CI run
+> 31048371247 reported a healthy one-click add against an API-seeded flow as "left
+> 3 nodes on the canvas instead of 1". Exact counts stay with the callers. No
 > assertion here was weakened and no timeout raised — a longer wait provably cannot
 > fix it, since this test already waited 15 s and lost 3/3. Provider-credential
 > churn, the standing hypothesis, is **refuted**: the drop reproduces solo with no
--- a/docs/core-functionality/model-provider/provider-management.md
+++ b/docs/core-functionality/model-provider/provider-management.md
@@ -1,6 +1,6 @@
 # Provider Management — modal, provider count, components, add/remove API key
 
-**Last validated:** Langflow 1.12.x
+**Last validated:** Langflow 1.12.x (nightly `1.12.0.dev17`)
 
 ---
 
@@ -163,7 +163,13 @@ sidebar entry to the canonical **Language Model** component:
    the route into provider configuration.
 4. *Trigger shows the selected model* — assert `model_model` text is a
    non-empty model name and is NOT a "Select a model" placeholder (the catalog
-   pre-selects a default, key-independent).
+   pre-selects a default). **Not key-independent**, as this step used to claim:
+   with no provider credential configured at all the node still mounts but its
+   Language Model field renders a **"Setup Provider"** CTA behind
+   `parameter-permission-gate` and `model_model` is absent, so all four tests
+   fail on an observable unrelated to what they assert (measured on
+   1.12.0.dev17, #1265). It is the credential that matters, not a funded key — a
+   drained `openai` alongside an active anthropic/google passes.
 
 > **Flake verdict & attributed sidebar barrier (issue #1265).** Test 1 flaked on
 > the 2026-07-15 and 2026-08-04 dailies with the same signature
@@ -195,6 +201,42 @@ sidebar entry to the canonical **Language Model** component:
 > `tests/helpers/other/page-entry-barrier.test.ts` against the real classifier.
 > The residual limitation is stated in that helper's header: the probe runs
 > after the budget is spent, so a wedge shorter than the wait reads healthy.
+
+> **Verdict on the no-node hard failure (issue #1304).** Test 4 hard-failed on the
+> 2026-08-05 daily (run 30997773754, all 3 attempts) waiting for
+> `[data-testid^="rf__node-"]`, i.e. `addLanguageModelNode` returned and the canvas
+> had no node. Root cause is **the shared sidebar add dropping its click**, not
+> this file and not the model selector: Langflow accepts the click on
+> `add-component-button-language-model`, never registers the add, and no flow write
+> follows. An instrumented scout on nightly `1.12.0.dev17` measured **4 of 20**
+> adds producing no node within 4 s, and in **all 4** an identical second
+> fill+click produced it — with the `+` button still visible, the search input
+> still holding the term, and zero `POST/PATCH /api/v1/flows` after the first click
+> in 3 of the 4. That is the swallowed-click class (#420/#966) one layer later, on
+> the surface #537 already recorded as re-rendering while its catalog streams in.
+> Pre-fix solo baseline: **2 of 11** runs at `--workers=1 --retries=0`, and one of
+> them failed on a *different* test of this file (test 3) — whichever test is
+> running when the drop happens is the one that reports it.
+> **Why it hard-failed that day rather than flaking:** the window WAS degraded,
+> which the triage could not see because the liveness recorder's probe-measured
+> windows (10:36:37→10:38:25, 10:42:39→10:44:11) by construction miss degradation
+> that does not fail the probe. Inside 100 s on the same shard the run also lost
+> `flow-functionality/stop-building.spec.ts:24` (on `div-generic-node`) and
+> `ui-ux/langflowShortcuts.spec.ts:47` ("the Chat Output component should be on the
+> canvas") — the **same mechanism under two more messages** — while 152–157 s
+> page-entry barrier failures straddled it. This file's own siblings passing on the
+> **same worker** 7 s and 17 s earlier rule out instance-wide breakage.
+> **The fix is in `tests/helpers/flows/add-component-from-sidebar.ts`, not here:**
+> it verifies the node count moved, re-issues the fill+click **once** when it did
+> not, and otherwise fails naming the swallowed click and the observed sidebar
+> state (unit-pinned in `add-component-from-sidebar.test.ts`, including that the
+> message is deliberately **not** infra-classifiable, per the #1262 rule). No
+> assertion here was weakened and no timeout raised — a longer wait provably cannot
+> fix it, since this test already waited 15 s and lost 3/3. Provider-credential
+> churn, the standing hypothesis, is **refuted**: the drop reproduces solo with no
+> neighbour, and the same class hit `edit-name-description-node.spec.ts:42` on that
+> daily with no provider-mutating spec running. #1265 stays a separate issue — its
+> observable is the sidebar never opening, upstream of any add.
 
 **`language-model-regression.spec.ts`**
 1. *OpenAI answers* (skip without `OPENAI_API_KEY`) — Basic Prompting +
--- a/tests/helpers/flows/add-component-from-sidebar.test.ts
+++ b/tests/helpers/flows/add-component-from-sidebar.test.ts
@@ -1,0 +1,116 @@
+// Unit tests for the sidebar-add post-condition (issue #1304).
+// Run with: npm run test:units
+//
+// What rides on these functions: whether a dropped sidebar click is REPAIRED and
+// named, or shows up as some caller's generic "the node is not visible" timeout.
+//
+// Measured on nightly 1.12.0.dev17 (issue #1304): 4 of 20 adds of the Language
+// Model component produced no node within 4 s, and in all 4 an identical second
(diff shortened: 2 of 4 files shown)


rafaelgiln:
## CI caught a design flaw in the first commit — fixed in d252050

The E2E lane was red for **one** reason, and it was mine, not ambient: the post-condition required `after === before + 1` and hard-failed anything else as a duplicated add.

```
4) agent-context-id-isolation.spec.ts:512 › Context ID isolation — retrieval layer (model-free)
   › add a Message History node pinned to the seeded session

Error: the sidebar add left 3 nodes on the canvas instead of 1 after 1 attempt(s) of 12000ms each
       on getByTestId("add-component-button-message-history") …
   at helpers/flows/add-component-from-sidebar.ts:190
```

`setupMessageHistoryNode` navigates to an **API-seeded flow** and waits only for `sidebar-search-input`, so the canvas is still mounting the flow's **own** nodes when the baseline is taken. One legitimate click took the count from 0 to 3, and a healthy add was reported as a duplicate — on both attempts. A baseline that can grow on its own is no evidence of a double add.

**The post-condition is now a set difference over node ids**: the add landed when an id that was not on the canvas before is there now. `duplicated` and `lost` are gone as outcomes — exact counts belong to the callers, which already assert them (`toHaveCount(1 | 2 | before + 1)`). Swallowed-click detection, the single bounded re-issue and the attributed message are unchanged, so nothing about the #1304 fix itself changed.

The failure is encoded as a unit test, along with the case a count delta gets wrong in the *other* direction (a node vanishing while a new one appears is a landed add, where a delta would call it swallowed and click again).

**Re-validated:** `npm run test:units` 449/449 · typecheck ✅ · eslint 0 errors · the spec CI failed on passes in 12 s · 15/15 on `singleton-components` + `canvas-connect-components` + `componentDelete` (the call sites asserting exact counts) · 4/4 on `modelInputComponent`.

### The other three entries in that run were not this PR

- `agent-context-id-continuity.spec.ts:405` (flaky, passed on retry) — `Agent credential never settled on the persisted flow (#751 guard, #1072)`, the #1274 credential cluster; this lane's `openai` key is drained (`You have no credits remaining`) and `google` lost `gemini-2.5-flash` to a retirement notice.
- `agent-context-id-isolation.spec.ts:557` (flaky, passed on retry) — same credential cluster.
- `parameters-panel-field-types.spec.ts:415` "table field edit persists" (flaky, passed on retry) — its own predicate timeout, unrelated to the sidebar add.

Also visible in that log and pre-existing: `🚨 Backend Error: 500 - /api/v1/flows/`, which the PR body already documents (`sqlite3.OperationalError: database is locked` on the frontend's bulk delete of the #1002 transient flow).

Issue #1178, opened 2026-07-31
triage-dispatch: the agent pays full price to triage a daily that produced zero results, on an unpinned Opus


rafaelgiln:
## Problem

The triage agent is the largest consumer of the shared `ANTHROPIC_API_KEY`, and on 2026-07-31 it paid its **maximum** price to triage a daily that produced **zero test results**.

Measured from the action's own log (run [30624377765](https://github.com/oriontech-me/langflow-e2e/actions/runs/30624377765)):

```
"model": "claude-opus-5[1m]"
"num_turns": 28,
"total_cost_usd": 1.6609247499999995
```

28 of the 30 allowed turns, on the 1M-context Opus, for **$1.66** — and the input it was reasoning over did not exist:

```
::warning::results.json artifact unavailable for run 30624101377
no results.json — history-only.
```

The triggering daily ([30624101377](https://github.com/oriontech-me/langflow-e2e/actions/runs/30624101377)) had all four shards fail without producing a blob, so the merge job died on `No report files found in .../all-blobs` and never uploaded `results.json`. With no digest, the agent reconstructed the run from scratch through `Bash` — which is what filled the transcript and consumed the turn budget.

Cost of the triage agent over the window in which the key drained:

| Date | Turns | Cost |
|---|---|---|
| 2026-07-29 | 25 | $2.06 |
| 2026-07-30 | 16 | $0.99 |
| 2026-07-31 | 28 | $1.66 |
| **Total** | | **$4.71** |

For comparison, the E2E suite's entire Anthropic spend on 07-31 was ~$0.25 (61k sonnet-5 + 41k haiku-4-5 tokens). The 1.8M opus-5 tokens on the console are dominated by cache reads, which is why 1.8M tokens cost $1.66 rather than $9 — **token count is not the bill here, and the fix is not a smaller model on its own.**

## Fix

Two independent changes, both in `triage-dispatch.yml`.

### 1. Gate the agent on a usable input

`propose` currently degrades to "history-only" when the artifact is absent and runs the agent anyway. There is nothing to triage in that state: a run that recorded zero test results has no failures to group and no clusters to dedup. The gate decides *before* the agent starts, reusing `analyze()` from `scripts/check-run-integrity.mjs` (the same `runguard` logic the daily already trusts) rather than re-implementing report parsing:

| Report state | Decision |
|---|---|
| artifact absent | **skip the agent** — the daily produced nothing |
[...]

PR #1368, opened 2026-08-07
ci(triage): bill the agent to the subscription, not the shared key

### LINKED ISSUE #1367: Triage agent costs ~$1.78/business day on the shared Anthropic key — move it to subscription auth
## What was measured

An Anthropic Console usage breakdown for **2026-08-07** showed **2,885,814 tokens on `claude-sonnet-5`** — 98.5% of the day's 2,928,587 total. The key is configured in `.env` and as `secrets.ANTHROPIC_API_KEY`, and is used only by this repo, so the question was where the volume came from.

It came from **one run**.

| Fact | Value |
|---|---|
| Anthropic agent runs that day | 1 — `triage-dispatch.yml` job `Propose triage plan`, run [31165533824](https://github.com/oriontech-me/langflow-e2e/actions/runs/31165533824) |
| Result envelope | `num_turns: 28`, `duration_ms: 443332`, **`total_cost_usd: 1.7802987`** |
| Model | `claude-sonnet-5`, pinned at `triage-dispatch.yml:119` / `:152` (#1178) |
| Spec-side provider tokens, same period | **8,741** for a whole daily (`reports/token-history.jsonl`), and on `gpt-4o-mini` / `gemini-flash-latest` — not Anthropic |

$1.78 is consistent with ~2.9M tokens under the Sonnet introductory band once cache reads dominate, which is what 28 turns re-reading CI logs produces.

Two corroborations that this is the whole story:

- `claude-haiku-4-5` at 41,551 tokens is the spec suite (local `.env` points at Anthropic/haiku while the OpenAI key is drained) — three orders of magnitude below the agent.
- `claude-opus-5` at **862** tokens proves local Claude Code is **not** billing this key. If it were, that line would be the largest one on the chart.

This is a fresh, larger data point for **axis A of #1183** (the triage agent measured at ~10x the entire spec suite): the historical mean was ~$1.05/run, this one was $1.78.

## Why it recurs

`propose` auto-fires on every red scheduled daily, and the daily is red on essentially every business day — 9 agent runs across the 9 business days from 07-24 to 08-07. So the standing cost is **~$1.78 per business day (~$37/month)**, and it steps up **~50% on 2026-09-01** when the Sonnet introductory band expires, with no change in consumption.

## Decision

Switch both agent jobs from `anthropic_api_key` to **`claude_code_oauth_token`**, the action's documented alternative (`anthropics/claude-code-action@v1`, `action.yml`):

```yaml
claude_code_oauth_token:
  description: "Claude Code OAuth token (alternative to anthropic_api_key)"
```

The job then draws on an already-paid Claude Code subscription. **Marginal dollar cost of the triage agent becomes zero.**

Two alternatives considered and not taken:

- **Drop the CI agent, run `/langflow-e2e-triage` locally.** Also $0, but it loses the property the propose phase exists for — the plan is already commented on the umbrella when someone opens the day.
- **Keep the API key and cut consumption** (gate the trigger on a changed failure signature; truncate what the agent reads; A/B haiku). Reduces, does not zero, and #1178 is explicit that a cheaper model must be validated rather than assumed.

## What this costs instead

Not free — **relocated**, and each of these is a real trade:

1. The run consumes the **subscription's rate-limit window**, shared with interactive sessions. The daily fires ~09:20 BRT and the measured run was 2.9M tokens.
2. The spend **leaves the Console dashboard**. Silver lining: `ANTHROPIC_API_KEY` remains for the specs' provider calls only, so a future spike there is unambiguously the specs.
3. **The OAuth token expires.** When it does, the step fails and a red daily goes untriaged until someone notices the failed workflow run.

## Required before merge

The secret does not exist yet (`gh secret list` shows no `CLAUDE_CODE_OAUTH_TOKEN`). Merging first would break both jobs.

```bash
claude setup-token
gh secret set CLAUDE_CODE_OAUTH_TOKEN --repo oriontech-me/langflow-e2e
```

rafaelgiln:
Closes #1367.

## What this changes

Two lines of auth in `.github/workflows/triage-dispatch.yml` — `propose` and `execute` both move from `anthropic_api_key` to `claude_code_oauth_token`. Nothing else: same triggers, same gates, same `--model claude-sonnet-5` pin, same `--allowedTools "Read,Bash"`, same turn caps.

## Why

The Console breakdown for 2026-08-07 showed **2,885,814 tokens on `claude-sonnet-5`**, 98.5% of the day. It was **one run** — `Propose triage plan`, [31165533824](https://github.com/oriontech-me/langflow-e2e/actions/runs/31165533824), `num_turns: 28`, **`total_cost_usd: 1.7802987`**, cache reads dominating a job whose entire input is CI logs.

The spec suite over the same period spent **8,741 tokens** (`reports/token-history.jsonl`), on `gpt-4o-mini` / `gemini-flash-latest`. The chart's `claude-haiku-4-5` 41,551 is the specs; its `claude-opus-5` **862** proves local Claude Code is not billing this key at all.

That is **axis A of #1183** measured again and higher — the historical mean was ~$1.05/run. `propose` auto-fires on every red scheduled daily and the daily is red on essentially every business day (9 agent runs across the 9 business days to 08-07), so the standing cost is **~$1.78/business day**, stepping up ~50% on **2026-09-01** when the Sonnet introductory band expires with no change in consumption.

`claude_code_oauth_token` is the action's own documented alternative to `anthropic_api_key` (`anthropics/claude-code-action@v1`, `action.yml`), drawing on a subscription that is already paid for. Marginal dollar cost of this job → **zero**.

## What it costs instead

Relocated, not eliminated. All three are recorded at the call site so they are not rediscovered as surprises:

1. The run consumes the **subscription's rate-limit window**, shared with interactive sessions. The daily fires ~09:20 BRT and this run was 2.9M tokens.
2. The spend **leaves the Console dashboard**. `ANTHROPIC_API_KEY` stays for the specs' provider calls, so a future spike there is unambiguously the specs.
3. **The OAuth token expires.** The step then fails and a red daily goes untriaged until someone reads the failed run. Re-mint with `claude setup-token`.

## ⚠️ Do not merge before the secret exists

`gh secret list` shows no `CLAUDE_CODE_OAUTH_TOKEN` today. Merging first breaks both agent jobs.

```bash
claude setup-token
gh secret set CLAUDE_CODE_OAUTH_TOKEN --repo oriontech-me/langflow-e2e
```

## Alternatives rejected

- **Drop the CI agent, run `/langflow-e2e-triage` locally** — also $0, but loses the reason propose exists: the plan is already on the umbrella when someone opens the day.
- **Keep the key, cut consumption** (gate the trigger on a changed failure signature, truncate the logs the agent reads, A/B haiku) — reduces without zeroing, and #1178 is explicit that a cheaper model must be validated rather than assumed.

## Verification

YAML parses and both jobs carry the new key, with no residual reference to the old one:

```
propose ['claude_code_oauth_token', 'prompt', 'claude_args']
execute ['claude_code_oauth_token', 'prompt', 'claude_args']
triggers: ['workflow_run', 'issue_comment', 'workflow_dispatch']
```

No unit guard references `anthropic_api_key` for this workflow (grepped `scripts/**/*.test.{mjs,ts}`), so nothing structural pins the old spelling.

**End-to-end proof is only available post-merge** — `workflow_run` and `issue_comment` triggers fire exclusively from the default-branch copy of the workflow file, as this workflow's own header comment records. Once the secret is set and this merges, the honest check is a `workflow_dispatch` of `propose` against an existing umbrella issue.

## Not covered here

Nothing detects the token expiring except a human reading a failed workflow run. Worth a follow-up if that turns out to bite.

### DIFF
--- a/.github/workflows/triage-dispatch.yml
+++ b/.github/workflows/triage-dispatch.yml
@@ -105,7 +105,28 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         with:
-          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          # Subscription auth, not the API key. This job is axis A of #1183 — the
+          # measured ~10x of the whole spec suite — and on 2026-08-07 one propose
+          # run billed $1.78 (28 turns, 2.9M tokens against `claude-sonnet-5`,
+          # cache reads dominating). It fires on essentially every red daily, so
+          # the recurring cost is ~$1.78 per business day and steps up ~50% on
+          # 2026-09-01 when the Sonnet introductory band expires, with no change
+          # in consumption. `claude_code_oauth_token` is the action's documented
+          # alternative to `anthropic_api_key`, and it draws on a Claude Code
+          # subscription that is already paid for, so the marginal dollar cost of
+          # this job becomes zero.
+          #
+          # What it costs instead, so nobody rediscovers it as a surprise:
+          #   - the run consumes the SUBSCRIPTION's rate-limit window, shared with
+          #     interactive sessions. The daily fires ~09:20 BRT.
+          #   - this spend leaves the Console usage dashboard. `ANTHROPIC_API_KEY`
+          #     still exists for the specs' provider calls (tens of thousands of
+          #     tokens a day, per reports/token-history.jsonl) — so a spike there
+          #     is now unambiguously the specs, never the agent.
+          #   - the OAuth token EXPIRES. When it does, this step fails and the red
+          #     daily goes untriaged until a human notices. Re-mint with
+          #     `claude setup-token` and `gh secret set CLAUDE_CODE_OAUTH_TOKEN`.
+          claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           prompt: "/langflow-e2e-triage --phase propose ${{ inputs.issue && format('--issue {0}', inputs.issue) || '' }}"
           # --model pinned (#1178): the action's default is `claude-opus-5[1m]`, and
           # nothing here needs Opus-tier reasoning at Opus-tier price. Sonnet 5
@@ -144,7 +165,10 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         with:
-          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          # Same subscription auth as propose, and for the same reason — see the
+          # long note on that step. This one runs at --max-turns 40, so it is the
+          # more expensive of the two whenever a human approves a plan.
+          claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           prompt: "/langflow-e2e-triage --phase execute --issue ${{ github.event.issue.number }}"
           # Same pin as propose (#1178). NOT gated on a report: execute is triggered
           # by a human "pode abrir" on a plan that already exists, so it has


rafaelgiln:
**Correction to this PR's Verification section**, recorded here rather than silently editing the body.

It said end-to-end proof was only available post-merge. That is wrong for one of the three triggers, and it is the one that mattered: `workflow_run` and `issue_comment` do fire exclusively from the default-branch copy, but **`workflow_dispatch` does not** — this workflow's own header comment says so explicitly ("Pre-merge, use the workflow_dispatch entry to test `propose` on a branch"), and the `propose` gate accepts a dispatch from any of the three trusted actors regardless of ref. So the pre-merge validation the PR-lane classifier asked for was available on the branch:

```
::warning:: CI-only change to .github/workflows/triage-dispatch.yml, which THIS lane does not
run — nothing here proves it works. Dispatch .github/workflows/triage-dispatch.yml on this
branch before merging (#1159).
```

The practical consequence is unchanged by the merge — **the new auth is still unproven**. `secrets.CLAUDE_CODE_OAUTH_TOKEN` exists (set 2026-08-07T12:19:48Z) and `main` now carries `claude_code_oauth_token` on both jobs, but no run has exercised it. First live exercise is whichever comes first: the next red scheduled daily's automatic `propose`, or a manual dispatch.

Validating on demand, against an existing umbrella:

```bash
gh workflow run triage-dispatch.yml --repo oriontech-me/langflow-e2e -f issue=<umbrella>
```

If the token is wrong or expired, the `anthropics/claude-code-action@v1` step fails at auth — loudly, in Actions — rather than degrading to a silent no-triage.

Issue #1275, opened 2026-08-04
run-e2e drops half of test_grep when it contains an alternation — the lane runs a silently narrower selection


Victor-w-Madeira:
`.github/actions/run-e2e/action.yml` combines the two filters by interpolating them into one lookahead pair, without parenthesising either side:

```bash
if [ -n "$TEST_TAG" ] && [ -n "$TEST_GREP" ]; then
  ARGS+=(--grep "(?=.*${TEST_TAG})(?=.*${TEST_GREP})")
```

Alternation has the lowest precedence inside a group, so a `test_grep` containing `|` splits the lookahead instead of being ANDed as a unit. Dispatching `test_tag=@stable`, `test_grep=@agents|@model-provider` produced:

```
--grep (?=.*@stable)(?=.*@agents|@model-provider)
```

which reads as `(?=(?:.*@agents)|(?:@model-provider))` — the second branch has to match at position 0, so it never matches a test title and the whole `@model-provider` half is dropped.

Measured locally:

| `--grep` | Selected |
|---|---|
| `(?=.*@stable)(?=.*(@agents\|@model-provider))` | **100 tests / 35 files** |
| `(?=.*@stable)(?=.*@agents\|@model-provider)` (what the action builds) | **75 tests / 27 files** |
| `(?=.*@stable)(?=.*@agents)` | **75 tests / 27 files** — identical |

The failure mode is the one this repo keeps legislating against: the run is **silently narrower** than the dispatch asked for, and the summary reports success over a smaller set (#1012's rule). It cost real coverage during the 1.11.2rc3 release validation — a dispatch aimed at the provider-parametrized specs ran only the `@agents` ones, and nothing in the log or the summary said so.

## Fix

Wrap each interpolated fragment before composing: `(?=.*(?:${TEST_TAG}))(?=.*(?:${TEST_GREP}))`. The destructive lane a few lines below builds its filter the same way (`FILTER="${FILTER}(?=.*${TEST_GREP})"`) and needs the same treatment.

Worth covering in `npm run test:scripts` if the composition moves into a script; as long as it lives in the YAML, a structural assertion on the built string is the cheapest guard — with the caveat that pinning a spelling is not pinning a behaviour (#1226).

## Where it came from

Found while dispatching `manual.yml` for the 1.11.2 release validation; the input was accepted, the job was green-path, and the narrowing only surfaced when the executed count was compared against the local selection.

PR #1373, opened 2026-08-07
fix(collect-models): stop one stalling provider from costing the PR lane its whole E2E run (#1370)

### LINKED ISSUE #1370: collect-models: the credential write is still missed after #1357, and it costs the PR lane its whole E2E run (recurrence of #1355)
Spun out of #1334's PR validation (#1369). Not a new mechanism — a recurrence of #1355 after #1357 shipped.

## Symptom

`pr-validation.yml`'s E2E job exits at its shared **`Collect models`** pre-flight, so **the lane never runs the PR's specs at all** — its step list goes `Collect models` → `Run destructive lane`, with no impacted-specs step in between.

```
build axis: ✅ openai
build axis: ✅ anthropic
build axis: ✅ google
⚠️  collect-models: no credential write observed for provider "anthropic" within 180s of clicking Save.
    The panel stays busy while a write is in flight, so the NEXT provider is what pays for this (#1355).
⚠️  collect-models: no credential write observed for provider "google" within 180s of clicking Save.
1 failed  →  Process completed with exit code 1
```

All three keys **validate** (`build axis: ✅` on each) — this is not a drained or revoked credential. The first attempt surfaced the downstream form of the same thing: `anthropic — no models collected from the providers panel`, which `collect-models.spec.ts:166` classifies as a **non-billing** hard failure and fails on, rather than the billing path that only warns (#952/#980).

## Occurrences (2026-08-07, all on `1.12.0.dev19`)

| Run | Branch | Outcome |
|---|---|---|
| [31175663512](https://github.com/oriontech-me/langflow-e2e/actions/runs/31175663512) 11:51 | `chore/quarantine-daily-1361` | `no credential write observed for provider "anthropic" within 180s` |
| [31188034419](https://github.com/oriontech-me/langflow-e2e/actions/runs/31188034419/job/92897516613) 14:31 | `test/issue-1334-oc-credential-axis` | `anthropic — no models collected from the providers panel`, job exit 1 |
| [same run, re-run](https://github.com/oriontech-me/langflow-e2e/actions/runs/31188034419/job/92900604201) 14:44 | same | `anthropic` **and** `google` both stalled, job exit 1 |
| [same run, 3rd attempt](https://github.com/oriontech-me/langflow-e2e/actions/runs/31188034419/job/92903638412) 14:55 | same | **passed** (5.0m), all three providers collected |

Two independent branches, and 2 of 3 attempts on one of them. The provider that stalls **moves between attempts** (anthropic; then anthropic + google) and a plain re-run clears it — the load-dependent shape, not a consequence of any diff.

#1355 was *"clicks a Save button still aria-busy from the previous provider"*, and #1357 replaced a fixed 60 s clock with waiting for the credential write, measured at 103 s. The warning text above is #1357's own — it fires when that wait **expires**, now at 180 s. So the mechanism #1355 described is still reachable; what is unmeasured is whether 180 s is simply short on a loaded runner, or whether the write can fail to be observed at all (in which case a longer clock never fixes it, which is the lesson #1357's own header records).

Not claimed here: that the threshold is the answer. Raising a timeout because a timeout expired is the move #1357 argued against.

## Why this matters beyond one red

`Collect models` is a **hard gate** on this lane. When it exits 1 the PR gets no E2E coverage of its own change, and the only signal is a red check whose cause is three steps away from the diff. #1369 merged on an `ambient-red` justification for exactly this ([comment](https://github.com/oriontech-me/langflow-e2e/pull/1369#issuecomment-5218556907)) — that reasoning was sound for one PR, but it is not a mechanism, and the next PR pays the same cost.

## Investigation directive

**Instrument before tuning.** (a) Establish what the 180 s wait actually observes on a stalling run — is the `POST|PATCH /api/v1/variables/` never issued, issued and never answered, or answered after the deadline? The distinction decides whether any clock helps. (b) Only then decide between a longer deadline, a per-provider retry, and making the sweep resilient to one provider stalling (the other two collected fine in every occurrence above). (c) Consider whether a `Collect models` failure should abort the PR lane at all: `daily-stable.yml` deliberately does **not** let it kill the shard (#980), and the asymmetry is undocumented.

## Deliverables (Done when)

- [ ] Root cause of the missed write confirmed with evidence from a stalling run, not inferred from the deadline expiring.
- [ ] The PR lane no longer loses its entire E2E coverage to one provider's Save stalling — by fix or by an explicit, documented decision that it should.
- [ ] The daily-vs-PR asymmetry on a red `Collect models` is either aligned or written down as intentional.
- [ ] Covered by `npm run test:scripts` / `npm run test:units` where the logic is ours.

rafaelgiln:
Closes #1370

## The problem

`Collect models` is a **hard gate** on `pr-validation.yml`: when it exits non-zero the impacted-specs step never runs and the PR gets **no E2E coverage of its own diff**. It did that twice on 2026-08-07 across two branches, and the warning it printed was not describing the failure it caused.

## What the investigation found

Read off the job logs of **all three attempts** of [run 31188034419](https://github.com/oriontech-me/langflow-e2e/actions/runs/31188034419), not inferred from the deadline expiring. Two premises died on the way — one the issue's, one mine.

**The issue says the stalling provider moves between attempts.** It does not. anthropic stalled **3 of 3**; google never stalled. What looked like a second provider was defect 1 lying.

**The deadline expiring is not the failure.** Attempt 3 printed the *same* two anthropic warnings and **passed**, 90 models, ~242 s after Save. The 180 s warning fires on a green run.

### 1 — A wait that never ran must not report a measurement

Both post-Save waits ended in `.catch(() => null)` / `.catch(() => false)`, so *"my deadline passed"* and *"the page closed under me"* were the same observation. On attempt 2 the spec hit its own 5-minute timeout and both pending waits rejected at once:

```
14:49:30.036  ⚠️ no credential write observed for provider "google" within 180s
14:49:30.048  ⚠️ provider "google" never showed the configured state ("Disconnect") within 60s
14:49:30.088  Error: locator.count: Target page, context or browser has been closed
```

A **180 s** wait and a **60 s** wait completing **12 ms apart**, at most 15.2 s after the click. Neither observed anything — and that is why the issue's own question ("is 180 s simply short?") is unanswerable from those runs.

The discriminator was measured directly against both surfaces, not remembered:

| ended by | `name` | message |
|---|---|---|
| context closed | `Error` | `… Target page, context or browser has been closed` |
| deadline | `TimeoutError` | `… Timeout 800ms exceeded …` |

Now classified instead of discarded, with a third `unknown` state so an unrecognised failure invents neither a measurement nor a cause (#1012).

### 2 — Per-provider ceilings the test's own budget cannot pay

180 s (credential write) + 60 s (configured state) + 15 s (toggles) is **255 s spent on one provider** inside a **300 s** test timeout (`playwright.config.ts:89`). A single stall consumes the run by arithmetic. Attempt 2 died at exactly 5 minutes; attempt 3 survived with **3 s** to spare.

`planSaveWait` gives the sweep the pair of bounds the toggle confirmation already has (#1197 §4.4) — a per-item timeout bounds ONE wait and can never see the sum. It returns a discriminated *"don't wait"* rather than a small number, because Playwright reads `timeout: 0` as **no timeout at all**.

### 3 — `no models collected` is a collector verdict, not a key verdict

A stalled save hands `validateProviderWithFallback` **zero candidates**, so the provider is recorded `inactive` having never been probed — and that landed in `hardFailures`, failing the spec as *"real key/account/config problem"* for a provider whose build axis reported ✅ and whose key was never touched. The #1011 mistake in a new place. It now carries a `collector stall: ` prefix (mirroring `build axis: `) and gets its own step.

## Why this is not a weakened assertion

The new step is scoped by `COLLECT_REQUIRED_PROVIDERS`, and this is the half worth reading carefully.

`pr-validation.yml` pins its own run to one provider (`select-pr-model-target.mjs --provider openai`, #1169, gated on the same `needs_models` as the sweep), so a stalled anthropic there changes **no spec that lane will execute** — and yet it cost the lane everything.

- **Unset** (daily, `manual.yml`, every local run) still requires **every env-keyed provider** — #570's guarantee unchanged where it bites.
- **Nothing else narrows**: a dead key, an image that cannot build a component, a total wipeout and a broken Model Providers UI all still fail for every provider on every lane.
- A name the run has no key for is **reported**, never dropped — silently requiring nothing is how a typo disables a gate for good.

`PR_LANE_PROVIDER` is declared once at job level: the gate and the pin must not drift, or the lane would gate on a provider it does not run and run one it does not gate. The daily's opposite trade (`continue-on-error`, #980) is now stated in that file rather than implied by the other one — deliverable 3.

## What it costs, stated rather than buried

With the budget in place, attempt 3's anthropic would have collected **zero** models instead of 13. It passed by luck, 3 s from the timeout. The trade is one stalling provider's collection against the whole run.

The underlying slowness (anthropic's write >240 s in CI, 103 s locally per #1357) is a backend cost this PR does **not** explain. It makes the lane survive it.

## Validation

**Langflow `1.12.0.dev19`** (nightly, `langflowai/langflow-nightly:latest`)

```
run 1/3 : clean expected=1  160s
run 2/3 : clean expected=1  123s
run 3/3 : clean expected=1   17s
typecheck=0  lint=0   units 539/539   scripts 797/797
FF coverage: complete · no FF-MUTATION markers in diff ✓ · final green run after reverts ✓
```

Run 1 started from a fresh container and exercised the full save path for all three providers; run 3 took the `alreadyConfigured` short-circuit.

**The first burst was rejected, and why matters.** It read `clean expected=1 / clean expected=0 / clean expected=0`. Runs 2 and 3 lasted `128292 ms` and `128257 ms` — **35 ms apart**, a fixed clock: `HEALTH_TIMEOUT_MS = 120_000` in `tests/globalSetup.ts:77`. Run 1's sweep wedged the local backend (#922/#927) and the next two aborted in `globalSetup` without getting a test to the browser. They were nonetheless recorded **clean**, because that classification checks `unexpected === 0` and never `expected > 0` — a zero-test run counting as proof of stability, the `runguard`/#1012 lesson missing one layer up. Container restarted, burst re-run on a clean target; the three `expected=1` above are that burst. Filed separately — it is out of this branch's scope.

**Force-fail:** every provider record rewritten to carry a collector-stall reason ⇒ the new step's `fatal` list non-empty → `unexpected=1`; reverted; final green confirmed.

**Unit mutation** (a test that cannot fail proves nothing): deleting the `aborted` branch kills 2 tests; disabling the budget's skip branch kills 3.

**Orphan flows:** 26, all starter templates seeded by the fresh container. Zero leaked.

## Coverage

18 unit tests in `collect-models.test.ts` plus a structural guard in `select-pr-model-target.test.mjs` that pins an **absence** — no provider name spelled twice outside comments — the same shape as the `--grep` composition guard (#1275). The stall itself needs a loaded CI runner and is not reproducible on demand, so what is tested is the decision: when it gives up, what it reports, and what it fails on.

### DIFF
--- a/.github/workflows/daily-stable.yml
+++ b/.github/workflows/daily-stable.yml
@@ -396,6 +396,18 @@ jobs:
       - name: Collect models
         id: collect_models
         run: npx playwright test tests/collect-models.spec.ts --reporter=line
+        # NOT a hard gate here, deliberately, and the asymmetry with
+        # pr-validation.yml is intentional rather than an oversight (#1370 asked
+        # for it in writing). There this step gates the job: a red sweep means
+        # the impacted specs never run, which is right, because a spec that needs
+        # a model must not run without one (#1152). Here a red sweep costs the
+        # provider's parametrized specs and nothing else, and losing a scheduled
+        # day of @stable coverage to a drained key is the worse trade (#980).
+        #
+        # COLLECT_REQUIRED_PROVIDERS is therefore deliberately UNSET below: this
+        # lane rotates providers by weekday (#1185) and owes multi-provider
+        # coverage, so every env-keyed provider stays required. Narrowing it is
+        # the PR lane's move, and only because that lane pins itself to one.
         continue-on-error: true
         env:
           CI: "true"
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -616,6 +616,14 @@ jobs:
       # Read by the sidecar only; the poller has its own TOKENS_MAX_SECONDS.
       TOKENS_BUDGET_MS: "15000"
 
+      # The single provider this lane runs its LLM specs against (#1169), named
+      # ONCE because two steps depend on it and they must not drift: "Collect
+      # models" below fails only on this provider, and "Pin the lane to a single
+      # provider's settled model" resolves this provider's settled model. Two
+      # literals would let a future change to one of them silently gate on a
+      # provider the lane no longer runs (#1370).
+      PR_LANE_PROVIDER: openai
+
     steps:
       - uses: actions/checkout@v7
 
@@ -700,6 +708,24 @@ jobs:
           # backend; the health gate below is what covers that case.
           PLAYWRIGHT_RETRIES: "0"
           PLAYWRIGHT_BASE_URL: "http://localhost:7860/"
+          # Which provider a COLLECTOR stall may fail this step on (#1370).
+          #
+          # A provider the collector never managed to configure is recorded
+          # `inactive` having never been probed at all — a collector verdict, not
+          # a key one — and every such provider used to fail this step. On a lane
+          # that pins itself to ONE provider (below), a stalled anthropic changes
+          # no spec this job will execute, and yet it exits here non-zero, which
+          # on THIS lane is a hard gate: the impacted-specs step never runs and
+          # the PR gets no E2E coverage of its own diff. Measured 2026-08-07 on
+          # run 31188034419, where anthropic stalled on all three attempts and
+          # two of them cost the whole lane.
+          #
+          # This narrows the COLLECTOR-stall verdict only. Every other failure —
+          # a dead key, a provider the image cannot build, a total wipeout, a
+          # broken Model Providers UI — still fails this step for every provider,
+          # unchanged. Unset elsewhere (daily-stable.yml, manual.yml, local), so
+          # every env-keyed provider stays required there.
+          COLLECT_REQUIRED_PROVIDERS: ${{ env.PR_LANE_PROVIDER }}
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
@@ -784,7 +810,9 @@ jobs:
       # with no sweep there is no settled model to read.
       - name: Pin the lane to a single provider's settled model
         if: needs.detect-specs.outputs.needs_models == 'true'
-        run: node scripts/select-pr-model-target.mjs --provider openai
+        # Same provider the Collect models step gates on, read from one place so
+        # the two cannot drift apart (#1370).
+        run: node scripts/select-pr-model-target.mjs --provider "$PR_LANE_PROVIDER"
 
       # In-run token consumption recorder, extended to this lane (#1183 — measure
       # the suite's real LLM spend per lane/provider). Same mechanism as
--- a/docs/collect-models.md
+++ b/docs/collect-models.md
@@ -275,7 +275,7 @@ turned one slow write into a **lost PR lane**, each with its own fix and its own
 unit coverage. Read them in this order; the third is the one that costs coverage.
 
 **1 — A wait that never ran must not report a measurement.** The two waits after
-Save (the credential response, and the panel reaching `Disconnect`) ended in
+the Save click (the credential response, and the panel reaching `Disconnect`) ended in
 `.catch(() => null)` / `.catch(() => false)`, which makes *"the deadline expired"*
 and *"the page was closed under me"* the same observation. On run
 [31188034419](https://github.com/oriontech-me/langflow-e2e/actions/runs/31188034419)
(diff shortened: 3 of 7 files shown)


rafaelgiln:
## The first CI run of this PR found the third instance of the same defect

Worth recording rather than amending away: the two waits this PR bounds behaved **exactly** as designed on [run 31197255019](https://github.com/oriontech-me/langflow-e2e/actions/runs/31197255019/job/92928489961), and the run still died.

```
⚠️  no credential write observed for provider "anthropic" within 175s of clicking Save.
⚠️  skipped waiting for provider "anthropic" to reach the configured state ("Disconnect")
    — the sweep's shared budget is down to 30s with 1 provider(s) still to configure,
      so waiting here would leave them nothing.
Error: collect-models: the "Save" button for provider "google" never became actionable
       after 60.1s over 231 poll(s).
```

anthropic's write was still in flight — the first warning says so — so google's `Save` stayed `aria-busy`. And `waitForButtonIdle` sits **before** both bounded waits, on a fixed 60 s clock **outside** the budget, with a caller that **throws**. One provider's backend cost ended the whole pre-flight: the exact thing this issue is about, one step further upstream than I had looked.

It also corrects an arithmetic claim in the first commit and in the spec doc. The real per-provider spend is **60 + 180 + 60 + 15 = 315 s** inside a **300 s** test, not the 255 s I wrote — the ceiling I had not counted is the one that threw.

`39924f0` fixes it the same way as the other two: the idle wait draws on the shared budget, and a `Save` that never becomes actionable is recorded as that provider's **stall** instead of thrown. That loses no signal — a stall still fails the gate on every lane that requires the provider, which on the daily is all of them. What it stops doing is turning one provider into zero coverage for the run. A provider whose Save was never clicked also stops waiting for `Disconnect`, which cannot appear for a save that was not issued.

One unit test pins both halves of the new verdict: recognisable as a collector stall, and **not** matching the billing/quota regex — without that it would downgrade to a warning and vanish on every lane.

Verified on a fresh `1.12.0.dev19` container: `1 passed (2.8m)`, all three providers active, 90 models. Unit lanes 540/540 and 797/797.

**Not treated as ambient.** The red was caused by this branch's own incompleteness, so it is fixed here rather than justified.

rafaelgiln:
## The green run reproduced the failure and survived it

[Run 31198738701](https://github.com/oriontech-me/langflow-e2e/actions/runs/31198738701/job/92933338025) hit the same stall that killed the two previous attempts, and finished:

```
⚠️  no credential write observed for provider "anthropic" within 176s of clicking Save.
⚠️  skipped waiting for provider "anthropic" to reach the configured state — budget down to 30s
⚠️  collect-models: the "Save" button for provider "google" never became actionable after 30.1s
⚠️  skipped waiting for provider "google" to reach the configured state — no Save was issued
❌ anthropic (no model) — collector stall: no credential write answered within 176s of clicking Save
❌ google  (no model) — collector stall: the "Save" button ... never became actionable after 30.1s
  1 passed (4.4m)
PR lane pinned to openai / gpt-4o-mini
```

Every new path fired: the budget clamped google's idle wait from 60 s to 30.1 s instead of letting it throw; `no Save was issued for this provider` is the new guard refusing to wait for a `Disconnect` that cannot appear; both providers carry a `collector stall:` reason naming the collector rather than the key; and the gate passed because the lane requires only the provider it pins.

## What it cost, so nobody reads it as a regression later

That run collected **41 models (openai only)** where a healthy run collects 90. anthropic's write held the panel for >176 s, so google's `Save` could not go idle before the budget ran out.

**No budget setting recovers google here.** Its Save stays `aria-busy` until anthropic's write returns, and that measured past 240 s — more than the spec's entire 300 s timeout. Within one run, when anthropic costs that much, google cannot also be configured. That is the backend cost, unchanged by this PR.

What did change is the floor. Before, the same conditions threw at google's `Save` **before any file was written**, so the run produced no `models.json` at all and the PR lane got nothing. Now openai is collected, written, pinned, and its specs run. Strictly more coverage in the same conditions, on both lanes — on the daily it is the difference between a warning with one provider and a warning with none.

Issue #1024, opened 2026-07-28
Daily dies at `prep`: playwright.config.ts writes to stdout, so `--list --reporter=json` is unparseable


Victor-w-Madeira:
`playwright.config.ts` prints a line to **stdout** at import time, and `daily-stable.yml`'s `prep` job redirects stdout into a JSON file. The file is therefore not JSON, `partition-shards.mjs matrix` throws, and **no shard matrix is produced — the whole daily run dies before a single test executes.**

Introduced by #1010 / PR #1021 (`ffac90b`, merged 2026-07-28), which announces the `@destructive` lane exclusion on every run:

```ts
// playwright.config.ts
if (!DESTRUCTIVE_LANE) {
  console.log(
    "[lane] @destructive tests are excluded from this run — run them with: PW_DESTRUCTIVE=1 npx playwright test --grep @destructive",
  );
}
```

The intent is right — an exclusion must never become a silent cap (#570's class) — but the channel is wrong.

## Reproduction

The `prep` step, verbatim from `daily-stable.yml:81`, run against `main` at `ffac90b`:

```console
$ npx playwright test --grep "@stable" --list --reporter=json > /tmp/stable-list.json
$ head -c 60 /tmp/stable-list.json
[lane] @destructive tests are excluded from this run — run

$ node scripts/partition-shards.mjs matrix /tmp/stable-list.json reports/spec-durations.json 4
Unexpected token 'l', "[lane] @des"... is not valid JSON
$ echo $?
1
```

## Blast radius

**Exactly one consumer**, and it is the gate for the whole daily: `daily-stable.yml:81` is the only place that pipes Playwright's stdout into a JSON file. Every other JSON consumer sets `PLAYWRIGHT_JSON_OUTPUT_NAME` / `PLAYWRIGHT_JSON`, where the reporter writes the file itself and a `console.log` cannot reach it.

**No PR check goes red**, which is why this merged clean: `pr-validation.yml` uses `--reporter=github` and `line`. The first thing to fail is the 05:00 BRT daily.

## What the existing guards do and do not do

The #1011 / #1012 machinery **holds** on this path — verified by tracing it, not assumed:

| Step | Outcome when `prep` fails |
|---|---|
| `test` | **skipped** (`needs: prep`) — note: `skipped`, not `failure` |
| `merge` | **runs** — `if: always()` (line 474) |
| Download `blob-*` | fails, nothing matches the pattern |
[...]

Issue #225, opened 2026-05-12
test-infra: pin browser locale to en-US in playwright.config.ts


Victor-w-Madeira:
## Context

The regression suite asserts on English strings throughout — error toast titles, button labels, status messages, etc. Examples in the prompt-template area alone:

- `"There is something wrong with this prompt"` (toast title from `errors.prompt` in `en.json`)
- `"Use Double Brackets"` (display name from upstream `BoolInput`)
- `"Input variables contain invalid characters or formats"` (upstream `ValueError` message)

`playwright.config.ts` does not pin the browser locale. The tests pass today because the Langflow frontend's default i18n bundle is English and the test runner's default locale matches whatever the host machine reports. Either of those defaults could change:

1. Langflow adds locale detection from `Accept-Language` and CI runners pick up a non-English locale
2. The project starts supporting a non-English default
3. A future Playwright update changes the default `locale` value

In any of those cases, the affected specs would fail with hard-to-diagnose i18n-mismatch errors.

## Scope

Add `locale: "en-US"` to the `use` block in `playwright.config.ts`:

```ts
use: {
  baseURL: BASE_URL,
  locale: "en-US",
  actionTimeout: 20000,
  // ...
},
```

This pins the browser context locale and the `Accept-Language` header to `en-US` for every test in every project, regardless of host machine settings.

## Acceptance

- [ ] `locale: "en-US"` added to the `use` block in `playwright.config.ts`
- [ ] Full regression suite still green
- [ ] CONTRIBUTING.md notes the convention so future tests don't try to override it for unrelated reasons

## Out of scope

- Adding multi-locale coverage (different concern; would require parameterised tests)
- Pinning timezone (separate question; `TZ` env var + `timezoneId` in Playwright)
- Refactoring existing assertions to use `data-i18n` keys instead of English strings — defensive but invasive

## Notes

Surfaced in the second-pass review of PR #223 (issue #221). One-line config change with broad protective effect; safe to apply as a small dedicated PR.

PR #1403, opened 2026-08-10
Parameterise the browser locale — prerequisite for non-English coverage

### LINKED ISSUE #1400: Parameterise browser locale — prerequisite for non-English coverage

**Type:** prerequisite (infra) — no coverage bullet of its own.

**Problem**
`CONTRIBUTING.md` pins the browser locale to `en-US` for every test in every project, because the suite asserts English strings throughout, and states that non-English coverage *"is a separate, parameterised concern — raise it as its own issue rather than editing the shared default."*

This is that issue. Until it lands, QA-CHECKLIST.md §18.1–18.2 (i18n, 5 bullets) cannot be written: §18.2 in particular cannot run under the pin, and a per-test `locale` override would violate the rule those bullets depend on.

**Scope**
- Parameterise the locale so a spec can opt into a non-English one without changing the shared default.
- Keep `en-US` as the default for every existing spec.

**Note:** touches `playwright.config.ts` / shared fixtures, so `pr-validation.yml` treats it as suite-wide and raises the full-run warning — dispatch `manual.yml` for the PR.

**Unblocks:** i18n §18.1–18.2, scheduled after this wave.

Victor-w-Madeira:
Closes #1400

## Why

`playwright.config.ts` hardcoded `locale: "en-US"` (#225) and `CONTRIBUTING.md` banned overriding it anywhere. The default is right — the suite asserts English strings throughout — but it left non-English coverage with no route at all: the only mechanism available (`test.use({ locale })`) was exactly what the rule banned, which is why the i18n checklist bullets could not be written.

## What lands

- **`tests/fixtures/locale.ts`** — the default, the validated per-spec opt-in (`withLocale`) and the run-wide `PW_LOCALE` resolution, as pure functions. It lives under `tests/fixtures/` deliberately: that prefix is the only one `impacted-specs-by-import.mjs` already treats as suite-wide, which is what a file governing `use.locale` for every spec must be.
- **`playwright.config.ts`** — `locale` is a resolved value instead of a literal. The default is unchanged.
- **Fail-closed both ways.** An unusable tag aborts the config naming the value and the POSIX slip (`pt_BR`) rather than falling back to en-US under a command line that asked for something else; a real override announces itself on stderr, never stdout — that stream is the daily's shard-matrix contract (#1024).
- **`tests/fixtures/locale-gate.spec.ts`** — the behavioural gate, sibling to `http-error-gate.spec.ts` / `flow-error-gate.spec.ts`: it pins that the resolved locale reaches a real browser context and that `withLocale()` overrides it there.
- **Docs** — `CONTRIBUTING.md` and `CLAUDE.md` record the default, the sanctioned opt-in, and the three-axis measurement below.

Usage:

```ts
test.describe("date rendering under a non-English locale", () => {
  test.use(withLocale("pt-BR"));
});
```

```bash
PW_LOCALE=pt-BR npx playwright test …   # whole run, announced on stderr
```

## The measurement, because the obvious reading is wrong

Measured on Langflow Nightly `1.12.0.dev20`, two browser contexts identical except `locale`. **The browser locale does not decide the language Langflow renders.** `src/frontend/src/i18n.ts` reads `localStorage.languagePreference` (default `"en"`) and never consults `navigator.language` — under pt-BR the UI stays English and `<html lang>` stays `en`.

| Axis | Changes with `locale`? |
|---|---|
| The language the UI renders (`localStorage.languagePreference`) | **No** |
| `Intl` formatting + the document/asset `Accept-Language` | **Yes** — `12/31/1969, 9:00:00 PM` → `31/12/1969, 21:00:00` |
| The locale the backend's `set_locale` middleware sees | **Partly, by leakage** — 17 of the 20 `/api/` requests the home screen makes send `en` (the frontend pins `Accept-Language: i18n.language` on the calls that pass through its axios interceptors); 3 carry the context locale, and the backend really does localise (`/api/v1/flows/basic_examples/` → `Sugestões básicas` under `pt`) |

That is why `withLocale()` returns **only** `locale`, and why `CONTRIBUTING.md` tells a spec whose subject is the backend's locale to set `Accept-Language` explicitly. Whoever writes the §18.x i18n bullets needs this: a spec asserting "the UI follows the browser locale" would be built on a false premise.

## Validation

- `npm run test:units` 571 ✓ · `npm run test:scripts` 797 ✓ · `typecheck` ✓ · `lint` 0 errors ✓ · `check:checklist-coverage` ✓
- `npx playwright test --grep @stable --list --reporter=json` still parses; `partition-shards.mjs matrix` → 4 shards (185 files). Same with `PW_LOCALE=pt-BR`; with `PW_LOCALE=pt_BR` it exits 1 with 0 bytes on stdout and the cause named.
- The 3 CI canary specs 10/10 green against 1.12.0.dev20; the new gate green in both modes (3 passed by default; 2 passed + 1 skipped under `PW_LOCALE=pt-BR`).
- **Mutation-checked**, since a guard that survives its own mutation guards nothing: reverting the config to a literal, dropping the validation throw, suppressing the notice, moving it to stdout, adding a project-level `locale`, and making `withLocale()` ignore its argument each fail a named test.
- An independent review of the first commit refuted one factual claim in the docs and found the wiring guard asserted the wrong level; `98bca32` fixes both, with the measurement in the commit body.

## Note for the reviewer

Touches `playwright.config.ts` + `tests/fixtures/**`, so `pr-validation.yml` classifies it suite-wide and raises the full-run warning — the issue predicted this. Final proof is a `manual.yml` dispatch.

`ROADMAP.md` still marks Wave 5 `◀ CURRENT` while Wave 5 closed with 0 open issues and Wave 6 is the only milestone with open `roadmap` items — pre-existing staleness, not this PR's, but worth a separate fix.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -240,3 +240,12 @@ Key settings in `playwright.config.ts`:
 - Fully parallel, 5-minute timeout per test
 - 3 retries locally, 2 retries in CI; trace captured on first retry
 - HTML reporter locally, blob reporter in CI
+- Browser locale `en-US` by default, resolved by `tests/fixtures/locale.ts` rather than
+  hardcoded (#1400): a spec opts out with `test.use(withLocale("pt-BR"))`, a whole run
+  with `PW_LOCALE`, and an invalid tag aborts the config instead of falling back to
+  English. **It does not decide the language Langflow renders** — measured on
+  `1.12.0.dev20`, the UI reads `localStorage.languagePreference` and never
+  `navigator.language`, and the frontend pins `Accept-Language: i18n.language` on every
+  `/api/**` call, so the context locale reaches neither the UI strings nor the backend's
+  `set_locale` middleware. It governs `Intl` formatting and the document `Accept-Language`
+  only; `CONTRIBUTING.md` → *Browser locale* has the three-axis table
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -245,7 +245,13 @@ Key settings in `playwright.config.ts`:
   with `PW_LOCALE`, and an invalid tag aborts the config instead of falling back to
   English. **It does not decide the language Langflow renders** — measured on
   `1.12.0.dev20`, the UI reads `localStorage.languagePreference` and never
-  `navigator.language`, and the frontend pins `Accept-Language: i18n.language` on every
-  `/api/**` call, so the context locale reaches neither the UI strings nor the backend's
-  `set_locale` middleware. It governs `Intl` formatting and the document `Accept-Language`
-  only; `CONTRIBUTING.md` → *Browser locale* has the three-axis table
+  `navigator.language`, so under `pt-BR` every string stays English and `<html lang>`
+  stays `en`. What it does govern is `Intl` formatting and the document
+  `Accept-Language`. The backend's `set_locale` middleware is a **partial** third axis,
+  not a "no": Langflow localises by `Accept-Language` (`/api/v1/flows/basic_examples/` →
+  `Sugestões básicas` under `pt`), and while the frontend pins the header on the calls
+  that pass through its axios interceptors — 17 of 20 on the home screen — the other 3
+  (a `/api/` subresource, two `/api/v9/invites/` XHRs) carry the context locale, so a
+  spec asserting on the backend's locale must set the header itself. `CONTRIBUTING.md` →
+  *Browser locale* has the three-axis table and the measurement. Pinned behaviourally by
+  `tests/fixtures/locale-gate.spec.ts`, the sibling of the two error-policy gates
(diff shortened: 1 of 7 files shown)


Victor-w-Madeira:
Copilot's finding is addressed in 60629d2.

It was right, and the same exposure hit a second test it did not name: `the daily's prep command produces parseable JSON` also inherits the environment and dies on `PW_LOCALE=pt_BR` — measured, 1 of 10 failing before the fix. Both are new exposure from this PR, since nothing in the config read `PW_LOCALE` before it.

The two hand-rolled spawns are now one hermetic reader (`readFromConfig`) that strips `PW_LOCALE`/`PW_DESTRUCTIVE` and runs from an empty cwd, so `dotenv` cannot pick them out of the git-ignored `.env` either. The prep test keeps `cwd: REPO_ROOT` — running as the daily runs is its whole point — and only sheds the two variables the daily never sets.

Verified 10/10 with `PW_LOCALE=pt_BR PW_DESTRUCTIVE=1` exported and 10/10 clean; `npm run test:units` 571/571.

PR #1404, opened 2026-08-10
fix(provider-setup): freeze the model catalog per run (#1386)

### LINKED ISSUE #1386: [Daily #1383] mcp-client-agent-gemini — the describe title is resolved at module load from a models.json the same shard rewrites mid-run
Spun out of daily-failure triage #1383 (run [31373880200](https://github.com/oriontech-me/langflow-e2e/actions/runs/31373880200), 2026-08-10). Same sweep as the collect-models skip cluster filed from this run, but a distinct defect: this one is latent and would fire on any run where the pre-flight leaves a provider empty and the in-shard `collect-models.spec.ts` then fills it.

**Upstream:** _not filed_

## Symptom

The test produced no run at all. Three attempts, `duration 0`, `workerIndex -1`, each ending in Playwright's `Test not found in the worker process. Make sure test title does not change.` — no browser was opened and no assertion in the spec was reached, so nothing here is evidence about Gemini or MCP.

`@stable` was auto-removed by the workflow (commit `c954cd9`) on the strength of a failure that never exercised the test.

|---|---|---|
| `tests/tests-automations/regression/mcp/client/mcp-client-agent-gemini-tool-regression.spec.ts:139` ("Gemini invokes the echo MCP tool (regression for fixed upstream #440)") | `the worker to find the test by its title` | `Test not found in the worker process. Make sure test title does not change.` |

Single spec, single signature. It is filed apart from the collect-models cluster because the fix is not the same one: even with a healthy sweep, any run whose shard list contains `collect-models.spec.ts` can rewrite `models.json` after the runner has already computed this file's titles.

Descriptive — the mechanism is visible in the shard-3 log and the spec source; whether the fix belongs to the spec, the resolver or the shard composition is open.

The suite title is computed **at module load**:

```ts
const geminiModel = resolveGeminiModel();
test.describe(`MCP Client – Gemini tool regression (#440) [${PROVIDER} / ${geminiModel ?? "default"}]`, ...)
```

`resolveGeminiModel()` reads `tests/helpers/provider-setup/data/models.json` off disk on every call. On shard 3 that file changed **during the run**:

- `09:25:47` — the pre-flight `Collect models` step finishes with `Models found (google): []` and writes `models.json saved with 41 models` (openai only).
- `09:26:19` — `collect-models.spec.ts`, which is **in shard 3's own spec list**, runs as a test and reports `provider "google" credential write took 11.3s (HTTP 201)`, followed by a populated `Models found (google): [`.

So the runner listed the test as `… [google / default]` (the reporter log confirms that exact title), and by the time a worker loaded the file the resolver had a google model to return. The titles no longer matched, and Playwright reported precisely that. All three retries reproduced it because the file stayed rewritten.

Two consequences worth stating. The failure is **not attributable to the spec's subject**, so today's auto-removal of `@stable` is collateral. And it is **order-dependent, not deterministic** — it needs a degraded pre-flight followed by a successful in-shard collect, which is why it appears on some dailies and not others; the same test's other hits in the 30-day window are recorded under different signatures.

## Investigation directive

The question is not why google was empty (that is the sibling collect-models issue) — it is why a suite title depends on mutable state read at collection time.

1. Find every spec whose `test`/`test.describe` title interpolates a value derived from `models.json` (`resolveGeminiModel` and its siblings under `tests/helpers/provider-setup/`). This spec is the one that fired; it is unlikely to be the only one exposed.
2. Decide where the invariant belongs. Options that differ in cost, not just style: keep the model out of the title and report it from inside the test; resolve the model once and freeze it for the process; or stop `collect-models.spec.ts` from sharing a shard with specs that read its output.
3. Check whether `collect-models.spec.ts` being partitioned into a shard's normal spec list is intended at all — it is a pre-flight that also runs as a test, and its side effect is a file other specs read at import time.
4. Confirm the spec's `@stable` removal was collateral (it never ran) before restoring the tag, and re-validate the Gemini/MCP assertion itself on a run where google is configured — the #440 regression guard has not been exercised on this run either way.

## Deliverables (Done when)

- [ ] The set of specs whose titles depend on `models.json` is enumerated, not just this one.
- [ ] A title computed at collection time can no longer be invalidated by a later write to `models.json` — with the chosen mechanism stated (frozen resolution, title without the model, or shard composition).
- [ ] `@stable` restored (auto-removed in `c954cd9` on a failure that never ran the test), after the Gemini/MCP assertion is seen passing on a run where google is configured.

rafaelgiln:
**Fixes #1386.**

## Problem

A test's identity in Playwright is its **title**, and the title is computed **twice, in two processes, at two moments**: the runner when it builds the test tree (the `load` task), each worker when it reloads the file it was handed. Nineteen specs derive that title from `models.json` **at module load** — 18 through `resolveTestTargets()` (`` `[${label}]` `` = `provider / model`) and `mcp-client-agent-gemini-tool-regression.spec.ts` through `resolveGeminiModel()`. So for those files the title is a pure function of the file content at read time, and anything writing that file between the two reads changes the test's identity underneath it.

There is such a writer, and it is **inside the run**: `tests/collect-models.spec.ts` is `@stable`, so it is collected by the daily's `--grep @stable --list` and partitioned into a shard's spec list like any other file (verified: 187 `@stable` files, `collect-models.spec.ts` among them).

On the daily of 2026-08-10 (run [31373880200](https://github.com/oriontech-me/langflow-e2e/actions/runs/31373880200)), shard 3:

- `09:25:47` — pre-flight `Collect models` finishes with `Models found (google): []` and writes `models.json` with openai only.
- `09:26:19` — the in-shard `collect-models.spec.ts` runs as a test, succeeds, and rewrites `models.json` **with** google.

The runner had listed `… [google / default]`; the worker resolved `… [google / gemini-flash-latest]`. Result, three attempts: `duration 0`, `workerIndex -1`, `Test not found in the worker process. Make sure test title does not change.` No browser opened, no assertion ran — and the auto-removal path took `@stable` off the spec (commit `c954cd9`) on the strength of that.

Reproduced in an isolated two-spec harness (no Langflow): a writer rewriting a JSON that a sibling interpolates into its `describe` title yields the byte-identical signature.

## Fix

`globalSetup` freezes the catalog's **raw text** into `PW_MODELS_SNAPSHOT` before the load task — `createGlobalSetupTasks` precedes `createLoadTask` in Playwright's CLI task list — workers inherit `process.env`, and every title-feeding resolver reads the snapshot instead of the file. The mid-run write still happens and is picked up by the next run; it can no longer change a title, whatever writes it and whichever shard it lands in.

New: `tests/helpers/provider-setup/catalog-snapshot.ts` (+ 12 unit tests). Wired into `test-targets.ts`, `resolve-gemini-model.ts`, `resolve-gpt-model.ts`.

**Rejected alternatives** (argued in the new file's header, not left to look like oversights):

- *Keep the model out of the title.* It cannot go: under `ALL_MODELS` / `MODEL_TEST_PROVIDER` a spec parametrizes one describe **per model**, so the model is what makes those titles distinct — dropping it collapses them into duplicates.
- *Keep `collect-models.spec.ts` out of the shard that reads its output.* A scheduling constraint the partitioner would have to keep true forever on every lane, useless for a local `npx playwright test tests/`, and not the invariant: the defect is that identity depends on mutable state, not that one particular file mutates it.

Four decisions worth reading before changing this:

- An **absent** `models.json` freezes `[]`. Without a sentinel, "no file" would fall through to a live read in the workers — the same mismatch from the other direction, on a run whose pre-flight failed entirely but whose in-shard collect succeeded.
- A **malformed** catalog is frozen **verbatim**, so #1035's deliberate throw stays in the reader instead of aborting the whole run inside `globalSetup`.
- `providers.json` is deliberately **not** frozen: it only ever feeds a `skipReason`, never a title or the number of targets, and `globalSetup` itself **writes** it on the credential-degradation path (#1058).
- The snapshot stands in for **one** path, so every `catalogPath:` unit-test seam keeps injecting its own fixture.

## Scope note

No spec logic changed. Label parity was confirmed on a live run: `Agent Tool Name Validation [openai / gpt-4o-mini]` — identical to the pre-change title. The only spec edit is restoring `@stable` on the Gemini/MCP test (auto-removed in `c954cd9` on a failure that never ran it), plus the record in its spec doc. The generated `QA-CHECKLIST.md` blocks are untouched — they regenerate on merge; the manual Part II bullet already existed (line 728).

**Also found, filed nowhere yet** (issue directive item 3, descriptive): `collect-models.spec.ts` runs **twice** on the shard that owns it — once as that job's pre-flight, once as a test. With this fix that is no longer a correctness problem, but the second execution is a duplicated sweep that can wedge the backend (#922/#927) with **no health gate after it**. Worth a follow-up issue.

## Validation (nightly 1.12.0.dev22, `--retries=0`)

- `npm run typecheck` ✅ · `npm run lint` ✅ (0 errors) · `npm run test:scripts` ✅ 797/797 · `npm run test:units` ✅ 564/564
- `npm run check:checklist-coverage` ✅ · QA-CHECKLIST guard ✅ (no generated block edited)
- `mcp-client-agent-gemini-tool-regression.spec.ts` — **3/3 clean** with `--workers=1 --retries=0` and google configured (30.5s / 27.7s / 34.5s)
- `agent-tool-name-validation.spec.ts` (`--grep "an invalid tool name blocks execution"`) — 1 passed, label unchanged
- **Force-fail, executed:**
  - `readCatalogText` mutated to ignore the snapshot → real run (writer + Gemini spec) **failed** with `Test not found in the worker process. Make sure test title does not change.`; unmutated, **2 passed**.
  - Same mutation against the unit lane → **4 of 12 failed** (exactly the frozen-read assertions). Reverted: 12/12, `grep -c FF-1386` = 0.
  - End-to-end scenario with `models.json` seeded openai-only and a writer running first: title stayed `[google / default]` on both sides and the test **executed** — the daily's exact scenario.
- Zero `🚨 Backend Error` on the three clean Gemini runs except one known ambient bulk `DELETE /api/v1/flows/` 500 (advisory, unrelated to this change).
- `--trace=on` skipped for the Gemini spec: it loads the Simple Agent template, the family where tracing hangs locally (documented limitation).
- Flow cleanup audited via `GET /api/v1/flows/`: one leaked `Simple Agent` from a run that died inside `load()` before the id was captured, purged by id (`DELETE → 200`). No orphan MCP servers.

---

## Follow-up commit — `fix(mcp): make the tool-count poll survive a slow action_count probe`

Surfaced while reproducing this PR on a second local instance, and it is a
**different** defect from the one above: the test ran (browser open, title
resolved `[google / gemini-flash-latest]`, 38.4 s) and died on
`apiRequestContext.get: Timeout 20000ms exceeded` for
`GET /api/v2/mcp/servers?action_count=true` — where #1386's signature is
`Test not found in the worker process`, `duration 0`, no browser at all.

That endpoint makes the backend **start every registered stdio MCP server** to
count its actions, so its cost scales with how many are registered and with how
cold the `npx` cache is: measured 9-15 s against an instance carrying only the
two servers Langflow registers for its own projects, and past the suite's 20 s
`actionTimeout` while this spec's own server is still spinning up. A probe that
times out **throws**, and `expect.poll` propagates the throw instead of polling
again — so one slow probe aborted the whole 90 s window on its first attempt.
The poll could not tolerate the very slowness it exists to wait out. CI does not
see it because its instance is fresh.

A failed probe is now "not ready yet": explicit 45 s per-probe timeout, 150 s
window, and the last probe failure printed alongside the assertion message
(#1012 — "the server never started" and "every probe timed out" are different
diagnoses). The assertion itself is unchanged.

Validation: 3 clean runs `--workers=1 --retries=0` (32.6 s / 31.3 s / 28.8 s) ·
typecheck ✅ · eslint ✅ (0 errors) · force-failed both ways — polling for a
server name that never appears fails with the new message, and pointing the
probe at a 404 path fails **and** prints `last probe … failed: HTTP 404`.

### DIFF
--- a/docs/mcp/client/mcp-client-agent-gemini-tool-regression.md
+++ b/docs/mcp/client/mcp-client-agent-gemini-tool-regression.md
@@ -60,6 +60,16 @@ check encodes the expected #440 state.
 > **`@stable` — promoted under #947** after #440 was confirmed fixed on
 > 1.12.0.dev5 and the positive assertion (Gemini invokes `echo`) ran clean
 > `--workers=1 --retries=0` with a per-test force-failure check.
+>
+> **Removed and restored under #1386.** The daily of 2026-08-10 (run 31373880200)
+> auto-removed the tag on a failure that never executed the test: this file's
+> `test.describe` title interpolates `resolveGeminiModel()`, resolved at module
+> load, and the shard's own `collect-models.spec.ts` rewrote `models.json` after
+> the runner had computed the title — so the worker reported `Test not found in
+> the worker process` with `duration 0` and `workerIndex -1`. The cause is fixed
+> at the source (the catalog is frozen for the whole run —
+> `tests/helpers/provider-setup/catalog-snapshot.ts`), and the assertion itself
+> was re-validated 3/3 with `--retries=0` on 1.12.0.dev22 with Google configured.
 
 ---
 
--- a/tests/globalSetup.ts
+++ b/tests/globalSetup.ts
@@ -13,6 +13,7 @@ import {
   writeProviderHealth,
 } from "./helpers/provider-setup/provider-health";
 import { preconfigureRoutedProvider } from "./helpers/provider-setup/preconfigure-routed-provider";
+import { freezeModelCatalog } from "./helpers/provider-setup/catalog-snapshot";
 import {
   catalogVerdict,
   type CatalogSnapshot,
@@ -336,7 +337,42 @@ async function reportCatalogDrift(ctx: APIRequestContext): Promise<void> {
   );
 }
 
+/**
+ * Freeze the model catalog for this run (#1386).
+ *
+ * Runs FIRST and touches no network, because it must hold even on the paths where the
+ * rest of this file gives up: it is what keeps the runner's view of `models.json` and
+ * every worker's view identical, and 18 specs derive their `test.describe` title from
+ * that file. `globalSetup` precedes the load task, so this lands before the first
+ * title is computed. See `helpers/provider-setup/catalog-snapshot.ts` for why the
+ * catalog is a run-scoped value rather than a file each process re-reads.
+ *
+ * Never fatal: an unreadable catalog is exactly the state the resolvers already
+ * handle, and reporting it here means the cause is named before the specs skip
+ * (#1012) rather than aborting a whole shard over a file that is gitignored by design.
+ */
+function freezeCatalog(): void {
+  const verdict = freezeModelCatalog();
+  if (verdict.reason) {
+    console.warn(`[preflight] WARNING: ${verdict.reason}`);
+  }
+  if (verdict.kind === "absent") {
+    console.log(
+      "[preflight] models.json absent — the EMPTY catalog is frozen for this run. " +
+        "Provider-parametrized specs will report the fallback target; run " +
+        "`npx playwright test tests/collect-models.spec.ts` first to populate it.",
+    );
+    return;
+  }
+  console.log(
+    `[preflight] model catalog frozen for this run` +
+      (verdict.count === undefined ? "" : ` — ${verdict.count} models`) +
+      ". A mid-run write to models.json can no longer change a test's title (#1386).",
+  );
+}
+
 export default async function globalSetup(): Promise<void> {
+  freezeCatalog();
   const ctx = await playwrightRequest.newContext({ baseURL: BASE_URL });
   try {
     await assertBackendHealthy(ctx);
--- a/tests/helpers/provider-setup/catalog-snapshot.test.ts
+++ b/tests/helpers/provider-setup/catalog-snapshot.test.ts
@@ -1,0 +1,199 @@
+// Unit tests for the run-scoped model-catalog snapshot (issue #1386).
+// Run with: npm run test:units
+//
+// What rides on this: the TITLE of 19 spec files. 18 derive it from
+// `resolveTestTargets()` and one from `resolveGeminiModel()`, and a title is a test's
+// identity — the runner computes it at collection, each worker recomputes it at load.
+// When those two disagree Playwright reports `Test not found in the worker process`
+// with `duration 0` and never opens a browser, which on the daily of 2026-08-10 was
+// then read by the auto-removal path as a hard failure and cost an `@stable` tag.
+//
+// The disagreement is not hypothetical and does not need a rare race: the writer is
+// `tests/collect-models.spec.ts`, which is `@stable` and therefore partitioned into a
+// shard's own spec list, and it rewrites `models.json` from inside the run.
+//
+// So the assertions below are all one property, stated from different sides: **what a
+// resolver returns must not depend on when it is called.**
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import {
+  CATALOG_SNAPSHOT_ENV,
+  MODELS_PATH,
+  freezeModelCatalog,
+  readCatalogText,
+} from "./catalog-snapshot";
+import { resolveGeminiModel } from "./resolve-gemini-model";
+import { resolveGptModel } from "./resolve-gpt-model";
+import { resolveTestTargets } from "./test-targets";
+
+const OPENAI_ONLY = [{ provider: "openai", model: "gpt-4o-mini" }];
+const WITH_GOOGLE = [
+  { provider: "openai", model: "gpt-4o-mini" },
+  { provider: "google", model: "gemini-flash-latest" },
+];
+
+/** A temp file standing in for `models.json`, seeded with `records`. */
+function tempCatalog(records: unknown, name = "models.json"): string {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), "catalog-snapshot-"));
+  const file = path.join(dir, name);
+  fs.writeFileSync(file, JSON.stringify(records));
+  return file;
+}
+
+/**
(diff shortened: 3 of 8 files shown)


rafaelgiln:
## The red `Run impacted E2E specs` is a drained OpenAI key, not this diff

Both runs (`93636738457`, rerun `93645667548`) fail the same single test — `openai-provider.spec.ts:201`, `configured OpenAI selects a GPT model in the Agent and executes the flow` — with 62 passed / 36 skipped alongside it.

The provider says so verbatim, once per attempt:

```
Error: You have no credits remaining. Add credits to continue using the API
at https://platform.openai.com/settings/organization/billing/.
```

The agent therefore produces no reply and `div-chat-message` never renders, which surfaces as `expect(locator).toBeVisible() failed` — a UI-shaped symptom for a billing cause. `Collect models` cannot catch it: it validates the key with a listing call that answers 200 without credit, so the provider is recorded `active` and the spec runs.

Three independent reasons this is not caused by the change under review:

1. The freeze worked on both runs — `[preflight] model catalog frozen for this run — 89 models`. `resolveGptModel()` read a full catalog.
2. `collect-models.spec.ts` is **not** in this lane's spec list (20 selected), so nothing rewrote `models.json` mid-run and the frozen value is byte-identical to what the previous code would have read from disk.
3. Every other spec that resolves a model through the frozen catalog passed, including all 18 that derive their title from it.

Prior art for the same test and signature, before this branch existed: daily failures on 2026-07-15 and 2026-08-06 (`expect(locator).toBeVisible() failed`) and 2026-08-05 (credential-settle). What is new today is only that the key is dry, so the failure is reproducible instead of intermittent.

No test-side action is available or appropriate here — the fix is credit on the account.

PR #1405, opened 2026-08-11
Cover the MCP single-server read-back and update API

### LINKED ISSUE #1397: Extend mcp-server.spec.ts — read back and update a registered server

**Backlog pointer:** QA-CHECKLIST.md §14.1 (1 bullet) → `mcp/server/mcp-server.spec.ts` (extends the existing spec)

**Scope**
- A registered server is read back individually via `GET /servers/{name}` with the same fields it was created with, and `PATCH /servers/{name}` updates them.
- Only the list, create, delete and 409/404 paths are covered today, and they already live in this file — the read-back belongs with them rather than in a new spec.

**Note:** the file is `@stable`. Editing it requires the full validation pipeline (typecheck, lint, linter skill, `--retries=0` run, force-fail, `--trace=on`, backend-error audit) before the PR.

Victor-w-Madeira:
Closes #1397

## Why

Every test in `mcp/server/mcp-server.spec.ts` registers through the Add MCP Server modal and spot-checks the result. `GET /api/v2/mcp/servers/{name}` and `PATCH /api/v2/mcp/servers/{name}` — the endpoints the modal's own edit path is built on — had no coverage of their own; only list / create / delete and the 409/404 pair (`mcp/client/mcp-server-registration-status-codes.spec.ts`) did. QA-CHECKLIST §14.1, one new bullet.

## What the two tests assert

**Read-back.** A server registered with `command` + `args` + `env` reads back through the single-server endpoint with a body **deep-equal** to what was posted — including the `env` pair, which round-trips through the encrypted column and comes back decrypted — carrying no `name` key (the name is owned by the URL path), and the same server is visible in the list endpoint.

**Update.** A `PATCH` carrying `args` **and nothing else** is visible on a subsequent `GET` while `command` and `env` — two keys the patch never mentioned — survive; an `env`-only patch leaves `command`/`args` intact and *replaces* the whole `env` object rather than deep-merging into it; a body `name` disagreeing with the URL is refused 422 (`name is immutable`) and changes nothing, not even as stray config.

Three design choices, each measured on nightly `1.12.0.dev20` rather than assumed:

- **No subprocess, no npm registry.** Registration persists the config without probing it, so both tests register a deliberately non-existent package — the same reasoning as the `uvx mcp-server-test` persistence test in the same file. That is what lets them ship `@stable` with nothing to flake on.
- **Every assertion reads back with a fresh `GET`**, never the write's own response body: `POST` and `PATCH` both echo the config, so an endpoint that validated without persisting would pass an echo-only check.
- **Pure `page.request`**, so the deliberate 422 never reaches the fixture's `page.on("response")` monitor and needs no `allowHttpErrors()`. Proven by contrast in a full-file run: the browser-driven contract test prints `🚨 Backend Error: 422`, these two print nothing.

## Two product behaviours found and deliberately NOT asserted

Both measured, both documented in the spec doc under *What this test does not cover*:

- `GET /api/v2/mcp/servers/{unknown}` answers **200 `null`** rather than the 404 its sibling `DELETE` returns (upstream #14005 fixed the delete path and left this one).
- `PATCH /api/v2/mcp/servers/{unknown}` answers **200 and creates the server** — a typo'd name silently registers a ghost.

Asserting today's behaviour would enshrine something that looks wrong; asserting the corrected behaviour would ship a durably red `@stable` test that triage strips within a day. **These want a product-finding issue** — happy to open one on request.

A third asymmetry is recorded the same way: a merge patch carrying `args` without `command` is validated with no command in scope, so `{"args": ["-y", …]}` is refused 422 while the identical args are accepted by `POST` alongside `command: npx`.

## Validation

Against Langflow Nightly **1.12.0.dev20**, per CONTRIBUTING:

- `typecheck` ✓ · `lint` 0 errors ✓ · `check:checklist-coverage` ✓ · `check-checklist-guard` ✓ · `validate:specs` exit 0 (and this doc left the "no `src/` paths" advisory list — its dependency entries now lead with the backticked path, the shape that checker reads).
- `--retries=0` runs green, 3× burst, `--trace=on` green, and `--workers=4 --repeat-each=3` → 6 passed with the server list back to its prior two entries.
- **Seven force-failure / mutation checks** across the two rounds — a mutated expectation, a dropped `env` on the POST, a PATCH that persists nothing, a rename sent with the matching name, a merge that drops `command`, an extra key in the stored config, and the reachability of the relocated `name` assertion — each failing for the right reason. (A first attempt mutated a constant used on *both* sides of the round trip, proved nothing, and was redone.)
- No `🚨 Backend Error` from either test; no leaked MCP servers afterwards.
- Full-file run: 7 passed, 1 skipped (the #1266 `test.fixme` quarantine), 1 failed — test 6 "Streamable HTTP … server-everything", which is not `@stable`, fails on the #1335 `add-mcp-server-button` signature, and which the spec doc already records as failing identically with and without a change. The independent reviewer's own full-file run had it **passing**, which fits a registry/timing difference; the diff is additive and 200 lines below it either way.

## Commits

1. `ed60c31` — the two tests, the spec doc, the checklist bullet.
2. `5b31038` — independent-review fixes: `@regression` dropped (it means "previously fixed bug"; these are new coverage), the cleanup registration moved **before** the status assertion so a commit-then-non-2xx POST cannot leak a server, the `name`-absence assertion moved above the deep equality where it can actually report, and the first PATCH reduced to `args` only after measurement refuted the comment justifying `command` alongside it.
3. `0a7ad8a` — whitespace only: `authoring-conventions.md` requires the test title on the `test(` line and asks for a retrofit when editing a file that wraps them. `--list` reports the same 9 tests with the same titles before and after.

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -744,6 +744,7 @@
 - [x] Tool list refreshes when a registered server is edited to run a different package → `mcp/server/mcp-server.spec.ts`
 - [x] stdio `command` must be a single executable — a command with an embedded argument is refused and the same registration split into `command` + `args` is accepted (upstream hardening `#14073`; #1091) → `mcp/server/mcp-server.spec.ts`
 - [x] The project's own Streamable HTTP endpoint registers as an MCP server and exposes its flows as tools → `mcp/server/mcp-server.spec.ts`
+- [x] A registered server is read back individually via `GET /servers/{name}` with the fields it was created with, and `PATCH /servers/{name}` updates them — merging per top-level key and refusing to rename (#1397) → `mcp/server/mcp-server.spec.ts`
 - [-] Resource exposed by server is accessible via URI — flow files are exposed as MCP resources: `resources/list` is `@stable`; `resources/read` blocked by a live Langflow regression on 1.12.x (`AttributeError: 'str' object has no attribute 'hex'`, filed upstream **LE-2012**) — kept as a guard, not promoted → `mcp/server/mcp-server-resources.spec.ts`
 - [ ] Install this project into an MCP client — `GET /{project_id}/installed` reports the current state, `POST /{project_id}/install` performs it, and the UI reflects both (the user-facing way an MCP Server is actually consumed; no bullet covered this before 2026-08-06)
 - [ ] `GET /{project_id}/composer-url` returns a URL that resolves, and the copy control in the UI yields the same value
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -744,12 +744,11 @@
 - [x] Tool list refreshes when a registered server is edited to run a different package → `mcp/server/mcp-server.spec.ts`
 - [x] stdio `command` must be a single executable — a command with an embedded argument is refused and the same registration split into `command` + `args` is accepted (upstream hardening `#14073`; #1091) → `mcp/server/mcp-server.spec.ts`
 - [x] The project's own Streamable HTTP endpoint registers as an MCP server and exposes its flows as tools → `mcp/server/mcp-server.spec.ts`
-- [x] A registered server is read back individually via `GET /servers/{name}` with the fields it was created with, and `PATCH /servers/{name}` updates them — merging per top-level key and refusing to rename (#1397) → `mcp/server/mcp-server.spec.ts`
 - [-] Resource exposed by server is accessible via URI — flow files are exposed as MCP resources: `resources/list` is `@stable`; `resources/read` blocked by a live Langflow regression on 1.12.x (`AttributeError: 'str' object has no attribute 'hex'`, filed upstream **LE-2012**) — kept as a guard, not promoted → `mcp/server/mcp-server-resources.spec.ts`
 - [ ] Install this project into an MCP client — `GET /{project_id}/installed` reports the current state, `POST /{project_id}/install` performs it, and the UI reflects both (the user-facing way an MCP Server is actually consumed; no bullet covered this before 2026-08-06)
 - [ ] `GET /{project_id}/composer-url` returns a URL that resolves, and the copy control in the UI yields the same value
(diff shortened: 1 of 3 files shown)


Victor-w-Madeira:
## `Run impacted E2E specs` is red — measured, and it is not this PR

The lane failed two tests of `mcp-server.spec.ts`: `mcp server tools should be refreshed when editing a server` (`@stable`) and `Streamable HTTP … server-everything` (not `@stable`). Neither is one of the two tests this PR adds — both of those passed.

I did not want to label it "known flake" without evidence, since the `@stable` one is green in the last 10 dailies. So I measured it against a **disposable Langflow container** (`podman run` on :7861, never used, no provider configured — the same shape the impacted-specs lane gives a PR that needs no models), recreating the container before each run so every measurement starts from identical state, at `--workers=2 --retries=0` like CI:

| ref | expected | unexpected | which failed |
|---|---|---|---|
| `main`, run A | 3 | **3** | HTTP/SSE persist, tools refreshed, Streamable HTTP |
| `main`, run B | 5 | **1** | Streamable HTTP |
| this branch | 6 | **2** | HTTP/SSE persist, Streamable HTTP |

Two readings:

1. **`main` alone fails 1–3 of its 7 tests** under these conditions, and the failing set varies between two identical fresh instances. This branch's 2 failures are a subset of `main`'s run A and sit inside that spread — the addition is not distinguishable from the file's own variance.
2. **Both tests this PR adds passed in every run** (they are in the `expected` count each time). They are pure API, no modal, no `npx`.

So the red is pre-existing instability of the file under 2 workers, not a regression from this diff — the case `CONTRIBUTING.md` treats as non-blocking for this lane.

Worth flagging separately, though: **#1335 is closed**, and it is the issue that tracked exactly this signature (`add-mcp-server-button` / the MCP-server dropdown not clickable). The daily hides it because CI runs with `retries: 2`, so a 1-in-3 flake mostly recovers and is recorded as flaky rather than failed. At `retries=0` it is plainly visible. That deserves its own `qa-infra` issue rather than riding along here.

Issue #1409, opened 2026-08-11
Two projects whose names share their first 26 characters cannot coexist — POST /api/v1/projects/ returns 409 naming an MCP server the user never created


Victor-w-Madeira:
Product finding surfaced while writing the per-project MCP exposure coverage for #1396. It is not MCP-specific from the user's point of view: it blocks **project creation** with an error about a subsystem they never touched.

## Measured

Langflow Nightly **1.12.0.dev20**, superuser, two ordinary project names:

```bash
T=$(curl -s localhost:7860/api/v1/auto_login | jq -r .access_token)
new() { curl -s -o /dev/stderr -w '%{http_code}\n' -X POST localhost:7860/api/v1/projects/ \
        -H "Authorization: Bearer $T" -H 'Content-Type: application/json' -d "{\"name\":\"$1\"}"; }

new "Marketing Automation Project Alpha"   # 201
new "Marketing Automation Project Beta"    # 409
```

```json
{"detail": "MCP server name conflict: 'lf-marketing_automation_proje' already exists for a different project. Cannot create MCP server for project 'Marketing Automation Project Beta' (ID: …)"}
```

The two names differ — they simply agree on their first 26 sanitized characters. Short, distinct names are unaffected.

## Mechanism

`src/backend/base/langflow/api/utils/mcp/config_utils.py:128`

```python
server_name = f"lf-{sanitize_mcp_name(project_name)[: (MAX_MCP_SERVER_NAME_LENGTH - 4)]}"
```

`MAX_MCP_SERVER_NAME_LENGTH` is 30 (`src/lfx/src/lfx/base/mcp/constants.py`), so the project name is cut to **26 characters** and prefixed with `lf-`. Every project auto-registers an MCP server entry under that derived name — visible in `GET /api/v2/mcp/servers` as e.g. `lf-starter_project`. On create, the same function looks the name up in the user's registered servers, finds it pointing at a different project's SSE URL, and the operation is refused (`:158-165`).

So the uniqueness constraint is applied to a **truncated derivation** of the name rather than to the name itself, and it is enforced on the *project*, not on the MCP entry.

## Why it matters

- It refuses a legitimate, unambiguous project name. Long descriptive project names are normal, and the first 26 characters colliding is easy — "Customer Support Automation — Tier 1" / "… Tier 2" is the same failure.
[...]

Issue #1408, opened 2026-08-11
Per-project MCP exposure is a discovery control only: a de-selected flow is unlisted but still callable over tools/call


Victor-w-Madeira:
Product finding surfaced while writing the per-project exposure coverage for #1396. Not a suite defect — the spec deliberately stops short of asserting it, and this issue is why.

## Measured

Langflow Nightly **1.12.0.dev20**, against a purpose-made project + a Chat Input → Chat Output passthrough flow:

| Step | Result |
|---|---|
| `PATCH /api/v1/mcp/project/{p}` → `mcp_enabled: true`, `action_name: X` | `tools/list` contains `X`; `tools/call X` runs the flow, `isError: false` |
| `PATCH` → `mcp_enabled: false` | `GET /api/v1/mcp/project/{p}` → `tools: []`, and `tools/list` → `[]` ✅ |
| `tools/call X` **after** de-selection | **200, `isError: false`, the flow runs and echoes the input** ❌ |
| `tools/call` on a name that never existed | `isError: true`, `"Flow with name '…' not found"` |

**It is not a stale cache.** A flow that was enabled, then de-selected, and **never called while enabled** executes just the same on the first `tools/call` it ever receives.

## Where it comes from

- `src/backend/base/langflow/api/v1/mcp_projects.py` — the project server registers
  `handle_list_project_tools` → `handle_list_tools(project_id=…, mcp_enabled_only=True)`. Listing filters correctly.
- `src/backend/base/langflow/api/v1/mcp_utils.py` → `handle_call_tool` resolves the flow by action name (`get_flow_snake_case(..., is_action=True)`), verifies `flow.folder_id == project_id`, and enforces `FlowAction.EXECUTE` — but **never reads `flow.mcp_enabled`**.

So the two halves of the same control disagree: `mcp_enabled` gates discovery, not invocation.

## Why it matters

"Expose this flow as a tool" is the only per-project control a user has over what a connected MCP client can reach, and the UI presents it as a checkbox per flow. After unchecking it, any client that already knows — or guesses — the action name keeps full access to run the flow, with whatever side effects that flow has. Nothing in the API or the UI signals that.

Note the call path already reasons about authorization: the `ensure_flow_permission(EXECUTE)` call right below carries a comment about stopping an external "viewer" from running a flow as a tool. The missing `mcp_enabled` check sits next to code written for exactly this class of concern, which is what makes an oversight the likelier reading over a deliberate design.

[...]

PR #1410, opened 2026-08-11
Cover per-project MCP tool exposure over the protocol

### LINKED ISSUE #1396: Create mcp-server-project-config.spec.ts — per-project tool exposure over the protocol

**Backlog pointer:** QA-CHECKLIST.md §14.1 (1 bullet) → `mcp/server/mcp-server-project-config.spec.ts`

**Scope**
- `GET`/`PATCH /{project_id}` selects which flows are exposed as tools.
- A de-selected flow disappears from `tools/list` **over the protocol**, not only from the UI listing.
- Reuse `tests/helpers/mcp/mcp-streamable-client.ts` for the protocol-side assertion rather than hand-rolling a client.

Victor-w-Madeira:
Closes #1396

## Why

`PATCH`/`GET /api/v1/mcp/project/{id}` decide which of a project's flows are exposed as MCP tools, and nothing asserted that the **protocol** agrees with that selection: `mcp-server-tab.spec.ts` covers the UI, and `mcp-server-protocol.spec.ts` covers a flow that is already exposed on the default project. QA-CHECKLIST §14.1.

## What lands

Two tests, pure API + JSON-RPC — no browser, no LLM key, no npm registry:

- **The selection round-trips.** A freshly created project exposes nothing; one `PATCH` exposes exactly the flow it names; `GET` reads back both the action pair and the flow's own name/description, which are not the same thing.
- **The protocol reflects it, both ways.** The action appears in `tools/list` under its action name and its **action** description, `tools/call` runs it and echoes a unique sentinel — so the listing is a working capability, not a string — and after `mcp_enabled: false` it is **gone from `tools/list`**, which is where the issue asks for the assertion. `GET` agrees, and the unfiltered listing still finds the flow, proving it was withdrawn rather than deleted.

It creates its **own project and flow**, so it cannot race `mcp-server-protocol.spec.ts` (which mutates the default project and carries a comment anticipating exactly this second spec). The protocol client is the repo's `mcp-streamable-client.ts`, as the issue requires. The flow is created by a file-local helper rather than by extending `createRunnableChatFlowViaApi` with a `folder_id`: that helper has 29 call sites across 21 specs, and the import-graph selection (#1054) would pull every one of them into this PR's E2E lane for two lines.

## Three product behaviours measured on the way — two shaped the spec, one is a filed defect

**A de-selected flow is unlisted but still callable.** `tools/call` on a flow whose `mcp_enabled` is false runs it and returns `isError: false`. Not a cache: a tool never called while enabled behaves the same, while an unknown name answers `Flow with name '…' not found`. `handle_list_project_tools` passes `mcp_enabled_only=True`; `handle_call_tool` resolves by action name, checks the project, enforces `FlowAction.EXECUTE` — and never reads `mcp_enabled`. **The per-project selection is a discovery control, not an invocation control.** Filed as **#1408**; recorded under *What this test does not cover* rather than asserted, since asserting the corrected behaviour ships a durably red `@stable` test that triage strips within a day.

**Two projects whose names share their first 26 characters cannot coexist** — `POST /api/v1/projects/` answers 409 naming an `lf-…` MCP server the user never created. This spec's first version hit it (`namePrefix: "e2e-mcp-project-config"` pushed the uniqueness suffix past the cut), and it only appeared under `--workers=4 --repeat-each=3`. It is not a test artefact: `Marketing Automation Project Alpha` creates and `… Beta` is refused. Filed as **#1409**.

**A truncated tool name is uninvocable.** `tools/list` caps the served name at 30 while the lookup does not, so a 37-character action is advertised under a name that `tools/call` answers `not found` for. Recorded in the spec doc; MCP tool naming deserves its own spec.

## Validation

Against Langflow Nightly **1.12.0.dev20**, per CONTRIBUTING:

- typecheck ✓ · lint 0 errors ✓ · both checklist guards ✓ · `validate:specs` counts the doc as declaring its dependencies ✓
- 3× `--retries=0` burst green (~0.9 s each) · `--trace=on` green · `--workers=4 --repeat-each=3` → 6 expected / 0 unexpected · alongside `mcp-server-protocol.spec.ts` at `--workers=4` → 4 passed, no interference
- **Eight force-failure / mutation checks** across both rounds, each failing for the right reason; no `🚨 Backend Error`; instance left byte-identical (1 project, 28 flows, 2 MCP servers), including after a run where teardown was forced to fail.

## Commits

1. `472cde9` — the spec, the spec doc, the checklist bullet.
2. `80914fd` — **independent-review fixes**, and the first two are the ones worth reading:
   - the de-selection assertion **could pass vacuously** — `resp.result?.tools ?? []` turned any failed JSON-RPC call into an empty list, so a broken protocol would have reported green. Both `tools/list` calls now verify the response first; the reviewer's mutation fails against the fix.
   - `?mcp_enabled=false` **does not list the disabled flows** — it removes the filter entirely, returning every flow in the project. My doc claimed a contract the endpoint does not have.
   - `GET` does not return `action_name` untouched — it sanitizes at cap 46 against the protocol's 30, so the `beforeAll` guard now checks the character class as well as the length.
   - a **duplicate checklist bullet**: §14.1 already carried this behaviour as `[ ]` and I appended a second. The existing one is now checked. *(The same defect was in PR #1405 and is fixed there — `b2f0bf7`.)*
   - teardown no longer swallows cleanup failures.

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -747,9 +747,8 @@
 - [-] Resource exposed by server is accessible via URI — flow files are exposed as MCP resources: `resources/list` is `@stable`; `resources/read` blocked by a live Langflow regression on 1.12.x (`AttributeError: 'str' object has no attribute 'hex'`, filed upstream **LE-2012**) — kept as a guard, not promoted → `mcp/server/mcp-server-resources.spec.ts`
 - [ ] Install this project into an MCP client — `GET /{project_id}/installed` reports the current state, `POST /{project_id}/install` performs it, and the UI reflects both (the user-facing way an MCP Server is actually consumed; no bullet covered this before 2026-08-06)
 - [ ] `GET /{project_id}/composer-url` returns a URL that resolves, and the copy control in the UI yields the same value
-- [ ] Per-project MCP configuration — `GET`/`PATCH /{project_id}` selects which flows are exposed as tools, and a de-selected flow disappears from `tools/list` over the protocol
+- [x] Per-project MCP configuration — `GET`/`PATCH /{project_id}` selects which flows are exposed as tools, and a de-selected flow disappears from `tools/list` over the protocol (invocation after de-selection is a separate finding, #1408) → `mcp/server/mcp-server-project-config.spec.ts`
 - [x] A registered server is read back individually via `GET /servers/{name}` with the same fields it was created with, and `PATCH /servers/{name}` updates them — merging per top-level key and refusing to rename (#1397) → `mcp/server/mcp-server.spec.ts`
-- [x] Per-project tool exposure — `PATCH`/`GET /api/v1/mcp/project/{id}` selects which flows are exposed, and a de-selected flow disappears from `tools/list` **over the protocol** (invocation after de-selection is a separate finding, #1408) → `mcp/server/mcp-server-project-config.spec.ts`
 - [ ] Prompt exposed by server returns correct template (no product surface on 1.11.x — MCP server `prompts/list` returns `[]`; #829)
 
 ---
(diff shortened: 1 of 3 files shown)


Victor-w-Madeira:
Both of Copilot's comments are addressed in 427f328 — they are two halves of the same wording defect.

The claim that a prefix longer than 6 makes *every* project collide is wrong: the derived `lf-${sanitize_mcp_name(name)[:26]}` truncates the 20-character `-${Date.now()}-${rand5}` suffix from the right, so ≤6 *guarantees* the whole suffix survives and past that the conflict risk grows with the prefix (at 7 characters four of the five random characters still survive). What was measured is the 22-character prefix the first version used, which left nothing past the first three digits of the timestamp — that one collides every time.

The spec doc headlined the same rule as "is six characters" and closed with "this spec's own prefix is five characters"; it is a ceiling, not a length. Both now read as the constraint, with this spec at five and one character of slack.

No behaviour change — the prefix stays `e2ecf`.

PR #1420, opened 2026-08-11
Ignore typescript >= 6, and replace the measurement that was wrong

### LINKED ISSUE #1419: dependabot.yml records a measurement that is wrong: TS 6 needs two tsconfig lines, not one
`.github/dependabot.yml` closes its `typescript` `ignore` entry with a measurement from #1228:

> the same tsconfig yields `TS5107 ... is deprecated and will stop functioning in TypeScript 7.0`, silenced by one `ignoreDeprecations: "6.0"` line

**That is wrong, and it is wrong in the direction that costs the reader an afternoon.** Measured against 6.0.3 with a real `npm ci` while reviewing Dependabot PR #1390: with only that line, `tsc --noEmit` exits 0 while **every ts-node-driven gate stays broken** — `test:units` 39 failures, plus `check:checklist-coverage`, `regressions:check` and `coverage:summary`, all reporting `TS2591 Cannot find name 'fs' / 'process' / '__filename'`. Under TS 6 ts-node no longer picks up `@types/node` automatically; isolated to `npx ts-node -e 'import * as fs from "fs"'`, which fails without `types: ["node"]` and passes with it.

Three variants measured:

| variant | `tsc` | ts-node |
|---|---|---|
| `ignoreDeprecations: "6.0"` only | pass | **39 failures** |
| `module` + `moduleResolution: node16` (the real migration) | pass | **39 failures** |
| `ignoreDeprecations: "6.0"` **+** `types: ["node"]` | pass | pass |

So TS 6 is adoptable at **two** tsconfig lines (tsc 0, `test:units` 552/552, `test:scripts` 796/796, all three ts-node gates green, lint unchanged) — not one. The `node16` variant is the dangerous one: `tsc` green, ts-node dead, which the TypeScript job alone cannot see.

**Two things to fix.**

1. **The note.** Replace the one-line claim with the table above, so the next reevaluation starts from what was measured rather than repeating the check.
2. **The range.** #1390 was closed on the evidence that TS 6 costs a `tsconfig.json` change every contributor inherits, for no feature this suite uses, and that `typescript-eslint@8.66` peers `typescript >=4.8.4 <6.1.0` — so 6.1 breaks `npm ci`, the install, not the typecheck. Closing the PR alone does not stick: Dependabot proposes 6.0.4 next week. The existing `ignore` entry should widen from `>=7` to `>=6`, with the exit condition named (a reason to want TS 6, or typescript-eslint admitting 6.1+).

Cost of widening, documented in the file's own header: `ignore` also suppresses Dependabot **security** updates in that range. For a dev-only compiler an advisory still surfaces as an alert — it just produces no PR.

Related and settled in the same sweep: the `@flakiness/playwright` note directly below says v2 "arrives as its own PR, which is where its two real breaking changes ... can be checked". It arrived (#1389) and they were checked, so that note has expired too and should record the outcome instead of the plan.

Victor-w-Madeira:
Closes #1419

Two notes in `.github/dependabot.yml` had outlived what they described. The first was not merely stale — it was **false**, and false in the direction that costs the reader an afternoon.

## The TS 6 note

It recorded, from #1228, that `TS5107 moduleResolution=node10 is deprecated` is *"silenced by one `ignoreDeprecations: "6.0"` line"*. It is silenced for `tsc` **only**. Measured against 6.0.3 with a real `npm ci` (in a throwaway worktree, so nothing was read off `npm ls`):

| variant | `tsc` | ts-node |
|---|---|---|
| `ignoreDeprecations: "6.0"` only | pass | **39 failures** |
| `module` + `moduleResolution: node16` (the real migration) | pass | **39 failures** |
| `ignoreDeprecations: "6.0"` **+** `types: ["node"]` | pass | pass |

The ts-node column is `test:units`, and the same break takes `check:checklist-coverage`, `regressions:check` and `coverage:summary` with it — `TS2591 Cannot find name 'fs' / 'process' / '__filename'`. Under TS 6 ts-node stops picking up `@types/node` automatically; isolated to `npx ts-node -e 'import * as fs from "fs"'`, which fails without `types: ["node"]` and passes with it.

Worth noting which variant is the dangerous one: the *"correct"* `moduleResolution` migration leaves `tsc` green and ts-node dead — a state the TypeScript job alone cannot see. The QA-CHECKLIST guard job catches it, because it runs through ts-node.

So TS 6 costs **two** tsconfig lines, not one, and with them everything passes (tsc 0, `test:units` 552/552, `test:scripts` 796/796, three ts-node gates green, lint unchanged). What made it not worth paying — the reason PR #1390 was closed — is now in the file: no language feature this suite uses; `types: ["node"]` disables automatic `@types` inclusion, harmless while `@types/node` is the only one and a silent trap the day a second arrives; and `typescript-eslint@8.66` peers `typescript >=4.8.4 <6.1.0`, so **6.1 breaks `npm ci`** — the install, not the typecheck.

**The range widens `>=7` → `>=6`** with the exit condition named, because closing #1390 alone does not stick: Dependabot proposes 6.0.4 next week. Cost of widening, documented in this file's own header: `ignore` also suppresses **security** updates in that range. For a dev-only compiler an advisory still surfaces as an alert — it just produces no PR.

## The @flakiness/playwright note

It said v2 *"arrives as its own PR, which is where its two real breaking changes ... can be checked"*. It arrived (#1389) and they were checked — against the installed package, not the changelog: the dropped `flakiness-playwright-shard` bin is unused (only `flakiness-playwright-timings` remains in its `bin`), and `preprocess()` — the hook wanting Playwright 1.62+ against the 1.58.2 pin — **returns immediately unless `shardBalancing` is configured**, as does the "requires 1.62+" warning, so the version gap is unreachable. `flakinessProject` and the `FK_ENV_*` vars `daily-stable.yml` relies on both survive. Smoke-tested on node 20.20.2 + PW 1.58.2: loads, runs, writes a report, exit 0, no warning.

The note now records the outcome instead of the plan — **including the gap**: the OIDC upload path is exercised by neither the smoke test (no local credential) nor the PR lane (a manifest-only change resolves to zero impacted specs), so the next daily is its first real exercise.

## The guard had to change, and that is the interesting part

`scripts/dependabot-config.test.mjs` asserted the literal `[7]`, so it **failed the PR implementing the decision it exists to protect** — exactly the failure its own header warns about, and the second time this file has had it (the first was the @flakiness/playwright guard tied to the repo's `node-version:` pins, which would have failed the correct change).

It now derives the bound from `package.json`: the ignore must start at the **next** major, because a TypeScript major is never a version bump here and Dependabot edits only the manifest and lockfile. That invariant does not expire — adopting TS 6 moves both numbers. A higher bound is the same hole from the other side (it would leave the majors in between open to the bot), so the assertion is equality, not `>`.

Two mutations killed: `>=7` (leaves TS 6 to the bot) and `>=5` (blocks the line the repo is on — trips this test *and* the pre-existing `no ignore range can block the major this repo is actually on`).

The `update-types: semver-major` assertion is **deliberately dropped**: it existed because that shorthand would also have blocked TS 6, and that objection is gone now that blocking TS 6 is the decision. Keeping a guard whose stated reason expired is the defect this PR is about.

## Validation

`test:scripts` 797/797, `test:units` 598/598, typecheck clean, lint unchanged (0 errors). No spec or doc touched, so no QA-CHECKLIST bullet applies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -93,7 +93,7 @@ updates:
       - dependency-name: "@playwright/test"
       - dependency-name: "playwright"
 
-      # typescript >= 7 ONLY, and scoped by version range rather than by
+      # typescript >= 6, and scoped by version range rather than by
       # `update-types: semver-major` on purpose. TS 7 is the native (Go) port, and it
       # is not a version bump for this repo — it is three migrations at once, each
       # measured against 7.0.2 on #1204:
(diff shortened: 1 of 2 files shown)


Victor-w-Madeira:
Addressed in 522240d — the finding is valid, and the same unguarded `declared.match(/(\d+)/)[1]` was already in the sibling test, so it is one mechanism rather than two patches: `declaredMajor(name)` asserts the dependency is declared and that its specifier carries a numeric major, and both tests read through it.

Neither state is theoretical, which is why it is worth the lines: `typescript` leaving package.json is on the table for TS 7 — replacing it is what this very ignore's rationale describes — and a specifier with no numeric major is `"latest"`, a git URL or `workspace:*`.

Measured, since the point is the message and not the verdict (both states failed before too, just illegibly):

| mutation | before | now |
|---|---|---|
| `typescript` deleted from package.json | `Cannot read properties of undefined (reading 'match')` | `dependabot.yml ignores "typescript", which package.json does not declare — remove the ignore entry or restore the dependency` |
| `"typescript": "latest"` | `Cannot read properties of null (reading '1')` | `package.json declares "typescript": "latest", which carries no numeric major` |

Both tests report it, not just the new one.

One more of the same class while in there: the version-range read in `no ignore range can block the major this repo is actually on` also did `match(...)[1]` unguarded, so an entry carrying an upper bound or a wildcard died on the plumbing too. It now names the entry and the range, and asks for the guard to be widened deliberately rather than left unable to judge the entry.

`test:scripts` 797/797, lint clean.

Issue #1424, opened 2026-08-11
[Daily #1417] openai-provider — the global-variable persist call answers non-2xx while validate-provider succeeds (recurrent 2×)


Victor-w-Madeira:
Spun out of daily-failure triage #1417 (run [31475108157](https://github.com/oriontech-me/langflow-e2e/actions/runs/31475108157), 2026-08-11).

**Upstream:** _not filed_

## Symptom

A `@stable` test flaked (recovered on the third attempt) saving the OpenAI key in Settings → Model Providers. The test arms two waiters before clicking Save and then asserts both responses are 2xx. **`validate-provider` succeeded and the persist call did not**: the failure is on `expect(persistResp.ok()).toBe(true)` at line 197, one line after the validate assertion passed. The persist request therefore *fired and was answered* — it did not time out — and it came back outside 2xx.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/model-provider/openai-provider.spec.ts:138` ("OpenAI API key is configured via Settings → Model Providers") | `POST/PATCH /api/v1/variables/ — response.ok()` | `Error: expect(received).toBe(expected) // Object.is equality` |

## Why these failures are one cause

One test, one signature, filed on the **recurrence criterion**: the same test failed under the same `error_signature` on 2026-07-13 and 2026-08-11, which the dataset reports as `same_signature: true` inside the 30-day window.

**The strength of that match has to be stated, because the signature is a generic one and the test contains two assertions that produce it.** `expect(received).toBe(expected) // Object.is equality` is emitted identically by `expect(validateResp.ok())` and `expect(persistResp.ok())` — two adjacent lines whose causes are opposites: the first is "the key does not authenticate against OpenAI", the second is "the key does not persist into Langflow". For **today's** occurrence the report resolves it: the error is at line 197, the persist assertion. For **2026-07-13** it cannot be resolved from the history, which records only file, line and signature — and the context there argues for caution rather than for a match: that daily had 12 hard failures and 17 flakes with the LLM-facing specs failing broadly, during the period when this account's OpenAI key was quota-drained (#772/#775), which is exactly the condition that makes the *validate* assertion fail.

So: the criterion is met mechanically and this issue is filed on it, but the first job of the investigation is to settle whether 2026-07-13 was the same assertion. If it was not, today is a **first occurrence of the persist-side sub-case** and this issue should be re-scoped accordingly rather than carrying a recurrence it does not have.

[...]

Issue #1423, opened 2026-08-11
[Daily #1417] the sidebar add-component click does not place the node on the canvas — the node never enters the DOM (2 specs, recurrent 3× / 2×)


Victor-w-Madeira:
Spun out of daily-failure triage #1417 (run [31475108157](https://github.com/oriontech-me/langflow-e2e/actions/runs/31475108157), 2026-08-11).

**Upstream:** _not filed_

## Symptom

Two `@stable` tests flaked (each recovered on retry) at the **same point of the same interaction**: each searches the flow sidebar, clicks that row's `add-component-button-<name>`, and then waits for the node it just asked for. In both, the wait ends as **`element(s) not found`** — not "present but not visible" — so the node never entered the DOM at all.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/observability-monitoring/flow-error-message.spec.ts:40` ("a misconfigured flow surfaces an appropriate build-error message") | `getByTestId('button_run_api request')` | `Error: expect(locator).toBeVisible() failed` |
| `tests/tests-automations/regression/core-components/chat-input-output-component-regression.spec.ts:127` ("Chat Input component — renders on canvas with Message output handle and Input Text field") | `getByTestId('title-Chat Input')` | `Error: expect(locator).toBeVisible() failed` |

## Why these failures are one cause

The two specs live in different directories, cover different features (a build-error message vs. the Chat Input component's interface) and add different components, so the grouping is argued rather than assumed. `Error: expect(locator).toBeVisible() failed` is the most generic string Playwright emits and would be far too weak to group on by itself. Three further dimensions are shared:

1. **The same action, at the same point.** Both fill `sidebar-search-input`, wait for that row's control to be present, click `add-component-button-<name>`, and then wait for the placed node. In both, the click itself did **not** fail — it is the assertion immediately after that does.
2. **The same observable, in its stronger form.** Both failures report `element(s) not found` after the full 15 s budget. That is a materially different state from a node that rendered late or off-screen: the awaited testid (`button_run_api request`, `title-Chat Input`) was never in the DOM. Whatever the click did, it did not place a node.
3. **Both recovered on the very next attempt**, so in both cases the same sequence succeeded seconds later on the same instance.

[...]

Issue #1422, opened 2026-08-11
[Daily #1417] mcp-server — the stdio server's tool never appears in the tool dropdown, so the refresh-on-edit test cannot complete


Victor-w-Madeira:
Spun out of daily-failure triage #1417 (run [31475108157](https://github.com/oriontech-me/langflow-e2e/actions/runs/31475108157), 2026-08-11).

**Upstream:** _not filed_

## Symptom

One `@stable` test hard-failed (final status, after 3 attempts) — **the tool dropdown opens but never lists the tool served by the stdio MCP server that was just registered**, so the test cannot get to the edit-and-refresh behaviour it exists to assert. All three attempts failed identically, at the same line, with the same signature: this is deterministic on this run, not a flake that ran out of retries.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/mcp/server/mcp-server.spec.ts:780` ("mcp server tools should be refreshed when editing a server") | `locator('[data-testid="sequentialthinking-0-option"]') — visible` | `TimeoutError: page.waitForSelector: Timeout 10000ms exceeded.` |

## Why these failures are one cause

One test, one cause. No grouping argument is being made across specs, and two deliberate exclusions are worth recording so this issue is not later widened by mistake:

- **Not the same cause as #1266** (`mcp-server.spec.ts:131`, currently `test.fixme`'d). That one is transport-level — `apiRequestContext.get: Timeout` on `GET /api/v2/mcp/servers?action_count=true` — filed as a load/reachability issue about the spec. This one is a UI wait on a dropdown option, with the backend answering throughout.
- **Not the same cause as #1332** (`the "Add MCP Server" button never renders`, Streamable-HTTP case). Here the modal opened, the server was created, and the wait that failed is two steps later.

The sibling flake in the same file at `:588` (`HTTP/SSE MCP server fields should persist after saving and editing`) is **not** folded in either: it is a strict-mode violation on a name locator that matched two elements, it is a first occurrence, and it recovered on retry. It is noted against this run, not filed.

## Preliminary read (descriptive — NOT a verdict)

Descriptive only — no verdict on mechanism.

[...]

Issue #871, opened 2026-07-21
Triage quarantine PRs go red on the impacted-specs gate — quarantine via test.fixme, not @stable removal alone


rafaelgiln:
## Problem

Triage quarantine PRs cannot go green through the `pr-validation.yml` **Run impacted E2E specs** gate.

The gate (`.github/workflows/pr-validation.yml`, job `e2e`) selects specs to run **by file diff**, ignoring tags:

```bash
SPECS=$(git diff --name-only --diff-filter=d "origin/$BASE_REF"...HEAD -- 'tests/**/*.spec.ts' ...)
npx playwright test $SPECS --reporter=github
```

So a triage PR whose only change is **removing `@stable`** from a broken/flaky test edits that spec file → the gate runs the whole file → the broken test executes → the gate goes red. The quarantine can never merge cleanly through this gate while the test is broken.

**Observed:** PR #870 (triage #864 — removed `@stable` from 3 recurrent flakes). The `renders on canvas` test in `chat-input-output-component-regression.spec.ts:101` failed final in the gate (click 5000ms timeout), turning the PR `UNSTABLE`. It was merged anyway (no branch protection on `main`), but merging a quarantine PR red is a bad signal and only works because nothing blocks it today.

## Root cause

`@stable` removal targets the **daily** (`daily-stable.yml` runs `@stable` only). It does **not** stop the test running in any other context — the PR impacted-specs gate, `test:features`, `adaptive-impacted.yml`, or manual full runs. So a `@stable`-only quarantine is **incomplete**: the broken test keeps producing red/noise everywhere except the daily.

## Recommendation — quarantine via `test.fixme` (not gate surgery)

Change the triage quarantine mechanism from *tag-removal-only* to **`@stable` removal + `test.fixme`**:

- `test.fixme(...)` is the Playwright-native quarantine primitive — the test is skipped in **every** run context (daily, PR gate, full suite, adaptive, manual), so quarantine PRs merge green and the broken test stops generating noise anywhere.
- Keep removing `@stable` **as well**: the `QA-CHECKLIST.md` Phase 0 "validated" list is generated from `@stable` `test()` calls (`scripts/stable-tests.ts`); leaving the tag on would keep counting a quarantined test as validated.
- **Restoration** (the consumer issue's deliverable) becomes a single clean step: remove the `fixme` and restore `@stable` together, then re-validate per `CONTRIBUTING.md`.

[...]

PR #1425, opened 2026-08-11
chore(triage): quarantine 3 recurrent flakes from the 2026-08-11 daily (#1417)

Victor-w-Madeira:
Prevention step of the triage for daily run [31475108157](https://github.com/oriontech-me/langflow-e2e/actions/runs/31475108157) (umbrella #1417). Three flakes recurred under the same `error_signature` inside the 30-day window, which under `CONTRIBUTING.md` → the criteria table means a dedicated issue **and** a quarantine via PR.

The guard did **not** trip on this run (1 hard failure against a threshold of 5), so the workflow auto-removed `@stable` from the hard failure only — flakes are never auto-removed.

## What changed

Both edits per test, always together — `@stable` removed **and** `test.fixme` added. Tag removal alone only stops the daily; the test keeps going red on the `pr-validation.yml` impacted-specs gate, which selects specs by file diff rather than by tag (#871).

| Test | Recurrence | Issue |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/observability-monitoring/flow-error-message.spec.ts:40` | 3× — 2026-07-16, 07-20, 08-11 | #1423 |
| `tests/tests-automations/regression/core-components/chat-input-output-component-regression.spec.ts:127` | 2× — 2026-07-17, 08-11 | #1423 |
| `tests/tests-automations/regression/core-functionality/model-provider/openai-provider.spec.ts:138` | 2× — 2026-07-13, 08-11 | #1424 |

Each `test.fixme` carries the reason, the recurrence dates and the issue number in a comment above it, so the quarantine explains itself at the call site.

## What this PR is not

It carries **no fix and no investigation** — the triage dispatches, the dedicated issue resolves. Lifting each quarantine (remove `test.fixme`, restore `@stable`, re-validate per `CONTRIBUTING.md`) is a deliverable of #1423 / #1424, so this PR deliberately uses no auto-close keyword.

The hard failure from the same run (`mcp-server.spec.ts:780`, #1422) is **not** here: the workflow already removed its `@stable` in `40f94bd`, and its `test.fixme` belongs to #1422 when it is worked.

## Validation

- `npm run typecheck` — clean
- `npx eslint` on the three specs — 0 errors (only the pre-existing `no-skipped-test` / `no-conditional-in-test` warnings that every `test.fixme` in the suite carries)
- `npm run check:checklist-coverage` — passes; every `@stable` and documented spec is still referenced by a Part II bullet, so no checklist edit is owed
- `scripts/check-checklist-guard.mjs` — no generated block touched
- `npx playwright test --list --grep @stable` on the three files: the quarantined tests no longer select, the untouched ones still do (`flow-error-message` now lists none, `chat-input-output` 5, `openai-provider` 1)

### DIFF
--- a/tests/tests-automations/regression/core-components/chat-input-output-component-regression.spec.ts
+++ b/tests/tests-automations/regression/core-components/chat-input-output-component-regression.spec.ts
@@ -124,9 +124,14 @@ async function runFlowAndOpenChatOutputInspection(page: Page): Promise<string> {
 // Test 1 — Chat Input rendering on canvas
 // =============================================================================
 
-test(
+// Quarantined at triage (daily #1417): recurrent flake — the sidebar
+// add-component click does not place the node, so `title-Chat Input` never
+// enters the DOM (`element(s) not found`, not "present but not visible").
+// Same signature on the 2026-07-17 and 08-11 dailies. Lifting the quarantine
+// (remove test.fixme + restore @stable) is a deliverable of #1423.
+test.fixme(
   "Chat Input component — renders on canvas with Message output handle and Input Text field",
-  { tag: ["@stable", "@regression", "@components"] },
+  { tag: ["@regression", "@components"] },
   async ({ page }) => {
     await addChatInputComponent(page);
 
--- a/tests/tests-automations/regression/core-functionality/model-provider/openai-provider.spec.ts
+++ b/tests/tests-automations/regression/core-functionality/model-provider/openai-provider.spec.ts
@@ -135,9 +135,16 @@ async function expectReplyContainsToken(
 test.describe.configure({ mode: "serial" });
 
 test.describe("OpenAI Provider", () => {
-  test(
+  // Quarantined at triage (daily #1417): recurrent flake — the POST/PATCH
+  // /api/v1/variables/ persist call is answered non-2xx while
+  // validate-provider succeeds, so the key authenticates but does not
+  // persist. Same signature on the 2026-07-13 and 08-11 dailies; that match
+  // needs settling, because both `ok()` assertions below emit it and the
+  // 07-13 daily fell in the quota-drained window (see #1424). Lifting the
+  // quarantine (remove test.fixme + restore @stable) is a deliverable of #1424.
+  test.fixme(
     "OpenAI API key is configured via Settings → Model Providers",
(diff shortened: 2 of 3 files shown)


Victor-w-Madeira:
## CI verdict — merging over a red impacted-specs lane

`Run impacted E2E specs` is red on run [31491130759](https://github.com/oriontech-me/langflow-e2e/actions/runs/31491130759). Every other check passes. The red is **transport-level and not attributable to this PR**:

```
1) openai-provider.spec.ts:209 › configured OpenAI selects a GPT model in the Agent and executes the flow
   Error: Agent credential never settled on the persisted flow (#751 guard, #1072).
     No read of the persisted flow succeeded, so the binding is UNKNOWN — nothing here says
     anything about the credential. Treat the read error above as the finding (a wedged or
     recycling backend, #1077), not the binding.
   TimeoutError: apiRequestContext.get: Timeout 20000ms exceeded.
   [backend-unreachable] the backend did not answer GET /api/v1/version within 5005ms
```

The guard names its own verdict: no read of the persisted flow succeeded, so this says nothing about the credential. The lane also logged three `500`s on `POST /api/v1/flows/` and a preflight error that `ANTHROPIC_API_KEY` was never imported as a global variable — i.e. `Collect models` did not complete cleanly and the backend was wedged or recycling for part of the run.

**The one hypothesis that would have made this PR the cause was checked and does not hold.** `openai-provider.spec.ts` is `test.describe.configure({ mode: "serial" })`, and the test quarantined here (`:138`) is the one that writes the OpenAI key through Settings → Model Providers — so it could plausibly be the precondition for `:209`. It is not: the same run's preflight named **only** `ANTHROPIC_API_KEY` as missing from Langflow's global variables, so the OpenAI key was imported by `Collect models` independently of test 1. And `test.fixme` skips rather than fails, so it cannot cascade under serial mode.

Run outcome: **3 passed, 3 skipped, 2 flaky, 1 failed**. The 3 skipped are exactly the three tests this PR quarantines — the change does what it says.

Merging per the impacted-specs convention (red from pre-existing/infra breakage is non-blocking when the PR's own work is not the cause). Leaving the three flakes live through tonight's daily would defeat the prevention this PR exists for.

PR #1435, opened 2026-08-12
test(security): assert tweaks cannot reach template fields as executable input (#1394)

### LINKED ISSUE #1394: Create tweaks-injection.spec.ts — tweaks cannot reach template fields as executable input

**Backlog pointer:** QA-CHECKLIST.md §17.4 (1 bullet) → `security/tweaks-injection.spec.ts`

**Scope**
- Tweak values passed to `POST /api/v1/run/{id}` cannot reach template fields as executable input.
- Tag `@regression`.

**Upstream:** langflow-ai/langflow#9319, #8672.

rafaelgiln:
Closes #1394.

Closes the `[ ]` gap in `QA-CHECKLIST.md` § 17.4 — the first spec in the new `security/` area. It is **not** a duplicate of `api/flows/api-run-with-tweaks.spec.ts`: that spec proves tweaks *do* override configuration, and its own "does not cover" section names code-field injection protection as the gap. This one proves the inverse boundary — the fields a tweak must **never** reach.

The backend refusal is **silent by design**: `apply_tweaks()` (`src/lfx/src/lfx/processing/process.py`) asks `is_protected_tweak_field()` (`src/lfx/src/lfx/utils/flow_validation.py`) and, on a protected field, logs a warning and `continue`s — the run still answers `200` with the flow author's value. There is no error to assert on, so every refusal below is paired with a benign tweak sent the same way. Without that control, a spec asserting only "the sentinel is absent" passes just as well when the tweaks mechanism is dead altogether.

Two protection routes exist and fail independently, so both are covered: **by field type/name** (`code`), and **by component + field name** — an executable field that serializes as a plain `str` on a `CODE_EXECUTION_COMPONENT_TYPES` node. The second is the exact bypass that made the first, name-only fix insufficient (the upstream comment in `apply_tweaks()` names it).

Upstream: `langflow-ai/langflow#9319` (the defect, closed as fixed 2025-09-15) and `#8672` (the interpolation request that defines the boundary's other side).

## 1. Covered tests

| 1 | `a code tweak cannot replace a component's implementation` | A `tweaks.<node>.code` carrying a valid `ChatInput` class that returns `PWNED-<unique>` is refused, keyed by **node id** and by **display name**: `200`, the echo is still the stored `"Hello"`, the sentinel is absent from the response, and the stored flow's `code` is untouched afterwards. Catches a regression that would let any caller holding run permission execute code of their choosing in-process. |
| 2 | `the refusal is field-scoped: an unprotected field on the same node still applies` | The same flow, same node, same two addressing modes, same request shape — only the targeted field differs (`input_value`): the output *does* become the tweak value. This is what makes Test 1 evidence rather than a tautology, and it catches the opposite regression (the guard widening until legitimate tweaks stop working). |
| 3 | `an executable field on a code-execution component is refused while the same request's benign tweak lands` | **One** run (`output_type: "debug"`) carrying three overrides on `Python Interpreter -> Chat Output`: `python_code`, `global_imports` (both plain `str`, invisible to the field-type guard) and `sender_name`. Asserts `sender_name` **did** change, the interpreter result is still the author's `AUTHOR-<unique>`, and the build log still reads `Successfully imported modules: ['math']` — `os` never entered the `exec` namespace. All four observations come from a single response, so no ordering or timing explanation is available. |

## 2. How the test was built

- **Pure API** — Playwright's `request` fixture, no browser, so there are no selectors to confirm. Auth is `getAuthToken` (Bearer) for the flow/catalog/key endpoints plus a temporary API key, because `POST /api/v1/run/{id}` authenticates with `x-api-key`.
- **Flow A** reuses `createRunnableChatFlowViaApi` (the `Chat Input -> Chat Output` passthrough, stored default `"Hello"`), so the injected class has an observable that differs from the author's by construction.
- **Flow B** is a new helper, `tests/helpers/flows/create-python-interpreter-flow-via-api.ts`, that builds `Python Interpreter -> Chat Output` from the **live** `GET /api/v1/all`. A committed fixture was rejected: the field names this spec pins (`python_code`, `global_imports`) are exactly what upstream may rename, and a frozen fixture would keep testing the old shape and stay green. A missing component fails naming it rather than reaching `POST /api/v1/flows/` as an unattributable 422.
- **The edge is hand-built** (Langflow's `œ`-escaped handle encoding) because a two-node flow with no edge is not a cheaper equivalent: an unreachable vertex never runs and is simply **absent** from `outputs` (measured on 1.12.0.dev23), which would make Test 3's refusal unobservable while the test stayed green. The helper's pure half — handle encoding, catalog lookup failure, no mutation of the caller's catalog — is covered by 6 unit tests in the `npm run test:units` lane.
- **Assertion rationale:** the refusal is silent, so absence of the sentinel is necessary but not sufficient. Each test therefore also shows a tweak that *did* land — in Test 3, inside the very same request.
- **Cleanup:** both flows are deleted id-scoped in `afterAll` with its own `request`, then the temporary API key, each step wrapped so a failure cannot skip the rest. Instance flow count after the final run: 2, both pre-existing and unrelated — zero orphans from this spec.

## 3. Dependencies

- **None external** — no LLM, no provider key, no network egress. The instance must run with `LANGFLOW_ALLOW_CUSTOM_COMPONENTS=true` (every CI lane and both start scripts set it); with it off, `run_python_repl()` refuses to execute and Test 3 fails on its *author* baseline rather than passing vacuously.
- `PythonREPLComponent` must be in the catalog (category `utilities`, a **core** family — survives the `lfx-bundles` M4 shim deletion).
- **Parallel-safe.** Flow names and node ids are unique per call; nothing is name-addressed across flows. The whole file runs in ~3 s.

## Validation (nightly 1.12.0.dev23, `--retries=0 --workers=1`)

- 3/3 clean burst runs — `expected=3 unexpected=0 flaky=0` each, zero `🚨 Backend Error`
- `npm run typecheck` ✅ · `npm run lint` ✅ (0 errors, no new warnings) · `npm run test:units` ✅ 608 pass / 0 fail · `npm run check:checklist-coverage` ✅
- **Force-fail, one behavioural mutation per test, all reverted** (final green run after revert ✅):
  - T1 — routed the sentinel through the unprotected `input_value` alongside the `code` tweak, i.e. simulated the tweak reaching the component → red
  - T2 — pointed the benign tweak at the protected `code` field, so the control no longer observes an applied tweak → red
  - T3 — the flow author's stored `python_code` emits a different string, which is what a honoured `python_code` tweak looks like from the response → red
- `@stable` intentionally absent on first delivery (added only after team validation); the checklist bullet ships as `[-]`.

Note on the resolved version: the instance runs `nightly:latest` as pulled locally (`1.12.0.dev23`). `1.12.0.dev24` was published ~8 h before this run and could not be pulled on this machine (Colima: `no space left on device`); the spec is API-only and touches no surface that moved in that window.

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -872,8 +872,9 @@
 
 #### 17.4 Tweaks Injection
 
-- [ ] Tweak values passed to `POST /api/v1/run/{id}` cannot reach template fields as
+- [-] Tweak values passed to `POST /api/v1/run/{id}` cannot reach template fields as
       executable input (`langflow-ai/langflow#9319`, `#8672`)
+      → security/tweaks-injection.spec.ts
 
 ---
 
--- a/docs/security/tweaks-injection.md
+++ b/docs/security/tweaks-injection.md
@@ -1,0 +1,121 @@
+# Tweaks Injection
+
+**Last validated:** Langflow 1.12.x
+
+---
+
+## What this test validates *(required)*
+
+Validates that a **tweak value sent to `POST /api/v1/run/{flow_id}` cannot reach a template field as executable input** — the security boundary restored upstream after `langflow-ai/langflow#9319` ("Potential security issue in tweaks", closed as fixed on 2025-09-15) and deliberately *not* widened by the request in `#8672` (tweaks interpolated into component parameters).
+
+`tweaks` is the documented way for an API caller to reconfigure a flow per request. Every Langflow component also carries its own Python source in the `code` template field, and several components carry executable code (or the sandbox boundary that constrains it) in *plain-text* fields — `python_code`, `function_code`, `global_imports`, `tool_code`, `filter_instruction`, `allow_dangerous_code`. Without a refusal, an authenticated run caller could rewrite what a flow *is* at run time instead of merely reconfiguring it: `#9319` demonstrated replacing `ChatInput`'s whole class body through `tweaks.<node>.code`, and the follow-up comment confirmed the same reaches the build endpoints.
+
+The backend refusal is **silent by design** — `apply_tweaks()` consults `is_protected_tweak_field()` and, on a protected field, logs a warning and `continue`s; the run still answers `200` with the flow author's value. There is therefore **no error to assert on**: the only trustworthy evidence is *behavioral* — the flow keeps producing the author's output while a benign tweak sent the same way, in the same shape, still takes effect. Every test below pairs the refusal with that control, because a spec that only asserts "the sentinel is absent" passes just as well if the tweaks mechanism is dead altogether.
+
+Two protection routes exist and fail independently, so both are covered:
+
+- **by field type / field name** — `field_type == "code"` or `field_name == "code"` (Test 1).
+- **by component + field name** — a code-execution component (`CODE_EXECUTION_COMPONENT_TYPES`) whose executable field serializes as a plain `str` and is therefore invisible to the type check (Test 3). This is the exact bypass that made the first, name-only fix insufficient — the upstream comment in `apply_tweaks()` names it.
+
+If these tests fail, an API caller holding only run permission on a flow can execute code of their choosing inside the Langflow process.
+
+---
+
+## Tags *(required)*
+
+`@api` `@regression`
+
+No **functional** tag applies: the tag table has no security area, and the closest siblings (`api/flows/api-run-with-tweaks.spec.ts`, `api/flows/api-run-flow.spec.ts`) also carry only cross-cutting tags. `@regression` is what issue #1394 asks for; `@api` marks the layer.
+
+`@stable` is intentionally absent on first delivery — it is added only after team validation (`CONTRIBUTING.md`). The checklist bullet ships as `[-]`.
+
+---
+
+## Step by step *(required)*
+
+The spec runs **3 independent tests** via Playwright's `request` fixture. No browser, no LLM, no provider key. Two flows are created in `beforeAll` and deleted in `afterAll`.
+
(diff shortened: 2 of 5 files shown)


rafaelgiln:
**Update — the three tests now ship `@stable`** (maintainer decision, `f767be4`). The PR description above still says `@stable` is absent on first delivery; this comment supersedes that paragraph, and the spec doc's **Tags** section has been rewritten to match.

Why the lane profile fits: the file is pure API — no browser, no LLM, no provider key — and all three tests together take ~3 s, so it costs `daily-stable.yml` almost nothing and cannot go red for a provider-outage reason. It is **not** `@destructive`: it creates and deletes only its own two flows and its own API key, so it belongs in the normal lane (`@stable` + `@destructive` would silently never run there, #1010).

Evidence for the promotion, selected the way the daily selects it (`--grep @stable`), nightly 1.12.0.dev23, `--retries=0 --workers=1`:

```
run 1 {"expected":3,"unexpected":0,"flaky":0,"skipped":0}
run 2 {"expected":3,"unexpected":0,"flaky":0,"skipped":0}
run 3 {"expected":3,"unexpected":0,"flaky":0,"skipped":0}
```

`npm run typecheck` ✅ · `npm run lint` ✅ (no warnings on the new files) · `npm run check:checklist-coverage` ✅. The force-fail proof recorded on `e303d83` still stands — the tag arrays touch no assertion.

The checklist bullet moves `[-]` → `[x]`. The generated `Phase 0 — Validated` block is deliberately **not** regenerated here; `update-coverage-summary.yml` does that on merge (#741).

Issue #1432, opened 2026-08-12
The HTTP error monitor discards the response body in silence when it cannot be read


Victor-w-Madeira:
## What happens

`tests/fixtures/fixtures.ts:300-306` records every monitored HTTP error and then tries to attach its body:

```ts
errors.push(entry);
try {
  entry.responseBody = await response.text();
  console.log(`   Response: ${entry.responseBody}`);
} catch (e) {
  entry.responseBody = "Could not read response";
}
```

When `response.text()` throws, the string `"Could not read response"` goes into the entry and **nothing is printed**. The log shows the `🚨 Backend Error` line and then moves on, indistinguishable from an error whose body was empty. Reading the log, there is no way to tell "the backend said nothing" from "we could not ask".

This is the rule #1012 exists for, applied to the fixture that is itself the suite's evidence trail: an unread body is unknown, not absent.

## Why it matters right now

#1424 is open on *"the global-variable persist call answers non-2xx while validate-provider succeeds"* and has no cause yet. The same defect has now fired on three consecutive dailies:

| Daily | Spec | Recorded |
|---|---|---|
| 2026-08-10 | `tests/tests-automations/regression/core-functionality/model-provider/azure-ai-foundry-provider-setup.spec.ts` | 1× `400 POST /api/v1/variables/` |
| 2026-08-11 | `tests/tests-automations/regression/core-functionality/model-provider/openai-provider.spec.ts` | 2× `400 POST /api/v1/variables/` |
| 2026-08-12 | `tests/tests-automations/regression/core-functionality/model-provider/azure-ai-foundry-provider-setup.spec.ts` | 1× `400 POST /api/v1/variables/` |

Not one of those four carries a body in the logs, while every `500` in the same runs does (`{"detail":"An internal error occurred while deleting flows."}`, `{"detail":"Could not update the flow."}`, and the `validate/prompt` validation messages). The status alone does not say what the backend rejected, which is most of why the issue is still descriptive after three occurrences.

## What is not yet known

Why the read fails on these specific responses. The catch is unconditional, so the reason is discarded along with the body — which is the defect. Candidates to distinguish once it is logged: a body already consumed elsewhere, a request the browser did not retain, a redirect or a cancelled request. The fix is to stop discarding the information, not to guess which of these it is.

## Proposed fix

- Log the failure where the body would have gone, naming the reason (`e.message`), so an unreadable body is visible in the same place a readable one is.
[...]

Issue #1431, opened 2026-08-12
model-provider-api-key waits on an accessible name that becomes a spinner — `Replace` is not a distinct button


Victor-w-Madeira:
Surfaced in the 2026-08-12 daily ([run 31581590030](https://github.com/oriontech-me/langflow-e2e/actions/runs/31581590030)), which was green — 493 passed, 0 failures, 6 flaky. First occurrence for this test, filed anyway because the cause did not need a second one: it is readable in the frontend source.

**Upstream:** n/a — this is a spec defect, not a product one.

## Symptom

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/llm-agents/model-provider-api-key.spec.ts:94` ("a configured provider exposes the key edit surface (Replace, no raw input)") | `getByRole('button', { name: 'Replace', exact: true })` | `Error: expect(locator).toBeDisabled() failed` … `Error: element(s) not found` |

Not a timeout on state — the locator matched **nothing** for the full 10 s. The two assertions ahead of it in the same step passed on the same attempt: the provider row read `N models` (`:88`) and `provider-variable-input-OPENAI_API_KEY` was visible (`:92`). So the panel was in its configured state; only the button was unfindable.

## Why this is one cause

Single test, single wait point. No grouping argument is being made — this is filed as its own cause precisely because it is *not* the swallowed-add cluster (#1365) that produced two of the same run's other flakes.

## What the code says

Read on `origin/release-1.12.0` — the line the nightly is cut from.

There is no "Replace" button. There is **one** button, `provider-save-button`, whose label is chosen at render time in `src/frontend/src/modals/modelProviderModal/components/ProviderConfigurationForm.tsx`:

```tsx
<Button
  onClick={onSave}
  data-testid="provider-save-button"
  loading={isLoading || isFetchingModels}
  disabled={!canSave || isLoading || isFetchingModels}
>
  {validationFailed
    ? t("modelProviders.retrySaveButton")
    : isAlreadyConfigured
      ? t("modelProviders.replaceButton")
      : t("modelProviders.saveButton")}
</Button>
```

with `const isLoading = isSaving || validationState === "validating" || isFetchingModels;`.

[...]

PR #1441, opened 2026-08-12
fix(model-provider): settle the provider panel before saving credentials (#1424)

### LINKED ISSUE #636: [Daily #629] Flaky: provider API-key save in Settings times out on waitForResponse (google + openai)
Surfaced in the 2026-07-10 daily triage (umbrella #629). These are **flaky** (recovered on retry), so `@stable` is retained per the daily triage policy — this issue tracks the investigation.

## Failing tests — shared symptom: configuring a provider API key in Settings never observes the network response
- `google-provider.spec.ts:90` — "Google API key is configured via Settings → Model Providers"
  - **first flaky occurrence** (2026-07-10) — `waitForResponse` timeout **60s**
- `openai-provider.spec.ts:91` — "OpenAI API key is configured via Settings → Model Providers"
  - flaky 2026-07-09, 2026-07-10 — `waitForResponse` timeout **30s**

## Run context
- Date: 2026-07-10 · Langflow: `langflowai/langflow-nightly:latest`
- Run: https://github.com/oriontech-me/langflow-e2e/actions/runs/29087610808

## Observed evidence (not diagnosis)
`TimeoutError: page.waitForResponse: Timeout {30s,60s} exceeded while waiting for event "response"` while saving the provider key on the Settings → Model Providers page. The expected save request/response was never observed within budget.

### Axis 1 — is it a Langflow product bug?
Rule this out FIRST:
- Does the Settings save actually hit the backend, and does the backend respond slowly / not at all / with an error under load? A save endpoint that hangs or 5xxs intermittently is a product defect.
- Check the network log for the failing runs: was the request even fired? Did it return a non-matching status/URL (so our `waitForResponse` predicate missed it)? Did it 5xx?

### Axis 2 — is the TEST exercising Langflow correctly?
- Is the `waitForResponse` predicate (URL + method + status) still matching the endpoint the current Langflow build uses to persist provider keys? An upstream route/verb change would make the predicate wait forever on a request that never matches — a **test** defect masquerading as a hang.
- Is waiting on a specific network response the right signal, or should the test assert on the resulting UI/persisted state instead?
- Is the save genuinely being triggered (right field, blur/submit), or is the interaction racing an async re-render?

**Why this rigor matters:** if the save endpoint is intermittently failing, "fixing" the test by extending the timeout would (a) hide a real backend reliability defect and (b) leave engineering unaware. If instead the predicate is stale after an upstream change, the test is silently no longer validating provider configuration — it would pass for the wrong reason once the timeout is relaxed.

## Done when
Network-level confirmation of what happens on save for the failing runs; then either a product bug is filed OR the predicate/assertion is corrected with justification. Confirm the 2026-07-09 openai recurrence shares this signature.

### LINKED ISSUE #1424: [Daily #1417] openai-provider — the global-variable persist call answers non-2xx while validate-provider succeeds (recurrent 2×)

rafaelgiln:
**Fixes #1424.** Closes #1424.

## Problem

Two `@stable` tests were quarantined for the same signature — a non-2xx persist on
`/api/v1/variables/` while `validate-provider` succeeded — across four dailies:

| Daily | Spec | Recorded |
|---|---|---|
| 2026-07-13 (run 29244611612) | `openai-provider.spec.ts` | `PATCH … 400 {"detail":"Invalid API key for OpenAI"}` |
| 2026-08-10 (run 31373880200) | `azure-ai-foundry-provider-setup.spec.ts` | `POST … 400 Could not validate Azure AI Foundry credentials: … Read timed out. (read timeout=10.0)` |
| 2026-08-11 (run 31475108157) | `openai-provider.spec.ts` | `POST … 400 {"detail":"Variable name already exists"}` **×2** |
| 2026-08-12 (run 31581590030) | `azure-ai-foundry-provider-setup.spec.ts` | same Azure read timeout |

Root cause, read off the backend and then reproduced: `POST`/`PATCH
/api/v1/variables/` is **not** a dumb persist. When the name is the provider's
primary credential variable (`get_model_provider_variable_mapping()`),
`langflow/api/v1/variable.py` calls `validate_model_provider_key`, which makes a
**real** call to the provider (`llm.invoke("test")` for OpenAI,
`request_azure_ai_foundry_model_entries` for Foundry) and turns any `ValueError`
into a **400 whose body carries the reason**. So one assert covered three causes
that mean opposite things:

- **`Variable name already exists` — ours.** The panel derives create-vs-update
  from the credential variables, so the single submit control reads `Save` until
  `GET /api/v1/variables/` resolves while the key input is already rendered
  (#1431). A click inside that window makes the frontend CREATE a name that is
  already stored, and the backend refuses it in 0.07 s with no live call.
  Reproduced on demand on 1.12.0.dev24 by delaying only that request: label
  `Save`, `POST → 400`, toast *"Error Saving Configuration — Variable name
  already exists"*, control relabelled `Retry Save`. With the panel settled the
  same save is a `PATCH → 200`.
- **`Invalid API key for OpenAI` — not ours.** The 2026-07-13 occurrence, settled
  from that run's `playwright-json-daily` artifact: attempts 1–2 were the already
  fixed #636 PATCH-only waiter timeout, attempt 3 failed on **`persistResp`**
  (line 187). So the recurrence claim in the issue **holds**, but by a different
  cause than 08-11's.
- **Azure's read timeout — not ours.** The KEY write is validated against the live
  resource with a **10 s** budget; the ENDPOINT write is not validated at all, so
  a refused key leaves the pair **half-configured with no rollback** — which is
  what the pair-poll then timed out on 30 s later, with the cause nowhere in the
  assertion. The poll was never the problem and is kept.

## Fix

- **`tests/helpers/provider-setup/provider-panel-save.ts` (new).**
  `awaitProviderPanelSettled` gates on `provider-save-button` being visible, not
  `aria-busy`, and carrying the label the **backend state** implies — a two-sided
  gate, since the caller passes what it read from `GET /api/v1/variables/`.
  `collectCredentialWrites` records **every** variables write with its body: a
  two-variable provider issues two and the old `waitForResponse` resolved on the
  first, which on both Azure dailies was the endpoint's `201`.
  `readResponseBodySafely` keeps "empty" distinguishable from "could not read"
  (#1432's defect, avoided at the source).
- **`tests/helpers/provider-setup/variable-write-refusal.ts` (new).** Classifies a
  refusal from its body — `duplicate` / `invalid-credential` / `transport` /
  `unknown` — and renders method + URL + status + reason for assertion messages.
  Order is load-bearing: Azure's timeout and its 401 share the same
  `Could not validate … credentials:` prefix, so transport is tested first.
  12 unit cases in `npm run test:units`, every body verbatim from a measurement.
- **Both specs.** The verb the state implies is now **asserted** (`PATCH` on a
  configured instance, `POST` otherwise), so a create-over-existing fails naming
  both verbs. `validate-provider` is read from its **body** (`valid: true`) —
  it answers 200 for a credential it rejected. The credential is asserted
  **readable back**, since a 2xx that stored nothing used to pass. A refusal that
  is ours stays a hard failure; the two environmental ones are retried once
  through the panel's own `Retry Save` and then `test.skip` **quoting the backend
  detail** (#980's trade, #1012's rule).
- **Quarantines lifted** (#1425 `openai-provider.spec.ts:138`, #1433
  `azure-ai-foundry-provider-setup.spec.ts:474`): `test.fixme` removed and
  `@stable` restored on both.

**Rejected alternatives.** Failing on every 400 with a better message — it hands
back a red daily on every drained-key or slow-Azure window, i.e. the streak that
opened this issue. Pre-gating openai Test 1 on `providers.json` health — measured
wrong: a **credit-less** key still saves (the quota error carries no
`401`/`authentication`/`api key` token, so the validator's lenient branch stores
it), so the gate would skip a test that passes, and the spec doc already argued
against it. Loosening the Azure pair-poll — the write was refused, not delayed.

## Scope note

Unchanged: the Azure pair-poll and every assert after the save (deployment
add/enable/persist, the reload check, the credential teardown), openai Test 2 in
full, and the other five Azure tests. No timeouts were raised. The `#636`
both-verbs waiter behaviour is preserved (it moved into
`waitForCredentialPersist`).

## Product observations (recorded, no ticket filed — for the team to decide)

1. The panel enables its submit control before it knows whether the credential
   exists, so a fast user gets a spurious *"Variable name already exists"*.
   Measured window on the 2026-08-12 daily: label `Save` for ~9 s (#1431).
2. A two-variable provider's writes are **not atomic**: a refused key leaves the
   endpoint stored, and the panel reports a configured-looking half state.
3. Validation is **order-dependent** — writing the KEY before its companion
   ENDPOINT stores the credential with **no validation at all** (201 in 0.86 s on
   dev24), so "stored" does not imply "validated".

## Validation (nightly 1.12.0.dev24, `--retries=0 --workers=1`)

- `npm run typecheck` ✅ (0 errors) · `npm run lint` ✅ (0 errors, warnings
  pre-existing) · `npm run test:units` ✅ **635 pass / 0 fail** ·
  `npm run check:checklist-coverage` ✅
- `openai-provider.spec.ts --grep "OpenAI API key is configured"` — **3/3 clean**
  (11.7 s / 9.9 s / 8.7 s), plus one more green after the mutation harness was
  reverted
- `azure-ai-foundry-provider-setup.spec.ts` — **4 passed, 2 skipped** (13.9 s).
  Tests 5–6 skip with the concrete reason: the Foundry probe answers `401`.
  **The `manual.yml` dispatch on this branch did NOT validate test 5** — it
  concluded `success` as a green ALL-SKIP (`Running 1 test` → `1 skipped`,
  [run 31649806352](https://github.com/oriontech-me/langflow-e2e/actions/runs/31649806352)),
  because the CI runner gets the same `401` from that endpoint. The secrets are
  wired (`manual.yml:236-238`); the credential itself is dead. Test 5's new code
  paths are therefore proven by the unit lane plus the locally-forced refusals
  below, and **not** by an end-to-end green — see the PR comment, and #1424 stays
  open on it.
- **Force-fail, executed:**
  - settle gate removed + `GET /variables/` delayed 8 s → FAILS:
    `OPENAI_API_KEY is already stored, so saving must PATCH it … Expected "PATCH" Received "POST"`
  - causal control — settle gate present, same 8 s delay → **passes** (20.0 s vs
    9.9 s: it waited out the window)
  - readback mutated to a non-existent name → FAILS
  - `validate-provider` body assert inverted → FAILS (`Expected false Received true`)
  - persist mocked `400 Variable name already exists` → **FAILS**
    (`reason that IS ours (duplicate)`) — the skip cannot swallow our own defect
  - persist mocked `400 … Read timed out` → **1 skipped** after two refusals
    (retry then skip), reason printed
  - azure sibling (`SEED_FLOOR 3 → 99`) → test 1 FAILS, so the edit did not defuse
    the file's other tests
- `--trace=on` ❌ not run — the known local hang (`CONTRIBUTING.md`); covered by
  the `--retries=0` bursts + the executed force-fails above
- Zero `🚨 Backend Error` on the green runs. The mocked force-fails intentionally
  log two (they are the mutation).

## Flow cleanup

Test 1 creates no flows. Instance checked after the runs: 26 flows = the seeded
starter set, and one leaked duplicate `Simple Agent` (from openai Test 2 dying on
`You have no credits remaining`, a pre-existing env condition on this host) was
purged by id. `AZURE_AI_FOUNDRY_*` variables absent afterwards — the Azure
teardown restored the account state.

### DIFF
--- a/docs/core-functionality/model-provider/azure-ai-foundry-provider-setup.md
+++ b/docs/core-functionality/model-provider/azure-ai-foundry-provider-setup.md
@@ -70,6 +70,13 @@ validating.
 phase (3 clean `--retries=0 --workers=1` runs) plus the force-fail proof per
 test, per `CONTRIBUTING.md`.
 
+**Test 5** was quarantined at triage of the 2026-08-12 daily (PR #1433 —
+`@stable` removed **and** `test.fixme` added) and **restored in #1424**, where
+the `400` on the credential write was traced to the write-time live validation
+of the Foundry endpoint (see step 3). The other five tests kept `@stable`
+throughout: they cover the unconfigured panel and a deployment path that does
+not depend on that write.
+
 **Why `@stable` on the credential-gated tests too.** Issue #1194 assumed the
 whole spec needed Azure credentials and therefore shipped untagged. The surface
 triage disproved the premise for tests 1–4, and the credentials turned out to
@@ -256,13 +263,40 @@ Live-scouted testids (1.12.0.dev14, re-checked on dev15 where the run is green, 
 1. Probe `GET <AZURE_AI_FOUNDRY_ENDPOINT>/models` with the `api-key` header
    straight from the test host; skip with the concrete reason when the key or
    endpoint is missing, the probe is non-200, or the payload has no `data` list.
-2. Open the Foundry panel, fill both variables, arm waiters on
-   `POST /api/v1/models/validate-provider` **and** on the variables write
-   (`POST /api/v1/variables/` or `PATCH /api/v1/variables/{id}` — the frontend
-   branches on existence, #636), click Save.
+2. Open the Foundry panel, **settle it** (`awaitProviderPanelSettled` with
+   `expectConfigured: false` — the ownership gate in step 0 already established
+   that nothing is stored, so `provider-save-button` must read `Save`, not be
+   `aria-busy`; clicking before `GET /api/v1/variables/` resolves is what makes
+   the openai sibling create over an existing name, #1431/#1424), fill both
+   variables, arm waiters on `POST /api/v1/models/validate-provider` **and** on
+   the variables write (`POST /api/v1/variables/` or `PATCH
+   /api/v1/variables/{id}` — the frontend branches on existence, #636), click
+   `provider-save-button`.
 3. **Assert (configure):** validate-provider body `valid === true` and the
    variables write is 2xx — armed before the click, so a pre-existing configured
    state cannot pass the test.
+
+   > **The `400` this step used to die on, and why it is now classified (#1424).**
+   > The KEY write is validated **live**: `create_variable` calls
+   > `validate_model_provider_key`, which for Foundry issues
+   > `request_azure_ai_foundry_model_entries(endpoint, key)` with a **10 s read
+   > timeout** and turns **any** exception into `400 {"detail": "Could not
+   > validate Azure AI Foundry credentials: …"}`. Both the 2026-08-10 and
+   > 2026-08-12 dailies carry exactly that, with `HTTPSConnectionPool(…): Read
+   > timed out. (read timeout=10.0)` — the resource answered the test host's probe
+   > and then took longer than 10 s to answer Langflow. The write is refused, the
+   > key is **not** stored, and because the ENDPOINT write is not validated at all
+   > (`get_model_provider_variable_mapping()` names only
+   > `AZURE_AI_FOUNDRY_API_KEY` as the provider's primary variable — measured on
+   > 1.12.0.dev24, where writing the KEY *first* is accepted in 0.86 s with no
+   > validation because the endpoint is not yet stored), the pair is left
+   > **half-configured with no rollback** — which is what the poll in step 4 then
+   > times out on. So the poll was never the problem. A refusal whose body says
+   > the provider could not be reached, or that the credential did not
+   > authenticate, is retried **once** through the panel's own `Retry Save` and
+   > then ends the test as a `test.skip` quoting the backend's exact `detail`
+   > (#980's trade, #1012's rule). Any other refusal — including `Variable name
+   > already exists` — stays a hard failure.
 4. **Assert (provider state):** poll `GET /api/v1/variables/` until **both**
    `AZURE_AI_FOUNDRY_API_KEY` and `AZURE_AI_FOUNDRY_ENDPOINT` are stored, then
    assert the panel renders `llm-toggle-*` controls.
--- a/docs/core-functionality/model-provider/openai-provider.md
+++ b/docs/core-functionality/model-provider/openai-provider.md
@@ -38,6 +38,11 @@ PR #775 — the key had no quota, not the test — and **restored in #992** once
 the quota was confirmed back (live HTTP 200 completion on `gpt-4o-mini`) and
 the test ran clean at `--retries=0`.
 
+**Test 1** was quarantined at triage of daily #1417 (PR #1425 — `@stable`
+removed **and** `test.fixme` added) for the recurrent 400 on the persist call,
+and **restored in #1424** once both causes behind that one assert were measured
+and the create-over-existing race was fixed (see Notes).
+
 It was auto-removed a **second** time on the 2026-08-06 daily (commit `cb3082d`,
 run 31093877484) for the same underlying reason — the account had no credits —
 and **restored in #1333**, again on a live HTTP 200 completion plus 5 clean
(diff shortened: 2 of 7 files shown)


rafaelgiln:
## CI results — and the honest half: the Azure dispatch is a green ALL-SKIP

### PR lane ([run 31649798864](https://github.com/oriontech-me/langflow-e2e/actions/runs/31649798864)) — all checks green, and the openai half is genuinely proven

The import graph selected **both** touched specs (8 tests): **5 passed, 3 skipped
(25.4 s)**. Which is which matters, so it is spelled out rather than left to the
totals:

| Test | CI outcome |
|---|---|
| `openai-provider.spec.ts` Test 1 — *OpenAI API key is configured …* | **passed** |
| `openai-provider.spec.ts` Test 2 — *… executes the flow* | skipped — `providerSkipGate`, the account has no credits |
| `azure-…-setup.spec.ts` tests 1–4 | **passed** (4) |
| `azure-…-setup.spec.ts` tests 5–6 | skipped — see below |

Test 1 passing **on this lane, with `collect-models` reporting `openai inactive —
You have no credits remaining`**, is the measurement the design rests on: a
credit-less key still authenticates and still saves, so pre-gating this test on
`providers.json` health would have skipped a test that passes. That was the
rejected alternative in the PR body; the lane just confirmed it.

### The Azure dispatch did NOT validate test 5 — the Foundry credential is dead

[Run 31649806352](https://github.com/oriontech-me/langflow-e2e/actions/runs/31649806352)
concluded **success**, and that success is worth nothing: `Running 1 test using 1
worker` → `1 skipped`. The annotation names the reason:

```
Azure AI Foundry not usable: GET https://langflow-test-123.openai.azure.com/openai/v1/models answered 401
```

The three secrets ARE wired into this job (`manual.yml:236-238`), so this is not a
wiring gap — the resource itself now refuses the key, from the GitHub runner and
from a local host alike (measured here minutes apart: `401 … invalid subscription
key or wrong API endpoint`).

**That reframes the Azure half of #1424.** The 2026-08-10 and 2026-08-12 dailies
recorded a *read timeout* against this endpoint at ~09:24Z; by 23:15Z the same
endpoint answers *401*. So the credential was degrading and is now dead, which
means:

- the `400 … Read timed out` those dailies hit is consistent with a resource on its
  way out, and no test change could have made it pass;
- **test 5 cannot be validated anywhere until the credential is refreshed** — it
  skips with its reason on every lane, including the daily;
- `@stable` restored on test 5 matches this file's existing contract (test 6 has
  always been `@stable` and skips the same way when the probe fails, per the spec
  doc's *"a skip is never a green: it is reported as a skip"*), but it is a
  **documented skip, not a proof**.

### What this means for merging

The openai half is fixed and proven on CI. The Azure half ships the same
classification fix, is proven only at the unit level plus locally-forced refusals,
and **#1424 stays open** until either the Foundry credential is refreshed and a
dispatch runs test 5 green, or the team decides that coverage does not come back.
The credential refresh is an environment task, not a test change — flagged on
#1424 rather than smuggled into this PR.

Issue #1460, opened 2026-08-14
[Daily #1458] modelInputComponent + memory-base-registration read a model option's own text and break on the sr-only "N of M" counter


Victor-w-Madeira:
Spun out of daily-failure triage #1458 (run [31786538844](https://github.com/oriontech-me/langflow-e2e/actions/runs/31786538844), 2026-08-14). Split from the option-matching cluster: this half is the two **spec-local** matchers, which touch no shared file.

**Upstream:** _not filed_

## Symptom

Two `@stable` tests hard-failed (3/3 attempts each) reading a model option's own text — one compares it against the
trigger, the other matches an option by exact accessible name.

Read off the trace of the Google hard failure (`1-trace.trace`, snapshot `after@call@98`, 90 option nodes captured while the
picker was open). Every model option now carries a second text node with its position in the list:

```
["DIV", {"data-testid": "Google Generative AI-gemini-flash-latest-option", "role": "option", ...},
  ["DIV", {"class": "truncate text-[13px]"}, "gemini-flash-latest"],
  ["SPAN", {"class": "sr-only"}, "22 of 90"],
  ...]
```

The span is `sr-only` — invisible to a user, but part of `textContent` (so `hasText`) and part of the accessible name (so
`getByRole("option", { name, exact: true })`).

`modelInputComponent` reads `firstOption.innerText()` and asserts the trigger equals it: expected
`claude-opus-5\n1 of 90`, received `claude-opus-5`. The trigger is the half that is **correct** — the option is what gained
the extra line, and the test's own comment states the premise that expired ("its inner text is the bare model name").
`memory-base-registration` matches the embedding option by exact accessible name; its trace shows that option as
`['gemini-embedding-2', '1 of 1']`, so the exact name no longer matches.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/llm-agents/modelInputComponent.spec.ts:140` ("the trigger shows the model the user selects") | `getByTestId("model_model") toHaveText(label read from [data-testid$="-option"].first())` | `Error: expect(locator).toHaveText(expected) failed` |
| `tests/tests-automations/regression/core-functionality/memory/memory-base-registration.spec.ts:232` ("registering a memory base from the Create Memory modal persists it") | `getByRole("option", { name: embedding.modelId, exact: true })` | `TimeoutError: locator.click: Timeout 20000ms exceeded.` |

## Why these failures are one cause

[...]

Issue #606, opened 2026-07-09
initialGPTsetup calls setupOpenAI without an explicit model — latent #596-class selection drop


rafaelgiln:
## Problem

`tests/helpers/other/initialGPTsetup.ts` calls `setupOpenAI(page)` with **no `modelTestId`**, so Step 7 of `setup-openai.ts` "selects the first available" model. This is the exact latent sibling of the model-selection drop race fixed for Google in #596 (PR #603) and for the Agent node in #491 (PR #578):

- the node's default resolves through the missing `__default_language_model__` variable and falls back to the first configured provider's model;
- a `custom_component/update` racing the selection can silently revert the node to that fallback;
- "first available" is also catalog-order dependent — a catalog reorder silently changes which model every `initialGPTsetup` consumer runs.

Every spec using `initialGPTsetup` (the OpenAI path of `language-model-regression.spec.ts` and others) is exposed.

## Fix

Apply the pattern proven in #596:

1. add a `resolveOpenAIModel()` helper next to `resolveGeminiModel()` (`tests/helpers/provider-setup/resolve-gemini-model.ts`) — deterministic preference list over `data/models.json`;
2. thread an optional `model` through `initialGPTsetup` → `setupOpenAI`, defaulting to the resolved model;
3. where a consumer executes the flow, add the bounded pre-run widget gate (re-apply if the `model_model` widget shows a non-OpenAI value, ×3, then hard assert).

## Validation constraint

OPENAI_API_KEY is currently **inactive locally** (org lacks the required model tier), so these paths skip locally — local validation is limited to typecheck/lint/force-fail on the gate logic; the behavioral proof comes from CI (daily-stable has the secret).

## Origin

Follow-up flagged on PR #603 and in the spec doc `docs/core-functionality/llm-agents/language-model-regression.md` (latent sibling note).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1463, opened 2026-08-14
fix(provider-setup): match a model option by identity, and prove the absence before skipping (#1459) (#1461)

### LINKED ISSUE #1459: [Daily #1458] provider-setup helpers match no model option — the picker's sr-only "N of M" counter defeats every anchored matcher (2 hard failures, ~30 silent skips)
Spun out of daily-failure triage #1458 (run [31786538844](https://github.com/oriontech-me/langflow-e2e/actions/runs/31786538844), 2026-08-14). Split from the option-matching cluster: this half is the **shared** `provider-setup` helpers (14 files reach the same pattern), kept separate from the two spec-local matchers so the shared-file change lands in its own PR.

**Upstream:** _not filed_

## Symptom

Two `@stable` tests hard-failed (3/3 attempts each) and ~30 more skipped, all of them through the shared
`provider-setup` helpers, whose model-option matchers anchor a regex on the option's **whole** text.

Read off the trace of the Google hard failure (`1-trace.trace`, snapshot `after@call@98`, 90 option nodes captured while the
picker was open). Every model option now carries a second text node with its position in the list:

```
["DIV", {"data-testid": "Google Generative AI-gemini-flash-latest-option", "role": "option", ...},
  ["DIV", {"class": "truncate text-[13px]"}, "gemini-flash-latest"],
  ["SPAN", {"class": "sr-only"}, "22 of 90"],
  ...]
```

The span is `sr-only` — invisible to a user, but part of `textContent` (so `hasText`) and part of the accessible name (so
`getByRole("option", { name, exact: true })`).

The anchored matchers therefore match nothing: `^gemini-flash-latest$` cannot match `gemini-flash-latest\n22 of 90`.
The two failures surface at different points of the same primitive — one throws the helper's own
`MODEL_NOT_AVAILABLE`, the other picks the polluted label up as the model name and then times out clicking it.

The same picker produced **30 of the run's 35 skips**, all of them the suite's own `MODEL_NOT_AVAILABLE` guard reporting a
model it could not match as a model that does not exist:

| Skip reason (verbatim) | Tests | Where the model actually was |
|---|---|---|
| `MODEL_NOT_AVAILABLE: "claude-haiku-4-5" not found in dropdown — model may not be supported.` | 27 | option **9 of 90** |
| `MODEL_NOT_AVAILABLE: "gemini-flash-latest" not found in dropdown — model may not be supported.` | 2 | option **22 of 90** |
| `MODEL_NOT_AVAILABLE: "gpt-4o-mini" not found in dropdown — model may not be supported.` | 1 | not checked in the trace |

For context, `reports/daily-history.jsonl` records 4–15 skips on every other daily since 2026-07-30; this run recorded 35.

The specs that skipped (no line numbers — a skip carries none):

- `tests/tests-automations/regression/core-functionality/llm-agents/agent-context-id-continuity.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-input-sources.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-max-iterations.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-max-tokens.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-multi-tool-selection.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-empty-refusal-response.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-multimodal-image-input.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-system-prompt.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-component-regression.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-tool-error-handling.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-tool-name-validation.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-context-id-isolation.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-current-date-tool.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-model-connection-isolation.spec.ts`
- `tests/tests-automations/regression/core-functionality/llm-agents/agent-structured-output.spec.ts`
- `tests/tests-automations/regression/core-functionality/model-provider/anthropic-provider.spec.ts`
- `tests/tests-automations/regression/core-functionality/model-provider/google-provider.spec.ts`
- `tests/tests-automations/regression/core-functionality/model-provider/openai-provider.spec.ts`
- `tests/tests-automations/regression/mcp/client/mcp-client-agent-gemini-tool-regression.spec.ts`

|---|---|---|
| `tests/tests-automations/regression/core-functionality/llm-agents/language-model-regression.spec.ts:135` ("language model must respond with Google provider") | `tests/helpers/provider-setup/setup-google.ts:121 — [data-testid$="-option"] hasText /^gemini-flash-latest$/` | `Error: MODEL_NOT_AVAILABLE: "gemini-flash-latest" not found in dropdown — model may not be supported.` |
| `tests/tests-automations/regression/core-functionality/llm-agents/memory-history-regression.spec.ts:261` ("message history context retention suite") | `tests/helpers/provider-setup/setup-language-model-openai.ts:82 — [data-testid$="-option"] anchored on a label read from the DOM` | `TimeoutError: locator.click: Timeout 20000ms exceeded.` |

The shared dimension is **the option's text**, not the specs and not the signature — the two signatures differ and are
grouped anyway, deliberately.

- `setup-google.ts:121`, `setup-openai.ts:107` and `setup-anthropic.ts` all resolve a pinned model with
  `page.locator('[data-testid$="-option"]', { hasText: new RegExp(`^${modelTestId}$`) })`, and all three throw the same
  `MODEL_NOT_AVAILABLE` when that returns nothing. That is the 30-skip wave plus the Google hard failure.
- `setup-language-model-openai.ts:40-84` reads the labels with `options.allInnerTexts()` and then clicks an option filtered
  by `^label$`. Its last-resort branch does not reject a multi-line label (the `gpt-*` branch above it does, and its comment
  says why), so `chosen` came back as `claude-opus-5\n1 of 69` and the anchored click could never resolve — a 20 s timeout
  instead of a thrown `MODEL_NOT_AVAILABLE`. Same cause, one layer later.

Both failures and all 30 skips resolve a model through one of those four helpers on a run where the trace proves the
requested models were present in the list. Nothing else is common to them: they span three providers, four spec
directories and two shards.

The two spec-local matchers that broke on the same counter are filed separately (they touch no shared file); the guard that
let 30 tests skip on a false premise is filed separately again, because its fix is a different one.

Descriptive only. What the trace establishes, and nothing beyond it:

- The counter is **per option** and carries the option's index (`1 of 90`, `9 of 90`, `22 of 90`, …), so it is a position
  announcement, not a badge on one entry. It sits in a `sr-only` span inside the option's inner wrapper.
- The requested models **were there**: `gemini-flash-latest` at 22/90, `claude-haiku-4-5` at 9/90. The catalog offered 90
  models across three providers.
- The **trigger** (`getByTestId("model_model")`) is clean — it rendered `claude-opus-5`, no counter. Only options changed.
- `if-else-component-regression.spec.ts:59`, `data-operations-component.spec.ts:173` and
  `data-operations-legacy-link.spec.ts:292` use the same `getByRole("option", { exact: true })` shape against non-model
  dropdowns and **all passed** on this run, so the counter appears scoped to the model list rather than to every cmdk
  select. Worth re-checking rather than assumed.

An `sr-only` position announcement is the shape of a deliberate accessibility addition, and if that is what it is then the
matchers are what must change — but that is the investigation's call, not this issue's. Nothing here rules out a partial or
regressed render either.

## Investigation directive

**Product first.** (a) On the current nightly, open a Language Model node's model picker and confirm whether the `sr-only`
`N of M` span is present and intended — find the upstream change that added it (frontend `ModelList` /
`getModelOptionTestId` are the entry points the spec comments already name) and record whether it is an accessibility
addition or a render artefact. (b) Establish the blast radius from the DOM, not from the diff: check whether the counter
also reaches the trigger, the embedding picker and other cmdk-based selects — 14 files in this repo reach the anchored
`[data-testid$="-option"]` pattern, 5 of them shared helpers (`setup-openai.ts`, `setup-anthropic.ts`, `setup-google.ts`,
`setup-ollama.ts`, `setup-language-model-openai.ts`). (c) Only once (a) says the counter is intended, decide how the
helpers should identify an option — the `data-testid` and `data-value` attributes both carry `{Provider}::{model}` exactly,
so a text match may not be the right key at all. (d) Rule out that the 30 skips hid a second, unrelated failure: re-run the
skipped specs once the matcher resolves and confirm they pass on their own merits.

## Deliverables (Done when)

- [ ] The ~30 specs that skipped with `MODEL_NOT_AVAILABLE` on this run resolve their pinned model again and run — verified by a `--retries=0` pass over the `@agents` / `@model-provider` set, not by the skip count alone.
- [ ] The shared-helper change lands in its own PR (5 helper files, 14 reachable specs), separate from the spec-local matchers tracked in the sibling issue.
- [ ] Cross-linked with the spec-local matcher issue and with the `MODEL_NOT_AVAILABLE` guard issue spun out of the same triage.

### LINKED ISSUE #1461: [Daily #1458] MODEL_NOT_AVAILABLE asserts a cause it never checked — a broken matcher skipped ~30 @stable tests in silence
Spun out of daily-failure triage #1458 (run [31786538844](https://github.com/oriontech-me/langflow-e2e/actions/runs/31786538844), 2026-08-14). Not a failing-test cluster: this is the reporting gap the cluster exposed. Filed separately because its fix is not the matcher's.

**Upstream:** _not filed_

## Symptom

`MODEL_NOT_AVAILABLE` **assumes** absence from a matcher returning nothing, and on this run that assumption was false for
every one of the ~30 tests it skipped.

The three provider-setup helpers resolve a pinned model with an anchored text match and, when it resolves nothing, skip
(or throw) with `MODEL_NOT_AVAILABLE: "<id>" not found in dropdown — model may not be supported.` — a message that names
a cause it never checked. The trace of the one call site that throws instead of skipping
(`setup-google.ts:121`, reached from the spec below) shows the model **present**, at option 22 of 90.

The same picker produced **30 of the run's 35 skips**, all of them the suite's own `MODEL_NOT_AVAILABLE` guard reporting a
model it could not match as a model that does not exist:

| Skip reason (verbatim) | Tests | Where the model actually was |
|---|---|---|
| `MODEL_NOT_AVAILABLE: "claude-haiku-4-5" not found in dropdown — model may not be supported.` | 27 | option **9 of 90** |
| `MODEL_NOT_AVAILABLE: "gemini-flash-latest" not found in dropdown — model may not be supported.` | 2 | option **22 of 90** |
| `MODEL_NOT_AVAILABLE: "gpt-4o-mini" not found in dropdown — model may not be supported.` | 1 | not checked in the trace |

For context, `reports/daily-history.jsonl` records 4–15 skips on every other daily since 2026-07-30; this run recorded 35.

So 30 `@stable` tests left the run without a verdict, under a message asserting the models were unsupported, on a day
whose report otherwise reads as 4 failures. The skip count is the only trace it left, and it is the field nobody watches.

|---|---|---|
| `tests/tests-automations/regression/core-functionality/llm-agents/language-model-regression.spec.ts:135` ("language model must respond with Google provider") | `tests/helpers/provider-setup/setup-google.ts:121 — [data-testid$="-option"] hasText /^gemini-flash-latest$/` | `Error: MODEL_NOT_AVAILABLE: "gemini-flash-latest" not found in dropdown — model may not be supported.` |

One mechanism, evidenced at the single call site that surfaces it as a failure rather than a skip.

The spec above and the ~30 skipped specs run the same code: a pinned model is looked up by anchored text, and a miss is
reported as `model may not be supported`. The spec above is in the table because it is the one that **throws** — that is
why the run captured a trace for it, and the trace is what proves the premise false. The skips carry no trace and no
signature by construction, which is the gap itself.

The matcher's own repair is tracked separately (two sibling issues from this triage). This issue is only about the guard
reporting a cause it did not establish, which would have made that matcher break visible on day one instead of showing up
as a skip count.

Descriptive only.

- The guard's message states a cause (`model may not be supported`) from a single negative observation (an anchored text
  match found nothing). Every other explanation — a changed option label, a picker that had not finished loading, a
  provider that failed to authenticate — produces the identical message.
- The failure is silent by design: `test.skip(true, e.message)` yields a green test. `reports/daily-history.jsonl` records
  only `totals.skipped`, so 30 lost tests appear as one number that moved from 4 to 35.
- The information needed to tell the two apart was **in the DOM at the moment of the check**: the option list was open and
  90 options were rendered. The helper had it and discarded it.
- This is the same class as the run-integrity and catalog-drift guards already in the repo (`#1012`): an unevaluated test is
  unknown, not clean.

Whether the right answer is a stricter guard, a louder report, or both is open.

## Investigation directive

(a) Establish what the helper can cheaply prove at the point of the check — the option list is open and enumerable, and the
provider modal's `llm-toggle-<model>` controls are a second independent source for "this model exists and is enabled"; a
miss that can be contradicted by either is not an absence. (b) Decide the failure mode: a model genuinely retired from the
catalog is a legitimate skip, while a model present-but-unmatchable is a suite defect and should be **loud** — at minimum
the message must carry what the helper actually saw (option count, and the near-misses) instead of a cause it inferred.
(c) Check whether the skip totals should be gated at all: this run's 35 against a 4–15 baseline is exactly the signal that
went unread, and `reports/daily-history.jsonl` already carries the series. (d) Confirm the three helpers
(`setup-openai.ts`, `setup-anthropic.ts`, `setup-google.ts`) are the whole surface, and whether `setup-ollama.ts` and the
`fallbackToRanking` path in `setup-openai.ts` need the same treatment.

## Deliverables (Done when)

- [ ] `MODEL_NOT_AVAILABLE` is raised only when absence is **established**, not inferred from one matcher missing — and the message carries the evidence (what the picker actually offered).
- [ ] A model that is present but unmatchable fails or reports loudly instead of skipping; a model genuinely retired still skips.
- [ ] A regression test covers the distinction, so a future matcher break cannot resolve to a silent skip again (`npm run test:units` or a spec, whichever fits the helper).
- [ ] The skip-total anomaly this run produced (35 against a 4–15 baseline) is either surfaced by the daily or explicitly recorded as out of scope, with the reason.

rafaelgiln:
Closes #1461
Closes #1459

## Problem

`MODEL_NOT_AVAILABLE` asserted a cause it never checked.

On [run 31786538844](https://github.com/oriontech-me/langflow-e2e/actions/runs/31786538844) (2026-08-14, `1.12.0.dev26`) the daily reported 4 failures — and lost **~30 `@stable` tests to silent skips**, all through the same primitive. The run's skip total (35, against a 4–15 baseline since 2026-07-30) was the only trace it left, and that is the field nobody watches.

**Trigger (upstream, legitimate).** dev26 added a per-option accessibility counter to the unified ModelInput picker. Read from the frontend bundle (`index-JOT9mUHU.js`):

```jsx
<div className="truncate text-[13px]">{f.name}</div>
<span className="sr-only">{o("modelInput.optionPosition", { current: c, total: a })}</span>
```

so an option's `textContent` became `"claude-haiku-4-5\n9 of 90"`. The identity attributes are unchanged (`data-value="Provider::model"`, `data-testid="Provider-model-option"`).

**Layer 1 — the matcher (#1459).** `setup-google` / `setup-openai` / `setup-anthropic` resolved a pinned model with `locator('[data-testid$="-option"]', { hasText: new RegExp('^' + model + '$') })`, and `setup-language-model-openai` read labels with `allInnerTexts()`. All of them stopped matching.

**Layer 2 — the guard (#1461).** The miss was reported as `MODEL_NOT_AVAILABLE: "<id>" not found in dropdown — model may not be supported.` Every caller turns that prefix into `test.skip`, so a broken matcher resolved to a green run. The trace of the one call site that throws instead shows the model **present, at option 22 of 90**.

## Change

Model options are resolved through a new shared module, `tests/helpers/provider-setup/model-option.ts`, by **identity** (`data-value` → `data-testid` fallback), never by rendered text. The verdict is split so only one outcome can become a skip:

| Verdict | When | Consequence |
|---|---|---|
| `match` | identity resolves | clicks by testid (falls back to `data-value`) |
| `absent` | the picker was enumerated **and** the model is in neither the options nor the provider panel's `llm-toggle-*` list | `MODEL_NOT_AVAILABLE` → skip. The message now carries the option count, the per-provider breakdown and the nearest offered ids |
| `unmatchable` | the picker offers the model (or the panel has its toggle) but the suite cannot select it | `MODEL_PICKER_DEFECT` → **fails**. That is our defect, not a missing model |
| `empty` | zero options rendered | `MODEL_PICKER_DEFECT` → **fails**. Zero options prove nothing about a model (#1012) |

Wired into all five helpers. Notes on the two non-obvious ones:

- **`setup-openai`'s `fallbackToRanking` (#606) still degrades on a stale `models.json` pin — but only for an ESTABLISHED absence.** An empty picker, or one that offers the model without letting us select it, throws through it: degrading there would hide a suite defect behind a healthy-looking run.
- **`setup-ollama` already matched by testid**, so its selection was never affected. Only its evidence strings changed — a failure no longer offers `llama3.2:1b\n3 of 4` as the tag to pull.

One spec-local matcher of the same shape is migrated with them: `anthropic-provider.spec.ts`'s `switchModel`, which was **masked** by the skip wave (the file skipped before reaching it) and goes red the moment the helper works again.

## Validation

- **Langflow `1.12.0.dev26`** — the build the daily ran, pulled and started on port 7861. `dev25` (the local default) has no counter, which is why this only surfaced on 2026-08-14.
- `anthropic-provider.spec.ts` — **3/3 clean** at `--retries=0 --workers=1`.
- `language-model-regression.spec.ts` (the issue's hard failure) — **3/3 clean**; its Google test passes. The two OpenAI tests skip locally on a drained key.
- `npm run test:units` — **655 pass**, including 13 new ones whose fixtures carry the real dev26 markup, counter included.
- `npm run typecheck` — 0. `npm run lint` — 0 errors.

**Force-fail (executed, red observed, revert proved, final green run):**

- `Anthropic API key is configured via Settings → Model Providers` — commented out the Save click, so neither armed waiter fires → failed.
- `configured Anthropic selects a Claude model in the Agent and executes the flow` — asserted `/gemini/i` on the model trigger instead of `/claude/i` → failed.
- `switches between Claude model families (Haiku → Sonnet → Opus)` — **restored the pre-#1459 anchored text matcher**, so the dev26 counter defeats it → failed. This is the force-fail for the fix itself.
- Unit lane, 4 mutations: matching on `rawText`; renaming `MODEL_PICKER_DEFECT` to the skip prefix; degrading `empty` to `absent`; ignoring the toggle source. Each killed exactly the tests that cover it; revert → 13/13 green.

One backend error was declared ambient in the pipeline record: a `500` on `DELETE /api/v1/flows/{id}` from the spec's own id-scoped cleanup, the known local SQLite-lock class under Colima. It appeared in a run where all 3 tests passed, and not in two earlier runs of the same code; this diff touches no cleanup path, and the fixture treats HTTP errors as advisory (#1084).

## Scope — what is deliberately NOT here

- **`mcp-client-agent.spec.ts`** carries the same anchored matcher, and its fix is not in this PR because it **cannot be validated on this machine**: it dies on `GET /api/v2/mcp/servers?action_count=true` (20 s) both **with and without** this diff — baseline identical, 38.37 s vs 38.36 s, the #1266 class where that call starts every registered MCP server. Left at today's behavior (the matcher misses and the spec falls back to the shared provider setup), tracked for #1459's remaining half rather than shipped unproven.
- **`modelInputComponent.spec.ts` / `memory-base-registration.spec.ts`** — the spec-local half, owned by #1460.
- **Gating the run's skip total** (the issue's item c) — out of scope, with the reason: once the guard stops inferring absence, a silent skip is no longer this defect's channel, and a totals gate belongs to the run-integrity mechanism (`scripts/check-run-integrity.mjs`), not to a provider helper.
- **`setup-openai` / `setup-language-model-openai` have no live local coverage in this PR** — the local `OPENAI_API_KEY` is drained (`no credits remaining`), so those paths were exercised only through the unit lane here; CI covers them.

## Found on the way (not fixed here)

`model-provider-model-toggle.spec.ts` derives the model name as `testid.replace(/-option$/, "")`, which yields `Anthropic-claude-haiku-4-5` while the toggles yield `claude-haiku-4-5` — so `targetModel` is always empty and the test **skips on every run** (it is the `"No non-selected dropdown option belongs to the test's provider"` skip in the same daily). Its negative assertion is anchored on the option text too, so it would pass vacuously even if it ran. Worth its own issue: fixing it wakes an assertion that has never executed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -257,6 +257,22 @@ for (const { label, options, skipReason } of targets) {
 }
 ```
 
+**Skip on that prefix and on nothing else.** `MODEL_NOT_AVAILABLE` is raised only
+for an absence the helper **established** — it enumerated the open picker, the
+model was not among the options, and the provider panel's `llm-toggle-*` list did
+not carry it either — and the message reports what it saw (option count,
+per-provider breakdown, nearest offered ids). Two sibling prefixes must never be
+caught: a model the picker IS offering that the suite cannot select, and an empty
+picker, both raise `MODEL_PICKER_DEFECT` and must fail the test.
+
+That split is the fix for #1461: the old guard inferred "model may not be
+supported" from one anchored text matcher returning nothing, so when 1.12.0.dev26
+added a `sr-only` "N of M" counter inside every option (#1459), ~30 `@stable`
+tests skipped green in a single daily and the only trace was the run's skip total
+(35 against a 4–15 baseline). Resolve a model through
+`tests/helpers/provider-setup/model-option.ts`, which matches the option's
+identity (`data-value` / `data-testid`) — never its rendered text.
+
 ### Running agent tests
 
 This is the most important step — and the most overlooked.
--- a/docs/core-functionality/model-provider/anthropic-provider.md
+++ b/docs/core-functionality/model-provider/anthropic-provider.md
@@ -201,6 +201,20 @@ nightly. `@model-provider` (area) · `@settings` (Test 1 navigates Settings) ·
   need configure + select + switch. A dedicated `anthropic-provider.spec.ts`
   under `model-provider/` follows the family precedent (§7.2/§7.4/§7.6 all
   chose dedicated specs) and gives §7.3 a provider-centric home.
+- **How a model option is matched (#1459/#1461, 1.12.0.dev26):** Test 3's switch
+  and the shared `provider-setup` helpers resolve a model by the option's
+  **identity** — `data-value` (`${provider}::${model}`), falling back to
+  `data-testid` (`${provider}-${model}-option`) — through
+  `tests/helpers/provider-setup/model-option.ts`. Never by the option's text:
+  dev26 added a `sr-only` position counter inside every option
+  (`modelInput.optionPosition` → `"9 of 90"`), so `textContent` is
+  `"claude-haiku-4-5\n9 of 90"` and an anchored `^model$` matcher resolves
+  nothing. When it does not resolve, the failure mode is now split: an absence
+  the helper **established** (picker enumerated, model in neither the options
+  nor the provider panel's `llm-toggle-*` list) raises `MODEL_NOT_AVAILABLE`
+  and skips, carrying the option count, the per-provider breakdown and the
+  nearest offered ids; a model the picker IS offering, or an empty picker,
+  raises `MODEL_PICKER_DEFECT` and **fails** — no spec may skip on it.
 - **Model resolution from `models.json`:** Haiku/Sonnet/Opus are resolved by
   family regex from the collected catalog (preferring current, non-deprecated
   names, e.g. `claude-haiku-4-5`, `claude-sonnet-5`, `claude-opus-4-8`), so
--- a/tests/helpers/provider-setup/model-option.test.ts
+++ b/tests/helpers/provider-setup/model-option.test.ts
@@ -1,0 +1,208 @@
+// Unit tests for the model-picker resolver (issues #1459 and #1461).
+// Run with: npm run test:units
+//
+// What rides on this function is a verdict, not a selector. Until #1459 the three
+// provider-setup helpers resolved a pinned model with
+// `hasText: new RegExp("^" + model + "$")`, and on 1.12.0.dev26 the picker began
+// rendering a per-option `sr-only` position counter INSIDE the option:
+//
+//   <div data-testid="Google Generative AI-gemini-flash-latest-option" ...>
+//     <div class="truncate text-[13px]">gemini-flash-latest</div>
+//     <span class="sr-only">22 of 90</span>
+//   </div>
+//
+// The counter is invisible to a user but part of textContent, so `^model$` matched
+// nothing — and the helper reported that single negative observation as
+// `MODEL_NOT_AVAILABLE: … model may not be supported`, which every caller turns
+// into `test.skip`. One daily (run 31786538844) lost ~30 `@stable` tests that way,
+// with 35 skips against a 4–15 baseline as the only trace.
+//
+// So the property under test is NOT "the matcher works". It is: a model the picker
+// actually offers can only ever resolve to a LOUD failure — never to the skip
+// prefix. The fixtures below therefore carry the real dev26 markup, counter
+// included.
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import {
+  enumerateModelOptions,
+  nearestModels,
+  resolveModelOption,
+  toModelOption,
+  type ModelOption,
+  type RawModelOption,
+} from "./model-option";
+
+/** Verbatim shape of a dev26 option, counter included. */
+function raw(provider: string, model: string, position: string, deprecated = false): RawModelOption {
+  return {
+    testId: `${provider}-${model}-option`,
+    value: `${provider}::${model}`,
+    visibleLabel: model,
+    rawText: `${model}\n${position}${deprecated ? "\nDeprecated" : ""}`,
+    deprecated,
+  };
+}
+
+function option(provider: string, model: string, position = "1 of 1"): ModelOption {
+  return toModelOption(raw(provider, model, position));
+}
+
+const DEV26_PICKER: ModelOption[] = [
+  option("Anthropic", "claude-opus-5", "1 of 90"),
+  option("Anthropic", "claude-haiku-4-5", "9 of 90"),
+  option("Google Generative AI", "gemini-flash-latest", "22 of 90"),
+  option("OpenAI", "gpt-4o-mini", "40 of 90"),
+  option("OpenAI", "gpt-4o-mini-search-preview", "41 of 90"),
+];
+
+test("toModelOption reads the identity from data-value, not from the text", () => {
+  const parsed = toModelOption(raw("Google Generative AI", "gemini-flash-latest", "22 of 90"));
+  assert.equal(parsed.provider, "Google Generative AI");
+  assert.equal(parsed.model, "gemini-flash-latest");
+  // The counter survives in rawText: it is evidence for the message, never a match key.
+  assert.equal(parsed.rawText, "gemini-flash-latest\n22 of 90");
+  assert.equal(parsed.visibleLabel, "gemini-flash-latest");
+});
+
+test("toModelOption falls back to the testid when data-value is gone", () => {
+  const parsed = toModelOption({
+    testId: "Anthropic-claude-haiku-4-5-option",
+    value: "",
+    visibleLabel: "claude-haiku-4-5",
+    rawText: "claude-haiku-4-5\n9 of 90",
+    deprecated: false,
+  });
+  assert.equal(parsed.provider, "Anthropic");
+  assert.equal(parsed.model, "claude-haiku-4-5");
+});
+
+test("toModelOption keeps a model id containing '::' intact", () => {
+  const parsed = toModelOption({
+    testId: "Ollama-llama3.2:1b-option",
+    value: "Ollama::llama3.2:1b",
+    visibleLabel: "llama3.2:1b",
+    rawText: "llama3.2:1b\n3 of 4",
+    deprecated: false,
+  });
+  assert.equal(parsed.model, "llama3.2:1b");
+});
+
+// The regression the whole file exists for.
+test("the dev26 counter no longer defeats the match", () => {
+  for (const model of ["claude-haiku-4-5", "gemini-flash-latest", "gpt-4o-mini"]) {
+    const verdict = resolveModelOption(model, DEV26_PICKER);
+    assert.equal(verdict.kind, "match", `${model} must resolve on a dev26 picker`);
+    if (verdict.kind === "match") assert.equal(verdict.option.model, model);
+  }
+});
+
+test("a longer model id is not matched by its prefix", () => {
+  const verdict = resolveModelOption("gpt-4o-mini", DEV26_PICKER);
+  assert.equal(verdict.kind, "match");
+  if (verdict.kind === "match") {
+    assert.equal(verdict.option.testId, "OpenAI-gpt-4o-mini-option");
+  }
+});
+
+test("an identity attribute that drifts FAILS loudly instead of skipping", () => {
+  // The #1459 break one layer deeper: the option is right there, labelled exactly
+  // as the test pinned it, but `data-value` now spells the id differently. The
+  // suite cannot establish an absence against its own evidence, so this is a
+  // failure — the outcome the old guard turned into `test.skip`.
+  const drifted: ModelOption = toModelOption({
+    testId: "Anthropic-claude-haiku-4-5-option",
+    value: "Anthropic::claude-haiku-4.5",
+    visibleLabel: "claude-haiku-4-5",
+    rawText: "claude-haiku-4-5\n9 of 90",
+    deprecated: false,
+  });
+  const verdict = resolveModelOption("claude-haiku-4-5", [drifted]);
+  assert.equal(verdict.kind, "unmatchable");
+  assert.ok(!verdict.message.startsWith("MODEL_NOT_AVAILABLE"), "must not carry the skip prefix");
+  assert.match(verdict.message, /IS offered by the model picker/);
+  assert.match(verdict.message, /visible label "claude-haiku-4-5"/);
+});
+
+test("a display label is a usable identity when the attributes are gone", () => {
+  // The complement of the test above, and the reason absence stays rare: with no
+  // parseable attribute left, the visible label still identifies the option and
+  // the testid still clicks it. Matching here is correct — failing loudly on a
+  // selectable model would be its own coverage loss.
+  const labelOnly = toModelOption({
+    testId: "model-option-9",
+    value: "",
+    visibleLabel: "claude-haiku-4-5",
+    rawText: "claude-haiku-4-5\n9 of 90",
+    deprecated: false,
+  });
+  const verdict = resolveModelOption("claude-haiku-4-5", [labelOnly]);
+  assert.equal(verdict.kind, "match");
+});
+
+test("a provider name carrying spaces still parses from the testid alone", () => {
+  const verdict = resolveModelOption("gemini-flash-latest", [
+    toModelOption({
+      testId: "Google Generative AI-gemini-flash-latest-option",
+      value: "",
+      visibleLabel: "gemini-flash-latest",
+      rawText: "gemini-flash-latest\n22 of 90",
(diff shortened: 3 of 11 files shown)


rafaelgiln:
## CI red is ambient — measured, not asserted

`Run impacted E2E specs` came back **1 failed, 1 flaky, 1 skipped, 40 passed (7.0m)**. Neither red belongs to this PR.

### 1 failed — `general-bugs-agent-sum-duplicate-message-playground.spec.ts:8` (`@release @components @workspace`)

```
  waiting for locator('[data-testid="header-icon"] svg[data-testid="icon-Check"]')
  at …/general-bugs-agent-sum-duplicate-message-playground.spec.ts:43:16
```

It is selected by this PR only because it imports `setup-anthropic`. **It fails identically without this PR's changes**, measured on `1.12.0.dev26` back to back:

| Tree | Result | Duration | Error |
|---|---|---|---|
| this branch | failed | 37.2 s | `waitForSelector … icon-Check`, 30 s |
| `provider-setup/` reverted to `main` | failed | 36.5 s | identical selector, identical timeout |

The run never completes, so the completion check never renders — the spec dies after the Playground send, not in model selection. Model selection itself is exercised and green in this same lane: `anthropic-provider.spec.ts` (3 tests, including the Haiku → Sonnet → Opus switch) and `language-model-regression.spec.ts` both passed here, and both go through the resolver this PR introduces.

This spec carries no `@stable`, so the daily never runs it and its breakage has been invisible — the class of `@release`-only specs already noted as broken on a clean `main`. It needs its own triage issue; it is not a verdict about this diff.

### 1 flaky — `loop-component-regression.spec.ts:76`

`TimeoutError: page.waitForSelector: Timeout 15000ms exceeded`, passed on retry. It touches no model picker and no provider helper.

### 1 skipped

An OpenAI-parametrized case; the lane pinned `openai / gpt-4o-mini` and the sweep reported `gemini-2.5-flash` retired for new users (unrelated to this PR).

The 40 passing specs include every surface this PR changes. The failing spec is a pre-existing red on `main` with the pre/post measurement above, and holding this PR would leave the ~30 `@stable` tests that #1461 is about skipping silently for another daily.

rafaelgiln:
The failing spec now has its own issue: #1465 (measurement table included there). The flaky `loop-component-regression` passed on retry and touches neither the picker nor a provider helper.

Issue #827, opened 2026-07-18
Tests: Agent executes multiple tools in sequence + Inspect tools used in Playground (§6.4)


rafaelgiln:
## What
Cover uncovered §6.4: Agent executing multiple tools in sequence, and inspecting the tools an Agent used from the Playground.

## Notes
Multi-tool **extends** `agent-multi-tool-selection.spec.ts`; tool-inspection is likely a **new** spec.

> **Gated:** author the spec now, but promote to `@stable` only after the clean non-guarded baseline (infra issue in this milestone) is achieved — this area is in the current flaky cluster (#773).

## Done when
- [ ] Spec doc(s); cases added/created; QA-CHECKLIST bullets flipped (promotion gated)

PR #1466, opened 2026-08-14
fix(llm-agents): reach the provider panel, and refuse to claim there was nothing to configure (#1465)

### LINKED ISSUE #1465: general-bugs-agent-sum-duplicate-message-playground — the Playground run never completes (broken on main, invisible because the spec is @release-only)
Found while validating PR #1463 (#1459/#1461). It is **not** caused by that PR — measured both ways, see below.

## Symptom

`tests/tests-automations/regression/core-functionality/llm-agents/general-bugs-agent-sum-duplicate-message-playground.spec.ts:8` (`@release @components @workspace`) fails at the completion gate:

```
  waiting for locator('[data-testid="header-icon"] svg[data-testid="icon-Check"]')
  at …/general-bugs-agent-sum-duplicate-message-playground.spec.ts:43:16
```

3/3 attempts on the PR lane ([run 31808413329](https://github.com/oriontech-me/langflow-e2e/actions/runs/31808413329/job/94792890497)), and 1/1 locally.

## It is pre-existing, not PR #1463

Measured back to back on `1.12.0.dev26` with the same instance, Anthropic configured and healthy:

| Tree | Result | Duration | Error |
|---|---|---|---|
| PR #1463 branch | failed | 37.2 s | `waitForSelector … icon-Check`, 30 s |
| `tests/helpers/provider-setup/` reverted to `main` | failed | 36.5 s | identical selector, identical timeout |

So the model selection this spec performs (`setupAnthropic(page)` with no pin) is not the blocker: the run is sent and the completion check never appears. In the same PR lane, `anthropic-provider.spec.ts` (3 tests, including the Haiku → Sonnet → Opus switch) and `language-model-regression.spec.ts` passed through the same helper.

## Why nobody noticed

The spec carries `@release` but **not** `@stable`, so `daily-stable.yml` never runs it. It only executes when a PR happens to touch something it imports — which is why it surfaced now, on a provider-helper change. This is the `@release`-only blind spot: a spec can be red on `main` for weeks and every generated coverage total still counts it.

## Investigation directive

- Establish where the run dies: does `POST /api/v2/workflows` return an error the spec never reads, or does the build complete while `header-icon`/`icon-Check` no longer renders on 1.12? The completion observable is a UI icon, so a rename would produce exactly this timeout with a healthy run behind it.
- Decide the verdict (product regression vs. expired UI premise vs. wait strategy) with evidence on the current nightly before touching the spec.
- If it is a UI premise that expired, the fix is a completion observable that cannot silently pass — and the spec must be force-failed to prove it.

## Done when

- [ ] Root cause established with evidence on the current nightly.
- [ ] The spec passes on multiple clean `--retries=0` runs, or is quarantined with the product defect filed upstream.
- [ ] Decide whether it earns `@stable` — an `@release` spec no lane runs is coverage on paper only.

rafaelgiln:
Closes #1465

## Problem

`general-bugs-agent-sum-duplicate-message-playground.spec.ts` — the regression guard for input duplication in the Agent Playground — had been failing on `main` with **no signal at all**. It carries `@release` but not `@stable`, so no scheduled lane runs it; it only executed again when PR #1463 touched a helper it imports, and the pre/post measurement there proved it was not that PR (37.2 s on the branch vs 36.5 s with `provider-setup/` reverted, identical selector).

Three causes, stacked. All measured on `1.12.0.dev26`.

### 1. The provider-panel entry point stopped being addressable

The Agent node's **Setup Provider** control is still a real `<button>`, but it now carries `aria-labelledby="node-Agent-<id>-field-model-label"`, so its accessible name is the *field's* label:

```
aria snapshot:  - button "Language Model required": Setup Provider

getByRole("button", { name: "Setup Provider" })          → 0 matches
getByRole("button", { name: /Setup Provider/i })         → 0 matches
locator("button").filter({ hasText: "Setup Provider" })  → 1 match   (no aria-hidden, no inert)
```

All three keyed helpers (`setup-openai`, `setup-anthropic`, `setup-google`) looked it up by role+name. On an instance where the provider is **not** configured, `model_model` does not exist either — the field renders the button instead — so neither entry point resolved.

### 2. …and the miss was reported as a conclusion

With both lookups empty, the helpers returned through:

```ts
if (!hasModelDropdown && !hasSetupButton) {
  console.log("No Agent node found on canvas — skipping … setup.");
  return;
}
```

The canvas had **1** Agent node at that moment (measured). Nothing was configured, the Agent ran with no model, and the backend answered:

```
ComponentBuildError: Error building Component Agent:
No model selected. Please select a language model from the available options.
```

which surfaced 30 s later as the spec's completion gate timing out — far from the cause. Same class as #1461: a single negative observation reported as a fact.

There is a second half to it: even with the right matcher, the probe ran **before** ReactFlow mounted. Measured: 0 Agent nodes right after the URL settles, 1 a few seconds later. A decision taken on that snapshot reports "nothing to configure" about a canvas still painting.

### 3. The spec's own observable had expired

It waited for `[data-testid="header-icon"] svg[data-testid="icon-Check"]`, then expanded that header to read the tool payload from `chat-code-tab`. On dev26 **both count 0** — the same 1.11 → 1.12 change that removed the expandable tool-inspection accordion (#827).

## Change

**`tests/helpers/provider-setup/provider-panel-entry.ts` (new)** owns reaching the panel:

- the not-configured control is matched by its **visible text**, which the label wiring cannot take away;
- the probe is **polled to a deadline**, never sampled once;
- the decision is a pure function with four outcomes, and only one of them is silent:

| Probe | Decision |
|---|---|
| `model_model` present | open via the dropdown → *Manage Model Providers* |
| Setup Provider button present | open it directly |
| no Agent node | return silently — the only state where "nothing to configure" is true |
| Agent node present, neither control | **throw** `PROVIDER_PANEL_UNREACHABLE`, naming the counts it observed |

The three helpers call it instead of hand-rolling the block.

**The spec** keeps its intent and gets a stronger, measured observable. The chat message testid embeds the message text, so the duplication signature is read where the bug actually shows:

```ts
await expect(page.getByTestId("chat-message-User-2+2")).toBeVisible();
await expect(page.getByTestId("chat-message-User-2+22+2")).toHaveCount(0);
await expect(page.getByTestId("chat-message-User-22+2")).toHaveCount(0);
expect(answer).toContain("4");
expect(answer).not.toContain("26");           // 2+22+2
```

Reading the payload off the run stream was considered and **rejected**: for `2+2` the Agent answers directly without calling the Calculator, so a tool-anchored assertion would depend on the model *choosing* to use one (#1187's lesson).

Two more corrections in the same file: it loads through `SimpleAgentTemplatePage`, which blocks until the **persisted** binding carries the pinned provider+model (#751/#1274) rather than continuing as soon as a helper returns; and it gained the id-scoped `afterEach` cleanup it never had — it created one flow per run and deleted none.

**The spec doc did not exist** and is added. That absence is part of the story: the premise this spec depended on (a `role=button` named "Setup Provider") was never written down anywhere a reviewer could check.

## Validation

- **Langflow `1.12.0.dev26`** (the build the daily runs), container on port 7861.
- `general-bugs-agent-sum-duplicate-message-playground.spec.ts` — **3/3 clean** at `--retries=0 --workers=1`, **zero backend errors**, plus the green run after the force-fail reverts.
- `npm run test:units` — **648 pass**, including 6 new ones on the entry decision.
- `npm run typecheck` — 0. `npm run lint` — 0 errors.
- Flow hygiene: the 17 flows my scouting created were purged (42 → 26 on the instance).

**Force-fail (executed, red observed, revert proved):**

- *spec* — asserted the **duplicated** user message (`chat-message-User-2+22+2`) instead of the clean one → failed, so the assertion really reads the input text.
- *unit* — relaxed `agentNodes === 0` to `>= 0`, i.e. let every state resolve to "nothing to configure" → the two tests that pin the refusal failed; revert → 6/6 green.
- *behavioral, reported as it happened* — reverting the polled probe to a single snapshot (the pre-fix shape) left the spec **green** (16.3 s). With the Page Object the canvas is already mounted by the time the helper runs, so this spec does not exercise the polling; its justification is the scout measurement (a bare helper call right after the template load sees 0 Agent nodes). Stated here rather than claimed as coverage.

## Notes for the reviewer

- **Overlaps PR #1463 by file, not by region.** That PR rewrites step 7 of the same three helpers (model-option selection); this one replaces steps 1–2 (panel entry). Both are branched off `origin/main`, so whichever merges second may need a trivial rebase.
- **`@stable` is not added here**, per the issue's framing. While it stays `@release`-only, no scheduled lane runs this spec — which is exactly how it stayed broken unnoticed. Worth a decision after it has some green dailies behind it.


rafaelgiln:
## CI red — this PR depends on #1463, and my earlier validation understated it

`Run impacted E2E specs` failed with **3 failed**, and one of them is this PR's own spec:

```
general-bugs-agent-sum-duplicate-message-playground.spec.ts:45
  Error: MODEL_NOT_AVAILABLE: "claude-haiku-4-5" not found in dropdown — model may not be supported.   (3/3 attempts)
```

That is the **#1459/#1461 defect**, whose fix lives in PR #1463 and is *not* on this branch — this one is cut from `origin/main`.

### Correction to my own validation

The 3/3 clean burst I reported ran with **no model pinned**, and I did not notice: the worktree had no `tests/helpers/provider-setup/data/models.json` (it is gitignored), so the spec's `resolveClaudeModel()` returned `undefined` and `SimpleAgentTemplatePage.load()` took its unpinned path — which resolves a model by substring and therefore survives the dev26 `sr-only` counter. CI has a real catalog, pins `claude-haiku-4-5`, and hits the anchored matcher.

Re-measured with the catalog present, same instance (`1.12.0.dev26`), same pin as CI:

| Tree | Pin | Result |
|---|---|---|
| this branch alone | `claude-haiku-4-5` | **failed** — `MODEL_NOT_AVAILABLE`, the CI signature |
| this branch + #1463's commit applied locally | `claude-haiku-4-5` | **passed**, 16.7 s |

So the two PRs unblock each other by construction, and the ordering is decided:

- **#1463 first** — it repairs the matcher used on the path CI takes (provider already configured by `Collect models` → the `model_model` dropdown).
- **#1466 second** — it repairs the entry point for the path CI never takes (provider *not* configured → the Setup Provider button), which is why its own break was invisible until now.

The conflict between them is real but trivial: three import blocks in the same three helpers. They touch different steps of the same functions — entry (steps 1–2) here, model selection (step 7) there — and resolving by keeping both sides typechecks clean, which is how the measurement above was produced.

### The other two failures

- `language-model-regression.spec.ts` (Google) — the same `MODEL_NOT_AVAILABLE` mechanism, also fixed by #1463.
- `agent-context-id-continuity.spec.ts` — died on the page-entry barrier (`mainpage_title` absent for 30 s while `GET /api/v1/version` answered 200 in 115–4315 ms), alongside `apiRequestContext` timeouts and repeated 500s on `/api/v1/flows/`. That is the saturated-instance signature, not a verdict about this diff.

Recommendation: merge #1463, then re-run this lane. I have not merged anything.

rafaelgiln:
## CI verdict after rebasing onto `main` (post-#1463)

Green. Recording the first red run so the history reads correctly.

**Run [31827978955](https://github.com/oriontech-me/langflow-e2e/actions/runs/31827978955) — first attempt:** 34 passed, 3 failed. All three failures in `language-model-regression.spec.ts`, none in the spec this PR touches (`general-bugs-agent-sum-duplicate-message-playground.spec.ts` passed).

| Test | Failure |
|---|---|
| `:240` provider dialog opens | `[backend-unreachable] … the backend did not answer GET /api/v1/version within 5005ms` — attributed by the barrier helper itself |
| `:216` provider switch OpenAI→Google | `GET /api/v1/auto_login` timeout 20000ms at spec line 67, before any `setupGoogle`; retry hit the barrier with `/version` at 4340ms |
| `:135` must respond with Google provider | `node_duration_chat output` not visible in 60s; retry died on the same `auto_login` timeout |

Two of the three carry the documented infra signature (#1074's VOID class).

**`:135` was already red on `main` before this PR.** Today's daily (run [31786538844](https://github.com/oriontech-me/langflow-e2e/actions/runs/31786538844), 05:00, pre-#1463 merge) recorded it as:

```
Error: MODEL_NOT_AVAILABLE: "gemini-flash-latest" not found in dropdown — model may not be supported.
```

That is the #1459 defect #1463 fixed. In this run the signature is gone — `Collect models` reports `✅ google` and `gemini-flash-latest` is in the catalog, so the option now resolves and the failure moved one layer deeper, to the run itself.

**This PR is not implicated in any of the three.** The dropdown branch of `setupGoogle` is byte-identical to `main` (`hideInspectorPanel` → click → `manage-model-providers`); this PR only swaps `getByRole` for a visible-text match on the `Setup Provider` button and adds polling. `language-model-regression` drives the **Language Model** node on a Basic Prompting flow with no Agent node on canvas, so the changed branches are never taken.

**Control:** re-ran the failed job on the same commit — [job 94864539131](https://github.com/oriontech-me/langflow-e2e/actions/runs/31827978955/job/94864539131): **35 passed, 0 failed, 2 flaky** (`:135` and `:240`, both green on retry). Ambient saturation, confirmed by control rather than asserted.

**Local, post-rebase:** `typecheck` 0, `lint` 0 errors, `test:units` 661 pass, and the touched spec 2/2 green on 1.12.0.dev26 (60.3s, 46.5s). Leaked flows purged — instance back to its 26 seeded starters.

Issue #1234, opened 2026-08-03
[Daily #1231] loop-component — the Research Translation Loop template card never appears in the all-templates sidebar (deterministic, 3/3)


Victor-w-Madeira:
Spun out of daily-failure triage #1231 (run [30809091241](https://github.com/oriontech-me/langflow-e2e/actions/runs/30809091241), 2026-08-03). Hard failure: `@stable` was already auto-removed by the workflow (commit `32ac9a1`).

**Upstream:** _not filed_

## Symptom

The Research Translation Loop template card never renders in the all-templates sidebar, so the test cannot instantiate the flow it exercises. The nav click itself succeeds — the failure is on the card that should follow it. Failed **3 of 3 attempts** (11:27:06, 11:27:50, 11:28:27 UTC), i.e. it did not recover on either retry.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-components/loop-component-regression.spec.ts:152` ("Loop component — Research Translation Loop template: full wiring and iterates over 2 ArXiv papers") | `[data-testid="template-research-translation-loop"]` | `TimeoutError: page.waitForSelector: Timeout 10000ms exceeded.` |

## Why these failures are one cause

Single spec, single wait point, one signature — nothing is being grouped here. It is filed on its own rather than folded into any other cluster of the run because no other failure on 2026-08-03 waits on the templates modal, and the two recurrent flakes of the day sit on a different surface (Settings → Global Variables, tracked separately).

The file declares `test.describe.configure({ mode: "serial" })` (line 13), so the sibling test at `tests/tests-automations/regression/core-components/loop-component-regression.spec.ts:366` ("stops after exhausting input DataFrame and emits aggregated done") was **skipped as serial collateral** of this failure — it is not an independent skip and needs no separate issue.

## Preliminary read (descriptive — NOT a verdict)

Deterministic, not flaky: three consecutive attempts spanning 97s all died on the same locator, so the retry budget did not absorb it. The guard did **not** trip (1 hard failure, threshold 6), and the day's backend liveness measurement puts shard 3 — where this ran — at **0 outages, 0 min down**, so the in-run recorder gives no support to an outage reading. `page.waitForSelector: Timeout` is deliberately **not** on the infra-signature exemption list (`scripts/lib/infra-signatures.ts`) precisely because a wedge and a real regression both produce it.

[...]

PR #1538, opened 2026-08-20
fix(regression): repair the wiped sidebar search term, not the wait around it (#1518)

### LINKED ISSUE #1518: [Daily #1517] Sidebar component list never renders the filtered entry after sidebar-search-input.fill (4 specs, recurrent across 3 dailies)
Spun out of daily-failure triage #1517 (run [32349515682](https://github.com/oriontech-me/langflow-e2e/actions/runs/32349515682), 2026-08-20). Filed from a **guard-tripped** day (7 hard failures > threshold 5, so the workflow auto-removed no `@stable`).

**Upstream:** _not filed_

## Symptom

Four `@stable` specs fail on the same three-step opening sequence: open a blank flow, type a term into `sidebar-search-input`, then wait for the filtered component entry (or its `+` button) to appear. **None of them reaches the click** — the sidebar entry never renders.

```ts
await page.getByTestId("sidebar-search-input").fill("chat input");
await expect(page.getByTestId("input_outputChat Input")).toBeVisible();  // <-- never appears
```

The literal error strings differ only because each spec waits in its own way — `page.waitForSelector`, `locator.click`, `expect(...).toBeVisible()`, `locator.hover`. The budgets exceeded are 10 s, 20 s, 20 s and 30 s, across four different components (Chat Input ×2, Chat Output, Loop).

|---|---|---|
| `tests/tests-automations/regression/ui-ux/langflowShortcuts.spec.ts:47` ("LangflowShortcuts") | `getByTestId('add-component-button-chat-output')` | `TimeoutError: locator.click: Timeout 20000ms exceeded.` |
| `tests/tests-automations/regression/core-components/loop-component-regression.spec.ts:76` ("Loop component — renders correctly with all handles and output inspection buttons") | `[data-testid="add-component-button-loop"]` | `TimeoutError: page.waitForSelector: Timeout 10000ms exceeded.` |
| `tests/tests-automations/regression/core-functionality/knowledge-ingestion-management/limit-file-size-upload.spec.ts:41` ("user should not be able to upload a file larger than the limit") | `getByTestId('input_outputChat Input')` | `Error: expect(locator).toBeVisible() failed` |
| `tests/tests-automations/regression/ui-ux/notifications.spec.ts:29` ("User should be able to interact notifications tab") | `getByTestId('input_outputChat Input')` | `TimeoutError: locator.hover: Timeout 20000ms exceeded.` |

Same surface, same step, same failure point — the differing signatures are an artifact of how each spec spells its wait, not of a different problem:

- All four run the identical opening sequence: `blank-flow` → `sidebar-search-input.fill(<term>)` → wait for the filtered entry. Every failure lands on that wait; **not one of them gets as far as clicking the `+`**.
- The two `input_outputChat Input` waits (`limit-file-size-upload`, `notifications`) are on the **byte-identical locator** after the byte-identical `fill("chat input")`.
- The two `add-component-button-*` waits (`langflowShortcuts`, `loop-component-regression`) are the same locator shape one DOM level in, for Chat Output and Loop.
- It is **not** the already-known swallowed-add class (#1304 / #1335 / #1365, all closed). There the click was accepted and no node landed; `addComponentFromSidebar` repairs exactly that, by re-issuing the click. That repair cannot help here: the helper's own wait for the entry runs **before** it, which is why `langflowShortcuts` goes through the hardened helper and still fails.
- Three of the four do not use the helper at all — they hand-roll `fill` + wait — so a helper-side defect cannot be the common factor either. The common factor is the sidebar.

As a cluster it recurs on **three consecutive dailies**: 2026-08-18 (`langflowShortcuts`, `notifications`), 2026-08-19 (`loop-component-regression`, `limit-file-size-upload`) and 2026-08-20 (all four).

**This surfaced on a mass-failure day** — the guard tripped at 7 hard failures (threshold 5) and the run's in-run liveness recorder measured a real mid-run backend wedge (#1030) on shard 1: two outages spanning `08:50:43→08:54:17` UTC, 3.1 min of unreachable backend. The triage verdict on the day is that it was **mixed**, and this cluster is on the non-environmental side. The reason is measured, not argued:

- `loop-component-regression` failed at **08:41:46** and `limit-file-size-upload` at **08:43:58** UTC — **seven to nine minutes before the first outage opened**. Neither can be collateral of a wedge that had not started.
- `langflowShortcuts` (08:55:20) and `notifications` (08:55:29) ran **after** the second outage closed at 08:54:17.
- The cluster reproduces on 2026-08-18 and 2026-08-19, dailies that were not mass-failure days.

A plausible mechanism, recorded as a lead and **not** as a verdict: the sidebar is documented in-repo as re-rendering while its component catalog streams in (#537, cited in `tests/helpers/flows/add-component-from-sidebar.ts`). A filter applied against a catalog that has not finished arriving would yield no match and no retry — which matches "the entry never appears" better than it matches a slow render. That has to be confirmed or ruled out on the current nightly before it is treated as the cause.

## Investigation directive

1. Reproduce on the current nightly with `--retries=0`, at least across the four specs, and capture whether the sidebar entry appears **late** or **never** — a trace with the sidebar DOM at timeout tells the two apart, and they need different fixes.
2. Instrument the catalog load: does `GET /api/v1/all` (or whatever the sidebar consumes on this build) resolve before the `fill` runs? If the filter is applied against a partially-loaded catalog, the fix is a wait on the catalog, not a longer timeout.
3. Decide whether the repair belongs in `tests/helpers/flows/add-component-from-sidebar.ts` — extending it to re-fill the search when the entry does not appear, alongside the existing swallowed-click repair — or whether this is a product defect in the sidebar and belongs upstream. Three of the four call sites hand-roll the sequence, so if the fix is test-side, converging them onto the helper is part of it.
4. Rule the outage out or in per spec against the run's liveness table before attributing anything to the wedge.

## Deliverables (Done when)

- [ ] The four specs converge on one confirmed cause, or the cluster is explicitly split with evidence for the split.
- [ ] If the fix is test-side, the three hand-rolled call sites (`loop-component-regression`, `limit-file-size-upload`, `notifications`) are considered for `addComponentFromSidebar` so the repair applies once rather than four times.

rafaelgiln:
**Fixes #1518.** Closes #1518

## Problem

Four `@stable` specs died on the same three-step opening sequence — open a blank flow, fill `sidebar-search-input`, wait for the filtered component entry — over budgets of 10 s / 20 s / 20 s / 30 s and across four different components. None of them reached the click. Recurrent on three consecutive dailies (2026-08-18, 08-19, 08-20) and quarantined at triage of daily #1517.

| Spec | Waited for | CI signature (run 32349515682) |
|---|---|---|
| `ui-ux/langflowShortcuts.spec.ts` | `add-component-button-chat-output` | `locator.click: Timeout 20000ms` — **unexpected, 3 of 3 attempts** |
| `core-components/loop-component-regression.spec.ts` | `add-component-button-loop` | `page.waitForSelector: Timeout 10000ms` — flaky, passed on retry |
| `.../limit-file-size-upload.spec.ts` | `input_outputChat Input` | `expect(locator).toBeVisible() failed` — flaky, passed on retry |
| `ui-ux/notifications.spec.ts` | `input_outputChat Input` | `locator.hover: Timeout 20000ms` — flaky, passed on retry |

**Root cause — the search TERM is wiped, the sidebar is not slow.** The fill races the flow page mount and loses; the mount resets `sidebar-search-input` to `""`, nothing ever re-applies the filter, so no `add-component-button-*` exists and the caller burns its whole budget on a target that will never render. Measured on nightly `1.12.0.dev33` with an instrumented scout (fill → poll → capture → re-fill):

| Measurement | Result |
|---|---|
| Ungated fill (the helper's own shape) | **4 of 22 lose the term (~18 %)** |
| With a `waitForURL` + input-visible gate | **0 of 25** |
| Readback the instant `fill()` returns, on a miss | **`""`** |
| Sidebar at +12 s and at +25 s after a miss | **zero entries**, input still `""` |
| Identical re-fill | **repairs in ~320 ms (2 of 2)** |
| Entry on a fill that survives | 324–4191 ms, median ~335 ms |

So nothing arrives late and a longer caller timeout provably cannot help.

**Two rival readings refuted by measurement, not argument.** *Reparenting* — `GET /api/v1/all` on dev33 still lists Chat Input and Chat Output under `input_output` and Loop under `flow_controls`, none `legacy`, so every category-prefixed testid in the four specs is still correct. *The issue's own lead, a catalog still streaming (#537)* — the catalog request completes before the fill in **every** run, hit or miss, and the gated shape resolves the entry in ~335 ms; the list is ready, the input is not. It is also not the `flow-builder-welcome` overlay class (at miss time the input is **visible**, and `loop-component-regression` waits `state:"attached"` and still times out), nor the swallowed-add class of #1304/#1335/#1365 (the click is never issued at all).

**Per-row pre-fix baseline** (`--retries=0 --workers=1`, dev33): `langflowShortcuts` **1 of 5 (20 %)** with CI's byte-identical signature; the other three **0 of 6** each. The one that reproduces bare is the one spec with **no gate at all** before the fill — and 0 of 6 at the measured ~18 % has probability 0.30, so it is not evidence of immunity. Corroborating: the three flaky rows **passed on the very next attempt**, and a retry re-fills from scratch — that retry *is* this repair, paid at the price of a whole test.

**Outage ruled out per row** against the run's liveness table: `loop-component-regression` failed 08:41:46Z and `limit-file-size-upload` 08:43:58Z, seven to nine minutes **before** the first outage opened at 08:50:43Z; `langflowShortcuts` (08:55:20Z) and `notifications` (08:55:29Z) ran **after** the second closed at 08:54:17Z. Verdict for all four rows: **test-defect**.

## Fix

**At the source** — `tests/helpers/flows/add-component-from-sidebar.ts`. `issueAdd`'s bare `fill` becomes `fillSearchUntilEntryRenders`: fill, then poll (200 ms) for either the target entry **attached** (done) or the input value **diverging from the term** (a reset — re-type immediately, up to `SEARCH_FILL_ATTEMPTS = 3`). A deadline reached with the term **still held** is not this race: it throws `noMatchingEntryMessage`, naming the term, the entry count and the category-prefix trap (#1040), so an absent or reparented component is never silently re-typed three times. `SEARCH_ENTRY_TIMEOUT_MS = 10000` is the measured budget, not a guess. Neither message is infra-classifiable (#1262), so a real sidebar regression stays eligible for `@stable` auto-removal.

**Rejected alternatives, explicitly.** *A longer timeout* — refuted by measurement: the sidebar still held zero entries 25 s after a lost fill. *Another readiness gate* — three of the four specs already had one and all three still failed in CI, because each measures the wrong observable: the creation `POST 201` is a network event and not a mount, `waitForURL` flips before React finishes, and the input is **visible throughout** the mount, which is exactly when it is still resettable. The only reliable observable is the term surviving in the input, which is what the repair reads.

**Convergence.** The three hand-rolled call sites now call `addComponentFromSidebar`, because the wiped observable is the **one** shared input and the repair belongs in one place. The hovers they drop (`flow_controlsLoop`, `input_output_chat input_draggable`, the `input_outputChat Input` row) are not needed — the `+` button measured visible with no hover on 25 of 25 scout entries.

**Same race class, also fixed.** `langflowShortcuts` captured its flow id in a fire-and-forget `page.on("response")` + `resp.json()` handler that lost to `afterEach`: on the failure path the test threw before the body resolved and the run leaked a `New Flow` on the shared instance (one observed and purged while cleaning up after this issue's own repro runs). It is now an awaited `waitForResponse`, as the other three specs already did.

**Quarantine lifted** on all four: `test.fixme` removed and `@stable` restored. The second `test.fixme` in `loop-component-regression.spec.ts` belongs to **#1234** (the ArXiv component is absent from the image) and is deliberately untouched.

## Scope note

No assertion was weakened, no selector loosened, no wait added, no scope narrowed. Every existing assertion in the six runnable tests is unchanged — the diff removes the hand-rolled *add* sequences and the quarantine, and adds the repair plus its unit tests. The `waitForSelector('[data-testid="title-Loop"]')` and the callers' own `toHaveCount` assertions stay exactly as they were. `SEARCH_ENTRY_TIMEOUT_MS` is a new budget on a new gate, not a raised one on an existing wait: the caller's own budgets are untouched.

## Validation (nightly `1.12.0.dev33`, `--retries=0`)

- `npm run typecheck` ✅ (0 errors) · `npm run lint` ✅ (0 errors, 1424 pre-existing warnings) · anti-pattern checklist ✅
- **Burst: 12 of 12 clean** — 3 runs per spec at `--retries=0 --workers=1`, `backendErrors=false` on every run, zero `🚨 Backend Error` ✅
- **Force-fail: 6 of 6 runnable tests**, each verified red, reverted, with a final green run ✅
  - `langflowShortcuts` · `expectNodes(2)` → `expectNodes(7)` after Copy+Paste
  - `notifications` · `notification_button` → `notification_button_ff`
  - `limit-file-size-upload` · `ceilingKb` with `*4096` instead of `*1024`
  - `loop` #1 · handle `done-right` → `done-left`
  - `loop` #2 · `text=Flow build failed` → `text=Flow build succeeded`
  - `loop` #4 · `expect(count).toBe(3)` → `toBe(9)` in the N=3 step
- **Unit lane: 720 of 720** (`npm run test:units`), including 7 new tests written test-first for `classifySearchFill`, `searchResetMessage` and `noMatchingEntryMessage` — watched failing on `TS2305` for the four missing exports before the implementation existed ✅
- QA-CHECKLIST coverage guard ✅ (194 `@stable` spec files, every one referenced by a Part II bullet)
- Instance left clean: 27 flows, **0 `New Flow` orphans** ✅
- `--trace=on` not run: it hangs UI specs locally under Colima (known limitation); the burst ran without it.

## Dependencies

None — pure UI. No provider, no API key: none of the six tests makes a model call. Parallel-safe, except `loop-component-regression.spec.ts`, which is `test.describe.configure({ mode: "serial" })` as it already was.


rafaelgiln:
## CI: the one red is ambient — it belongs to #1460, not to this PR

`Run impacted E2E specs` finished **65 passed, 3 skipped, 1 failed**. The single failure is

```
tests/tests-automations/regression/core-functionality/llm-agents/modelInputComponent.spec.ts:140
  › ModelInputComponent › the trigger shows the model the user selects
  @release @components @workspace @model-provider

Error: expect(locator).toHaveText(expected) failed
  Locator: getByTestId('model_model')
  - Expected  - 1
  + Received  + 0
    claude-opus-5
  - 1 of 59
```

3 of 3 attempts, same signature.

### It is the mechanism #1460 already documents, at the same line

#1460 — *"modelInputComponent + memory-base-registration read a model option's own text and break on the sr-only `N of M` counter"* — reports this exact failure at this exact line, from daily run 31786538844 (2026-08-14):

> `modelInputComponent` reads `firstOption.innerText()` and asserts the trigger equals it: expected `claude-opus-5\n1 of 90`, received `claude-opus-5`. **The trigger is the half that is correct** — the option is what gained the extra line.

The only difference here is the catalog size (`1 of 59` vs `1 of 90`). Since 1.12.0.dev26 every model option renders a position counter inside itself:

```html
<div data-testid="Anthropic-claude-opus-5-option" role="option">
  <div class="truncate text-[13px]">claude-opus-5</div>
  <span class="sr-only">1 of 59</span>
</div>
```

The span is `sr-only` — invisible to a user, but part of `textContent`. So the value the test reads from the option is polluted and the **expected** string carries an extra line; the trigger renders correctly.

### Evidence that this PR did not cause it

- The offending line is **untouched on `origin/main`**: `modelInputComponent.spec.ts:202` still reads `const modelLabel = (await firstOption.innerText()).trim();`. This PR's diff touches **0 lines** of that spec.
- `@stable` was **already auto-removed** from this test by `f6f4c39` for daily `31786538844` — a known hard failure on `main` since 2026-08-14, six days before this PR.
- `git log f6f4c39..origin/main -- <that spec>` is **empty**: no fix ever landed.
- This PR changes the **sidebar add** path (`add-component-from-sidebar.ts`). It cannot alter the `textContent` of a model-picker option. The trace confirms the add itself worked — the Language Model node exists (`node-LanguageModelComponent-UtX73-field-model-label`) and the failure is two interactions later, after the picker was opened and an option clicked.
- The test is **not `@stable`**, so the daily never runs it. It reaches this PR only because the impacted-specs lane selects by **import graph, not by tag** (#871/#1054), and the helper this PR fixes is imported by **68 specs** — the #1423 cost, one issue over.

### Two things worth flagging beyond this PR

1. **#1460 was closed today (11:55Z) as COMPLETED with 0 closing PRs, and the fix never landed.** Both of its call sites are still on the broken read: `modelInputComponent.spec.ts:202` (`innerText()`) and `memory-base-registration.spec.ts:261` (`getByRole("option", { name, exact: true })`). #1504 still lists both as *"#1460 (open)"*.
2. **The remedy already exists and is unused.** `tests/helpers/provider-setup/model-option.ts` exposes `visibleLabel`, which strips `sr-only` and badge nodes inside the page — built for #1459, the sibling of the same triage. Routing these two call sites through it is the whole fix.

Commented on #1460 to that effect. Deliberately **not** fixed here: it is another issue's file, it would enter this PR without having gone through its FORCE_FAIL phase (already closed), and validating it needs a provider-configured instance.

The 20 specs this lane selected exercise the changed add path and **65 of 66 executed tests pass**, including all four specs this PR un-quarantines. The failing test does not touch the sidebar add: it fails on an assertion whose *expected* value is read from a DOM node that gained an `sr-only` child two nightlies ago. Merging changes nothing about it — it was red on `main` before this branch existed and will stay red until #1460's fix lands.

One caveat this PR does not hide: the lane **capped at 20 of 68 impacted specs — 48 were not run** (annotated on the run). The narrow risk is a call site whose target only attaches after a hover; that was measured and refuted — the `+` button was attached and visible with no hover on 25 of 25 scout entries, and `entryPresent` is evaluated before the input value, so an app that normalises the term cannot trigger a spurious re-fill once the entry has rendered. A `manual.yml` dispatch with `test_tag=@stable` is the honest full-coverage check for a helper of this reach.

Issue #1546, opened 2026-08-21
[Daily #1544] POST /api/v1/flows/download/ drops the credential variable binding — SecretStrInput comes back load_from_db true with value null (recurrent 2×)


Victor-w-Madeira:
Spun out of daily-failure triage #1544 (run [32464002123](https://github.com/oriontech-me/langflow-e2e/actions/runs/32464002123), 2026-08-21). Filed on a guard-tripped day because the cluster is cross-day recurrent and sits outside every measured outage window.

**Upstream:** _not filed_

## Symptom

A `@stable @api @regression` test hard-failed on all three attempts with an identical assertion diff. The test exports a flow the way the UI does and asserts that a credential-typed global variable reaches the export **as a binding** (the variable name) and **never** as the secret value. The secret half passes — `expect(body).not.toContain(sentinel)` never fired. What fails is the binding half: the exported `SecretStrInput` field comes back as

```json
"secret_token": { "_input_type": "SecretStrInput", "load_from_db": true, "password": true, "type": "str", "value": null }
```

so the field is still marked as DB-bound (`load_from_db: true`) while the name of the variable it is bound to is absent from the whole payload. Both `POST /api/v1/flows/download/` and its sibling surface in the same step were checked; the failing label is the download endpoint.

**This is a data-fidelity finding on export, not a secret leak** — worth stating up front because the spec's file name reads the other way.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/security/credential-secret-exposure.spec.ts:299` ("the exported flow carries the credential binding, never the secret") | `POST /api/v1/flows/download/ response body contains the variable name` | `Error: POST /api/v1/flows/download/ lost the variable binding` |

## Why these failures are one cause

One test, one signature — the argument here is about **why it is filed at all on a mass-failure day**, and why it is not read as wedge collateral.

It ran on shard 2, which did carry the wedge, but the three attempts ran at **09:07:05, 09:07:08 and 09:07:12** UTC and each took **under one second**. Shard 2's last measured outage window closed at **09:02:56** — more than four minutes earlier. The failure is also not transport-shaped: every attempt received a complete, well-formed flow JSON and failed a `toContain` on its contents, so the backend answered and answered fully.

[...]

PR #1551, opened 2026-08-21
test(security): quarantine the credential-export binding test for #1546

Victor-w-Madeira:
Quarantines one `@stable` test at triage of daily #1544, as prevention. Not a fix — the investigation is #1546.

## What was quarantined

`tests/tests-automations/regression/security/credential-secret-exposure.spec.ts:299` — *the exported flow carries the credential binding, never the secret*

Both edits, together: `@stable` removed **and** `test.fixme` added. Tag removal alone only stops the daily; the test would keep going red on the `pr-validation.yml` impacted-specs gate, which selects by file diff rather than by tag (#871).

## Why

Hard failure on all three attempts of run [32464002123](https://github.com/oriontech-me/langflow-e2e/actions/runs/32464002123) (2026-08-21), same signature as 2026-08-20:

```
Error: POST /api/v1/flows/download/ lost the variable binding
```

The exported `SecretStrInput` field returns `load_from_db: true` with `value: null` — DB-bound, but the variable name is gone from the payload. The secret is **not** leaked (`expect(body).not.toContain(sentinel)` never fired).

The 2026-08-21 daily tripped the mass-failure guard, so the workflow auto-removed nothing. This test is on the non-environmental side of the day's mixed verdict: it ran on shard 2, which did carry the wedge, but its three attempts ran at 09:07:05 / 09:07:08 / 09:07:12 UTC — over four minutes after shard 2's last outage window closed (09:02:56) — each in under a second, each receiving a complete flow JSON and failing a `toContain` on its contents.

## Not closing #1546

Lifting the quarantine (remove `test.fixme` + restore `@stable`) is a deliverable **of** #1546, so this PR deliberately carries no auto-close keyword.

Refs #1546, #1544

### DIFF
--- a/tests/tests-automations/regression/security/credential-secret-exposure.spec.ts
+++ b/tests/tests-automations/regression/security/credential-secret-exposure.spec.ts
@@ -296,9 +296,16 @@ test.describe("Credential secret exposure", () => {
     },
   );
 
-  test(
+  // Quarantined at triage (daily #1544): hard failure on all three attempts —
+  // `POST /api/v1/flows/download/` returns the SecretStrInput field as
+  // `load_from_db: true` with `value: null`, so the variable name the binding
+  // points at is absent from the export (the secret itself is not leaked). Same
+  // signature on the 2026-08-20 and 08-21 dailies, both outside every measured
+  // outage window. Lifting the quarantine (remove test.fixme + restore @stable)
+  // is a deliverable of #1546.
+  test.fixme(
     "the exported flow carries the credential binding, never the secret",
-    { tag: ["@stable", "@api", "@regression"] },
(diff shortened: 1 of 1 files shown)


Victor-w-Madeira:
Superseded by #1554, which landed the same quarantine on `main` (77c2e32): identical `test.fixme` + `@stable` removal, with a stronger comment — it names the upstream cause (langflow-ai/langflow#14639, fc3810da0, merged 2026-08-19 into release-1.12.0), the mechanism (the metadata-driven scrubber applied without its `variable_references` mode), the 5/5 reproduction on 1.12.0.dev33, and the bug doc under `docs/upstream-bugs/`. Nothing here to merge on top of it; #1546 remains the owner of lifting the quarantine.

Issue #1567, opened 2026-08-24
Cover the #14538 tweak surfaces left out of #1566 — LANGFLOW_TWEAKS_POLICY, the api_editable allowlist, and the graph-path floor


Victor-w-Madeira:
Follow-up to #1566, which adapts `security/tweaks-injection.spec.ts` to the loud-refusal contract but is deliberately scoped to the **protected-field floor** on `POST /api/v1/run/{flow_id}`. `langflow-ai/langflow#14538` (`4f3b9f3772`, 2026-08-21, on `release-1.12.0`) shipped three further surfaces that no spec in this suite touches.

## What is uncovered

**1. `LANGFLOW_TWEAKS_POLICY` — a new operator-facing deployment policy.** Three modes, resolved by `_resolve_tweak_policy()` in `lfx/processing/process.py`:

| Mode | Behaviour | Refusal reason returned |
|---|---|---|
| `permissive` (default, what every lane runs) | today's behaviour — only the protected floor refuses | `The field is protected and keeps the value set by the flow author.` |
| `declared` | honours the per-flow allowlist the author sets via `api_editable`, and **only** on flows where at least one field is toggled | `This flow declares which fields the API may set. Only fields marked editable via API accept a tweak.` |
| `off` | refuses every tweak **and** every component-targeted input | `This deployment does not accept tweaks.` / `This deployment does not accept component-targeted inputs.` |

An unknown value falls back to `permissive` with a `logger.warning` — a silent widening if an operator typos the setting, and a candidate assertion of its own.

**2. The per-flow `api_editable` allowlist** (`flow_declares_api_editable()`), which is what `declared` reads. Upstream states `declared` **cannot** expose a `code` field, i.e. the floor still applies above the allowlist — that is the security-relevant claim to pin.

**3. The graph-level run path.** Streaming and background run modes build the graph before applying tweaks, so they went through `process_tweaks_on_graph` / `apply_tweaks_on_vertex` rather than `process_tweaks`. Upstream's own commit message says that path *"filtered only the literal key `code` in `api/build.py`, so it accepted tweaks the sync mode refused: a `global_imports` override widening the exec sandbox was applied."* It is now claimed to enforce the same floor. **This is the highest-value half of this issue** — it is a real bypass that existed, and #1566 covers only the sync endpoint, so nothing in the suite would catch it returning.

[...]

Issue #1298, opened 2026-08-05
20 spec-doc dependency paths do not resolve upstream, and nothing catches a path that does not exist


Victor-w-Madeira:
Found while measuring #1040's hardcoded-path claim, and split out because it is a docs sweep plus a guard, not part of the drift detector.

## The defect

`scripts/validate-spec-deps.ts` checks that a spec doc's mandatory **External dependencies** section exists and is populated. It never checks that the paths in it **resolve**. So a path that does not exist is silent — the same failure mode as #1092, where a monitored path that had never existed on any ref made `git log --since -- <path>` print nothing, indistinguishable from "nothing changed here".

Measured on the upstream clone against `origin/release-1.10.0`, `1.11.2`, `1.12.0` and `main`:

| Class | Count | Why it does not resolve |
|---|---|---|
| `src/backend/base/langflow/components/…` | **18 occurrences / 16 files** | The pre-`lfx` tree, absent since *feat: introduce lfx package* (#9133, 2025-09-02) |
| `src/lfx/components/…` | **2 files** | Missing the real `src/lfx/src/lfx/` prefix — right family, wrong path |

Examples of the first class: `docs/core-components/tool-mode.md`, `docs/core-components/webhook-component-regression.md`, `docs/core-functionality/llm-agents/agent-max-iterations.md`, `docs/flow-functionality/general-bugs-shard-3836.md`, `README.md`. The second: `docs/core-functionality/llm-agents/agent-max-tokens.md`, `agent-n-messages-limit.md`.

A 21st of the same class was already fixed under #1040 (`docs/core-components/agent-component-regression.md` recorded `components/agents/agent.py`, which existed from #9133 until #10180 on 2025-11-12 — stale, not fictional).

## Why it matters

The **External dependencies** section is what a reviewer follows to check a spec against the product, and what a future reader greps when triaging. A path that silently resolves to nothing turns that into a dead end, and the reader has no signal distinguishing "this moved" from "this was never right".

## Scope

- [ ] Correct the 20 paths (both classes) to the current spelling. Mechanical; verify each against a real ref, not by pattern.
- [ ] Add the guard. It should **extend `scripts/watch-upstream-areas.mjs --mode=check`**, which already validates the 13-area path table against a real ref for exactly this reason (#1092), rather than duplicating that machinery in `validate-spec-deps.ts`.
[...]

PR #1570, opened 2026-08-24
test(security): assert the loud tweak refusal and restore @stable (#1566)

### LINKED ISSUE #1566: [Daily #1565] security/tweaks-injection — POST /api/v1/run answers 422 where the spec expects 200 when a tweak targets a protected/executable field (2 tests)
Spun out of daily-failure triage #1565 (run [32707807321](https://github.com/oriontech-me/langflow-e2e/actions/runs/32707807321), 2026-08-24). Both tests had `@stable` auto-removed by the daily workflow (commit `12bfd89` on `main`); restoring it is a deliverable of this issue.

**Upstream:** _not filed_

## Symptom

Two `@stable` API tests hard-failed with the same shape — **`POST /api/v1/run/{flow_id}` answers `422` where the spec asserts `200`** — at the point where the request carries a tweak on a **protected/executable** template field (`code` on the Chat Input; `python_code` + `global_imports` on the Python Interpreter). The third test in the file, which sends a **benign-only** tweak (`input_value`), passed.

|---|---|---|
| `tests-automations/regression/security/tweaks-injection.spec.ts:156` ("a code tweak cannot replace a component's implementation") | `expect(res.status()).toBe(200) — spec line 173, tweak { code }` | `Error: expect(received).toBe(expected) // Object.is equality` |
| `tests-automations/regression/security/tweaks-injection.spec.ts:246` ("an executable field on a code-execution component is refused while the same request's benign tweak lands") | `expect(res.status()).toBe(200) — spec line 272, tweaks { python_code, global_imports } + benign sender_name` | `Error: expect(received).toBe(expected) // Object.is equality` |

Same spec file, same endpoint (`POST /api/v1/run/{flow_id}`), same normalized signature (`Error: expect(received).toBe(expected) // Object.is equality`, Expected `200` / Received `422`), and both fail at the **first status assertion of the request that carries a protected field** — before any security assertion in the test runs. The sibling test in the same file (`the refusal is field-scoped: an unprotected field on the same node still applies`, line 213) posts to the same endpoint with a **benign-only** tweak and passed, so the shared dimension is not the endpoint, the flow, the API key or the request plumbing: it is the presence of a protected/executable field in the tweak payload. One cause, two tests.

Not a mass-failure day — the guard did not trip (2 hard failures, 1 infra-excluded flake, 526 passed on the whole run), so this does not read as environmental collateral. Both are **first occurrences** in the 30-day window (`recurrence.count: 1`, no prior same-signature hit on either test).

The spec was written against an explicitly **silent** backend refusal, and says so in its header comment: *"The backend refusal is SILENT by design: apply_tweaks() asks is_protected_tweak_field() and, on a protected field, logs a warning and skips it — the run still answers 200 with the flow author's value."* The whole paired-benign-control design of tests 1 and 3 exists because there was no error to assert on. A `422` is therefore the shape a change **from silent-skip to loud-refusal** would produce, and there is a named upstream candidate for exactly that: `langflow-ai/langflow` PR **#14538** (`4f3b9f3772`, 2026-08-21, on `release-1.12.0` — the line the nightly is cut from), *"feat: add deployment tweak policy and close the graph-path tweak bypass"*, whose commit message states *"Refused tweaks now raise instead of logging a warning and returning 200 … The API maps that to 422 naming the refused keys"* and *"Decide first, mutate second. A refusal must leave the payload untouched."* That is a **lead read off the upstream tree, not a verdict measured here** — it has not been confirmed against the running nightly, and confirming it is the first item of the directive below.

## Investigation directive

Investigate independently, **product first**, against the current nightly:

1. **Confirm the observed contract.** Post the two failing requests by hand and record the actual status **and body**. If the refusal is now loud, the body should be `{"detail": {"error": "Refused tweaks", "code": "TWEAKS_REFUSED", "message": <reason>, "fields": [<refused keys>]}}` — capture it verbatim rather than inferring it from the upstream source.
2. **Rule out a security regression before adapting anything.** The security property under test is that a tweak cannot reach a template field as executable input. Prove it still holds on the new path: with a refusal, nothing runs (so the sentinel cannot appear); confirm no mode still applies the field — in particular the graph-level path (streaming / background run) that `langflow-ai/langflow#14538` says previously accepted a `global_imports` override the sync path refused.
3. **Rule out request-shape drift.** Only after 1–2, check whether the 422 is instead a payload-validation change unrelated to the protected-field floor (e.g. the `tweaks` schema), which would point at the spec's request shape rather than at the refusal contract.
4. **Then decide the spec's shape.** If the loud refusal is confirmed as the new contract: assert the 422 plus `code: "TWEAKS_REFUSED"` and the refused `fields` (a stronger observable than sentinel-absence), and note that test 3's premise — *"an executable field … is refused **while the same request's benign tweak lands**"* — becomes impossible if a refusal aborts the whole request, so the benign control has to move out of that request (the passing test at line 213 already is that control). Record whether `LANGFLOW_TWEAKS_POLICY` (`permissive` default / `declared` / `off`) and the per-flow `api_editable` allowlist affect what the suite asserts, and whether they deserve their own coverage issue.
5. Reproduce on 3 environments before restoring `@stable`, per `CONTRIBUTING.md`.

## Deliverables (Done when)

- [ ] The **actual 422 body** from the current nightly recorded here verbatim (status, `code`, `fields`), not inferred from upstream source.
- [ ] A verdict recorded per `references/triage-verdicts.md`: product behaviour change (contract) vs. product regression vs. spec drift — and, if a behaviour change, that the **security property still holds** on every run mode, including the graph-level (streaming/background) path.
- [ ] `@stable` restored on both tests (auto-removed at `12bfd89`) and the spec doc under `docs/` updated — the header comment asserting the refusal is *silent by design*, the **Validation criterion**, and `Last validated`.
- [ ] If the new surface (`LANGFLOW_TWEAKS_POLICY` `declared`/`off`, the `api_editable` allowlist, graph-path enforcement) is worth covering, a separate roadmap issue is filed rather than folded in here.

Victor-w-Madeira:
**Closes #1566.**

## Problem

Two `@stable` tests in `security/tweaks-injection.spec.ts` hard-failed on the 2026-08-24 daily (run [32707807321](https://github.com/oriontech-me/langflow-e2e/actions/runs/32707807321), triage umbrella #1565), both with `Error: expect(received).toBe(expected)` — expected `200`, received `422` — at the first status assertion of the request carrying a protected field. `@stable` was auto-removed by the workflow at `12bfd89`. First occurrence for both; guard not tripped (526 passed / 2 hard failures / 1 infra-excluded flake), so not environmental collateral.

**Verdict: the product changed intentionally** (verdict 3 of `triage-verdicts.md`) — not a regression. `langflow-ai/langflow` PR 14538 (`4f3b9f3772`, 2026-08-21, on `release-1.12.0`, the line the nightly is cut from) replaced the silent refusal with a loud one, and the spec was written explicitly against the silent contract — its header comment said so.

Measured on `1.12.0.dev37`, not inferred from the upstream tree:

| Request | Result |
|---|---|
| `tweaks: { <chatInput>: { code } }`, by node id **or** display name | `422`, `fields: ["code"]` |
| `tweaks: { <python>: { python_code, global_imports }, <chatOutput>: { sender_name } }` | `422`, `fields: ["global_imports","python_code"]` — `sender_name` **not** in `fields` and **not** applied |
| `{ python_code }` / `{ global_imports }` alone | `422`, `fields: ["python_code"]` / `["global_imports"]` |
| `{ input_value }` / `{ sender_name }` alone | `200`, applied |
| a field the node's template does not declare | `200` — **skipped, not refused** |

Body shape: `{"detail":{"error":"Refused tweaks","code":"TWEAKS_REFUSED","message":<reason>,"fields":[<sorted keys>]}}`.

## Fix

The refusal is now **request-scoped**, so the file's original design — pairing every refusal with a benign tweak **inside the same request**, the only control available when there was no error to assert on — is impossible: a refusal runs nothing at all. The control moved to a **second request**, and the assertions got stronger on both halves.

- **Test 1** (`a code tweak is refused with a 422 naming the field, and the flow is left untouched`) asserts the refusal itself — `422`, `detail.error`, `detail.code === "TWEAKS_REFUSED"`, `detail.fields` exactly `["code"]` — for **both** addressing modes `apply_tweaks` accepts, since a refusal firing for only one of them is the bypass shape. New step: a **tweak-free run** answering `200` with the author's `"Hello"`, which is the direct assertion of `process_tweaks()`'s *"decide first, mutate second"* and doubles as the old in-request control.
- **Test 3** had to be renamed — its premise (`… while the same request's benign tweak lands`) is now impossible by design. It asserts that **both** protected keys are named (`toEqual`, not `toContain`: a refusal reporting only the first is the defect to catch) and that `sender_name` is **absent** from `fields`; the author's code, the unwidened sandbox log and the benign field being tweakable are then read off **one** `200` response from the second request.
- **Test 2** (untouched by the failure, still green) gains the **ignored-vs-refused** boundary the change introduced: a field the template does not declare answers `200`, so a caller cannot read every `200` as "my tweak applied".

**Rejected explicitly:** asserting `detail.message`. `_refusal_reason()` returns a different string per `LANGFLOW_TWEAKS_POLICY` (`permissive` — the product default and what every lane runs — vs `declared` vs `off`), so pinning it would fail a correctly-behaving instance that is merely configured differently. `code` and `fields` are stable across all three.

No escape hatch was added for the deliberate `422`s: the fixture's HTTP-error monitor attaches to the `page` fixture's responses and this file requests only `request`, so no `page` is instantiated and these refusals never reach the advisory log. Measured — the runs below printed no backend-error line.

## Scope note

- The security property under test is unchanged and still asserted: a tweak cannot reach a template field as executable input. What changed is *how it is observed* — the refusal directly, instead of the absence of a sentinel.
- `@stable` restored on Tests 1 and 3 only; Test 2 never lost it. The `QA-CHECKLIST.md` §17.4 bullet stays `[x]` — **no checklist edit, and `coverage:summary` was not run**.
- **Out of scope, tracked in #1567:** `LANGFLOW_TWEAKS_POLICY`'s `declared` / `off` modes, the per-flow `api_editable` allowlist, and the graph-path floor (`apply_tweaks_on_vertex`) — the streaming/background path that PR 14538 reports previously accepted a `global_imports` override the sync path refused. The policy is instance-global, which would make such a spec `@destructive` and unable to carry `@stable` (#1010).
- The spec doc records the measured contract, the three-test redesign, and the policy caveat under *Preconditions*.

## Validation (nightly `1.12.0.dev37`, `--retries=0`)

- `npm run typecheck` ✅ · `npm run lint` ✅ (0 errors, no finding in the touched file) · `npm run test:units` ✅ 742/742
- QA-CHECKLIST guard ✅ (no generated block edited) · `npm run check:checklist-coverage` ✅
- **5 clean runs**, no flake ✅ — 2.3s / 1.3s / 1.3s / 1.1s / 1.1s, 3 passed each
- `--trace=on` ✅ 1.6s (no hang; API-only spec, so the trace carries the HTTP calls and there are no screenshots to compare)
- zero `🚨 Backend Error` ✅
- Teardown ✅ — after ~12 runs, `GET /api/v1/flows/` returns **0** flows and the only API key left is Langflow's own MCP starter key
- **Force-fail: 8 mutations, each applied in isolation (`--grep`), observed failing, reverted** (`grep FFMARK` → 0, final run green):

| # | Mutation | Observed failure |
|---|---|---|
| FF1 | T1: tweak `code` → `input_value` (refusal stops firing) | expected `422`, received `200` |
| FF2 | T2: benign `input_value` → `code` | expected `200`, received `422` |
| FF3 | T2: undeclared field → `code` | expected `200`, received `422` |
| FF4 | T3: `python_code` dropped from the combined request | `fields` `toEqual` diff `-1/+0` — proves a refusal naming only one key fails |
| FF5 | T3: expected log → `['math', 'os']` | received `["Successfully imported modules: ['math']", …]` |
| FF6 | T3: 2nd request's `sender_name` → fixed literal | expected `BENIGN2-…`, received `FFMARK6` |
| FF7 | T3: expected author result → `AUTHOR-…-FFMARK7` | received `AUTHOR-…` |
| FF8 | helper: expected `code` → `FFMARK8` | received `TWEAKS_REFUSED` |

Checklist step 3 (`--debug` walk) is the reviewer's:

```bash
PLAYWRIGHT_BASE_URL=http://localhost:7870 \
  npx playwright test tests/tests-automations/regression/security/tweaks-injection.spec.ts \
  --workers=1 --debug
```


Victor-w-Madeira:
## Review round — two reviewers, one substantive change applied

Reviewed twice: the harness reviewer (inherits this session's context) and an **independent reviewer with clean context**, given the live nightly at `:7870` and the upstream clone, and instructed to *refute* this PR's claims. Neither could refute the asserted contract — the independent one re-derived the 422 shape, the `sorted(set(refused))` ordering, decide-before-mutate and the ignored-vs-refused boundary from `origin/release-1.12.0` and from 9 hand-built live probes, and ran the spec at 5 workers with `--repeat-each=3` (9/9).

One finding was substantive, and it was proved by execution rather than argument.

### Test 3 was green in a world where the refused tweak landed AND persisted

The second request re-tweaked `sender_name` — the very field whose *non-landing* the test title claims — and Test 3 had no stored-flow check at all. The independent reviewer ran the test body verbatim with a `PATCH` in between that persists the refused `sender_name` into the stored flow, i.e. a revert to apply-as-you-go + raise-at-the-end. **The test passed.**

That is the exact hazard upstream documents in `_refused_tweak_names()`:

> *"Refusal has to be decided for the whole request before anything is applied. Applying as we go and raising at the end leaves the accepted half written, and the graph the run paths hand us is cached and reused, so that half survives into later runs that send no tweaks at all."*

**Fixed in `644a21b`.** Test 3 is now four steps in a deliberately load-bearing order, commented as such:

1. the combined request → `422`, `fields` exactly `["global_imports","python_code"]`;
2. a request with **no tweaks at all** → the author's code ran, the import log is `['math']`, and `sender_name` is **not** the refused value (compared against the refused value, not the author's literal `"AI"`, so a template-default change cannot satisfy it silently);
3. **new** — `GET /api/v1/flows/{flowB}`: the persisted `python_code`, `global_imports` and `sender_name` are all still the author's, with every node lookup asserted present before its value is compared;
4. only **then** the benign re-tweak, as the tweakability control.

Re-verified with the reviewer's own mutation, as **FF9**: with the accepted half persisted by hand, the test now **fails** — `Expected: not "BENIGN-…"` in step 2. The hole is closed, measured the same way it was found.

### Also applied

| Finding | Change |
|---|---|
| Three assertions could not fail — the sentinel on a `422` body (twice; once the status holds, that body is handler constants plus refused *key names*, never a fresh `PWNED-<ts>` value) and a `not.toContain("sender_name")` already implied by the `toEqual` beside it | removed; the doc's Validation criterion stops counting them as evidence |
| Test 1's stored-flow check was vacuous on a lookup miss (`?? ""` → `JSON.stringify("")` contains no sentinel) | node and field asserted **present** first. **FF11**: pointing the lookup at a nonexistent id now fails with `node ChatInput-b6UCc not found on the stored flow`, where before it passed |
| The `declared`-policy claim was wrong in the safe-sounding direction | `is_tweak_refused_by_policy()` only bites under `declared` when the flow declares an `api_editable` field, and neither fixture flow does — so `declared` is byte-identical to `permissive` here. Corrected in the doc and the spec header; the decision not to pin `detail.message` now rests on `off` alone, which is a real divergence |
| Undocumented environment precondition | `LANGFLOW_SANDBOX_BACKEND` unset/`none` is now a Precondition: the asserted `Successfully imported modules: ['math']` log is emitted only on the in-process path, so a sandbox backend fails Test 3 on a **correct** instance. Also recorded that `['math']` is the catalog default for `global_imports` |
| Test 1 step 4 was mis-attributed | it does not assert "decide first, mutate second" — a tweak-free run never reaches `process_tweaks` (`simple_run_flow` skips it and may serve a pre-built graph). Reframed as the endpoint-health + stored-flow-intact control it actually is |

### One finding rejected, with evidence

The harness reviewer claimed the doc's justification for keeping `lfx/exceptions/tweaks.py` out of the dependency list was false — that no path-resolution guard exists, since `scripts/validate-spec-deps.ts` always exits 0 and `CLAUDE.md` says the section is only checked for being populated. It looked at the wrong script and at a stale doc. The guard that failed this PR is `pr-validation.yml` → *Resolve every External-dependencies path against upstream* → `node scripts/watch-upstream-areas.mjs --mode=check-docs --ref origin/main`, and it failed with `docs/security/tweaks-injection.md:144 — dependency path ... does not exist upstream ... (issue #1298)`; it passes after `392a9b8`. `CLAUDE.md`'s claim that paths are never resolved is out of date — worth correcting separately, since it has now misled a reviewer.

Its accompanying point **was** right and is applied: the note conflated `main.py`'s *path* resolving on both refs with the *handler* existing on `main`. It does not.

### Scope update — #1567 is closed

The independent reviewer caught that #1567 was closed as completed by **PR #1571** (merged 2026-08-24 14:58), which adds `security/tweaks-graph-path-floor.spec.ts` covering the graph run path. The body above still reads as if that were open work. So the gap this spec alone would leave — passing while the graph path applied a protected tweak — **is covered on `main` now**, and the doc points at the sibling instead of "tracked separately".

### Revalidation after the changes

`1.12.0.dev37` · **5/5** clean at `--retries=0` (1.1–1.3 s) · **9/9** at 5 workers with `--repeat-each=3` · typecheck ✅ · eslint ✅ (no finding in the touched files) · dependency-path guard ✅ (491 paths / 259 docs, none failing) · zero `🚨 Backend Error` · force-fails **11** total, all executed and reverted (`grep FFMARK` → 0), including FF9 and FF11 above and re-verification of FF1 and FF4 after the edits.

One structural note carried, not changed: Test 1 cannot by itself distinguish "protected tweaks are refused" from "**all** tweaks are refused" — that discrimination lives in Test 2, and `@stable` is removed per test, so a day that de-stabilises Test 2 alone would leave Test 1 green against a wholly broken tweaks mechanism. Pre-existing and identical to the arrangement before this PR; recorded here rather than folded in.

PR #1605, opened 2026-08-25
fix: resolve echo base URL with || so an empty env var falls back

Victor-w-Madeira:
Restores the fix that came from PR #2 on the internal mirror (`github.ibm.com:Langflow/e2e-qa`), cherry-picked onto the current `main`. Original authorship preserved: Eric Hare.

## Context

PR #2 was merged directly into the destination repository. The destination is a read-only mirror (see `NOTES.md`): the sync force-pushes `+refs/heads/*` with `--prune`, so the 2026-08-25 sync overwrote that merge (`6a732121` → `c9aa6f00`) and the fix was lost there.

The content was recovered from `refs/pull/2/head` (`3cabd19b`), which the push does not touch. This PR reintroduces the change through the origin repository, which is the correct entry point — the sync propagates it to the mirror from here.

## The change

The six specs that resolve the echo endpoint used `??`, which only falls through on `null`/`undefined`. `.env.example` ships `HTTPBIN_BASE_URL=` and the setup instructions say to copy it, so dotenv sets it to `""` — a value `??` accepts. Every fresh clone therefore resolved an empty base URL.

The consequence is worst in `api-request-component-regression.spec.ts`, where `new URL(ECHO_BASE)` runs at module load: it throws `TypeError: Invalid URL` during collection, which aborts the whole run (`npx playwright test --list` reports 0 tests in 0 files). The agent specs fail more quietly, with the target becoming a relative `/json` or `/uuid`.

Swaps `??` for `||` across the 6 files. The diff was verified to be identical to the original commit — only the surrounding context changed from the rebase.

### DIFF
--- a/tests/helpers/other/private-echo-endpoint.ts
+++ b/tests/helpers/other/private-echo-endpoint.ts
@@ -66,7 +66,7 @@ export type PrivateEchoEndpoint = { url: string } | { skipReason: string };
  * The returned URL has no trailing slash, so a caller can append a path.
  */
 export function privateEchoEndpoint(env: Record<string, string | undefined>): PrivateEchoEndpoint {
-  const base = (env.ECHO_BASE_URL ?? env.HTTPBIN_BASE_URL ?? "").replace(/\/$/, "");
+  const base = (env.ECHO_BASE_URL || env.HTTPBIN_BASE_URL || "").replace(/\/$/, "");
 
   if (!base) {
     return {
--- a/tests/tests-automations/regression/api/flows/api-component-regression.spec.ts
+++ b/tests/tests-automations/regression/api/flows/api-component-regression.spec.ts
@@ -28,8 +28,8 @@ import {
 // well as the daily. Retiring this legacy spec into the consolidated one is still
 // tracked separately.
 const ECHO_BASE = (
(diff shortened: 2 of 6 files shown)


Victor-w-Madeira:
Rebased onto `main` to resolve the conflict with #1610, which migrated `ssrf-url-validation.spec.ts` onto the `privateEchoEndpoint` helper.

The sixth site moved rather than disappeared. The spec's own resolution is gone, but the helper carried the same `??` chain over with it:

```ts
const base = (env.ECHO_BASE_URL ?? env.HTTPBIN_BASE_URL ?? "").replace(/\/$/, "");
```

So the spec is back to `main`'s version untouched, and the fix now applies to `tests/helpers/other/private-echo-endpoint.ts` instead.

Worth noting the failure mode there is milder than in the other five, but it is still real. The `!base` guard catches the empty string, so the spec skips with a clear reason rather than throwing — but the fallback still does not fall back: with `ECHO_BASE_URL=` set and `HTTPBIN_BASE_URL` pointing at something real, `??` keeps the empty string and the helper skips a test that could have run.

Still 6 files and +11/-11; no `??` remains in the fallback chain anywhere under `tests/`.

Issue #1604, opened 2026-08-25
Promote the two security specs shipped without @stable — 10 tests that watch nothing today


rafaelgiln:
## Problem

`security/model-provider-base-url-ssrf.spec.ts` (#1595, 6 tests) and `security/credential-secret-across-edges.spec.ts` (#1598, 4 tests) shipped **without `@stable`**. `daily-stable.yml` filters on `--grep "@stable"`, so those 10 tests watch nothing: they run only when a PR happens to touch them via the import graph, or when someone dispatches `manual.yml`.

They cover two upstream security fixes — the provider base-URL SSRF seam (#14640 + the #14704 follow-up gaps) and secret masking across a graph edge (#14216). A regression in either between two nightlies is invisible to every scheduled lane.

Area impact: `security/` currently has **12 of its 25 tests** in the daily. This takes it to 22.

## This is a correction, not a promotion

Both spec docs state the same justification, and it is wrong:

> **No `@stable` yet**, per the rule that the tag is added only after team validation.

**That is not the rule.** `CONTRIBUTING.md` → *Tag @stable — validated tests*:

> ### Standard: every new test enters with @stable
> `@stable` is the default for any new test. The test **enters with the tag in the PR itself**, together with the documentation file in `docs/`.

The four documented exceptions are: inherited-not-yet-reviewed, temporarily-removed-while-failing, utility specs (permanent absence, reason stated in the doc's **Tags** section), and `@destructive`. Neither spec is any of those — both are newly authored, both have complete spec docs, both assert causal API observables, and neither carries a lane selector.

The justification is also **circular**, which #1349 already settled for the A2A batch: without `@stable` the spec never enters the daily, so it can never accumulate the history its own doc cites as the precondition for the tag.

So the deliverable is not just the tag — the false sentence comes out of both docs.

## Evidence — local burst, 2026-08-25

Against `lf-dev38` = **`1.12.0.dev38`**, the freshest published nightly at the time of the run (`langflowai/langflow-nightly:latest` resolves to `1.12.0.dev38`). `ECHO_BASE_URL` pointed at a `go-httpbin` container on `172.17.0.5:8080`, with the instance's `LANGFLOW_SSRF_ALLOWED_HOSTS=172.16.0.0/12,10.0.0.0/8,192.168.0.0/16`.

```bash
PLAYWRIGHT_BASE_URL=http://localhost:7875/ ECHO_BASE_URL=http://172.17.0.5:8080 \
[...]

Issue #1593, opened 2026-08-25
pipeline: FORCE_FAIL's final green run assumes one Langflow instance per issue


rafaelgiln:
Found while working #1583. **Not a defect in the specs it blocked** — those are validated; this is a limitation of the deterministic pipeline that forced #1583 to finish through the prose orchestrator instead.

## What happens

`FORCE_FAIL`'s final-green step (`.claude/skills/langflow-e2e-issue-deterministic/pipeline/cli.ts`, the `s.phase === 'FORCE_FAIL'` block) loops over **every** `touchedSpecFiles()` entry and runs each against whatever single instance `PLAYWRIGHT_BASE_URL` points at:

```
for (const { file } of required) {
  outcome = runUntilClean([file, '--retries=0', '--workers=1'], file, 1, [], notes, ffAmbient)
  if (outcome.realFailure) fail(`final green run failed for ${file} after FF reverts`)
}
```

`next` accepts `--spec` in `VALIDATE` but **not** here, so there is no way to give one file its own instance. #1583 needs four (stock / trusted / untrusted / required) and the step reports:

```
✖ final green run failed for tests/tests-automations/regression/api/flows/serving-end-user-identity-default.spec.ts after FF reverts
```

## Why this is not a file-split problem

#1583's deliverable is a **pair**: the same `X-End-User-Id` header proven **inert** on a default instance and **decisive** on a configured one. That needs at least two instances by definition, so collapsing the four spec files into two — or into one plus one — fails identically. Any future issue whose contract spans configurations (an `@enterprise` variant matrix, a governance knob, a lane's fail-closed row) hits the same wall.

## The trap to avoid in the fix

Running the step with the lane selector **unset** makes it pass, and that is the wrong answer: the three `@serving` specs then select **zero tests** (correctly — `grepInvert` is doing its job), and `classifyRun` counts a zero-test run as `clean`. The gate would go green having executed nothing. Whatever the fix is, it must not be reachable by that route — and the zero-test-counts-as-clean behaviour is arguably worth its own guard regardless.

## Sketch of a fix

Accept a per-file instance map, e.g. `next <NNN> --final-green-spec <file>` recording one file at a time (mirroring how `VALIDATE`'s burst already accumulates clean runs per target across invocations), so the phase closes only when every touched file has its own green run — each measured on the instance that file needs. `VALIDATE` already has exactly this shape; `FORCE_FAIL` does not.

## Evidence #1583 produced by hand, which the gate could not express

[...]

Issue #1252, opened 2026-08-04
Duration-balanced sharding has never had data — `reports/spec-durations.json` was never committed, because its gate needs a fully green daily


Victor-w-Madeira:
Found while working #1187, which needs a wall-clock ceiling for the daily and
reasons from `reports/spec-durations.json`. The file does not exist.

## Duration-balanced sharding has never had data

`reports/spec-durations.json` has **never been committed to `main`**:

```
$ git log --oneline -- reports/spec-durations.json
(no output)
```

So `daily-stable.yml`'s prep job runs

```
node scripts/partition-shards.mjs matrix /tmp/stable-list.json reports/spec-durations.json "$N"
```

against a **missing path** on every run. `partition-shards.mjs` documents that as a
legitimate cold start (`may be "-" or a missing path (cold start) -> empty
durations`), so it degrades to a count balance and says nothing. The degradation is
silent by design — correctly, for the first run — but it has been *every* run since
the mechanism shipped in `81c2488` (#936 / PR #949, 2026-07-24).

## Why: the producing gate is all-or-nothing on a suite that is never fully green

```yaml
- name: Refresh spec durations (green runs only)
  if: needs.test.result == 'success' && github.event_name == 'schedule'
     && steps.runguard.outputs.empty == 'false' && steps.runguard.outputs.partial == 'false'
```

Every scheduled daily since the mechanism shipped, from `reports/daily-history.jsonl`:

| Date | passed | failed | flaky | Gate |
|---|---|---|---|---|
| 2026-07-24 | 388 | 5 | 8 | ✗ `test` red |
| 2026-07-27 | 379 | 2 | 10 | ✗ `test` red |
| 2026-07-28 | 0 | 0 | 0 | ✗ barred by `runguard.empty` (#1012, correctly) |
| 2026-07-29 | 205 | 10 | 12 | ✗ `test` red |
| 2026-07-30 | 466 | 2 | 10 | ✗ `test` red |
| 2026-08-03 | 426 | 1 | 4 | ✗ `test` red |

Six for six. Across the whole 23-entry history only **2026-07-03** ever had
`failed: 0` — three weeks before the mechanism existed.

The reasoning behind the gate is sound and should be preserved: a red run's per-spec
times are distorted by retries and timeouts, and poisoning the partition is worse
than not balancing it. What was not weighed is that **a suite with 1–10 hard failures
per day never produces a clean run**, so the gate is not strict — it is closed. One
[...]

Issue #1226, opened 2026-08-01
Follow-up #1216: the impacted-specs summary counts specs the provider verdict excluded


Victor-w-Madeira:
Follow-up to #1216 (PR #1223), from that PR's review. Three defects, all in the
surface #1216 just built, none blocking — grouped because they are one small PR.

## 1. The step summary's headline count is written before the verdict (the real one)

`pr-validation.yml:252` emits

```
- running in this PR: **$((TOTAL - DROPPED))** (of which `@stable`: $STABLE)
```

and the provider-coverage verdict only runs at `:316`. So on any run that excludes a
spec, that line counts the excluded spec as executed — and `$STABLE`
(`.stableSelected`) overstates too, because the spec that motivated #1216
(`core-components/agent-component-regression.spec.ts`) is `@stable`.

The exclusion `::warning::` and its summary block are appended below, so nothing is
*silent*. But the number a reviewer reads **first** says 7 ran when 6 did, and
reconciling it against a warning further down is left to them. That is the same
class of defect #1216 exists to remove: a figure that reads as coverage it did not
deliver (#1012's rule).

Fix: either move that bullet below the verdict, or emit a corrected
`running after the provider verdict: **N**` line beside the exclusion block. The
count must come from the same list `specs=` is emitted from, not from
`TOTAL - DROPPED`.

## 2. A hedged assertion in the fail-closed lane

`scripts/provider-dependent-specs.test.mjs:1080`

```js
assert.ok(malformed.status === 2 || malformed.status === 0);
```

Verified: the exit is deterministically **2**. `JSON.parse('"not json at all"')`
yields a string, `.selected` is `undefined`, and `decideProviderCoverage` throws
`selected must be an array`:

```
$ echo '"not json at all"' | node scripts/provider-dependent-specs.mjs --stdin --changed-file=/tmp/cf.txt
::error::provider-coverage verdict failed: Error: selected must be an array
exit=2
```

So the `|| status === 0` accepts exactly the pass-through the script exists to
forbid. Either assert `2`, or delete the case — `trulyMalformed` on the next lines
already covers the parse failure properly.

## 3. `--changed-file=` is parsed with `split("=")[1]`

`scripts/provider-dependent-specs.mjs` truncates a path containing `=`. It is
`/tmp/changed.txt` today, so this cannot fire in CI — but
`slice(indexOf("=") + 1)` costs nothing and removes the trap for whoever passes a
different path locally.

## Out of scope

[...]

PR #1608, opened 2026-08-26
chore(coverage): exclude enterprise/ from the TOTAL, and give the generator its first unit test (#1607)

### LINKED ISSUE #1607: Coverage Summary: exclude enterprise/ from the TOTAL, and give the generator its first unit test
## Decision this implements

**Enterprise surfaces are out of CI scope. Only OSS is tested on a schedule.** (Team decision, 2026-08-26.)

That makes the Coverage Summary's `**TOTAL**` answer a question nobody asked. The total reads as *"what fraction of what we intend to run is validated"*, and 71 `enterprise/` bullets can no longer reach `[x]` — not because the specs are unfinished (19 files, 62 tests, all merged and green) but because the lane they would need has been deliberately declined. Counting them in the denominator guarantees the metric never closes, and it drifts further every time an OSS bullet is validated.

Today: **441 / 660 = 67 %**. Excluding `enterprise/`: **441 / 589 = 75 %**.

## Why not a new bullet symbol

The obvious fix — a sixth symbol, `[o]` for *automated, deliberately unscheduled* — is the wrong shape and much more expensive:

- `BULLET_RE = /^- \[([x\- ~!])\] /` is a fixed character class, and `classify()` branches on it
- the table would grow a **sixth column** on top of five
- the legend is duplicated in `CLAUDE.md`, `CONTRIBUTING.md` and the `langflow-e2e` skill
- `scripts/check-checklist-guard.mjs` matches generated line shapes

Five places changed to express a fact that is not about the bullet at all. *"Not scheduled by decision"* is a property of the **module**, not of each individual test's state — every one of those 56 `[-]` bullets is accurately `[-]`: the test is automated and not validated. What changed is whether we intend to validate it.

## Scope

**1. `excludeFromTotal` on `ModuleConfig`** in `scripts/coverage-summary.ts`, set on the `enterprise/` entry only. `governance/` and `serving/` stay **in** — they are OSS (governance is the operator surface 1.12 ships in the nightly, serving is the same image under different env vars), and both are slated to gain scheduled lanes.

**2. The exclusion must be visible in the rendered table, never only in code.** A silent exclusion is how the number starts lying again the day the decision reverses — the same failure mode as the `mode=count` warning nobody read (#1252) and the `@stable` justification that survived in two spec docs (#1604). So:

- the module row keeps its real counts and gains the reason in its label
- the total row names what it excludes

```
| `enterprise/` — Enterprise-only Surfaces (not scheduled — decision) | 71 | 0 | 56 | 9 | 6 |
| **TOTAL (OSS — excludes `enterprise/`)** | 589 | 441 (75%) | … |
```

The work is not hidden: 19 files and 62 tests stay counted and visible on their own row. Only the denominator changes.

**3. The generator gets its first unit test.** Found while scoping this: `scripts/coverage-summary.ts` **exports nothing and has no `*.test.ts`**. It generates the table the roadmap and leadership read, runs on every push to `main` via `update-coverage-summary.yml`, and is covered by neither unit lane — the same gap #1017 created `test:units` for and #1593 just closed for the issue pipeline. The two existing `scripts/*.test.ts` files cover `remove-stable-from-failures` and `update-component-catalog-baseline`; nothing imports this one.

So the sum and the row rendering move into an exported pure function with tests on its **output**, not on the shape of the code (#1226: a guard that pins a spelling pins no behaviour). Minimum cases: an excluded module is absent from the total but present as a row; a non-excluded module counts; excluding every module does not divide by zero; the total label names the exclusion; with no module excluded the output is byte-identical to today's.

## Out of scope

- **`ROADMAP.md`'s hand-copied snapshot** (`76% (402/532)`) is already stale and will disagree further. It says *"refresh from `QA-CHECKLIST.md` when it changes — do not hand-maintain"*, and the checklist only regenerates on merge, so refreshing it here would mean writing numbers that do not exist yet. Post-merge step, tracked here but done after.
- **No `QA-CHECKLIST.md` edit in the PR.** The label and total live in the generator; `update-coverage-summary.yml` regenerates the block on merge. Committing regenerated counts is what the `pr-validation.yml` guard rejects (#741).
- **Reversing the decision** is a one-line change to the same field, which is the point of putting it there.

## Done when

- [ ] `excludeFromTotal` set on `enterprise/` only; `governance/` and `serving/` still counted
- [ ] The rendered table names the exclusion on both the row and the total
- [ ] `scripts/coverage-summary.test.ts` exists and runs in `npm run test:units`
- [ ] `npm run coverage:summary` is idempotent — a second run produces no diff

rafaelgiln:
## Type

- [ ] `feat` — new test or new helper/page object
- [ ] `fix` — fix for a broken or flaky test
- [ ] `docs` — documentation update

---

- [ ] **Current wave** — a `roadmap`-labeled issue in the current wave's milestone: #___
- [x] **Exception** — off-wave work backed by an issue: a follow-up, a `daily-failure` triage issue, or a `community`-labeled regression. Issue **#1607** (`qa-infra`, `follow-up`)

---

## What this PR does

**Enterprise surfaces are out of CI scope by team decision (2026-08-26): only OSS is tested on a schedule.**

That makes the Coverage Summary's `**TOTAL**` answer a question nobody asked. The total reads as *"what fraction of what we intend to run is validated"*, and 71 `enterprise/` bullets can no longer reach `[x]` — not because the specs are unfinished (19 files, 62 tests, merged and green) but because the lane they would need has been declined. Counting them in the denominator guaranteed the metric never closes, and it drifts further every time an OSS bullet is validated.

**68 % → 77 %** (451/660 → 451/589), with the work still visible on its own row.

### Why not a sixth bullet symbol

The obvious fix — `[o]` for *automated, deliberately unscheduled* — is the wrong shape and far more expensive: `BULLET_RE` is a fixed character class, `classify()` branches on it, the table would grow a **sixth** column, the legend is duplicated across `CLAUDE.md`, `CONTRIBUTING.md` and the `langflow-e2e` skill, and `check-checklist-guard.mjs` matches generated line shapes. Five places to express something that is not about the bullet at all: every one of those 56 `[-]` is accurately *automated, not validated*. What changed is whether we **intend** to validate it — a property of the module. Hence `excludeFromTotal` on `ModuleConfig`.

Reversing the decision is deleting one field.

### The exclusion is rendered, never silent

This is the load-bearing half. A silent exclusion is how the number starts lying again the day the decision reverses — the same failure mode as the `mode=count` warning nobody read (#1252) and the `@stable` justification that survived in two spec docs (#1604).

```
| `enterprise/` — Enterprise-only Surfaces (not scheduled — decision) | 71 | 0 | 56 | 9 | 6 |
| **TOTAL (OSS — excludes `enterprise/`)** | **589** | **451 (77%)** | **59 (10%)** | **16 (3%)** | **63 (11%)** |
```

`governance/` (14) and `serving/` (13) stay **counted**: both are OSS — the nightly ships and *enforces* 25 of EE's governance routes, and serving is the same image under different env vars — and both are slated to gain scheduled lanes.

### The generator gets its first unit test

Found while scoping: `scripts/coverage-summary.ts` **exported nothing and had no `*.test.ts`**, despite generating the table the roadmap and leadership read and running on every push to `main` via `update-coverage-summary.yml`. The same gap #1017 created `test:units` for, and #1593 just closed for the issue pipeline. The sum and row rendering moved into an exported pure `renderCoverageTable`, asserted on its **output** rather than on the shape of the code (#1226).

---

## Tests covered

`scripts/coverage-summary.test.ts` — new, 9 tests in `npm run test:units`:

|---|---|---|
| 1 | an excluded module keeps its own row, with its real counts | the work does not vanish — 71/0/56/9/6 still renders |
| 2 | an excluded module is absent from the TOTAL | the denominator actually drops it (13, not 84) |
| 3 | the total row NAMES what it excludes | the exclusion can never be silent |
| 4 | with nothing excluded the total counts every module, and says nothing extra | the no-op guarantee — invisible until a module opts in |
| 5 | excluding every module reports zero rather than dividing by zero | no `NaN` in a table that always renders |
| 6 | module rows and counts stay aligned by index | a zip bug would silently attribute one module's numbers to another |
| 7 | the exclusion note uses the module KEY, not its whole em-dashed label | see the first defect below |
| 8 | two excluded modules are both named, comma-separated | the plural path |
| 9 | importing the module does NOT run the generator | see the second defect below |

---

## How the tests were built

Two defects the implementation itself produced, both caught here and pinned:

**1. The total rendered two em-dashes.** Splicing the whole label produced `TOTAL (OSS — excludes \`enterprise/\` — Enterprise-only Surfaces)`. It survived the first fixture because that fixture used `label: "ent"` — a shape `MODULES` never contains. Exactly the #1216 trap, where a hand-built fixture the real resolver never emits let the bug through. Fixed to the key (`label.split(" — ")[0]`) and pinned with a realistic label.

**2. Making the script importable left `main()` bare at module scope**, so merely **loading** it rewrote `QA-CHECKLIST.md` — on every `npm run test:units`, CI included. A unit lane that mutates a tracked file as a side effect of importing is worse than no lane. Guarded with the repo's existing `require.main === module` idiom, the same one `remove-stable-from-failures.ts` and `update-component-catalog-baseline.ts` already use.

The test for it asserts the child process's **stdout**, not the checklist's bytes — and that matters. The first version compared the file before and after, and **could not fail**: with the guard removed, the import at the top of the test file has already run the generator by the time the test reads its baseline, so the baseline is the rewritten content and the child then finds nothing to do. Reading the output also means the test writes nothing at all.

Verified in both directions: importing runs nothing, invoking still regenerates, and a second invocation produces no diff.

---

## Dependencies

None. No Langflow instance, no `.spec.ts`, no new package.

---

- **`QA-CHECKLIST.md` is deliberately not in this diff.** The label and the total live in the generator; `update-coverage-summary.yml` regenerates the block on merge. Committing regenerated counts is what the `pr-validation.yml` guard rejects (#741).
- **`ROADMAP.md`'s hand-copied snapshot** (`76% (402/532)`) is already stale and will disagree further. It says *"refresh from `QA-CHECKLIST.md` when it changes — do not hand-maintain"*, and the checklist only regenerates on merge, so refreshing it here would mean writing numbers that do not exist yet. Post-merge step.
- **The scheduled lanes for `@destructive` and `@serving`** — the OSS half of the same conversation, 22 bullets, tracked separately.

---

## Known limitations

`renderCoverageTable` decorates the **rendered** label only; `m.label` itself is untouched, because it is the key the Phase tables are matched by (`labelToIndex`). Changing the field would break them with *"Phase N row … does not match any module in MODULES"*. Recorded as a comment at the site.

---

## Related issue

Closes #1607

---

## Validation

- [x] **Forced a failure — 7 mutations, 7 red, reverted byte-identical:**
  - `M1 ignore excludeFromTotal in the sum` → **5** failed
  - `M2 drop the excluded module's row` → 1 failed
  - `M3 total no longer names the exclusion` → 3 failed
  - `M4 splice the whole label instead of the key` → 2 failed
  - `M5 always add the note, even with nothing excluded` → 1 failed
  - `M6 misalign counts against modules` → 2 failed
  - `M7 remove the `require.main` guard` → 1 failed
- [x] **Rendered output inspected against the real file** before reverting it — the two rows above are what merge will produce.
- [x] **Idempotent** — a second `npm run coverage:summary` produces no diff.
- [x] **Phase tables still resolve** — no `does not match any module in MODULES` error.
- [x] `npm run typecheck` 0 · `npm run lint` 0 · `test:units` **797/797** · `test:scripts` 857 · `test:pipeline` 119/119 · QA-CHECKLIST guard clean
- [ ] ~`--trace=on` / `--debug` / backend errors~ — not applicable: no Playwright spec, no Langflow instance.
- [ ] ~Updated `QA-CHECKLIST.md`~ — deliberately not; see above.

### DIFF
--- a/scripts/coverage-summary.test.ts
+++ b/scripts/coverage-summary.test.ts
@@ -1,0 +1,127 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import * as path from "path";
+import { execFileSync } from "child_process";
+import { renderCoverageTable, type Counts, type ModuleConfig } from "./coverage-summary";
+
+// The generator's first unit test (#1607). It had none: the script exported
+// nothing, so the table the roadmap and leadership read — regenerated on every
+// push to `main` by update-coverage-summary.yml — was covered by neither unit
+// lane. Assertions are on the RENDERED OUTPUT, never on the shape of the code
+// (#1226: a guard that pins a spelling pins no behaviour).
+
+const c = (validated: number, needsValidation: number, partial: number, notAutomated: number): Counts =>
+  ({ validated, needsValidation, partial, notAutomated });
+
+const mod = (label: string, excludeFromTotal = false): ModuleConfig =>
+  ({ label, sectionStart: `## ${label}`, ...(excludeFromTotal ? { excludeFromTotal } : {}) });
+
+test("an excluded module keeps its own row, with its real counts", () => {
+  // The work must stay visible: 19 files and 62 tests do not vanish because the
+  // lane was declined. Only the denominator changes.
+  const rows = renderCoverageTable(
+    [mod("oss"), mod("ent", true)],
+    [c(10, 2, 0, 1), c(0, 56, 9, 6)],
+  );
+  const entRow = rows.find(r => r.includes("| ent"))!;
+  assert.ok(entRow, "the excluded module must still have a row");
+  assert.match(entRow, /\| 71 \| 0 \| 56 \| 9 \| 6 \|/);
+});
+
+test("an excluded module is absent from the TOTAL", () => {
+  const rows = renderCoverageTable(
+    [mod("oss"), mod("ent", true)],
+    [c(10, 2, 0, 1), c(0, 56, 9, 6)],
+  );
+  const total = rows.find(r => r.includes("TOTAL"))!;
+  // 13, not 84: the excluded module's 71 bullets are out of the denominator.
+  assert.match(total, /\*\*13\*\*/);
+  assert.match(total, /10 \(77%\)/);
+});
+
+test("the total row NAMES what it excludes — the exclusion is never silent", () => {
+  // The half that matters. A silent exclusion is how the number starts lying
+  // again the day the decision reverses (the mode=count lesson, #1252).
+  const total = renderCoverageTable(
+    [mod("oss"), mod("ent", true)],
+    [c(10, 2, 0, 1), c(0, 56, 9, 6)],
+  ).find(r => r.includes("TOTAL"))!;
+  assert.match(total, /excludes/i);
+  assert.match(total, /ent/);
+});
+
+test("with nothing excluded the total counts every module, and says nothing extra", () => {
+  // The no-op guarantee: this change must be invisible until a module opts in.
+  const rows = renderCoverageTable([mod("a"), mod("b")], [c(1, 1, 0, 0), c(2, 0, 0, 0)]);
+  const total = rows.find(r => r.includes("TOTAL"))!;
+  assert.match(total, /\*\*4\*\*/);
+  assert.equal(/excludes/i.test(total), false, "no exclusion note when nothing is excluded");
+  assert.match(total, /\| \*\*TOTAL\*\* \|/);
+});
+
+test("excluding every module reports zero rather than dividing by zero", () => {
+  const total = renderCoverageTable([mod("a", true)], [c(3, 1, 0, 0)]).find(r => r.includes("TOTAL"))!;
+  assert.match(total, /\*\*0\*\*/);
+  assert.doesNotMatch(total, /NaN/);
+});
+
+test("module rows and counts stay aligned by index", () => {
+  // A zip bug here would silently attribute one module's numbers to another —
+  // invisible in a table that always renders.
+  const rows = renderCoverageTable(
+    [mod("first"), mod("second")],
+    [c(1, 0, 0, 0), c(0, 0, 0, 9)],
+  );
+  assert.match(rows.find(r => r.includes("| first"))!, /\| 1 \| 1 \| 0 \| 0 \| 0 \|/);
+  assert.match(rows.find(r => r.includes("| second"))!, /\| 9 \| 0 \| 0 \| 0 \| 9 \|/);
+});
+
+test("the exclusion note uses the module KEY, not its whole em-dashed label", () => {
+  // The fixture this file first used was `label: "ent"` — a shape the real
+  // MODULES array never contains. Every real label is "`dir/` — Description",
+  // and splicing the whole thing into a sentence that already has an em-dash
+  // renders "TOTAL (OSS — excludes `enterprise/` — Enterprise-only Surfaces)".
+  const total = renderCoverageTable(
+    [mod("`oss/` — Open Source"), mod("`enterprise/` — Enterprise-only Surfaces", true)],
+    [c(10, 2, 0, 1), c(0, 56, 9, 6)],
+  ).find(r => r.includes("TOTAL"))!;
+  assert.match(total, /excludes `enterprise\/`\)/);
+  assert.doesNotMatch(total, /Enterprise-only Surfaces\)/);
+});
+
+test("two excluded modules are both named, comma-separated", () => {
+  const total = renderCoverageTable(
+    [mod("`a/` — A", true), mod("`b/` — B", true), mod("`c/` — C")],
+    [c(1, 0, 0, 0), c(1, 0, 0, 0), c(5, 0, 0, 0)],
+  ).find(r => r.includes("TOTAL"))!;
+  assert.match(total, /excludes `a\/`, `b\/`\)/);
+  assert.match(total, /\*\*5\*\*/);
+});
+
+test("importing the module does NOT run the generator", () => {
+  // The defect this pins was live for one commit: making the script importable
+  // left `main()` bare at module scope, so merely LOADING it rewrote a tracked
+  // file — on every `npm run test:units`, CI included. A unit lane that mutates
+  // the repo as a side effect of importing is worse than no lane.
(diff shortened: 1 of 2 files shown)


rafaelgiln:
**The lane's own warning cannot be followed — recorded rather than ignored.**

`Detect changed specs` emitted, correctly:

```
##[warning]CI-only change to scripts/coverage-summary.ts, which THIS lane does not run —
nothing here proves it works. Dispatch .github/workflows/update-coverage-summary.yml on
this branch before merging (#1159).
```

I tried:

```
$ gh workflow run update-coverage-summary.yml --ref chore/issue-1607-total-oss-excludes-enterprise
could not create workflow dispatch event: HTTP 422: Workflow does not have 'workflow_dispatch' trigger
```

That workflow is `on: push: branches: [main]` only. The remedy is impossible on any branch. Filed as **#1609** — measured scope is exactly one workflow reached by three scripts (`coverage-summary.ts`, `stable-tests.ts`, `lib/stable-tests.ts`), which happen to be the ones that generate the release signal.

**What I did instead**, since the prescribed proof is unavailable:

| check | result |
|---|---|
| ran the generator against the real `QA-CHECKLIST.md` and inspected the rendered rows | the two rows in the PR body are verbatim what merge will produce |
| second run | no diff — idempotent |
| Phase tables | resolve; no `does not match any module in MODULES` |
| import side effect | child process prints nothing (the `require.main` guard); invoking still regenerates |
| unit lane | 797/797, 9 of them new on this file |
| force-fail | 7 mutations, 7 red, reverted byte-identical |

Then reverted the file, so the generated block is not in this diff (#741).

**Residual risk, stated plainly:** nothing in CI exercises this generator before merge. If it throws on `main`, the `update-coverage-summary` job fails visibly and the checklist stays stale — it does not corrupt the file, because the write is the last step and only happens when the output differs. The post-merge run is worth watching.

PR #1619, opened 2026-08-26
fix(llm-agents): assert the tool step the max-iterations cap depends on (#1264)

### LINKED ISSUE #1264: [Daily #1258] agent-max-iterations — the agent answers normally instead of hitting the max-iterations limit (no outage cover)
Spun out of daily-failure triage #1258 (run [30901311395](https://github.com/oriontech-me/langflow-e2e/actions/runs/30901311395), 2026-08-04).

**Upstream:** _not filed_

## Symptom

A `@stable @regression` test hard-failed (final status, after retries) on an **assertion about content, not a timeout**: with `max_iterations` set to force the limit, the agent produced an ordinary answer instead of the expected limit message. The chat message rendered fine — the assertion polled a resolved locator 34 times and read a real, wrong value.

|---|---|---|
| `tests/tests-automations/regression/core-functionality/llm-agents/agent-max-iterations.spec.ts:134` ("agent stops when max iterations is reached") | `getByTestId('div-chat-message').last() to match /model call limits exceeded/i — received "I'll fetch that URL for you and get the version value."` | `Error: expect(locator).toContainText(expected) failed` |

One spec, one cause. The argument to state is why this is **isolated** from the rest of the run rather than grouped with it.

Every other hard failure on this run is a wait that never resolved (a page that did not bootstrap, a row that never became clickable, a predicate that never settled). This one is the opposite shape: the UI worked. The call log shows the locator `getByTestId('div-chat-message').last()` **resolving 34 times** to a real rendered message, and the assertion failing on its *content* — the received string is `"I'll fetch that URL for you and get the version value."` where `/model call limits exceeded/i` was expected.

It is also the **only** hard failure on this run with no environmental cover: it ran on **shard 1**, which the umbrella's backend-liveness table records as **0 outages, 0 min of unreachable backend**. The three other hard failures each overlap a recorded outage window; this one cannot.

`agent-max-iterations.spec.ts:163` ("causal control — a high max iterations does not hit the limit") was **skipped as a consequence** in the same file — cascading, covered here, not filed separately. That the causal control never ran matters: the pair is what distinguishes "the limit was not enforced" from "the limit message changed".

**This is the run's strongest signal and the least explainable by the day's infrastructure.** The run carried a measured mid-run backend wedge (#1030/#1048) — 13 outages, 12.4 min across 3 of 4 shards — but shard 1 recorded **zero** outages, and the failure is a content assertion over a healthy UI rather than a transport or wait failure. Neither the wedge nor the exemption list is relevant here.

Provider context: this run resolved `anthropic` / `claude-haiku-4-5` for this test (the daily rotates one provider per weekday, #1185). The received answer shows the agent electing to use its URL tool and reporting intent — i.e. it was operating normally, just not hitting the limit the test forces.

Recurrence, descriptive only: `recurrence.same_signature` is **false** for today's signature — the test has failed on 7 dailies in the 30-day window (07-13, 07-14, 07-17, 07-20, 07-22, 07-29, 08-04) but with varying observables, so this is not a confirmed single recurring cause. Read it as a long-unsettled test whose today's failure has an unusually specific shape.

The guard did **not** trip on this run (4 hard failures), so no day-level environmental verdict is claimed. `@stable` was auto-removed by the workflow, as designed.

## Investigation directive

1. On the current nightly, reproduce the configured `max_iterations` limit and confirm whether the Agent **enforces** it at all — does the run stop, and does it surface the limit to the user? A limit that is silently not enforced is a product regression and is the prime suspect, because the UI demonstrably worked here.
2. If it *is* enforced, confirm **what text** it surfaces now. The assertion expects `/model call limits exceeded/i`; upstream wording drift would make this a behaviour change (update test **and** spec doc), not a test bug.
3. Only after both, consider the test's own determinism: whether the prompt/`max_iterations` combination reliably forces the limit for `claude-haiku-4-5`, and whether the assertion can read an intermediate message rather than the final one (`.last()` on a stream that is still growing).

Run the causal control (`:163`) too — it was skipped by the cascade, and the pair is what separates "not enforced" from "wording changed".

## Deliverables (Done when)

- [ ] Explicitly answered: does the Agent enforce `max_iterations` on the current nightly, and what text does it surface when it does?
- [ ] The causal control at `:163` is run and passing, not just the primary test.
- [ ] If the limit wording changed upstream: the spec doc under `docs/` is updated alongside the assertion.

rafaelgiln:
**Closes #1264.**

## Problem

`agent stops when max iterations is reached` hard-failed 3/3 attempts on the 2026-08-13 daily ([run 31685261355](https://github.com/oriontech-me/langflow-e2e/actions/runs/31685261355)) — **with #1439 already merged** (`9792bed` is an ancestor of that run's `headSha`). It had failed on 7 dailies in the 30-day window; `@stable` was auto-removed twice (`d441780`, `1bb9425`), so the enforcement half of the causal pair has not run in the daily since 08-13.

The standing explanation does not survive the evidence. #1439 blamed an SSRF-blocked fetch target putting every run on the tool-error path, where LangGraph's `recursion_limit` was said to fire before the model-call cap — but that run's own job log records `ECHO_BASE_URL: http://172.18.0.5:8080`. **The target was reachable.**

The real dependence is structural, and it is in the middleware:

```python
# langchain/agents/middleware/model_call_limit.py
before_model: run_count >= run_limit  →  jump to "end" + inject the AIMessage
after_model:  run_count += 1
```

`before_model` #1 sees `0 >= 1` (false) and the first call happens. The graph then re-enters `before_model` **only through the tools node** — so with `run_limit = 1` the limit message exists only if that first call emitted a `tool_use` block. **The second model call is elected by the model**, and nothing in the product forces it: `lfx/components/models_and_agents/agent.py` declares `max_iterations` with `range_spec(min=1)` (a cap of 0 is impossible by design) and exposes **no `tool_choice`** input — Langflow dropped the legacy `create_granite_agent` path precisely because it hardcoded `tool_choice='required'`, which the WatsonX API rejects.

On 08-13 `claude-haiku-4-5` answered in prose. That run's token-attribution artifact (`tokens-4`) records `calls: 1`, `status: ok`, 934 prompt / 77–82 completion tokens, with the bubble reading *"I'll fetch that URL for you."* The cap never had a second `before_model` to fire on, and the assertion reported that as a broken cap — for two triage passes.

**The product is correct**, measured on `1.12.0.dev39` with `google/gemini-3.6-flash`, `max_iterations = 1` and a reachable target: exactly one chat bubble reading `Model call limits exceeded: run limit (1/1)`, with an executed `tool_use` (`fetch_content`) in its *Agent Steps* block. So the cap **is** enforced, the wording is **unchanged**, and the injected message **is** the last bubble — which also refutes the competing reading that the cap fired but failed to render last.

## Fix

**1. Test 1 asserts the precondition separately.** `expectToolLoopEntered(request, flowId)` polls `GET /api/v1/monitor/messages?flow_id=<id>` for a `tool_use` entry in the persisted AI message, before the limit-message assertions. A model that declines now fails with *"the model answered without calling any tool … This is model non-compliance with the Agent Instructions, NOT a broken max_iterations (#1264)"*, printing the persisted reply.

**2. The Agent Instructions state an ordering rule.** The previous wording ("To answer any question about a URL you MUST call the URL fetch tool") is satisfiable by announcing the intent, which is what happened. The new wording ("Your FIRST action MUST be a tool call — never reply with text before you have called a tool") cannot *force* the call; it only removes the reading under which a bare announcement is compliant. That is why the assertion, not the prompt, is the guarantee.

Rejected alternatives, explicitly:

- **Making the second call structural.** There is none. No `tool_choice` on the component; `max_iterations` floors at 1, so `before_model` #1 can never exceed the limit. The structured-output path is a different output handle and bypasses the agent loop when there are no tools.
- **Reading the tool step from the DOM.** The *Agent Steps* disclosure is collapsed and its text is not in the bubble's `innerText` — measured, a passing bubble is 43 characters, the limit message alone. The monitor API is also the route the sibling `agent-multi-tool-selection.spec.ts` already uses for tool observables, so this reuses `expectToolSelectionPersisted`'s shape rather than hand-rolling one.
- **Pinning the tool name to `fetch_content`.** The cap is reached by *entering the tool loop*, whichever of the template's two tools the model picks; pinning the name adds a second election dependency for no gain. Tool selection is the sibling spec's assertion. The names actually called are printed in the failure text.
- **Certifying the tag with an N/N run gate.** #1187 already established that a run-count gate cannot certify an election-dependent assertion. The tag follows a measured rate, recorded in the spec doc.

## Scope note

Test 2 (causal control) is untouched apart from the shared `SYSTEM_PROMPT` constant — its UUID assertion already proves a successful fetch, so it needs no precondition of its own. No assertion was weakened, no selector loosened, no wait or retry added: the change *adds* an assertion. The reachable echo target stays (cost hygiene). Cleanup is the existing id-scoped `trackCreatedFlows` tracker (#1108/#1346), unchanged; zero orphan flows after validation.

The record is also corrected in the spec doc and `QA-CHECKLIST.md`: the 733,990-token tool-error runaway is real history but **model-specific and does not reproduce on dev39** — `gemini-3.5-flash` is retired (404), and with the same blocked target the agent now makes one tool call and reports the SSRF error in plain text at `max_iterations = 5`, with the cap still firing first at 1. **There is no measured product defect left behind that finding to file upstream**, which is what the reopening was tracking.

## Validation (nightly `1.12.0.dev39`, `--retries=0`)

- typecheck ✅ · eslint ✅ (0 errors) · QA-CHECKLIST guard ✅ · coverage guard ✅
- **3 clean runs, no flake** ✅ — `clean expected=2 unexpected=0 flaky=0 skipped=0` each, `--workers=1 --retries=0`
- **Final green after reverts** ✅ — `clean expected=2 unexpected=0 flaky=0`
- zero `🚨 Backend Error` ✅ · zero orphan flows ✅

**Force-fail — 3 mutations, all red, all reverted:**

| Test | Mutation | Result |
|---|---|---|
| Test 1 | `max_iterations = HIGH_LIMIT (5)` instead of `1` → cap never fires | `unexpected=1` ✅ |
| Test 1 | `expectToolLoopEntered` filters `tool_use` by an impossible name → the **new** precondition fails even with the cap message present | `unexpected=1` ✅ |
| Test 2 | `max_iterations = 1` instead of `HIGH_LIMIT` → the limit message appears | `unexpected=1` ✅ |

**Per-provider measurement behind the restored `@stable`** (`manual.yml`, `retries=0`, both tests per run):

| When | Provider / model | Runs | Result |
|---|---|---|---|
| Pre-fix (unmodified spec), CI | `openai` / `gpt-4o-mini` | 3 | 3/3 clean |
| Pre-fix (unmodified spec), CI | `google` / `gemini-3.5-flash` | 3 | 3/3 clean |
| Post-fix, local `1.12.0.dev39` | `google` / `gemini-3.5-flash` | 3 | 3/3 clean |

9 runs, 18 tests, zero failures.

> ⚠️ **Stated, not implied (#1012): `anthropic` — the provider that failed — could not be measured anywhere.** Its key is out of credit in CI as well as locally (three `provider=anthropic` dispatches exited 1 at *Resolve the run's provider selection* with `Your credit balance is too low`). `daily-stable.yml`'s weekday rotation advances past an inactive provider, so anthropic (Tue/Fri) does not run in the daily while the key is dry — which is why restoring the tag costs nothing today, and also why the risk returns the day the credential is funded. If Test 1 then reds on an anthropic day, the failure text names the cause and it belongs in the spec doc's Tags section as a measured rate, not in a new product issue.


rafaelgiln:
## CI note — this PR merged with **no checks at all**, and that is worth recording

Not a red check and not a flake: **GitHub never created an Actions check suite for `2a1a7207`.**

```
GET /commits/2a1a72077885e8fdb05c15e917aff7dbb94e8c04/check-suites
  total=5 — claude, render, netlify, railway-app, flakiness-io, ALL status=queued, latest_check_runs_count=0
  (no github-actions suite present)
GET /commits/2a1a72077885e8fdb05c15e917aff7dbb94e8c04/check-runs   → total=0
GET /commits/2a1a72077885e8fdb05c15e917aff7dbb94e8c04/status       → pending
```

`gh run list --commit 2a1a7207` returns nothing, and `gh pr checks 1619` reports *"no checks reported on the branch"*.

**The cause is outside this PR.** `pr-validation.yml` is `active` (it is not in the repo's `disabled_manually` list, which holds only *Adaptive Impacted Tests*, *Langflow Source File Watcher*, *Nightly E2E Regression* and *Weekly Stable E2E*), its trigger is a plain `pull_request: branches: [main]` with **no path filter**, and it ran normally on other PRs earlier the same day — most recently `32977421393` on `fix/issue-1460-memory-base-option-identity` at 13:58Z. Actions itself was healthy in the same window: six `Manual E2E Run` dispatches ran at 14:36–14:45Z, and `Update Coverage Summary` fired on the merge push at 15:23:25Z. So the workflow, its trigger and the runner were all fine — the `pull_request` event simply never produced a run for this head.

### What stands in place of the PR lane

Every gate `pr-validation.yml` would have run was executed locally against `1.12.0.dev39` before the merge, plus the three unit lanes re-run afterwards on the merged tree:

| PR-lane job | Local result |
|---|---|
| TypeScript check (`tsc --noEmit`) | ✅ clean |
| ESLint | ✅ 0 errors |
| `npm run test:scripts` | ✅ **858 pass / 0 fail** |
| `npm run test:units` | ✅ **797 pass / 0 fail** |
| `npm run test:pipeline` | ✅ **119 pass / 0 fail** |
| QA-CHECKLIST guard (`check-checklist-guard.mjs`) | ✅ no generated block edited |
| QA-CHECKLIST coverage guard | ✅ every `@stable`/documented spec referenced |
| Impacted-specs E2E | ✅ the only impacted spec is the one this PR changes — 3 clean runs at `--workers=1 --retries=0` (`expected=2 unexpected=0 flaky=0 skipped=0`), plus 3 force-fail mutations red and a final green run after reverts |

The impacted-specs job is the one worth naming explicitly: this diff touches a single `.spec.ts` and no helper, POM or fixture, so `impacted-specs-by-import.mjs` resolves to exactly that file — which is the file the 3-run burst above covers directly.

### Why the merge still stands

The change adds an assertion and weakens none, is confined to one spec plus its doc and two hand-written checklist bullets, and carries local evidence for every gate the absent lane would have applied. Nothing here depends on a check that did not run.

**Follow-up worth a look, separate from this issue:** a `pull_request` event silently producing no Actions run is a hole in the merge gate, not a property of this PR. If it recurs, it belongs in its own `qa-infra` issue — a PR that merges green-by-absence is indistinguishable from one that merges green.

PR #1652, opened 2026-09-01
fix(i18n): create the missing-key gap instead of borrowing it from the bundle (#1646)

### LINKED ISSUE #1646: [Daily #1642] locale-resilience — the English-fallback probe key no longer falls back, and the spec's own expiry hint is the first thing to check
Spun out of daily-failure triage #1642 (run [33410643882](https://github.com/oriontech-me/langflow-e2e/actions/runs/33410643882), 2026-08-31). First failure of this test since the spec landed on 2026-08-21 (PR #1541); it was green on the 7 dailies in between.

**Upstream:** _not filed_

## Symptom

The test asserts that a key missing from the `pt` bundle falls back to English while its siblings stay
translated, and probes that with the label `Vector Database` (`memory.dbProviderLabel`). All 3 attempts failed
the `toBeVisible` on `getByRole('dialog').getByText('Vector Database').first()` with `element(s) not found`
after 15 s. The spec carries its own expiry hint in the failure message: if upstream has since **translated**
that key, the probe is no longer a missing key and the test is measuring the wrong thing.

|---|---|---|
| `tests/tests-automations/regression/i18n/locale-resilience.spec.ts:241` ("a missing key falls back to English beside siblings the bundle translates") | `getByRole('dialog').getByText('Vector Database').first()` | `Error: Vector Database did not fall back to English — if upstream has translated memory.dbProviderLabel, re-measure by diffing the en translation object in the container's langflow/frontend/assets/index-*.js against assets/pt-*.js and pick` |

Single spec, single test. Filed separately from the run's other `toBeVisible`/`toBe` failures on purpose: the
signature here is specific (the spec's own assertion message, not a bare matcher failure), and it shares no
locator, step or area with them. Grouping on the generic half would be the signature collision the
enrich/create rule warns about.

This surfaced on a **mass-failure day**: 10 hard failures crossed the guard threshold of 5, so the workflow
auto-removed **no** `@stable` tag and the tags on this run survived by default.

**The triage's verdict on the day is that it was environmental IN PART, and demonstrably not in whole.**
The in-run liveness recorder measured a real mid-run wedge — **14 outages, 16.7 min of unreachable backend,
on 4 of 4 shards**, shard 4 alone at 41.9% of its 24.2-minute span across 8 windows of 8–104 s. Cross-referencing
each failing attempt's window against that shard's `backend-liveness.jsonl` (probe share down, ±20 s padding):
4 of the 10 hard failures had **every** attempt inside a heavy outage — `api-health-check.spec.ts:7` at
50/100/100%, `model-provider-model-toggle.spec.ts:296` at 44/49%, `google-provider.spec.ts:165` at 48/47/44%,
`mcp-client-agent-gemini-tool-regression.spec.ts:139` at 40/46/47%. The remaining 6, including the one(s) on
this issue, had their **final** attempt with the backend measured at **0–3% down**. Shard 1 measured 21% of its
probes down and passed every test, so a high down-share is on its own not a cause.

**That is a verdict on the day, not on this cluster.** The environmental signal is consistent with several of
these symptoms and must be ruled out during investigation — it is not asserted as the cause here.

**Read specific to this cluster (descriptive):**

- Measured at **0% / 0% / 3%** of probes down across the three attempts, on shard 2. The wedge does not explain
  it.
- This is the one failure of the run whose spec **predicted its own failure mode in writing** and left the
  re-measurement procedure in the error string. That makes "the probe key expired" the cheapest branch to test
  first — but it is a branch, not a conclusion: an actual regression in the fallback mechanism, or a change to the
  dialog, produces the same `element(s) not found`.
- The sibling spec in the same area, `i18n/language-selection.spec.ts:229`, also failed this run — as a **flake**
  that recovered on retry, on a different signature (`expect(locator).not.toHaveAttribute`). Not folded in here:
  it recovered, and the flake criterion does not make it actionable.
- The spec's own comment at `:245` records that on 1.12.0.dev33 all six non-English bundles were missing the same
  five keys, which dates the measurement the probe was chosen from.

Because the day-level verdict above judges these failures **non-environmental** (final attempt against a backend
measured 0–3% down), `@stable` was **not** kept: the workflow removed nothing on a guard-tripped run, so the tests
were **manually quarantined at triage** — `@stable` removed **and** `test.fixme` added — so they stop running in
every context (daily, PR impacted-specs gate, full suite) until this issue is worked.

## Investigation directive

Independent branches. Here the spec names the cheapest one, so run it first — but run the others regardless:

1. **Probe expiry (the spec's own hint)** — diff the `en` translation object in the running container's
   `langflow/frontend/assets/index-*.js` against `assets/pt-*.js`. If `memory.dbProviderLabel` is now translated,
   the probe expired: pick another key present only in `en`, and record the date of the new measurement in the
   spec so the next expiry is diagnosable the same way.
2. **Product** — if the key is still missing from `pt`, the fallback mechanism itself is the suspect: check what
   the dialog renders in that slot under `pt` and whether it falls back at all.
3. **Test design** — confirm the dialog is the same dialog (a moved or renamed container gives `element(s) not
   found` with the fallback working perfectly).
4. **Environment** — ruled out by measurement (0/0/3% down). Do not attribute without new evidence.

## Deliverables (Done when)

- [ ] Whichever branch resolves it, the spec records the date and the container build the probe key was measured against, so the next expiry is diagnosable without re-deriving the procedure.
- [ ] If the fallback mechanism itself regressed, the upstream ticket is filed and linked in the **Upstream:** line above.

rafaelgiln:
**Fixes #1646.** Closes #1646.

## Problem

The §18.2 fallback test failed **3/3 attempts** in daily [33410643882](https://github.com/oriontech-me/langflow-e2e/actions/runs/33410643882) (2026-08-31) — `getByRole('dialog').getByText('Vector Database')` not found after 15 s. First failure since the spec landed on 2026-08-21 (#1541); green on the 7 dailies in between.

The spec **predicted this in writing** and left the re-measurement procedure in its own error string. That procedure works, and the answer is that the probe expired. Measured on the current nightly `1.12.0.dev44`, diffing the container's shipped assets exactly as the hint says:

| key | `en` | `pt` |
|---|---|---|
| `memory.dbProviderLabel` | `Vector Database` | **`Banco de dados vetorial`** |
| `memory.dbProviderDescription` | `Where this memory base stores vectors…` | **`Onde esta base de memória armazena os vetores…`** |

Both keys are now present in **all six** non-English bundles. There is nothing left to fall back for — the dialog opens and renders, in Portuguese, which is why the English matcher found nothing. **The fallback mechanism is intact**, so branch 3 of the issue's directive (a moved or renamed dialog) is refuted by the same observation, and branch 4 (environment) was already ruled out by the issue's own liveness cross-reference (0% / 0% / 3% of probes down across the three attempts, on shard 2, while shard 1 measured 21% down and passed everything).

Re-measured **5/5 failing (100%, 0 environment voids)** on `1.12.0.dev44` before the fix, so this is a hard deterministic failure and not a flake.

The gap set did not merely shrink, it **turned over**: 5 en-only keys on `1.12.0.dev33` — the measurement the probe was chosen from — against **17** on `dev44`, none of them these two.

## Fix

The load-bearing number is the **lifetime**: landed 08-21, green on 7 dailies, red on 08-31. **Ten days.** Picking another en-only key from the current 17 restarts the same clock, so the design changes instead.

**The test now creates the gap rather than borrowing it.** It intercepts the separately served `pt-*.js` chunk (measured: `HTTP 200`, 146 373 bytes), strips the two `"key":"value"` pairs from the body, and fulfils with the patched text before the browser parses it.

**Every assertion is unchanged** — same dialog, same English strings, same Portuguese siblings (`Criar Memória`, `Tamanho do lote`), same raw-key check. Only the precondition moves under the test's control. It also exercises `fallbackLng` more directly than a bundle gap upstream may close at any moment ever did: a gap in the shipped bundle was never §18.2's contract, the fallback behaviour is.

Two things are added, both load-bearing:

- **A strip-count gate.** If a future build drops these keys from `pt` again, the strip matches nothing while the dialog still renders English, and the test would pass without ever exercising its own mechanism — the silent-pass shape #1012 exists to stop. The test asserts **exactly two** pairs removed from a chunk it really intercepted, and fails naming the count. This is asserted *before* the dialog is opened.
- **A replacement failure message.** The old one told the reader to re-measure the bundle gap, which is no longer how the gap is produced; the new one names the interception and the count, so the next failure points at the route rather than at a phantom i18n regression.

The **decoy rule is kept and re-framed**: it now governs which key may be *removed*, since a key whose call site supplies its own inline `defaultValue` (`shortcuts.modifierOnly`) proves nothing about `fallbackLng` whether the gap is the bundle's or the test's.

`page.route` was chosen over patching a local bundle because the chunk is a real served asset — the interception is the only point where the test can control what i18next sees without touching the image.

## Scope note

- **Quarantine lifted**, which is the issue's own deliverable: `test.fixme` back to `test` and `@stable` back in the tag array, reverting exactly what triage commit `377a251b` applied.
- Tests 1 and 2 of the file (the normaliser ladder and the browser-locale axis) are **untouched** — they are in scope for validation and force-fail only because the whole-file rule says so.
- No new helper, no new POM, no new tag. No LLM and no provider key: test 3 opens the Create Memory modal and never selects an embedding model.
- Cleanup unchanged — one empty flow per test via `POST /api/v1/flows/`, editor unmounted then deleted id-scoped in `afterEach`.
- **Nothing filed upstream**, deliberately: translating a key is correct product behaviour, not a regression, so the issue's `Upstream: not filed` line stands.
- The `QA-CHECKLIST.md` §18 bullet is updated because its text had become false — it claimed the two keys are ones "only `en` carries". Manual Part II bullet only; no generated block touched.

## Validation (nightly `1.12.0.dev44`, `--retries=0 --workers=1`)

- `npm run typecheck` ✅ 0 errors · `npm run lint` ✅ 0 errors
- QA-CHECKLIST guard ✅ (`no auto-generated blocks were edited`) · coverage guard ✅ (`every @stable and every documented spec is referenced by a Part II bullet`)
- **Pre-fix baseline: 5/5 failed (100%), 0 voided on environment signatures** — recorded on the unmodified spec with the quarantine lifted, ~20–23 s per run
- Determinism burst, whole file — **3/3 clean**, `expected=3 unexpected=0 flaky=0 skipped=0`
- Final green after reverts — **1/1 clean**, `expected=3`, ~21 s
- Zero `🚨 Backend Error` ✅ · zero orphan flows on the instance ✅
- `--trace=on` not run: it hangs locally well beyond the template family (#518). Covered by the bursts and the force-fails instead.

### Force-fail — 3/3 executed, scope is the whole file

- **FF:** `stripPtKeys` builds a regex matching nothing, so the chunk is served untouched — the pre-#1646 precondition → *a missing key falls back to English…* **failed** (`unexpected=1`). It failed on the **strip-count gate**, which is the proof that the new guard fires rather than the visible assertion catching it downstream.
- **FF:** the ladder assertion demands the wrong resolved language on every row (`en` ↔ `pt`) → *the application boots into a shipped language…* **failed** (`unexpected=1`)
- **FF:** the load-bearing assertion demands `localStorage.languagePreference === "nb-NO"`, i.e. that the browser locale *did* become a stored preference → *the application boots in English and never adopts…* **failed** (`unexpected=1`)
- **Revert proven:** `git diff | grep -c FF-MUTATION` → **0**, followed by the final green run above.

### DIFF
--- a/QA-CHECKLIST.md
+++ b/QA-CHECKLIST.md
@@ -1009,9 +1009,12 @@
       → i18n/locale-resilience.spec.ts
 - [x] A locale bundle missing individual keys falls back to English for those keys instead of
       failing the render — asserted in the Create Memory modal under `pt`, on the two
-      `memory.dbProvider*` keys only `en` carries (`shortcuts.modifierOnly` is a decoy: its
-      call site passes an inline `defaultValue`, so it renders English even with the
-      fallback broken)
+      `memory.dbProvider*` keys the test itself strips from the served `pt` chunk. The gap is
+      created rather than borrowed since #1646: upstream translated those keys 10 days after
+      the spec landed, so a probe chosen from the shipped bundle's gaps expires on upstream's
+      schedule (`shortcuts.modifierOnly` remains a decoy and must not be the key removed — its
+      call site passes an inline `defaultValue`, so it renders English even with the fallback
+      broken)
       → i18n/locale-resilience.spec.ts
 
 ---
--- a/docs/i18n/locale-resilience.md
+++ b/docs/i18n/locale-resilience.md
@@ -1,6 +1,6 @@
 # i18n — Locale resilience
 
-**Last validated:** Langflow 1.12.x (measured on nightly `1.12.0.dev33`)
+**Last validated:** Langflow 1.12.x (measured on nightly `1.12.0.dev44`)
 
 ---
 
@@ -25,12 +25,14 @@ individual keys.
    reads `en`, and `localStorage.languagePreference` is **unset** — the browser
    locale never becomes a language preference.
 3. **should fall back to English for a key the active bundle lacks, next to
-   siblings it translates** — with the interface in Portuguese, the flow
-   editor's Create Memory modal renders `Criar Memória`, `Nome` and
-   `Tamanho do lote` in Portuguese while the `Vector Database` label and its
-   description — two keys the Portuguese bundle does not carry — render their
-   English text. No raw i18n key (`memory.dbProviderLabel`) appears anywhere in
-   the dialog.
+   siblings it translates** — with the interface in Portuguese, and with
+   `memory.dbProviderLabel` and `memory.dbProviderDescription` **removed from the
+   `pt` chunk the browser is served**, the flow editor's Create Memory modal
+   renders `Criar Memória`, `Nome` and `Tamanho do lote` in Portuguese while
+   those two keys render their English text (`Vector Database` and its
+   description). No raw i18n key (`memory.dbProviderLabel`) appears anywhere in
+   the dialog. The gap is **created by the test**, not borrowed from whatever the
+   shipped bundle happens to be missing that week — see **Validation criterion**.
 
 ---
 
@@ -83,9 +85,11 @@ id-scoped.
   becomes a boot risk again and test 1's table becomes reachable from the
   browser. It uses `withLocale()` — the sanctioned opt-in — never a bare
   `test.use({ locale })`.
-- **Test 3's vehicle is chosen, not convenient — and one obvious candidate is a
-  decoy.** On `1.12.0.dev33` all six non-English bundles are missing the **same
-  five** keys that `en` carries: `shortcuts.modifierOnly`, `memory.backendLabel`,
+- **Test 3's vehicle is chosen, not convenient, and one obvious candidate is a
+  decoy.** This still governs which key the test may REMOVE: a key whose call
+  site supplies its own English default proves nothing about `fallbackLng`,
+  whether the gap is the bundle's or the test's. On `1.12.0.dev33` all six
+  non-English bundles were missing the **same five** keys that `en` carries: `shortcuts.modifierOnly`, `memory.backendLabel`,
   `memory.dbProviderLabel`, `memory.dbProviderDescription` and
   `memory.dbProviderNotConfigured`. `shortcuts.modifierOnly` **must not be
   used**: its call site passes an inline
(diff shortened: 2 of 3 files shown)


rafaelgiln:
## CI is green, but the E2E lane recorded `1 flaky` — disclosed, not swept

All 7 checks passed and the job exited 0, but the run summary reads `2 passed` while `Running 3 tests` — the `1 failed` line was lost to `\r` interleaving, which is exactly the mis-parse `langflow-e2e` warns about. Reading the block instead of the summary:

```
1) locale-resilience.spec.ts:116:7 › the application boots into a shipped language …
   Error: expect(received).toBe(expected)
   Call Log: - Timeout 30000ms exceeded while waiting on the predicate
     at helpers/flows/open-new-flow-templates-modal.ts:106
       at dismissWelcomeOverlayAndWaitForModal (…:106:3)
       at addFlowToTestOnEmptyLangflow (…:11:3)
       at awaitBootstrapTest (…:26:5)
1 flaky
```

Attempt 0 failed, the retry passed. Alongside it: `📋 Found 1 backend error(s)` — `{"detail":"Name must be unique"}` on flow creation, advisory by contract.

**It is not this PR's change.** The failure is in the shared bootstrap helper, reached from test 1, which lives in a *different* `describe` block from the one this PR touches — the `page.route` added here is installed in the third block's `beforeEach` only and cannot reach it. Test 1's body is untouched by this diff; it appears here solely because the whole-file rule puts it in scope for validation and force-fail.

**It is pre-existing and cross-spec.** Searching the per-run Playwright JSON of the recent dailies for the `overlayOrModalAppeared` / `dismissWelcomeOverlayAndWaitForModal` signature returns hits on **4 separate dailies** across **7 different spec files** — `model-provider-model-toggle`, `agent-multimodal-image-input`, `agent-system-prompt`, `agent-tool-name-validation` (×2), `agent-config-persistence`, `memory-history-regression`, `agent-max-iterations` — and **none** of them is `locale-resilience.spec.ts`. The lane ran this file with 2 workers, and `Name must be unique` is the shape of two workers bootstrapping the same template concurrently.

**Locally the file is clean 4/4 at `--workers=1 --retries=0`** on nightly `1.12.0.dev44`: the 3-run determinism burst (`expected=3` each) plus the final green after the force-fail reverts.

Merging on green. The bootstrap flake needs its own issue — it belongs to `open-new-flow-templates-modal.ts`, not to §18.2, and folding it in here would be the signature collision the triage rules warn about.

Issue #1666, opened 2026-09-01
[Daily #1665] collect-models confirms 0 of 74 enable writes on all four shards, warns, and the whole run proceeds on the MIN_DEFAULT_MODELS cold path


Victor-w-Madeira:
Spun out of daily-failure triage #1665 (run [33511210195](https://github.com/oriontech-me/langflow-e2e/actions/runs/33511210195), 2026-09-01). Run-level mechanism, not a per-spec assertion: it is what put every spec on the cold path this run. The picker/panel disagreement it produced is tracked separately on #1649.

**Upstream:** _not filed_

## Symptom

`collect-models` reported **0 of 74** enable writes confirmed by the server, on **all four shards**, for **every** provider — and only as a `⚠️`, so the run proceeded with each provider left on its `MIN_DEFAULT_MODELS = 5` default (the *cold path*).

Identical on 4/4 shards, in both sweeps of each shard (the `Collect models` pre-flight and `globalSetup`):

```
⚠️  [OpenAI] collect-models: the server confirms only 0 of 36 model(s) this sweep tried to enable.
⚠️  [Anthropic] collect-models: the server confirms only 0 of 8 model(s) this sweep tried to enable.
⚠️  [Google Generative AI] collect-models: the server confirms only 0 of 30 model(s) this sweep tried to enable.
```

That server-side confirmation is new — it landed with #1651 on 2026-08-31 23:18 — so **this is the first daily that could report it at all**. Whether the condition is new is therefore unknown, and the issue must not assume either way.

The spec row below is the failure whose own error *is* the cold path: on attempt 1 it spent the full 60 s `model_model` budget that #1651 raised (from 15 s) precisely because the correct flush path costs ~30 s. Its recorded last-attempt signature is the page-entry barrier, from a third attempt on a backend that was measurably degraded — both attempts are in the table's single signature cell by construction, which is why the prose above matters.

| Spec (line) | Waits for | Signature |
|---|---|---|
[...]

Issue #1589, opened 2026-08-25
The infra-signature exemption reads only the last attempt — an intermittent wedge un-exempts its own collateral


Victor-w-Madeira:
Spun out of daily-failure triage #1588 (run [32827671203](https://github.com/oriontech-me/langflow-e2e/actions/runs/32827671203), 2026-08-25). Not a per-spec cause issue — the subject is the exemption's attempt-selection rule, which is why it carries `qa-infra` and not `daily-failure`. Adjacent to #1549 but separable: that issue is about the barrier's own *message*, this one is about **which attempt** the exemption reads.

## What was measured

The infra-signature exemption (#1031/#1310) classifies a hard failure from the **last** failed attempt. Both `scripts/remove-stable-from-failures.ts` and `scripts/append-weekly-history.mjs` say so, and give the reason:

> The error of the LAST failed attempt — the same result `build-run-payload.mjs` picks for `error_signature`, so the exemption and the history file talk about the same attempt. Last, not first: retries are what a wedge burns, and the final attempt is the one that decided the verdict.

On run 32827671203 that rule met an **intermittent** wedge, and the reasoning does not hold there. All 7 hard failures were on shard 3, the only red job; its in-run liveness recorder measured **156 of 894 probes down (17.4 %)** across 15 outage windows plus 45 sub-threshold blips, with the median `/api/v1/version` at **142 ms against shard 2's 8 ms**. The wedge did not burn the retries — it *cycled through* them, in windows as short as 6–8 s inside a 30 s budget.

Running `classifyInfraError()` over **every** attempt of each hard failure:

| Spec:line | a0 | a1 | a2 | Exempt as scored (last) |
|---|---|---|---|---|
| `core-functionality/llm-agents/language-model-regression.spec.ts:246` | **`api-request-timeout`** | — | — | no |
| `core-functionality/llm-agents/language-model-regression.spec.ts:270` | — | — | — | no |
| `core-functionality/llm-agents/llm-invalid-api-key-ui.spec.ts:20` | — | **`api-request-timeout`** | — | no |
| `core-functionality/llm-agents/llm-invalid-api-key-ui.spec.ts:88` | — | **`api-request-timeout`** | — | no |
| `core-functionality/llm-agents/memory-history-regression.spec.ts:237` | **`api-request-timeout`** | — | — | no |
| `core-functionality/llm-agents/memory-history-regression.spec.ts:346` | — | — | — | no |
[...]

Issue #1648, opened 2026-08-31
[Daily #1642] provider-item-OpenAI never becomes interactable — one hard failure inside the wedge, and two flakes on a clean PR lane with the backend healthy


rafaelgiln:
Spun out of daily-failure triage #1642 (run [33410643882](https://github.com/oriontech-me/langflow-e2e/actions/runs/33410643882), 2026-08-31). Filed after the fact: the daily triage first parked this on #1589 as un-exempted wedge collateral, and the evidence that reopened it arrived from the validation lane of the triage's own quarantine PR #1647.

**Upstream:** _not filed_

## Symptom

Three failures on 2026-08-31 wait on the same control — `getByTestId('provider-item-OpenAI')`, the OpenAI row of the
provider list — and time out because it never becomes interactable. The element is reached from two different call
sites, which is what makes it a control problem rather than a spec problem:

- `model-provider-model-toggle.spec.ts` reaches it through its own `openProviderModelList()` helper at line 181,
  where the daily recorded `TimeoutError: locator.waitFor: Timeout 10000ms exceeded` on `... to be visible` for
  attempt 0 (the Signature cell below carries the last attempt's line, as recorded in the history row).
- The two context-id specs reach it through the shared `tests/helpers/provider-setup/setup-openai.ts` (lines 31 and
  107) via `SimpleAgentTemplatePage.load()`, where it timed out as `locator.click: Timeout 20000ms exceeded`.

A shared helper and a per-spec helper failing on the same testid is the observation this issue exists to explain.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/tests-automations/regression/core-functionality/llm-agents/model-provider-model-toggle.spec.ts:296` ("disabling a model removes it from a component model dropdown") | `getByTestId('provider-item-OpenAI') to be visible (10s)` | `TimeoutError: locator.click: Timeout 20000ms exceeded.` |
| `tests/tests-automations/regression/core-functionality/llm-agents/agent-context-id-continuity.spec.ts:407` ("agent run persists every session message tagged with the custom context_id [openai / gpt-4o-mini]") | `getByTestId('provider-item-OpenAI') click, via setupOpenAI (20s)` | `TimeoutError: locator.click: Timeout 20000ms exceeded.` |
[...]

PR #1672, opened 2026-09-01
fix(provider-setup): name the stalled enable write instead of blaming the model picker (#1649)

### LINKED ISSUE #1649: [Daily #1642] MODEL_PICKER_DEFECT — the picker offers exactly 5 options per provider while the panel reports 49 enabled (5 occurrences, all on 2026-08-31)
Split out of #1648 during its DEBUG phase (run [33417956992](https://github.com/oriontech-me/langflow-e2e/actions/runs/33417956992), PR #1647's validation lane, 2026-08-31). #1648's symptom table listed this failure as a `provider-item-OpenAI` click timeout; reading the lane's log shows it is not — it never reached that control. It gets its own issue rather than staying miscounted in #1648.

**Upstream:** _not filed_

## Symptom

`MODEL_PICKER_DEFECT` — the assertion #1461 added so a picker/panel disagreement is a failure and not a skip — fired **five times on 2026-08-31** and on no other day in the 25 dailies scanned. The provider panel reports `gpt-4o-mini` enabled; the model picker does not offer it, and offers only **5** OpenAI options against a panel showing 49.

The option counts are the informative part: every daily occurrence enumerated **exactly 5 per provider**.

| Date | Where | Spec | Options enumerated |
|---|---|---|---|
| 2026-08-31 | daily [33410643882](https://github.com/oriontech-me/langflow-e2e/actions/runs/33410643882) | `tests/tests-automations/regression/core-functionality/llm-agents/agent-component-regression.spec.ts` ("agent interaction suite") | 15 (Anthropic: 5, Google Generative AI: 5, OpenAI: 5) |
| 2026-08-31 | daily [33410643882](https://github.com/oriontech-me/langflow-e2e/actions/runs/33410643882) | `tests/tests-automations/regression/core-functionality/llm-agents/agent-current-date-tool.spec.ts` ("toggle OFF: the date tool is removed from the agent's toolkit") | 15 (Anthropic: 5, Google Generative AI: 5, OpenAI: 5) |
| 2026-08-31 | daily [33410643882](https://github.com/oriontech-me/langflow-e2e/actions/runs/33410643882) | `tests/tests-automations/regression/core-functionality/llm-agents/agent-input-sources.spec.ts` ("input via the Agent's direct field drives the agent response") | 15 (Anthropic: 5, Google Generative AI: 5, OpenAI: 5) |
| 2026-08-31 | daily [33410643882](https://github.com/oriontech-me/langflow-e2e/actions/runs/33410643882) | `tests/tests-automations/regression/core-functionality/llm-agents/agent-max-iterations.spec.ts` ("agent stops when max iterations is reached") | 15 (Anthropic: 5, Google Generative AI: 5, OpenAI: 5) |
| 2026-08-31 | PR #1647 lane [33417956992](https://github.com/oriontech-me/langflow-e2e/actions/runs/33417956992) | `tests/tests-automations/regression/core-functionality/llm-agents/agent-context-id-continuity.spec.ts:407` ("agent run persists every session message tagged with the custom context_id [openai / gpt-4o-mini]") | 31 (Anthropic: 5, Google Generative AI: 21, OpenAI: 5) |

All five were `attempt#0` and all five **recovered on retry**, so every lane stayed green and no `@stable` tag moved. The row below is the one #1648 listed; the other four are named here because they carry the same signature and were invisible until this scan.

|---|---|---|
| `tests/tests-automations/regression/core-functionality/llm-agents/agent-context-id-continuity.spec.ts:407` ("agent run persists every session message tagged with the custom context_id [openai / gpt-4o-mini]") | `selectPinnedModelOption` reading the open model picker, via `tests/helpers/provider-setup/setup-openai.ts:107` | `Error: MODEL_PICKER_DEFECT: "gpt-4o-mini" is ENABLED in the provider panel (llm-toggle-gpt-4o-mini) for OpenAI but the model picker does not offer it — 31 option(s) enumerated (Anthropic: 5, Google Generative AI: 21, OpenAI: 5). The two sources disagree, so the model is not absent: either the picker did not refresh after the panel closed, or the option list is filtered. Reported as a FAILURE, not a skip (#1461).` |

The signature is copied from the run log, not from `reports/daily-history.jsonl` — this occurrence is on a PR lane, which writes no history row.

Five occurrences, one day, one verbatim signature, one assertion, one call site (`setup-openai.ts:107` → `selectPinnedModelOption` → `model-option.ts:410`), and a shared numeric fingerprint: **OpenAI enumerates 5 options** in all five, and four of them enumerate 5 for *every* provider. A per-provider cap of exactly 5 is not the shape of random partial data.

There is also a **separation** argument, and it is why this issue exists at all: #1648 grouped the fifth occurrence with two `provider-item-OpenAI` timeouts on the shared claim of "the same testid, two call sites". This failure touches neither.

1. **Different control.** `provider-item-OpenAI` is the provider row in the list; this fails on the **model picker's option list**, read by `enumerateModelOptions` after the panel is closed.
2. **Different call site.** `setup-openai.ts:31` (the provider row) vs `setup-openai.ts:107` (`selectPinnedModelOption`). #1648's "clicks at lines 31 and 107" reads 107 as a second click on the provider row; it is not a click at all.
3. **Different failure mode.** The provider-row failures are locator timeouts whose call log never resolves the element. This one **resolved everything it asked for** and then compared two populated sources that disagreed — the assertion fired on content, not on absence.

The lane this occurred on was **not healthy**, and that context belongs here even though it is not a conclusion. `.github/actions/wait-for-backend` reported `Backend answered /api/v1/version on attempt 8 after 92s` (`last failure TIMEOUT`) immediately before the specs ran, and `/api/v1/auto_login` timed out at 20 s again 35 s into the run. A picker that enumerates 5 of 49 OpenAI options is consistent with a partially-answered catalog — but so is a genuine refresh-ordering defect in the picker, which is exactly what #1461 wrote this assertion to catch, and the counts (`Anthropic: 5, Google Generative AI: 21, OpenAI: 5`) are not obviously a truncation of anything.

The test **recovered on retry**, so the lane was green and `@stable` is unchanged.

## Investigation directive

Independent branches, product first:

1. **Product** — on the current nightly, close the provider panel with N models enabled for OpenAI and enumerate the picker repeatedly. Establish whether the picker ever settles at a count below the panel's, and how long it takes to converge.
2. **Ordering** — `setup-openai.ts` closes the panel (step 6) and then waits for `model_model` to be visible before clicking it. Confirm whether visible-and-clickable is sufficient, or whether the option list can be populated from a pre-close snapshot.
3. **Degraded catalog** — reproduce with `GET /api/v1/models` delayed or partially answered and compare the enumerated counts against this occurrence's. #1648 proved that delaying that request reproduces its own symptom byte-for-byte; the same harness applies here.
4. **Environment** — the 92 s wedge is context, not a verdict. Do not attribute without measuring branch 1.

## Deliverables (Done when)

- [ ] Root cause confirmed (product regression vs. test/wait-strategy vs. environment), with evidence on the current nightly.
- [ ] The spec passes reliably (multiple clean `--retries=0` runs).
- [ ] **Quarantine lifted** in the fix PR — remove `test.fixme` **and** restore `@stable`, re-validated per `CONTRIBUTING.md`. *(Nothing to lift: this test is not quarantined — the `test.fixme` at `agent-context-id-continuity.spec.ts:377` belongs to #1643.)*
- [ ] Whether `MODEL_PICKER_DEFECT` has fired on earlier dailies is established from the per-run Playwright JSON, not from `error_signature` — #1648 measured that the history stores only the normalized first error line, which is why a control can recur for a month invisibly (#1626).

rafaelgiln:
Closes #1649.

## Problem

`MODEL_PICKER_DEFECT` fired 3× again on the first daily after #1651 ([33511210195](https://github.com/oriontech-me/langflow-e2e/actions/runs/33511210195), 2026-09-01), one of them a hard failure, with the numeric fingerprint this issue was opened on intact — `15 option(s) enumerated (Anthropic: 5, Google Generative AI: 5, OpenAI: 5)`. Only the provider moved: `gpt-4o-mini`/OpenAI → `gemini-3.5-flash`/Google.

The cause is not the picker, and #1651's gate already knew it. That gate **gave up eight times** on that run (shards 1, 2 and 4), every one reading the same shape:

```
⚠️  provider panel: the model-toggle batch did not settle in time —
    30 toggle(s) clicked, 1 write(s) started, 0 finished.
```

`POST /api/v1/models/enabled_models` was issued and had not answered 90 s later. The listener counts `requestfailed` too, so that is a hang, not a network error. The gate then closes the panel anyway — by design: it warns, it does not throw — the close-path flush never calls `refreshAllModelInputs`, and the picker keeps the pre-toggle enabled set, which on a freshly configured provider is `MIN_DEFAULT_MODELS = 5`.

**The picker is correct there; the panel is what lies.** `aria-checked` is `useModelToggleQueue`'s optimistic cache — it flips at click time, before any request — and it is exactly the source feeding the loud message's *"is ENABLED in the provider panel"*. Corroborated on the same run by `collect-models`' own server-side read (`0 of 30/36/8` enable writes confirmed, on all four shards) and by `5` being `MIN_DEFAULT_MODELS` (`lfx/base/models/model_utils.py`).

So the gate observed the real cause and printed it to a log line that no failure message, no `error_signature` and no triage dataset correlates, while the failure 90 s later named two hypotheses nobody had measured. That is why this issue carried two verdicts and a reopen.

Two of the six occurrences are the same cause wearing a different face: after a give-up, the `model_model` wait exhausts its 60 s and dies as a bare `locator.waitFor: Timeout 60000ms exceeded` (`agent-current-date-tool.spec.ts:260` attempt 1, `agent-context-id-continuity.spec.ts:407` attempt 0).

**Verdict: `test-defect` — in attribution.** Reproduced causally rather than by re-running, since re-running cannot falsify a timing defect: on a healthy `1.12.0.dev45` container a scout held that POST for 120 s with `page.route`, everything else unchanged, and produced the daily's evidence byte for byte (`36 toggle(s) clicked, 1 write(s) started, 0 finished` → `MODEL_PICKER_DEFECT … 5 option(s) enumerated (OpenAI: 5)`), first attempt.

## Fix

Carry the batch's own observation to whatever fails next, so a stalled write is named instead of the picker.

- **`model-toggle-batch.ts`** — `ToggleBatchResult` now returns `writesStarted`/`writesFinished` (the give-up message already printed them; the caller could not read them, which is the whole of the reopen). New `MODEL_TOGGLE_WRITE_STALLED`, plus two **pure** functions: `writeStallReason()` and `modelTriggerStallMessage()`.
- **`model-option.ts`** — `ResolveContext.toggleWrite` is the third source; a new `write-stalled` verdict is returned **inside** the existing `checked.includes(requested)` branch, after `empty`/`match`/`unmatchable`. `openModelPickerAfterPanelClose()` shares the post-close open across the three setups and re-throws a `model_model` failure as the same named stall, keeping Playwright's original message inside it.
- **`setup-{openai,google,anthropic}.ts`** — capture the batch result they were discarding, pass it as `toggleWrite`, and open the picker through the shared helper. The three blocks were copy-paste, and had already drifted: #1651 landed the same budgets in each with three differently-worded comments.

**Rejected alternatives**, both declined deliberately: failing fast on a give-up (skips the picker read that proves the disagreement, and a write landing at 91 s would be mis-reported), and seeding enabled models through the API instead of the UI toggles (a redesign of the setup contract that overlaps #1666's open decisions). A new `scripts/lib/infra-signature-patterns.json` entry is **not** included — none of its five patterns matches a locator timeout or `PROVIDER_LIST_STALLED` either; that is #1589's gap, and the new message is shaped to be matchable there.

## Scope note

No assertion weakened, no budget raised, nothing became a skip. The message never starts with `MODEL_NOT_AVAILABLE`, so no caller can turn it into a `test.skip`, and `write-stalled` deliberately does **not** return through `absentBehavior: "return"` — that hatch exists for a stale pin from `models.json` (#606), and degrading on an instance that could not accept the write would hide exactly the state this issue was reopened for. A picker miss after a **settled** batch is still `MODEL_PICKER_DEFECT`, unchanged, and an unobserved batch leaves every existing verdict byte-identical (#1012).

**This does not make a saturated day green, and is not meant to.** On such a day these specs still fail — they name the instance instead of accusing the picker. The run-level cause that puts a spec on this path at all (`collect-models` confirming 0 of 74 enable writes) is **#1666**, untouched here.

No `.spec.ts` is created or edited, so no tag changes and no `QA-CHECKLIST.md` bullet moves.

## Validation (nightly `1.12.0.dev45`, `--workers=1 --retries=0`)

- typecheck ✅ (0) · eslint ✅ (0 errors) · QA-CHECKLIST guard ✅ · coverage guard ✅
- `npm run test:units` — **870 passed, 0 failed** (11 new)
- `agent-current-date-tool.spec.ts` — **3/3 clean**, `expected=2 skipped=0 flaky=0`, zero `🚨 Backend Error`. Run 1 was the real cold path (fresh container, no Google credential): 173 s against 34 s for the warm ones.
- `model-provider-model-toggle.spec.ts` — **3/3 clean**, `expected=2 skipped=0 flaky=0`, zero `🚨 Backend Error`.
- Pre-fix baseline: **0/5** (`repro-run`, `google / gemini-3.5-flash`). The defect cannot reproduce on a healthy instance by construction — it requires the write to stall — hence the induced-delay proof above.
- End-to-end red/green pair on the live instance: the identical 120 s hold printed `MODEL_PICKER_DEFECT` before this change and `MODEL_TOGGLE_WRITE_STALLED` after.
- **Force-fail** — the diff carries no `.spec.ts`, so the mechanical gate has nothing to mutate and passes vacuously (the shape #1651 hit too). It was done behaviourally against the 11 new unit tests, three mutations, each reverted, each red on a distinct set:
  - M1 removed the `gave-up` guard in `writeStallReason` → red: *a picker miss after a SETTLED write is still MODEL_PICKER_DEFECT* · *a settled batch is never a stall* · *with no stall the model_model failure is left exactly as it was*
  - M2 made the carrier never recognise a stall → red: *a picker miss after a STALLED write blames the write* · *a gave-up batch yields a reason naming the write* · *the model_model message blames the instance*
  - M3 consulted the stall **before** the picker's own verdicts → red: *a stall never rewrites a verdict about a picker that DOES offer the model* · *a listed model with its toggle OFF stays a setup failure* · *a picker miss after a STALLED write blames the write*

  A first attempt at M2 was discarded rather than banked: it left unreachable code and failed at file level (828 tests, 2 file failures) — a compile error, not the behavioural red the gate exists for. After revert: 0 `FF-MUTATION` markers, 870/870, typecheck 0, lint 0.

## Symptom rows

Every row gets its own verdict; all are `test-defect` (attribution). `agent-current-date-tool.spec.ts:260` is also listed on **#1666**, which owns the run-level cold-path cause; what changes here is only how its attempt-1 `model_model` timeout is reported.

## Out of scope, found on the way

- `tests/helpers/provider-setup/setup-language-model-openai.ts` still closes the panel with no flush wait, and its comment asserts the very claim #1649 disproved. Consequence is milder (it ranks rather than pins, so it can silently select a default model instead of failing) — left out to keep this diff to attribution only.
- `model-provider-model-toggle.spec.ts` had one hard failure and one flake on the 2026-09-01 daily with **no issue opened**; its `getByText('Model Providers')` signature (`helpers/ui/go-to-settings.ts:15`) reproduced once in six local runs, outside this diff and predating it.

### DIFF
--- a/docs/core-functionality/llm-agents/agent-context-id-continuity.md
+++ b/docs/core-functionality/llm-agents/agent-context-id-continuity.md
@@ -1,6 +1,6 @@
 # Agent context_id — continuity between session messages
 
-**Last validated:** Langflow 1.12.x (nightly `1.12.0.dev44`)
+**Last validated:** Langflow 1.12.x (nightly `1.12.0.dev45`)
 
 ---
 
@@ -72,8 +72,13 @@ surface; `@components` — Message History node drives the retrieval assert.
   unflushed batch leaves the model picker on the pre-toggle set — the
   `MODEL_PICKER_DEFECT` this spec reported five times on 2026-08-31 (#1649). The
   wait and its ~30 s refresh budget live in `tests/helpers/provider-setup/`; the
-  measurement is in the agent-area `CLAUDE.md` § 5. **This does not change what
-  this test validates** — it is a precondition of reaching the assertions at all.
+  measurement is in the agent-area `CLAUDE.md` § 5. When the enable write is
+  issued and never answers inside that wait's budget — the 2026-09-01 shape, and
+  what this spec hit again on run 33511210195 — the failure is
+  `MODEL_TOGGLE_WRITE_STALLED` instead: an INSTANCE stall naming the write, not a
+  picker defect, because the picker is then correctly showing the five models the
+  server actually has (agent-area `CLAUDE.md` § 5.1). **Neither changes what this
+  test validates** — both are preconditions of reaching the assertions at all.
 
 ---
 
--- a/tests/helpers/provider-setup/model-option.test.ts
+++ b/tests/helpers/provider-setup/model-option.test.ts
@@ -366,3 +366,104 @@ test("with only listedModels observed the message says so instead of claiming EN
   assert.match(verdict.message, /listed by the provider panel/);
   assert.ok(!/is ENABLED in the provider panel/.test(verdict.message));
 });
+
+// --- #1649 (reopened): a stalled write is not a picker defect ---
+//
+// The loud "is ENABLED in the provider panel" verdict reads `aria-checked`, which is
+// `useModelToggleQueue`'s OPTIMISTIC cache: it flips at click time, before any
+// request. So when the batched write never answers, the panel claims the model is on
+// while the server still has only its `MIN_DEFAULT_MODELS` five — and the picker,
+// rendering the server's truth, is the honest source. On the 2026-09-01 daily that
+// state produced four `MODEL_PICKER_DEFECT` failures naming two hypotheses nobody
+// had measured, while the gate had printed the actual cause 90 s earlier.
+//
+// The stall verdict is consulted INSIDE the checked-includes-requested branch, after
+// `empty`/`match`/`unmatchable`: a picker that DID offer the model is never a stall,
+// and a stall must not rewrite a verdict about one.
+
+const STALLED = {
+  clicked: 30,
+  verdict: "gave-up" as const,
+  writesStarted: 1,
+  writesFinished: 0,
+};
+
+test("a picker miss after a STALLED write blames the write, not the picker", () => {
+  const verdict = resolveModelOption("gemini-3.5-flash", [option("OpenAI", "gpt-4o-mini")], {
+    listedModels: ["gemini-3.5-flash", "gemini-3.1-flash-lite"],
+    checkedModels: ["gemini-3.5-flash"],
+    providerLabel: "Google Generative AI",
+    toggleWrite: STALLED,
+  });
+  assert.equal(verdict.kind, "write-stalled");
+  // Loud, and never a skip: the whole point of #1461's assertion survives.
+  assert.ok(!verdict.message.startsWith("MODEL_NOT_AVAILABLE"));
+  assert.match(verdict.message, /^MODEL_TOGGLE_WRITE_STALLED:/);
+  // It must NOT keep asserting the panel's claim as a fact about the server.
+  assert.ok(!/is ENABLED in the provider panel/.test(verdict.message));
+  assert.match(verdict.message, /OPTIMISTIC/);
+  assert.match(verdict.message, /1 write\(s\) started, 0 finished/);
+  // The picker's own counts stay in the message: they are what shows the server had
+  // the five defaults, which is the reading that makes the picker correct.
+  assert.match(verdict.message, /1 option\(s\)/);
+  assert.match(verdict.message, /MIN_DEFAULT_MODELS/);
+  assert.match(verdict.message, /INSTANCE stall/);
+});
+
+test("a picker miss after a SETTLED write is still MODEL_PICKER_DEFECT", () => {
+  // The load-bearing negative. If the stall branch widened to every give-up-shaped
+  // context, the genuine picker/panel disagreement #1461 exists to catch would be
+  // relabelled as an environment problem and stop being investigated.
+  const verdict = resolveModelOption("gemini-3.5-flash", [option("OpenAI", "gpt-4o-mini")], {
+    listedModels: ["gemini-3.5-flash"],
+    checkedModels: ["gemini-3.5-flash"],
+    providerLabel: "Google Generative AI",
+    toggleWrite: { clicked: 30, verdict: "settled", writesStarted: 1, writesFinished: 1 },
+  });
+  assert.equal(verdict.kind, "unmatchable");
+  assert.match(verdict.message, /^MODEL_PICKER_DEFECT:/);
+  assert.match(verdict.message, /is ENABLED in the provider panel/);
+});
+
+test("an unobserved batch leaves every existing verdict byte-identical", () => {
+  // Callers that never ran the gate (and every caller before this change) pass no
+  // `toggleWrite`. An unobserved source must not be read as a negative one (#1012),
+  // so the message they get is the one they got before.
+  const withoutContext = resolveModelOption(
+    "gemini-3.5-flash",
+    [option("OpenAI", "gpt-4o-mini")],
+    {
+      listedModels: ["gemini-3.5-flash"],
+      checkedModels: ["gemini-3.5-flash"],
+      providerLabel: "Google Generative AI",
+    },
+  );
+  assert.equal(withoutContext.kind, "unmatchable");
+  assert.match(withoutContext.message, /^MODEL_PICKER_DEFECT:/);
+});
+
+test("a stall never rewrites a verdict about a picker that DOES offer the model", () => {
+  // `match` and `unmatchable`-by-identity are decided before the panel sources are
+  // consulted at all. A stalled write says nothing about a model the picker is
+  // offering, and reporting an instance stall there would hide the #1459 class of
+  // suite defect (identity no longer resolving).
+  const matched = resolveModelOption("gpt-4o-mini", [option("OpenAI", "gpt-4o-mini")], {
+    checkedModels: ["gpt-4o-mini"],
+    providerLabel: "OpenAI",
+    toggleWrite: STALLED,
+  });
+  assert.equal(matched.kind, "match");
+});
+
+test("a listed model with its toggle OFF stays a setup failure even under a stall", () => {
+  // The optimistic cache is what makes a stall look enabled; a toggle reading OFF is
+  // therefore NOT the stall's signature, and `MODEL_NOT_ENABLED` — which already
+  // names the debounce cause — remains the right verdict.
+  const verdict = resolveModelOption("gemini-3.5-flash", [option("OpenAI", "gpt-4o-mini")], {
+    listedModels: ["gemini-3.5-flash", "gemini-3.1-flash-lite"],
+    checkedModels: ["gemini-3.1-flash-lite"],
+    providerLabel: "Google Generative AI",
+    toggleWrite: STALLED,
+  });
+  assert.equal(verdict.kind, "not-enabled");
+});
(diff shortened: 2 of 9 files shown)


rafaelgiln:
### CI: green, and the two flakies read faithfully

All 7 checks passed. The E2E lane really ran — `Running 45 tests using 2 workers` → **38 passed, 2 flaky, 5 skipped (11.0m)** — so this is not a green that executed nothing. Both flakies recovered on retry; neither is caused by this change, and one of them will *look* like it is, which is why it is written down here.

**1. `agent-context-id-continuity.spec.ts:407`** — failed at `waitForProviderRow` (`provider-list-state.ts:237`, from `setup-openai.ts:40`). That is #1648's `PROVIDER_LIST_STALLED` path: the provider list never arrived. It happens *before* anything this PR touches (line 40 vs line 103).

**2. `language-model-regression.spec.ts:246`** — failed at `openModelPickerAfterPanelClose` (`model-option.ts:518`):

```
TimeoutError: locator.waitFor: Timeout 60000ms exceeded.
Call log:
  - waiting for getByTestId('model_model') to be visible
```

The stack frame is new, the behaviour is not. Before this PR the identical wait lived inline at `setup-openai.ts:104` with the identical 60 s budget and the identical locator; the refactor moved it into a shared helper and changed nothing about it. What the log shows is in fact the branch this PR adds working as designed: the catch consults `writeStallReason`, which returned `null` because the batch did **not** give up, so Playwright's own error is re-thrown untouched, call log included — *"No stall to blame: a trigger that never returns on a healthy flush is a real defect and must keep surfacing as Playwright's own error"*. Had the batch stalled, the same run would have printed `MODEL_TOGGLE_WRITE_STALLED` naming the write instead.

So the honest reading of #2 is: an instance that did not bring the trigger back within 60 s while the toggle batch had settled — the class of `model_model` timeout that has existed since #1651 raised that budget from 15 s, unchanged by this PR and deliberately not widened by it.

Local validation on `1.12.0.dev45` was `--retries=0`: 3/3 clean on `agent-current-date-tool.spec.ts` and 3/3 clean on `model-provider-model-toggle.spec.ts`, plus 870/870 units.

Issue #1679, opened 2026-09-02
[Daily #1676] enableAndSettleModelToggles still clicks every model toggle — the batch outlives its 90 s budget by the per-model validation #1666 measured, and takes the worker down with it


Victor-w-Madeira:
Spun out of daily-failure triage #1676 (run [33630411848](https://github.com/oriontech-me/langflow-e2e/actions/runs/33630411848), 2026-09-02). Third consecutive daily of one cause under three different diagnostic names, so the same-signature recurrence criterion never fired and all four came back `actionable: false`. Not quarantined at triage: the criterion did not require it, and the fix is in our own helper rather than in the product, so quarantining four `@stable` tests would hide the largest open lever on #1077 (the triage recorded that judgement explicitly).

**Upstream:** _not filed_

## Symptom

Four `@stable` agent specs failed their first attempt inside `setupGoogle`, all with the identical message:

```
MODEL_TOGGLE_WRITE_STALLED: the model picker's trigger never became usable after the provider panel
closed for Google Generative AI — the enable write never answered — 29 toggle(s) clicked, 1 write(s)
started, 0 finished before the flush budget expired, so POST /api/v1/models/enabled_models did not
land. The panel's post-close refresh runs in that write's own onSettled, which never fired, so this
is an INSTANCE stall
```

`enableAndSettleModelToggles` (`tests/helpers/provider-setup/model-toggle-batch.ts:282-290`) clicks **every** unchecked `:visible` toggle in the open provider panel — 29 of them for Google — and then waits up to 90 s (`model-toggle-batch.ts:254`) for the product's batched write to settle. It does not settle.

**The cost is not confined to these four tests: the same write appears to take the whole backend down.** Measured on this run, from the four `liveness-N` artifacts and the service-container log:

| event | time (UTC) |
|---|---|
| the four Google specs enter `setupGoogle` and start clicking | 12:39:15 – 12:39:24 |
| the umbrella's outage windows open, on all four shards | 12:39:30 / 12:39:32 / 12:39:38 / 12:41:03 |
| gunicorn `WORKER TIMEOUT` → `Worker was sent SIGKILL` (shards 1, 2 twice, 3) | 12:41:11 / 12:41:11 / 12:41:19 / 12:43:28 |

Shard 4's window opens 4 s after its only Google spec started. The four shards are independent containers, so a shared infra hiccup does not explain a synchronized onset — the common factor is the phase of the run.

[...]

Issue #1159, opened 2026-07-30
pr-validation: a change to .github/** or scripts/** selects ZERO specs, so CI-only mechanisms merge with no runtime proof


rafaelgiln:
## What happens

`detect-specs` resolves the E2E work by import graph (#1054). A diff confined to
`.github/workflows/**`, `.github/actions/**` or `scripts/**` imports into no spec,
so the resolver correctly reports:

```
Impacted specs: 0 (0 direct, 0 transitive); running 0 minus 0 dropped
→ Run impacted E2E specs: skipping
```

The PR then merges having proven the change **parses**, never that it **runs**.

## Where it was observed

PR #1157 (#1045) extracted the post-collect-models health gate into
`.github/actions/wait-for-backend` and rewired four workflows to it —
`daily-stable`, `pr-validation`, `manual`, `weekly-stable`. Every check on the PR
was green and the E2E lane was `skipping`. The action had never executed anywhere
at merge time.

Proof came only from a manual `manual.yml` dispatch on the branch
([run 30589339315](https://github.com/oriontech-me/langflow-e2e/actions/runs/30589339315)) — which did pass, and which nothing in CI
required. Had the composite action been wired wrong (a bad `uses:` path, a
missing input, `node` absent on a lane), the first evidence would have been the
next daily failing in a step that had nothing to do with the specs — the exact
attribution problem the gate itself exists to prevent, reintroduced one layer up.

This is not hypothetical for the class: #1019/#1044/#1045 were all fixes to CI
mechanisms, and each was merged on structural review plus a hand-dispatched run
that a reviewer had to remember to ask for.

## Why the existing guards do not cover it

- `#1054` fixed the *opposite* miss (a changed helper reaching 112 specs and
  selecting none). Its rule is "walk importers transitively" — correct, and
  silent by construction for a file no spec can import.
- The suite-wide escape hatch covers `tests/fixtures/**`, `playwright.config.ts`
  and the global hooks only. A workflow is not in it, and making it suite-wide
  would be wrong: a YAML comment must not run 240 specs.
- `npm run test:scripts` / `test:units` cover the *logic* a script implements.
  They cannot cover the wiring — that the workflow passes the inputs the action
  declares, on a lane where the interpreter exists.

## Possible levers (not a decision)

1. **Canary lane.** When the diff touches `.github/workflows/**`,
   `.github/actions/**` or `scripts/**`, select a small fixed spec set (2-3 fast,
[...]

PR #1685, opened 2026-09-03
fix(history): close the review findings on the liveness block (#1077)

### LINKED ISSUE #1077: daily-stable: give the mid-run worker wedge a baseline that outlives artifact retention

Victor-w-Madeira:
Closes #1077.

## What this PR is now

**One commit: the review fixes.** The mechanism itself is already on `main`.

PR #1687 was branched off this branch — an accident of two agents sharing one working tree — and was **squash-merged**, so it carried this PR's first four commits into `main` under its own title. `b53e4eb1`'s body opens with `* feat(history): record the in-run backend liveness on each daily history line (#1077)`. `scripts/lib/backend-history.mjs`, `scripts/backfill-backend-history.mjs`, their tests, the `daily-stable.yml` wiring, the schema docs and the five backfilled history rows are therefore live already — **at the state before two review passes and Copilot were addressed**. Squashing changes the SHAs, so those commits are not ancestors of `main` and git saw the files as add/add; this branch has been rebased onto `main` and the four now drop as already-upstream.

Issue #1077 is still open, because the squash carried no `Closes`. This PR closes it.

## Why the issue existed

The mid-run worker wedge is measured by the in-run recorder into per-shard `liveness-N` artifacts uploaded with `retention-days: 7`. Measured 2026-09-02: the dailies of 08-24, 08-25 and 08-26 were already `expired=true`. Every measurement accumulated on #1077 since July is a table typed by hand into a comment, re-derived by a human inside a one-week window, and the four axes its benchmark compares on appeared on no committed record. The `backend` block puts them on the one file in this repo that is append-only and committed back to `main`.

The baseline it captured, now on `main`:

```
2026-08-27  outages=0   down=0s       shards=4/4  passed=545
2026-08-28  outages=10  down=838.2s   shards=4/4  passed=536
2026-08-31  outages=14  down=1004.2s  shards=4/4  passed=523
2026-09-01  outages=19  down=1524.1s  shards=4/4  passed=524
2026-09-02  outages=12  down=864s     shards=4/4  passed=534
```

Five consecutive scheduled dailies, every shard executing tests, `shards_measured == shard_total` on all four. **2026-08-27 is the near-control**: 0 outage windows on four shards at roughly half the summed span of the wedged days — the wedge is not a constant of the runner. Applying and measuring a lever against this is **#1686**, blocked on #1679.

## What the commit fixes

Two independent review passes with clean context (both against the real artifacts, never fixtures) plus Copilot. **Two findings overlap, and both land on one asymmetry: the ad-hoc backfill script was hardened against a state the appender — the path that runs unattended on every daily forever — was not.**

- **A stray `*.json` in the liveness directory invented a shard.** `readSummaries` takes any file that parses to an object. Measured: a `manifest.json` beside a real summary yields `shards_reported: 2` with a phantom `{shard: "?", measured: false}` while one shard uploaded — which *fills* the gap between `shards_reported` and `shard_total` that a shard whose job died is supposed to show. The filter lives in `buildBackendBlock`, **before** `attribute()`, never in a caller: the shard join is positional over the array `attribute()` mapped, so filtering afterwards would desynchronise the indexes and trade a visible phantom for two shards silently swapping totals. The backfill's own copy of the filter is deleted with it.
- **`--shard-total` is required and strictly parsed.** It defaulted to `4`. A 2-shard repro backfilled under that default records `shards_reported: 2` against `shard_total: 4` — "two shards died", permanently — and the `shards_reported > shard_total` refusal cannot see it, since it only catches the excess. Sharper still: `Number(x) || null` turned `abc`, `4x`, `0` and an omitted value into `shard_total: null`, which **short-circuits that refusal** and writes the state the code's own comment calls the one that makes a silent shard undetectable. `--shard-total=6` also fell through to the default.
- **The reconciliation invariant is enforced, not asserted.** `unassigned` only ever detected the under-count, so two shards listing the same file both counted it. Every file is now counted once per run; a duplicate is attributed to the first shard and named in `double_claimed` (empty in practice — `partition-shards.mjs` emits disjoint lists). A test asserts the sum **through the appender** against the line's own totals, rather than against per-shard expectations written by hand, which cannot notice the two halves drifting together.
- **Three refusals had no test and survived mutation**: the missing-summary branch (without it a run dies later with "No history line for run(s)", pointing the operator at the history file when the cause is a missing download), the missing `--artifacts`, and an unreadable `--history` path, which raised a raw `node:fs` stack at the one script whose whole job is not to damage that file.
- **The structural import guard was #1226's failure mode verbatim.** It passed against `import{x}from"…"`, against a multi-line import, and against a static import of `report-backend-outages.mjs` — equally fatal, since `backend-history.mjs` pulls it in. Measured with that plus a module-scope throw: guard green, appender `exit=1`, **no history file at all**, on every lane. It now matches either module of the chain over whitespace-normalised text.
- **"Untouched lines pass through as their original bytes" was unfalsifiable** — the fixture's untouched line was itself a `JSON.stringify`, so re-serialising read identically. It is deliberately non-canonical now.

## Validation

| check | result |
|---|---|
| `npm run test:scripts` | 1027 pass / 0 fail (42 new; 1 pre-existing skip) |
| `npm run lint` | 0 errors (1413 pre-existing warnings) |
| Re-derive the five rows **now on `main`** with the fixed code | byte-identical |
| Mutation pass: 18 on the block builder, 8 on the backfill refusals, 3 import spellings | all caught |

```bash
npm run test:scripts && npm run typecheck && npm run lint
node --test scripts/lib/backend-history.test.mjs scripts/backfill-backend-history.test.mjs
jq -r 'select(.backend) | "\(.date)  outages=\(.backend.outages_total)  down=\(.backend.down_seconds_total)s  passed=\(.totals.passed)"' reports/daily-history.jsonl
```

**Nothing about the recorded baseline moves.** The five rows on `main` were written by the pre-fix code and re-derive byte-identically under the fixed code.

## Known gap, recorded rather than fixed

History lines are written on `schedule` only, so a lever measured through a `workflow_dispatch` of `daily-stable.yml` produces no durable row. Either compare across ≥2 scheduled dailies — what #1686's last *Done when* asks for anyway — or extend the append step first. Noted in `failure-modes.md`, which is already on `main`.

The series also cannot be extended backwards: everything before 2026-08-27 is gone.

The `daily-stable.yml` wiring still has **no CI proof** and cannot get one — it is one `env:` block on a step that runs only on `schedule`, so the PR lane never reaches it and a dispatch skips it. First real proof is the next scheduled daily: the `Append daily history` step log ends with `outages=… down=…s shards_measured=N/4`.

### DIFF
--- a/reports/README.md
+++ b/reports/README.md
@@ -200,11 +200,14 @@ The series is **not continuous**. Document any missing weeks here rather than ba
         "totals": { "passed": 128, "failed": 0, "flaky": 4, "skipped": 2 }
       }
     ],
-    "unassigned": { "passed": 0, "failed": 2, "flaky": 0, "skipped": 0 }
+    "unassigned": { "passed": 0, "failed": 2, "flaky": 0, "skipped": 0 },
                                              // OPTIONAL: specs no shard claimed. Omitted when
                                              // empty — which is the normal case, and is why the
                                              // rest of this example is the real 2026-09-02 row
-                                             // while this one line is not.
+                                             // while these last two lines are not.
+    "double_claimed": ["a.spec.ts"]          // OPTIONAL: specs more than one shard listed.
+                                             // Omitted when empty; present means the partition
+                                             // is wrong.
   }
 }
 ```
@@ -227,7 +230,9 @@ The series is **not continuous**. Document any missing weeks here rather than ba
   - **Absent ≠ zero, and `shards_measured: 0` ≠ healthy.** The block is **omitted entirely** by a lane that never measured (the weekly, any local run, anything that does not set `LIVENESS_DIR`). A run whose shards *did* report but recorded no probes carries the block with `shards_measured: 0` — that is UNKNOWN, and must never be read as a backend that stayed up (#1012). `shards_reported` below `shard_total` is a shard whose job died before writing a summary; `shard_total: null` means the run declared no count, which is the one state that makes such a shard undetectable.
   - `blips_total` counts single-probe failures **below** the 2-probe window threshold, so they are excluded from `down_seconds_total`. A non-zero value means the down-share on that row is a **floor**, not a ceiling — the caveat that changed a triage conclusion on 2026-09-01 (#1665).
   - `span_seconds` is the **recorder's** observed span (first probe → last probe), not the shard job's duration: the recorder starts after the post-`collect-models` health gate and is killed in teardown. It is the window the down-share is computed against, and the only wall clock the artifacts can support.
-  - `attempts` / `failing` / `collateral` count test **attempts** (retries included — the cost of a wedge is precisely that each collateral test burns its whole retry budget), while `shards[].totals` counts **tests**, in the same vocabulary as the line's run-level `totals`. The per-shard totals **sum to that run-level `totals`**; when they cannot, the difference is named in `unassigned` rather than silently lost.
+  - `attempts` / `failing` / `collateral` count test **attempts** (retries included — the cost of a wedge is precisely that each collateral test burns its whole retry budget), while `shards[].totals` counts **tests**, in the same vocabulary as the line's run-level `totals`.
+  - **The per-shard totals plus `unassigned` sum to the line's run-level `totals`, always.** That is enforced, not hoped for: every spec file is counted exactly once across the whole run. A file no shard listed lands in `unassigned`; a file **two** shards listed is attributed to the first and named in `double_claimed` rather than counted twice — `unassigned` can only ever detect the under-count, so without that rule the sum would break in the one direction nothing checks. `double_claimed` is empty and omitted in practice, because `partition-shards.mjs` emits disjoint lists; its presence means the partition itself is wrong.
+  - A file under the liveness directory that is **not** a shard summary is ignored rather than counted as a shard, with the number ignored printed. Left in, it produces a phantom `{shard: "?"}` record that inflates `shards_reported` — filling exactly the gap against `shard_total` that a shard whose job died is supposed to show.
   - Absent from history written before #1077.
 - `param` (optional, additive to schema v1) is the parameterization label a model-parameterized spec carries on its `describe` title — e.g. `"google / gemini-2.5-flash"` or `"model:gpt-4o-mini"`. The triage dataset builder uses it to group failures by provider variant and surface **provider-wide** clusters (same provider failing across ≥2 spec files → likely an environment/package cause, not per-test rot; #899). Absent for non-parameterized specs and for history written before #899.
 
--- a/scripts/backfill-backend-history.mjs
+++ b/scripts/backfill-backend-history.mjs
@@ -23,7 +23,8 @@
 //   node scripts/backfill-backend-history.mjs \
 //     --history reports/daily-history.jsonl \
 //     --artifacts <dir> \
-//     [--shard-total 4] [--dry-run]
+//     --shard-total <n> \
+//     [--dry-run]
 //
 // `<dir>` holds one subdirectory per run id, laid out as the run's artifacts
 // download:  <dir>/<run_id>/liveness-N/backend-liveness.json
@@ -36,18 +37,45 @@ import path from "node:path";
 import { buildBackendBlock } from "./lib/backend-history.mjs";
 import { readSummaries } from "./report-backend-outages.mjs";
 
+// Accepts both `--flag value` and `--flag=value`. The `=` form used to fall
+// through to the fallback silently, which for `--shard-total=6` meant writing
+// the DEFAULT while the refusal below reported a number the caller never typed.
 function arg(name, fallback = null) {
-  const i = process.argv.indexOf(`--${name}`);
-  return i === -1 ? fallback : process.argv[i + 1];
+  const argv = process.argv;
+  const i = argv.indexOf(`--${name}`);
+  if (i !== -1 && i + 1 < argv.length && !argv[i + 1].startsWith("--")) return argv[i + 1];
+  const inline = argv.find((a) => a.startsWith(`--${name}=`));
+  if (inline) return inline.slice(name.length + 3);
+  return fallback;
 }
 
 const historyPath = arg("history", "reports/daily-history.jsonl");
 const artifactsDir = arg("artifacts");
-const shardTotal = Number(arg("shard-total", "4")) || null;
 const dryRun = process.argv.includes("--dry-run");
 
 if (!artifactsDir) {
   console.error("--artifacts <dir> is required.");
+  process.exit(2);
+}
+
+// REQUIRED, and parsed strictly. It used to default to 4 — a guess about runs
+// the caller may not have checked, and `shard_total` is the one field whose
+// entire purpose is making a shard that uploaded NOTHING visible. A 2-shard
+// repro backfilled under that default records `shards_reported: 2` against
+// `shard_total: 4`, which reads as "two shards died" forever, and the
+// `shards_reported > shard_total` refusal below cannot see it — it only catches
+// the excess. An unparseable value is refused for the sharper version of the
+// same reason: `Number("abc") || null` yielded `null`, which SHORT-CIRCUITS that
+// refusal and writes the state this file's own comments call the one that makes
+// a silent shard undetectable.
+const shardTotalRaw = arg("shard-total");
+if (shardTotalRaw === null) {
+  console.error("--shard-total <n> is required — the run's real shard count, not a guess.");
+  process.exit(2);
+}
+const shardTotal = Number(shardTotalRaw);
+if (!Number.isInteger(shardTotal) || shardTotal < 1) {
+  console.error(`--shard-total must be a positive integer; got "${shardTotalRaw}".`);
   process.exit(2);
 }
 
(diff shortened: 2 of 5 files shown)


Victor-w-Madeira:
Note for review — the `Run impacted E2E specs` job reports `skipping`, with the CI-only-change guard (#1159) asking for a dispatch:

> CI-only change to `.github/workflows/daily-stable.yml`, `scripts/append-weekly-history.mjs`, which THIS lane does not run — nothing here proves it works. Dispatch `daily-stable.yml`, `weekly-stable.yml` on this branch before merging.

The verdict is correct (`dispatch`, not a silent skip — the classifier is doing its job), **but the remedy it names cannot cover this change**, and that is worth stating rather than leaving a reviewer to chase it:

- `Append daily history` is `if: always() && github.event_name == 'schedule'`, so a `workflow_dispatch` of `daily-stable.yml` **skips the step entirely** — it would burn ~35 min and prove nothing about the diff.
- `weekly-stable.yml` is disabled (kept as a fallback), and the appender change is a no-op there by construction: no `LIVENESS_DIR`, no block, byte-identical line. That path is covered by `scripts/lib/backend-history.test.mjs` → *"a lane that sets no LIVENESS_DIR writes the same line it always did"*.

This is the same gap the PR records in `failure-modes.md` — history lines are written on `schedule` only — reached from the other side.

What was proven instead, locally, against the real artifacts of the five backfilled runs rather than fixtures: the block re-derives byte-identically with the shipped code; a module-scope `throw` in the builder leaves `exit=0` with the line written and the block absent; and the merge-job wiring was checked by hand — `all-liveness` is the same path `Download shard liveness data` writes and the outage reporter reads, and `needs.prep.outputs.shard_total` is in scope for the `merge` job.

Final proof remains the **next scheduled daily after merge**: the `Append daily history` step log ends with `outages=… down=…s shards_measured=N/4`, and the `chore(history)` commit it pushes to `main` carries `.backend` on the new row. If that row is missing or reads `shards_measured=0/4`, revert commit 1 — the backfilled rows in commit 2 stand on their own.

Victor-w-Madeira:
Both Copilot comments are addressed in `12e2f73f`, together with a second independent review pass. Answering each directly:

**`--shard-total` defaulting to `4`** — taken, with option (b): it is now **required**, and parsed strictly. Your reading of the risk was right and the sharper case sat one line below it: `Number(x) || null` turned `abc`, `4x`, `0` and an omitted value into `shard_total: null`, which **short-circuits** the `shards_reported > shard_total` refusal entirely and writes the exact state the code's own comment calls the one that makes a silent shard undetectable. Also fixed while there: `--shard-total=6` fell through to the default, so the refusal could name a number the caller never typed — `arg()` parses both forms now. Six tests cover it (required, five unparseable values, the `=` form).

**`readSummaries()` passed through unfiltered** — taken, and it was the better of the two catches. One correction to the suggested placement, which matters: the filter went into `buildBackendBlock`, **before** `attribute()`, not into `backendBlockFromDir`. The shard join is positional over the array `attribute()` mapped, so a filter applied after that call would desynchronise the indexes and trade a visible phantom shard for two shards silently swapping totals — a strictly worse failure than the one being fixed.

The `.filter()` the backfill script carried is deleted with it, so the two callers of the builder can no longer disagree about what counts as a shard. That asymmetry was the real defect underneath both of your comments: the **ad-hoc** script, run by hand under supervision, was hardened — while the **appender**, which runs unattended on every daily forever, was not. Dropped files are counted and named in the log rather than filtered silently (#1012 is this repo's rule about that).

Verification: `npm run test:scripts` 983/983, typecheck clean, 0 lint errors, and the five backfilled rows **re-derive byte-identically** with the corrected code — so nothing about the recorded baseline moved.

Requesting a fresh review.

Issue #1300, opened 2026-08-05
Token recorder: the four gaps #1183's measurement exposed — no provider rollup, the PR lane's populated path never observed, local `.env` invisible, no settled 5-day baseline


rafaelgiln:
Follow-up of #1183 (measurement complete, closed). That issue produced the number; these are the
four residuals it exposed, and none of them is "measure more" — each is a separate mechanism.

## 1. There is no `by_provider` rollup — the schema records models only

#1183's final answer owed a **per-provider** figure and delivered it as a hand rollup of `by_model`
ids in a comment. `grep -n provider scripts/watch-tokens.mjs` returns nothing: the summarizer emits
`by_model` and `by_spec`, never a provider bucket.

Hand-rolling it worked for three lines and does not scale to a trend, and it is exactly the kind of
derivation that goes wrong silently: `gpt-5-mini` and `gpt-4o-mini` are both openai,
`gemini-flash-latest` is google, `claude-haiku-4-5` and `claude-sonnet-5` are anthropic — a prefix
rule gets all five right and gets the next Azure-hosted deployment wrong (#1281 already priced
`gpt-5-mini` as an Azure AI Foundry deployment, so the model id and the billing account are already
not the same axis here).

Either add `by_provider` derived from the same span data, or record the decision that `by_model`
supersedes it and fix the wording that asks for a provider split. Do not leave a consumer to infer
the provider from a model id.

## 2. The PR lane's populated path has never been observed

#1210 wired the recorder into `pr-validation.yml` and `manual.yml` in measurement-only mode. Its own
first run recorded `0 trace(s)` and said so honestly, which was the correct result for a canary run
of three LLM-free specs — and that was recorded at the time as needing "the next PR touching
`llm-agents/` or `model-provider/`" to close.

Five days later it is still open. Nine consecutive `pr-validation` runs were checked
(30965499480 → 30987740791): **every artifact contains `token-attrib.jsonl` and none contains
`token-probes.jsonl`**, i.e. zero traces on each. All four token steps were green in all nine. So
what is proven on this lane is the empty path and the honest-zero message; the populated path is
unproven, and a lane that has only ever measured zero cannot be trusted to measure non-zero.

Note the two files are independent: attribution rows exist (56 of them in run 30987740791), so the
`deleteFlow` hook is working on this lane. It is the poller that has never seen an LLM call here.

Closing this needs one PR whose impacted specs actually call a model — then record the run id and
[...]

Issue #1715, opened 2026-09-04
run-e2e.sh aborts when HOME is unset, which is every scheduler that is not cron


Victor-w-Madeira:
## What happens

`scripts/run-e2e.sh` dies on its first line of real work when `HOME` is not set:

```
./scripts/run-e2e.sh: line 190: HOME: unbound variable
```

Line 190 is the `uv` PATH export, and its own comment explains the assumption:

> `uv` lives in ~/.local/bin and cron does not load it: in a non-interactive shell
> `command -v uv` fails

The reasoning is right and the assumption is narrower than the comment implies.
**cron sets `HOME`** from the passwd entry; **systemd does not set it for a system
service**. Under `set -u`, an unset `HOME` is not a degraded PATH — it is an immediate
exit before the preflight runs.

## How it surfaced

The scheduled lane on the QA runner host was moved from cron to a systemd timer,
because Debian's cron (3.0pl1) has no `CRON_TZ` and that host's clock is not UTC, so a
cron schedule can only be written in local time and moves an hour by itself at the DST
change. The first execution under the timer failed on line 190 having done nothing
else. Under cron the same wrapper had just run clean, twice.

## Why it is worth fixing in the repository

The script is the one piece of this lane that is meant to be substrate-agnostic: it
already refuses to assume the tunnel, the target's name or the clone's location. `HOME`
is the one environment fact it does assume, and it assumes it from one specific
scheduler. Anything that is not cron — a systemd timer, a container entrypoint, a CI
step with a minimal environment — trips it.

## Proposed fix

- Default the value where it is used, e.g. `${HOME:-$(getent passwd "$(id -u)" | cut -d: -f6)}`,
  falling back to the passwd entry rather than to a hardcoded path.
- Prefer a resolved fallback over a literal `/root`: the script is used by hand on
  developer machines too, and a literal would silently point the PATH at a directory
  that is not theirs.
- Cover it in `npm run test:scripts`: run the PATH block with `HOME` unset and assert
  it neither aborts nor produces an empty first PATH entry.

## Workaround in place

The unit declares `Environment=HOME=/root` and the machine's wrapper defaults it, so the
scheduled lane is not blocked. This issue is about the script not depending on either.

PR #1716, opened 2026-09-04
Trace the way daily-stable.yml traces, so the VM lane can pass the traces specs

### LINKED ISSUE #1714: start-langflow-source.sh pins tracing off, so the VM lane cannot pass the traces specs
## What happened

The first full `@stable` run on the QA VMs finished **red** while the same day's Actions
daily finished **green**. Nine tests failed, and **eight of them are the
traces/observability family**; six more tests skipped inside that same family.

| | VM run (1 shard) | Actions daily, same day |
|---|---|---|
| passed | 570 | — |
| failed | **9** | 0 |
| flaky | 3 | — |
| skipped | 18 | — |
| duration | 49.2 min | — |
| verdict | red | success |

The report was complete — every shard blob present, neither empty nor partial — so this
is a test verdict, not an infrastructure abort.

## Root cause

One of the failures names it in its own assertion message:

> expected >= 2 trace(s) after a successful run, got 0. A 200 with no traces after a run
> is what `LANGFLOW_DEACTIVATE_TRACING=true` produces — this file needs an instance with
> tracing enabled

The two lanes disagree on that variable:

| Lane | Value | Where |
|---|---|---|
| Actions daily | `"false"` — tracing **on** | `.github/workflows/daily-stable.yml` |
| VM, from source | `true` — tracing **off**, hardcoded | `scripts/start-langflow-source.sh` |

The source starter inherited the line from `scripts/start-langflow-pip.sh`, where it is a
deliberate choice for local use: a developer's own token spend should not enter the token
history (#1300, #1183). That reasoning is sound for an interactive instance and wrong for
the scheduled lane, which the workflow runs with tracing on.

Consequence: those specs cannot pass on the VM at all. The `socket hang up` errors on the
other trace routes are the same mechanism reaching the client differently.

## Why this is an environment difference, not a product failure

Both sides ran the same product — the run resolved, placed and served the same published
nightly the Actions lane pulls, and recorded the version it served. What differs is one
environment variable that the starter fixes instead of accepting from its caller.

Stage 1 of the VM migration exists to find exactly this class of difference, and its rule
is to fix the environment rather than the test.

## Proposed fix

- Make the value a **parameter** of `start-langflow-source.sh` rather than a literal,
  defaulting to what the workflow passes, so the scheduled lane matches CI without the
  caller having to know the variable exists.
- Keep the local behaviour available: the pip starter's rationale still holds for
  interactive use, so the fix is the parameter, not the removal.
- **Pin the parity by test**, the way `start-echo-source.sh` pins the echo version and
  `start-ollama-source.sh` pins the Ollama one: read the value out of
  `.github/workflows/daily-stable.yml` and assert the starter's default agrees with it, so
  the two cannot drift apart silently again.

## Acceptance

- A VM run **through `scripts/run-e2e.sh`**, with no variable set by the caller, serves an
  instance with tracing enabled, and the traces family passes there.
- A unit test in the `npm run test:scripts` lane fails if the workflow's value and the
  starter's default stop agreeing.

## Evidence

- VM run `smoke-20260904` on the QA runner host — 570 passed / 9 failed / 3 flaky /
  18 skipped, red verdict, report complete.
- Actions daily the same day — run `33873006780`, success.

Victor-w-Madeira:
Closes #1714.

## What this fixes

The first full `@stable` run on the QA VMs came back **red** while the same day's
Actions daily came back **green**. **All nine failures** were the tracing family —
`api-monitor-traces.spec.ts` contributed two of them — and one named the cause in its own
assertion message:

> A 200 with no traces after a run is what `LANGFLOW_DEACTIVATE_TRACING=true` produces

The two lanes disagreed on one variable — `"false"` in `daily-stable.yml`, a hardcoded
`true` in `start-langflow-source.sh`. That is an environment difference, and the rule for
those is to fix the environment rather than the test.

## Where the fix lives, and why not in the starter

The starter only stops **pinning** the value; its default stays `true`, which is the right
answer for a developer's own instance (#1300, #1183). `scripts/run-e2e.sh` sets it,
because mirroring `daily-stable.yml` is that file's entire job — it already owns the
browser platform override and the `uv` PATH for the same reason.

To be precise about the justification, since the first version of this PR got it wrong:
the env-block parity test compares variable **names** (plus SSRF and `--workers`), so it
would *not* have caught a moved default. The property is real, the mechanism cited was
not — so the new starter test now **reads** pip's value, which pins the pair in both
directions and makes the claim true.

## Properties, each pinned by mutation

| Property | Mutation that must fail |
|---|---|
| The lane's value is **read out of** `daily-stable.yml`, never copied | default back to `true` → parity test fails |
| It is read **past comments**, and every occurrence must agree | a `#` line spelling the old value → guard stays **green** (no false positive) |
| The guard measures the **default**, not the ambient environment | delete the line *with the variable exported* → fails (it did not, before review) |
| The value **crosses the ssh boundary** | drop it from the remote command → boundary test fails |
| A value that is neither `true` nor `false` **stops the run** | remove the check → fails |
| The lane **does not append** to `reports/token-history.jsonl` | remove `TOKENS_SUPPRESS_HISTORY=1` → fails |
| The starter takes the value from its **caller**, matching pip's default | flip either starter's default → fails |

## What review changed, and why it mattered

- **The lane would have started writing the daily's token series.** With tracing on there
  is spend to record, and `--summarize` appends to a **git-tracked** file inside a clone
  this run does not own. The scheduled wrapper pulls with `--ff-only` precisely so local
  work fails the pull instead of being rewritten — so a line a day would have broken that
  guarantee daily, and once committing is on it would be two lines a day for one scope.
  `TOKENS_SUPPRESS_HISTORY=1` now, for the neighbouring reason the PR and manual lanes
  carry it (#1183).
- **The guard read the workflow as raw text** — no `stripComments()`, first match only.
  Both failures #1300's review already paid for, and the failure mode is the bad kind: it
  fails a *correct* workflow and invites restoring `true`.
- **The guard passed its own mutation** when the variable was exported, because
  `sourced()` forwards `process.env`.
- **A caller-supplied value went into the remote command bare.** `0` or `FALSE` means
  tracing off while looking deliberate — #1714's own failure class, through the caller.
- One of these was found by mutation rather than by review: the starter test's read of
  pip's default was unanchored, and the pip starter's **comment** spells the same
  assignment two lines above the real one, so it stayed green after the setting changed.

## Checks

- `npm run test:scripts` — **1096 passing**, 1 skipped, 0 failing, 7 new tests.
- Measured on the real machines: the divergence came out of a full run (570 passed /
  9 failed / 3 flaky / 18 skipped, report complete) with the same day's Actions daily
  green. Post-fix verification of the traces family on the VMs is running; result to
  follow in a comment, since no Actions lane exercises this path.

## Follow-ups, not in this PR

- **#1715** — `run-e2e.sh` aborts when `HOME` is unset, which is every scheduler that is
  not cron. Workaround in place on the machine.
- **The class this belongs to.** Review found three more of the daily's service variables
  that never reach the VM: `LANGFLOW_ALLOW_CUSTOM_COMPONENTS`, `LANGFLOW_SQLITE_PRAGMAS`
  and `LANGFLOW_WORKER_TIMEOUT`. Filed separately, with the severity measured rather than
  assumed.

### DIFF
--- a/reports/README.md
+++ b/reports/README.md
@@ -23,12 +23,9 @@ carry. Four known blind spots (#1211), stated here rather than only in a PR body
 comment so a reader of a number finds its limits in the same place:
 
 - **A local instance records nothing, and developer spend is OUT OF SCOPE for this series
-  (#1300).** Both start scripts *default* `LANGFLOW_DEACTIVATE_TRACING` to `true`, so the poller
-  has no traces to read at all when developing against a local container. The value is a
-  parameter rather than a literal because the scheduled VM lane needs the opposite and is not a
-  local instance: `scripts/run-e2e.sh` runs it with tracing ON, the way `daily-stable.yml` does,
-  and on the same CI keys — so its spend belongs in this series exactly as the Actions lane's
-  does, and the attribution problem below is not reintroduced (#1714). That is a **decision**, not a
+  (#1300).** The Docker and pip start scripts set `LANGFLOW_DEACTIVATE_TRACING=true` as a
+  literal, and `start-langflow-source.sh` defaults it to `true`, so the poller has no traces to
+  read at all when developing locally. That is a **decision**, not a
   pending fix: flipping the flag locally would produce traces nobody can attribute, because the CI
   secret and a developer's `.env` draw on one account balance and #1183's key-separation
   recommendation (`ANTHROPIC_API_KEY_AGENT` / `ANTHROPIC_API_KEY_CI`) is still unimplemented — so a
@@ -36,6 +33,13 @@ comment so a reader of a number finds its limits in the same place:
   consequence to carry: **this file is CI spend, and must never be quoted as total account
   spend.** The gap between the two is whatever developers spent locally, which only the provider
   console can answer, and only once the keys are split.
+- **The scheduled VM lane traces, and still writes nothing here (#1714).** It is not a local
+  instance: `scripts/run-e2e.sh` overrides that default and runs with tracing ON, the way
+  `daily-stable.yml` does, because the traces specs assert against a traced instance. Its
+  summary call carries `TOKENS_SUPPRESS_HISTORY=1` all the same — while both dailies run, the
+  Actions one owns this series, and a second line a day for the same `@stable` sweep would
+  double-count the trend and halve the anomaly window. The VM lane's own series lives outside
+  that clone until the switch makes it the only writer.
 - **`Collect models` spend enters the totals with no spec name.** That pre-flight sweep drives the
   same Langflow instance the token poller watches, but it is not a spec run, so its traces land in
   `unattributed`, never in `by_spec`.
--- a/reports/README.md
+++ b/reports/README.md
@@ -23,8 +23,12 @@ carry. Four known blind spots (#1211), stated here rather than only in a PR body
 comment so a reader of a number finds its limits in the same place:
 
 - **A local instance records nothing, and developer spend is OUT OF SCOPE for this series
-  (#1300).** Both start scripts set `LANGFLOW_DEACTIVATE_TRACING=true`, so the poller has no
-  traces to read at all when developing against a local container. That is a **decision**, not a
+  (#1300).** Both start scripts *default* `LANGFLOW_DEACTIVATE_TRACING` to `true`, so the poller
+  has no traces to read at all when developing against a local container. The value is a
+  parameter rather than a literal because the scheduled VM lane needs the opposite and is not a
+  local instance: `scripts/run-e2e.sh` runs it with tracing ON, the way `daily-stable.yml` does,
+  and on the same CI keys — so its spend belongs in this series exactly as the Actions lane's
+  does, and the attribution problem below is not reintroduced (#1714). That is a **decision**, not a
   pending fix: flipping the flag locally would produce traces nobody can attribute, because the CI
   secret and a developer's `.env` draw on one account balance and #1183's key-separation
   recommendation (`ANTHROPIC_API_KEY_AGENT` / `ANTHROPIC_API_KEY_CI`) is still unimplemented — so a
--- a/scripts/run-e2e.sh
+++ b/scripts/run-e2e.sh
@@ -194,6 +194,15 @@ export PATH="$HOME/.local/bin:$PATH"
 # globalSetup — see the migration's divergence list, entry 1.
 export PLAYWRIGHT_HOST_PLATFORM_OVERRIDE="${PLAYWRIGHT_HOST_PLATFORM_OVERRIDE:-ubuntu24.04-x64}"
 
+# Tracing ON, because daily-stable.yml runs with it on and the traces/observability
+# specs assert against a traced instance. Both starters default it OFF — right for a
+# developer's own instance, wrong for the lane that has to match CI — so the value
+# belongs here, in the file whose whole job is mirroring that workflow. Measured, not
+# assumed: on 2026-09-04 eight @stable specs failed on the VM while the same day's
+# Actions daily was green, for no reason other than this variable (#1714). Read the
+# workflow before changing it — run-e2e.test.mjs asserts the two agree.
+LANGFLOW_DEACTIVATE_TRACING="${LANGFLOW_DEACTIVATE_TRACING:-false}"
+
 # ---------------------------------------------------------------------------
 # UTILITIES
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


Victor-w-Madeira:
Reviewed the diff plus mutation runs against the real files (separate worktree on the PR head, `node --test scripts/run-e2e.test.mjs scripts/start-langflow-source.test.mjs scripts/watch-tokens.test.mjs`). Every figure below is measured, not read.

The fix is in the right place and the property it pins — the value crossing the ssh boundary — is the right one. What needs changing is one **unforeseen side effect** (the lane starts writing the token series) and **two of the three guards not pinning what they claim to pin**.

---

## 1. The VM lane now writes `reports/token-history.jsonl`

`run-e2e.sh:858` runs `watch-tokens.mjs --summarize` with no `TOKENS_SUPPRESS_HISTORY`, and the summarizer skips the history line only on a run with zero traces. Today this lane records zero, so it writes nothing. With tracing ON it appends one line per run to a **git-tracked** file, in the same series and the same anomaly baseline (20-line window) the Actions daily feeds with the same day's same `@stable` sweep.

- While `COMMIT_HISTORY=0`, the line is an uncommitted modification in the VM's checkout — the next `git fetch`/pull discards it or conflicts. That is also why this PR's new README sentence ("*its spend belongs in this series exactly as the Actions lane's does*") is not true yet.
- When step 09 turns committing on, there are **two lines per day for one scope**: double-counted trend, and the effective anomaly window halved.

This is the same reason pr-validation/manual carry `TOKENS_SUPPRESS_HISTORY` (#1183) — there on scope grounds, here on duplication grounds. Suggest setting `TOKENS_SUPPRESS_HISTORY=1` on this lane's summarize call (or gating the write on `COMMIT_HISTORY`, alongside the other "no consequence yet" switches), and rewording the README to what the lane actually does today.

## 2. The parity guard reads the workflow as raw text

It reintroduces both failures the #1300 review already paid for: no `stripComments()`, and first-occurrence only. **Measured** — adding one prose comment above the setting in `daily-stable.yml` (`# Was LANGFLOW_DEACTIVATE_TRACING: "true" until #1300 …`):

```
not ok 10 - the lane runs tracing the way the workflow does, read out of the workflow
  error: this lane has to trace the way daily-stable.yml does (true)
```

It fails a **correct** workflow, and the message invites the next reader to "fix" `run-e2e.sh` by restoring `true` — reintroducing #1714 with the test green. `watch-tokens.test.mjs:1276-1290` documents exactly this risk ("*a comment merely SPELLING … could both mask a real value and fail a correct workflow*") and already solves it with `stripComments()` + `matchAll` (every occurrence must be defensible) + `evaluateWorkflowValue` from `scripts/lib/gh-expression.mjs` for the day the value becomes an expression, as it already is on pr-validation. Reusing the first two is the minimum.

## 3. The guard passes its own mutation when the variable is in the environment

`sourced()` forwards `{...process.env}`. With the `LANGFLOW_DEACTIVATE_TRACING="${…:-false}"` line deleted from `run-e2e.sh`:

| environment | result |
|---|---|
| clean | `# fail 1` — caught |
| `LANGFLOW_DEACTIVATE_TRACING=false` exported | `# pass 2 / fail 0` — **passes the mutation** |

The likely place for that export is the VM itself, or the shell of whoever was debugging #1714. One-line fix: `sourced('echo …', { LANGFLOW_DEACTIVATE_TRACING: "" })` — `:-` treats empty as unset, so the default is what gets measured.

## 4. The default's justification is attributed to a test that does not check it

The PR body and both new comments say the env block is *asserted against* `start-langflow-pip.sh`'s, and that moving the default on one side "would break that property". Measured:

- flipping the **source** starter's default to `false` fails **only** the new test (#15); the parity test does not object;
- flipping the **pip** starter's default to `false` fails **nothing** (103 pass / 0 fail).

The parity test compares variable **names**, with values only for SSRF and `--workers` ("*the two defaults whose exact value is load-bearing*"). The design property is legitimate; the mechanism cited does not enforce it, and the pip→source direction is unpinned. Idiomatic fix: have the new test **read** pip's default instead of hardcoding `true` — the same "read from that file" the parity test already uses. That makes the comment true and pins the pair.

## Nits

- **Caller value interpolated unquoted into the ssh command.** This is the first *caller-supplied* variable to go in bare (the others are computed internally). It crosses no privilege boundary, but `0`, `FALSE`, or a value with a space goes in silently — and "tracing silently off" is #1714's own failure class. A `case "$LANGFLOW_DEACTIVATE_TRACING" in true|false) ;; *) die …` at the top covers both.
- **`reports/README.md`:** after the insertion, "That is a **decision**, not a pending fix" has lost its antecedent — it refers to the local default but now follows the sentence about the VM lane.
- **Issue acceptance:** criterion 1 asks for "*a VM run with **the starter's default** serves an instance with tracing enabled*", and the starter's default is still `true` — for a good, well-argued reason, but the criterion should be edited so nobody reads it as met. And: was there a VM run **after** the fix confirming the traces family green? The body cites only the pre-fix measurement plus `test:scripts`, and no Actions lane exercises this path (`Run impacted E2E specs: skipping` is correct — nothing under `tests/` imports these scripts).

## Follow-up, separate issue

#1714 is one instance; three siblings remain. `daily-stable.yml` sets `LANGFLOW_ALLOW_CUSTOM_COMPONENTS`, `LANGFLOW_SQLITE_PRAGMAS` and `LANGFLOW_WORKER_TIMEOUT`, and **none of the three** reaches the VM (absent from `run-e2e.sh`, `start-langflow-source.sh` and `start-langflow-pip.sh`).

- `LANGFLOW_ALLOW_CUSTOM_COMPONENTS` is the worst and probably explains the run's **9th** failure: the workflow's own comment says "*the nightly image ships LANGFLOW_ALLOW_CUSTOM_COMPONENTS=false (a deliberate default)*", and `docs/core-components/full-custom-component.md:125` says the same of the product default — so `full-custom-component`, `code-execution-endpoints` and the other specs declaring that precondition cannot pass on the VM.
- `LANGFLOW_SQLITE_PRAGMAS` (foreign_keys:ON) is the one that changes a verdict **without** failing: without it, a cascade/FK defect is invisible on the VM and visible on Actions.

If mirroring `daily-stable.yml` is that file's entire job, the guard worth having is `watch-upstream-areas.mjs`'s fail-closed shape: enumerate the daily's service env and require every variable to either reach the target **or** be recorded out of scope with a reason.

## What is good

- Choosing `run-e2e.sh` over the starter is **better** than the issue's proposal: it keeps #1300/#1183 intact for a developer's instance and puts the value in the file whose job is mirroring the workflow.
- The "crosses the ssh boundary" test pins the right property — its failure mode *looks* fixed from the calling machine, and this is the only way to know the value travelled.
- Incidentally this also closes #1300's blind spot on this lane. Worth recording that `watch-tokens.test.mjs`'s `pollerLanes()` derives over `.github/workflows/*.yml` only, so its claim ("*a fourth lane inherits this check the day it is written*") is already false for a lane that is not a workflow — which is how the VM lane escaped it.

Issue #1720, opened 2026-09-04
With tracing on, the source instance wedges under the traces family — the fourth divergence, and the first about substrate


Victor-w-Madeira:
Found while verifying #1716 on the machines. It is the fourth entry on stage 1's
divergence list, and the first one that is about the **substrate** rather than a setting.

## What happens

With `LANGFLOW_DEACTIVATE_TRACING=false` — the value #1716 gives the lane, because
`daily-stable.yml` runs it and nine `@stable` tests need it — the **source** instance
wedges under the traces family's load. The suite's own auth helper names it:

```
auth: /api/v1/auto_login did not answer (apiRequestContext.get: Timeout 20000ms exceeded.)
  — retry 3/3 in 20000ms. The backend is wedged or recycling (see #1077).
```

Counted in one run: 24 POST timeouts, 6 GET, 5 DELETE, all at the 20 s cap, plus orphan
flows and an orphan API key left behind because their cleanup DELETE timed out too.

**It does not stay inside the traces family.** The timeouts land on
`/api/v1/auto_login`, which every API spec authenticates through — `security/credential-secret-exposure`
failed that way, outside the observability directory. #927 already established that this
wedge is **process-wide**, not worker contention.

## Why it is not a missing variable — three runs, same specs, same machine

| Run | Failed | Passed | Wall |
|---|---|---|---|
| tracing ON (the #1716 fix alone) | 7 (+2 flaky) | 7 | 5.1 min |
| tracing ON + the workflow's `LANGFLOW_SQLITE_PRAGMAS` + the echo endpoint up | 9 | 6 | 10.4 min |
| tracing ON + the workflow's `LANGFLOW_WORKER_TIMEOUT=120` | 8 | 7 | 6.0 min |

Neither of the two variables #1717 lists changed the outcome, and the count **oscillates**
between 7 and 9 across runs. A missing setting produces a deterministic failure; this
produces a different number every time, which is a wedge signature.

Both lanes run `--workers 1` (`daily-stable.yml` pins it deliberately, #927), so worker
count is not the difference either. What differs is what serves the requests: a container
image on one side, `uv run` against a source checkout on the other, writing traces into
SQLite.

## What was done meanwhile, and why it is not a fix

The scheduled lane on the QA runner sets `LANGFLOW_DEACTIVATE_TRACING=true` in its
machine-side wrapper, with the reason recorded next to it. The repository keeps the
correct default from #1716; the machine carries the exception explicitly.

[...]

PR #1722, opened 2026-09-05
Keep the daily's three series in a ledger outside the clone

Victor-w-Madeira:
## Why

`reports/daily-history.jsonl`, `reports/token-history.jsonl` and `reports/spec-durations.json` only grow because `daily-stable.yml` writes them, and the migration's next etapa turns that lane off. Nothing else appends. The gap would run from there to the day the VM lane commits: `build-triage-dataset.mjs` would read a base that starts partial, and the shard matrix — which balances by measured duration — would degrade with it, both for a reason no run reports.

So the VM lane starts keeping the three now, while the Actions series is still complete enough to seed from.

## Where they go, and why it is not `reports/`

Outside the clone. `reports/` is tracked, so a run that writes there leaves the tree dirty, and the machine's next `git pull --ff-only` refuses the following morning — the protection working exactly as designed, against us. The series moves; the protection stays.

That decision is checked rather than trusted. `ledger_dir_is_outside_repo()` resolves both sides through `pwd -P`, so a symlink pointing back into the working tree is caught too, and it answers for a path that does not exist yet — preflight has to refuse **before** creating anything, or the trace left by the rejected check is a directory inside the clone.

Default is `$XDG_STATE_HOME/langflow-e2e`, then `$HOME/.local/state/...`. With neither set the value is empty and preflight **dies** rather than guessing a path: a scheduled run that quietly keeps no series is indistinguishable, months later, from a machine that was down. Preflight is also where it is checked, not publish — an unwritable ledger discovered at publish costs the whole hour the run just spent; discovered in preflight it costs one variable.

## Two switches retired

**`COMMIT_HISTORY` was named for something this script does not do.** The code under it only ever wrote a file — no commit, no push — so the run series was gated on a decision belonging to a later etapa, and the hole in the data would have been exactly as wide as the wait. Committing is what stays behind, and it stays behind as **absent code**, not as a switch set to zero: a switch reads as implemented and disabled, and invites someone to flip it on a machine with no write credentials. A test pins the absence.

**`REFRESH_DURATIONS` goes for the same reason in reverse** — it was off because it wrote into the tracked tree, and it no longer does.

## The token series

The summary stops carrying `TOKENS_SUPPRESS_HISTORY` on a scheduled run. #1183's argument excludes the PR and manual lanes because their scope is not the daily's sweep; this lane's scope **is** the daily's sweep. What must not happen is the two ending up in one series, and `WORKFLOW` is what prevents it: without it the summarizer writes `workflow: "unknown"`, which in a merged file reads as an Actions row that lost its label rather than as a VM one. `tokens_history_env()` exists so that composition is unit-tested rather than read — the two outcomes are mutually exclusive by construction, and the label cannot be dropped while the path is set.

The history line also gains `LIVENESS_DIR` and `SHARD_TOTAL`, which the workflow passes and this lane did not: without the expected count a shard that died before writing its summary vanishes from the row instead of reading as a gap (#1012), and those fields are exactly where the wedge in #1720 shows up.

## What deliberately did NOT change

The **read** side. The matrix still balances on the tracked `reports/spec-durations.json`. While both dailies run the product is a comparison, and a matrix balanced by VM-measured timings puts specs on different shards than the Actions lane does — so a failure's neighbours, and the load its backend was under, would differ for a reason that has nothing to do with the product. `USE_LEDGER_DURATIONS` is the next etapa's switch to throw.

## Seeding

A ledger file that does not exist yet is copied once from the tracked one. Day zero is not free: the durations file is what balances the matrix, and the token summary's anomaly baseline is a median over recent entries — both answer badly from three lines, and badly in the direction of looking fine. One-way and once; after the copy the tracked file is never read again by this lane, and never written.

## Tests

12 new cases in `scripts/run-e2e.test.mjs` (52 in that file, 1135 in `test:scripts`). The ones that matter:

- a ledger inside the clone is refused however the path is spelled — the clone, `reports/`, a path that does not exist yet, a relative path, and a symlink that resolves back in
- preflight refuses such a ledger **and creates nothing on the way out**
- preflight refuses when there is nowhere to put one, and the message names `KEEP_LEDGER=0` as the way out
- only a scheduled run keeps the series
- the spend line carries its label, and the two outcomes cannot both happen
- seeding is one-way and once — re-seeding would replay the Actions series over the VM's own, every single day
- nothing in `phase_publish` writes into the tracked series: the three paths may appear there only as the source `ledger_seed` copies *from*
- writing back to the repository is absent, not switched off

The switch-absence test strips comments before reading the file, for the reason #1716 paid for: a comment that merely spells the thing under test fails a correct file, and the paragraph explaining why `COMMIT_HISTORY` was removed is exactly such a comment.

Closes the migration's step 17.

### DIFF
--- a/reports/README.md
+++ b/reports/README.md
@@ -39,7 +39,8 @@ comment so a reader of a number finds its limits in the same place:
   summary call does not write **here**, all the same: while both dailies run, the Actions one
   owns this file, and a second line a day for the same `@stable` sweep would double-count the
   trend and halve the anomaly window. It writes its own series to a **ledger outside the clone**
-  (`LEDGER_DIR`, default `$XDG_STATE_HOME/langflow-e2e`), one file per series with the same
+  (`LEDGER_DIR`, defaulting to `$XDG_STATE_HOME/langflow-e2e` and, when that is unset,
+  `$HOME/.local/state/langflow-e2e`), one file per series with the same
   schemas as the three here, and every line it writes carries `workflow: "daily-stable-vm"` — so
   when the two are merged, which era a row came from is still readable. Outside the clone because
   `reports/` is tracked: a line written here by the VM lane is an uncommitted change that the
--- a/reports/README.md
+++ b/reports/README.md
@@ -36,10 +36,14 @@ comment so a reader of a number finds its limits in the same place:
 - **The scheduled VM lane traces, and still writes nothing here (#1714).** It is not a local
   instance: `scripts/run-e2e.sh` overrides that default and runs with tracing ON, the way
   `daily-stable.yml` does, because the traces specs assert against a traced instance. Its
-  summary call carries `TOKENS_SUPPRESS_HISTORY=1` all the same — while both dailies run, the
-  Actions one owns this series, and a second line a day for the same `@stable` sweep would
-  double-count the trend and halve the anomaly window. The VM lane's own series lives outside
-  that clone until the switch makes it the only writer.
+  summary call does not write **here**, all the same: while both dailies run, the Actions one
+  owns this file, and a second line a day for the same `@stable` sweep would double-count the
+  trend and halve the anomaly window. It writes its own series to a **ledger outside the clone**
+  (`LEDGER_DIR`, default `$XDG_STATE_HOME/langflow-e2e`), one file per series with the same
+  schemas as the three here, and every line it writes carries `workflow: "daily-stable-vm"` — so
+  when the two are merged, which era a row came from is still readable. Outside the clone because
+  `reports/` is tracked: a line written here by the VM lane is an uncommitted change that the
+  machine's next `git pull --ff-only` refuses, silently, the following morning.
 - **`Collect models` spend enters the totals with no spec name.** That pre-flight sweep drives the
   same Langflow instance the token poller watches, but it is not a spec run, so its traces land in
   `unattributed`, never in `by_spec`.
--- a/scripts/run-e2e.sh
+++ b/scripts/run-e2e.sh
@@ -552,12 +552,17 @@ durations_table() {
 
 # Where this run's spend line goes, as one KEY=VALUE per line. A function so the
 # composition is testable rather than read: the two outcomes are mutually exclusive by
-# construction, and WORKFLOW can never be dropped while the history path is set —
-# without it the summarizer writes `workflow: "unknown"`, which reads as an Actions row
-# that lost its label rather than as a VM one.
+# construction, and neither label can be dropped while the history path is set.
+#
+# WORKFLOW, because without it the summarizer writes `workflow: "unknown"`, which reads
+# as an Actions row that lost its label rather than as a VM one. GITHUB_RUN_ID because
+# the summarizer takes the run's identity from the environment, which in Actions is
+# simply there and here is not: a smoke found the spend row landing with
+# `run_id: null` while the history row for the SAME run carried the id, and a row that
+# cannot be joined back to its run is dropped by every consumer that groups by run.
 tokens_history_env() {
   if ledger_active; then
-    printf '%s\n' "TOKENS_HISTORY=$LEDGER_TOKENS" "WORKFLOW=$WORKFLOW_ID"
+    printf '%s\n' "TOKENS_HISTORY=$LEDGER_TOKENS" "WORKFLOW=$WORKFLOW_ID" "GITHUB_RUN_ID=$RUN_ID"
   else
     printf '%s\n' "TOKENS_SUPPRESS_HISTORY=1"
   fi
--- a/scripts/run-e2e.sh
+++ b/scripts/run-e2e.sh
@@ -122,6 +122,47 @@ REPORT_URL="$RUN_URL_BASE/$RUN_ID/playwright-report/index.html"
 EVENT_NAME="${EVENT_NAME:-schedule}"
 WORKFLOW_ID="${WORKFLOW_ID:-daily-stable-vm}"
 
+# ---------------------------------------------------------------------------
+# THE LEDGER — the three series this lane has to keep
+# ---------------------------------------------------------------------------
+# reports/daily-history.jsonl, reports/token-history.jsonl and
+# reports/spec-durations.json only grow because the Actions daily writes them, and the
+# next etapa turns that daily off. Nothing else appends, so the gap would run from
+# there to the day this lane commits: build-triage-dataset.mjs would read a base that
+# starts partial, and the shard matrix — which balances by measured duration — would
+# degrade with it, both for a reason no run reports. This lane starts keeping the three
+# now, while the Actions series is still complete enough to seed from.
+#
+# OUTSIDE the clone, and that is not a preference. `reports/` is tracked, so a run that
+# writes there leaves the tree dirty and the wrapper's next `git pull --ff-only`
+# refuses — and that refusal is a guarantee worth keeping, not an obstacle: it is what
+# stops a machine from silently discarding local work. So the series moves, not the
+# protection. ledger_dir_is_outside_repo() checks it rather than trusting the default,
+# because the cost of getting it wrong is paid once a day, quietly, by the pull.
+#
+# WORKFLOW_ID above is what keeps the two eras apart inside one file: every line this
+# lane writes carries `daily-stable-vm`, so the Actions rows stay distinguishable after
+# the two series are merged.
+KEEP_LEDGER="${KEEP_LEDGER:-1}"
+# `$XDG_STATE_HOME` is the right shelf for this: state that must survive between runs,
+# is not a cache (losing it loses history), and is not config. Empty rather than a
+# fallback path when neither variable is set — an unwritable ledger is refused in
+# preflight, where it costs a variable, instead of guessed at into a directory nobody
+# looks in.
+LEDGER_HOME="${XDG_STATE_HOME:-${HOME:+$HOME/.local/state}}"
+LEDGER_DIR="${LEDGER_DIR:-${LEDGER_HOME:+$LEDGER_HOME/langflow-e2e}}"
+LEDGER_HISTORY="${LEDGER_HISTORY:-${LEDGER_DIR:+$LEDGER_DIR/daily-history.jsonl}}"
+LEDGER_TOKENS="${LEDGER_TOKENS:-${LEDGER_DIR:+$LEDGER_DIR/token-history.jsonl}}"
+LEDGER_DURATIONS="${LEDGER_DURATIONS:-${LEDGER_DIR:+$LEDGER_DIR/spec-durations.json}}"
+# The READ side, and it stays off here on purpose. While both dailies run, the product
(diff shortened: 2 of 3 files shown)


Victor-w-Madeira (commented):
## Review

The decision is right and the argument for it is the strongest part of the PR: `reports/` is tracked, the pull's refusal is a guarantee rather than an obstacle, so the series moves and the protection stays. Retiring `COMMIT_HISTORY` as *absent code* rather than as a switch set to zero is the right call for the same reason, and `tokens_history_env()` making the two outcomes mutually exclusive by construction is exactly where that belongs.

Reviewed against the branch at `73a80e17` on top of `origin/main` (`799afd5b`, so #1721 is in the base), plus `watch-tokens.mjs`, `append-weekly-history.mjs` and `reports/README.md`. Everything below was measured on the branch, not read off it.

Five things. The first three are holes in the pins, and all three are cheap.

### 1. The preflight check the whole argument rests on is unpinned

The PR's case for checking at preflight is explicit — "an unwritable ledger discovered at publish costs the whole hour the run just spent; discovered in preflight it costs one variable." That check is the two lines nothing tests:

```
-    mkdir -p "$LEDGER_DIR" || die "cannot create the ledger directory ($LEDGER_DIR)."
-    [ -w "$LEDGER_DIR" ] || die "the ledger directory is not writable ($LEDGER_DIR)."
```

```
$ node --test scripts/run-e2e.test.mjs
# tests 52
# pass 52
# fail 0
```

Deleted outright, 52/52 green. Every `preflight_ledger` test is a refusal — inside the clone, nowhere to put it, `KEEP_LEDGER=0` — and the path that actually runs every scheduled night is exercised by none of them. A tmpdir and two assertions (the directory exists afterwards, `ledger: <dir>` is logged) close it, and the same test is where the `-w` branch gets a `chmod 500` case.

### 2. The placement guard covers `LEDGER_DIR`; the three files that actually get written are unchecked

`LEDGER_HISTORY`, `LEDGER_TOKENS` and `LEDGER_DURATIONS` default under `LEDGER_DIR` but are independently overridable, and `ledger_dir_is_outside_repo` never sees them:

```
$ LEDGER_DIR=/tmp/led-ok LEDGER_HISTORY="$PWD/reports/daily-history.jsonl" \
    bash -c 'source scripts/run-e2e.sh; preflight_ledger; echo "exit=$?"'
    ledger: /tmp/led-ok
exit=0
```

Preflight green, and `HISTORY_FILE` then points at the tracked file — the dirty tree this change exists to prevent, reached through a variable the script itself exposes. The `nothing in the publish phase writes into the tracked series` test cannot catch it either: it greps the *script text*, and the script text says `"$LEDGER_HISTORY"`.

"That decision is checked rather than trusted" is the standard the PR sets for `LEDGER_DIR`; the three paths are still trusted. Either run the resolved paths' parents through the same check in preflight (a loop over the three, no new mechanism), or drop the per-file overrides and derive them — they are undocumented in the usage block, so nothing depends on them yet.

### 3. The absence pin misses the spelling a script that operates on a clone by path would use

`writing back to the repository is absent, not merely switched off` matches `/\bgit\s+(commit|push|add)\b/`, so `git` followed by anything else is invisible. Added inside `phase_publish`:

```
+    git -C "$REPO_DIR" add -A
+    git -C "$REPO_DIR" commit -m "chore(history): daily [skip ci]"
+    git -C "$REPO_DIR" push origin main
```

```
# tests 52
# pass 52
# fail 0
```

`git -C` is not an exotic spelling here — it is how this script talks to the clone everywhere else, and it is what the later etapa will most likely be written with. (Worth noting the sibling test *does* fire if the `add` names a tracked path by hand; `add -A` is what gets past both.) `/\bgit\b[^\n]*\b(commit|push|add)\b/` costs one character class and keeps the assertion's meaning.

### 4. `KEEP_LEDGER` takes any value, and anything but `1` keeps nothing

```
$ KEEP_LEDGER=yes bash -c 'source scripts/run-e2e.sh; preflight_ledger; ledger_active; echo "active=$?"'
    ledger: not kept for this run (KEEP_LEDGER=yes, event=schedule)
active=1
```

The `elif` that makes an empty `LEDGER_DIR` fatal — "a scheduled run that quietly keeps no series is indistinguishable, months later, from a machine that was down" — is guarded on `KEEP_LEDGER = "1"`, so a typo'd truthy walks straight past it into the `else`. The `info` line names the value, which is better than nothing, but the whole point of that `die` is that nobody reads a log line months later.

This is the class `require_bool` was added to this same file for, in the PR merged into the base of this one: *"these flags read anything that is not 'true' as false, so `0`, `FALSE` or a value carrying a space would go in silently."* One call (`0`/`1` rather than `true`/`false`) makes the two consistent.

### 5. `ledger_seed`'s copy is fail-hard while every neighbour is fail-soft

```
$ bash -c 'source scripts/run-e2e.sh; ledger_seed /nonexistent/led.jsonl reports/daily-history.jsonl; echo SURVIVED'
cp: /nonexistent/led.jsonl: No such file or directory
   (SURVIVED is never printed)
```

`cp … && info …` is the function's last command, so a failed copy returns non-zero, and under `set -euo pipefail` the call — a plain command in the body of `if ledger_active`, not a condition — takes the run down mid-publish. The three appenders around it are all `|| warn "… (not blocking)"`, and the durations seed runs *first*, so a full hour of testing ends with no verdict because a `cp` failed. Preflight makes this unlikely (the directory was writable a moment ago), which is exactly why the realistic trigger — a disk that filled during the run — deserves the same `|| warn` the appenders get.

### Verified, and not findings

- `TOKENS_HISTORY` (`watch-tokens.mjs:334`) and `WORKFLOW` (`:455`) are the real knobs, and the **anomaly baseline reads the same `historyPath` it writes** (`:451` / `:521`) — so pointing the variable at the ledger moves read and write together, and the seed is what keeps day zero from answering "fine" off three lines. That is the load-bearing half of the seeding argument and it holds.
- `LIVENESS_DIR` / `SHARD_TOTAL` land where intended (`append-weekly-history.mjs:256-262`), and `Number("") || null` makes the `${SHARD_TOTAL:-}` guard safe when publish runs without prep.
- `ledger_dir_is_outside_repo` is correct on the cases that matter, including the ones the tests do not cover: `LEDGER_DIR` equal to `REPO_DIR` (the `*` in `"$repo"/*` matches empty), a trailing slash, and an ancestor-walk that bottoms out at `/`. The `cd` sits in a command substitution, so no cwd escapes.
- `npm run test:scripts`: **1136 tests, 1135 pass, 0 fail**, as claimed.
- The README's own guidance at line 313 — "prefer appending to an existing file with a discriminator field (e.g. `"workflow": "nightly"`)" — is exactly the shape `daily-stable-vm` takes, so the eventual merge has a documented precedent rather than an ad-hoc one.
- One nuance behind the framing, which does not change the decision: `git pull --ff-only` refuses on a locally-modified tracked file only when the incoming update also touches it. It does today, because the Actions daily commits these three back every morning — which means the *collision* disappears at exactly the moment this lane becomes the sole writer. The durability argument (do not dirty a tracked tree a run does not own) is what carries the decision past that point, and it is the one the code comments lead with. Worth keeping in that order.

### Smaller

- `ledger_dir_is_outside_repo` declares `local dir="$1"` and never uses it.
- `reports/README.md`'s **Sources** table (lines 13-14) still reads as though `daily-stable.yml` is the only writer of the daily series; the ledger is mentioned only in the token caveat further down. One cell, and it is the first thing a reader of that file looks at.
- The body says "Closes the migration's step 17", and there is no issue for it — `gh issue list --search "step 17 in:title"` and `"ledger in:title"` both come back empty. #1716 and #1721 each closed a real issue; this one gives a reviewer nothing to check acceptance against, and the repo's own PR checklist asks for a linked wave item or an approved exception.

Victor-w-Madeira:
## Smoke on the QA machine

Run through the real code path, on a worktree of this branch, against a **copy** of the 2026-09-04 full run (600 results, 1 shard, 3 blobs) — `phase_merge` then `phase_publish`, with `LEDGER_DIR` pointed at a scratch directory. Cleaned up afterwards: worktree removed, scratch directories gone, `/root/e2e-qa` back on `main` with a clean tree.

**The four preflight branches, exercised on the machine.**

| case | result |
|---|---|
| `LEDGER_DIR` inside the clone | refused, exit 1 — **and the directory was not created** |
| neither `LEDGER_DIR` nor `HOME`/`XDG_STATE_HOME` | refused, exit 1, naming `KEEP_LEDGER=0` as the way out |
| `KEEP_LEDGER=0` | passes, `ledger: not kept for this run` |
| a real ledger | `ledger: /root/…`, created |

**The three series, after one publish.**

```
daily-history.jsonl   47 daily-stable  +  1 daily-stable-vm
token-history.jsonl   25 daily-stable  +  1 daily-stable-vm
spec-durations.json   225 files
```

Both seeded from the tracked copies on first write, both trees clean afterwards, `git status --porcelain reports/` empty.

The history row carries the liveness block that `LIVENESS_DIR` + `SHARD_TOTAL` buy, and it is not decoration — it is where the wedge in #1720 shows up:

```json
"backend": {"shard_total":1,"shards_measured":1,"wedged":true,
            "outages_total":2,"down_seconds_total":40,"blips_total":12,
            "collateral_attempts":2}
```

### What the smoke found that the reviews did not

**The spend row was landing with `run_id: null`.** The summarizer takes the run's identity from the environment, which in Actions is simply there; here `phase_publish` was not passing it. The daily-history row for the same run had the id, so the two series would have disagreed about which run they described, and every consumer that groups by run would drop the VM's spend rows — while the file looked complete. Fixed in fef789d0, both labels now produced by `tokens_history_env()` so a test pins them; re-verified on the machine: `{'workflow': 'daily-stable-vm', 'run_id': 'smoke-20260904'}`, totals `{traces: 1, total_tokens: 88}`.

The 09-04 run itself recorded **no** spend line, correctly — it ran with tracing off and the summarizer said so (`no traces recorded`). The token path was proven separately, with a synthetic probe through the same `tokens_history_env` + `env` idiom the publish phase uses.

### A limitation this step does not remove

While #1720's exception stands, the wrapper on the `qa` exports `LANGFLOW_DEACTIVATE_TRACING=true`. No traces means no spend line, so **`token-history.jsonl` will accumulate nothing from the VM** until that issue closes — the file is seeded and correct, and the VM contributes no rows. The other two series accumulate from Monday. Said here rather than discovered later from a series that looks maintained and is not.

### Also self-caught before review

Two `set -e` hazards I introduced or inherited: a failing `cp` in `ledger_seed` and a failing `mv` of the durations table would each have aborted `phase_publish` — after the run, before the verdict. Both fail soft now (a2719456), which is what every other step in that phase already does. And `durations_table()` came out of `phase_prep` so that `USE_LEDGER_DURATIONS=1` with no table there yet says so instead of falling back in silence.

### Monday

`e2e-daily.timer` fires Mon 04:00 EDT (08:00 UTC). `HOME=/root` is set both in the unit and in the wrapper, so `LEDGER_DIR` resolves to `/root/.local/state/langflow-e2e`, which does not exist yet — the run creates and seeds it. Checked before merging, because the new preflight gate is fatal.

PR #1724, opened 2026-09-05
Close the ledger's placement hole, and the pins that did not fire

### LINKED ISSUE #1723: VM ledger: the placement guard has a hole, and three of its pins do not fire
Follow-ups from the review of #1722, all measured on that branch rather than read off it. Two are defects in the guarantee the PR claims; two are holes in what pins it.

### 1. The three series paths escape the placement guard

`LEDGER_HISTORY`, `LEDGER_TOKENS` and `LEDGER_DURATIONS` default under `LEDGER_DIR` but are independently overridable, and `ledger_dir_is_outside_repo` never sees them:

```
$ LEDGER_DIR=/tmp/led-ok LEDGER_HISTORY="$PWD/reports/daily-history.jsonl" \
    bash -c 'source scripts/run-e2e.sh; preflight_ledger; echo "exit=$?"'
    ledger: /tmp/led-ok
exit=0
```

Preflight green, and `HISTORY_FILE` then points at the tracked file — the dirty tree the change exists to prevent, reached through a variable the script itself exposes. The `nothing in the publish phase writes into the tracked series` test cannot catch it: it reads the script text, and the text says `"$LEDGER_HISTORY"`.

"That decision is checked rather than trusted" is the standard set for `LEDGER_DIR`; these three are still trusted.

### 2. `KEEP_LEDGER` accepts any value, and anything but `1` keeps nothing

```
$ KEEP_LEDGER=yes bash -c 'source scripts/run-e2e.sh; preflight_ledger; ledger_active; echo "active=$?"'
    ledger: not kept for this run (KEEP_LEDGER=yes, event=schedule)
active=1
```

The `elif` that makes an empty `LEDGER_DIR` fatal is guarded on `KEEP_LEDGER = "1"`, so a typo'd truthy walks past the `die` into the `else`. The `info` line names the value, but the whole point of that `die` is that nobody reads a log line months later. Same class `require_bool` was added to this file for.

### 3. The preflight branch that runs every night is pinned by nothing

Deleting both lines outright leaves the suite green:

```
-    mkdir -p "$LEDGER_DIR" || die "cannot create the ledger directory ($LEDGER_DIR)."
-    [ -w "$LEDGER_DIR" ] || die "the ledger directory is not writable ($LEDGER_DIR)."
```

Every `preflight_ledger` test is a refusal — inside the clone, nowhere to put it, `KEEP_LEDGER=0`. The accepting path is covered by none of them, and the PR's whole argument for checking at preflight rests on it.

### 4. The absence pin misses the spelling the later etapa will use

`writing back to the repository is absent, not merely switched off` matches `/\bgit\s+(commit|push|add)\b/`, so `git` followed by a flag is invisible:

```
+    git -C "$REPO_DIR" add -A
+    git -C "$REPO_DIR" commit -m "chore(history): daily [skip ci]"
+    git -C "$REPO_DIR" push origin main
```

Suite still green. `git -C` is how a script that operates on a clone by path is written, and it is what the commit in a later stage will most likely be.

### Already fixed after the reviewed commit

`ledger_seed`'s fail-hard `cp` and the unused `local dir="$1"` were fixed in `a2719456` and `2f2831c4`, both after `73a80e17`.

Victor-w-Madeira:
Closes #1723 — the four open findings from the review of #1722. Two are defects in the guarantee that PR claimed; two are holes in what pins it. Every one was demonstrated by mutation in the review, and each fix is verified by putting that same mutation back.

## 1. The three series paths escaped the placement guard

`LEDGER_HISTORY`, `LEDGER_TOKENS` and `LEDGER_DURATIONS` defaulted under `LEDGER_DIR` but were independently overridable, and `ledger_dir_is_outside_repo` never saw them:

```
$ LEDGER_DIR=/tmp/led-ok LEDGER_HISTORY="$PWD/reports/daily-history.jsonl" \
    bash -c 'source scripts/run-e2e.sh; preflight_ledger; echo "exit=$?"'
    ledger: /tmp/led-ok
exit=0
```

Preflight green, `HISTORY_FILE` aimed at the tracked file — the dirty tree the whole mechanism exists to prevent, reached through a variable the script itself exposed. The publish-phase test could not catch it: it reads the script text, and the text says `"$LEDGER_HISTORY"`.

**They are derived now, not read from the environment.** Running their parents through the same check would also have closed it; deriving removes the state instead, and `LEDGER_DIR` is the one knob a machine ever needs. The overrides were undocumented, so nothing depended on them. After:

```
LEDGER_DIR=/tmp/led-ok LEDGER_HISTORY=<repo>/reports/daily-history.jsonl
  →  HISTORY -> /tmp/led-ok/daily-history.jsonl
```

## 2. `KEEP_LEDGER` took any value, and anything but `1` kept nothing

The `elif` that makes an empty `LEDGER_DIR` fatal was guarded on `KEEP_LEDGER = "1"`, so a typo'd truthy walked past the `die` into the `else` — the `die` that exists precisely because a scheduled run quietly keeping no series is indistinguishable, months later, from a machine that was down.

`require_flag` is the `0`/`1` twin of the `require_bool` this file already has. Applied to the two ledger switches and deliberately **not** to the publication ones: a typo'd `CREATE_ISSUE=yes` leaves that switch OFF, the safe direction, while a typo'd `KEEP_LEDGER=yes` disables a fatal gate. Empty stays "unset" — `${KEEP_LEDGER:-1}` is the file's convention everywhere.

## 3. The preflight branch that runs every night was pinned by nothing

Deleting `mkdir -p` and the `-w` check outright left the file at 52/52 green. Every `preflight_ledger` test was a refusal; the accepting path — the one the argument for checking at preflight rests on — was covered by none of them. Now it asserts the directory exists afterwards and that the run logs where it is, plus a `chmod 500` case for the writability check, **skipped under root**, where `-w` is true for anything and the assertion would be theatre.

## 4. The absence pin missed the spelling the later etapa will use

`/\bgit\s+(commit|push|add)\b/` never saw `git -C "$REPO_DIR" add -A`. Widening it needed one more thing to stay honest: the insides of double-quoted strings are stripped along with comments, because a `warn` that tells an **operator** to run `git -C <clone> checkout … # cycle parity, not the commit` is this script asking a human to move a clone, not this script writing to a repository — the widened pattern flagged that line until the strings came out. Same lesson as #1716, one layer over: read code, not text that merely spells it.

## Verification

Each fix was checked by restoring the mutation that exposed it:

| mutation | before | after |
|---|---|---|
| one per-file override put back | green | **fails** `the three series paths are derived` |
| `mkdir` + `-w` deleted | green (52/52) | **fails** both new preflight tests |
| `git -C … add/commit/push` injected into `phase_publish` | green | **fails** `writing back to the repository is absent` |
| `KEEP_LEDGER=yes` | kept nothing, exit 0 | exits 1, naming the value |

`npm run test:scripts`: **1143 tests, 1142 pass, 0 fail** (59 in `run-e2e.test.mjs`, 7 new).

## Docs, alongside the four

`reports/README.md`'s Sources table still read as though `daily-stable.yml` were the only writer of the daily series, with the ledger mentioned only in a token caveat much further down — and that table is the first thing a reader of that file looks at. From the #1722 review rather than from #1723's four items, hence the separate commit.

## Already fixed before this branch

`ledger_seed`'s fail-hard `cp` and the unused `local dir="$1"` were addressed in `a2719456` and `2f2831c4`, both after the commit the review was cut against.

## Not in this PR

The review's closing note — that #1722 said "Closes the migration's step 17" with no issue behind it, which leaves a reviewer nothing to check acceptance against. #1723 is the linked item for this one.


Victor-w-Madeira (commented):
## Review

All four items land, and each one was re-verified here by restoring the mutation that exposed it rather than by reading the diff. **One change requested**, on item 4: the widening is right, but the stripping that accompanies it re-opens the same guard on the spelling this file uses most.

### Verified against the branch (worktree at `8c380616`, `node --test`)

| mutation restored | result |
|---|---|
| `LEDGER_HISTORY="${LEDGER_HISTORY:-…}"` put back | **fails** `the three series paths are derived` |
| `mkdir -p` + `[ -w … ]` deleted from `preflight_ledger` | **fails** both new preflight tests (45, 46) |
| bare `git -C "$REPO_DIR" add -A` in `phase_publish` | **fails** `writing back to the repository is absent` |
| clean tree | 59/59 `run-e2e.test.mjs`; `npm run test:scripts` 1143 tests / 1142 pass / 1 skip / 0 fail |

The two claims the PR body makes about the environment also hold on inspection: `LEDGER_HISTORY|LEDGER_TOKENS|LEDGER_DURATIONS` appear nowhere in the tree outside `run-e2e.sh` and its own test, so removing the overrides breaks no caller; and `reports/README.md`'s new row is accurate — `WORKFLOW="$WORKFLOW_ID"` reaches `append-weekly-history.mjs`'s `workflow` field, so the rows really do carry `daily-stable-vm`.

---

### 1. Requested change — the quote-strip blinds the absence pin to `eval` / `target_ssh`

`.map((l) => l.replace(/"[^"]*"/g, '""'))` removes the insides of **every** double-quoted string, and in this file a double-quoted string is very often *code that runs*, not a message. `target_ssh "…"` is the dominant idiom here — 10+ call sites — and it is exactly how a lane that commits from a VM would be written.

Repro, on this branch with nothing else changed:

```
phase_publish() {
+  eval "git -C $REPO_DIR add -A && git -C $REPO_DIR commit -m x && git push"
```

```
# tests 59
# pass 59
# fail 0
```

So the net of item 4 is one spelling gained (`git -C`) and another lost (`eval "git …"`, `target_ssh "… && git push"`, `bash -c "git commit …"`) — and the lost one is closer to how this script already talks to the far side of an ssh than the one that was gained. Same for the `COMMIT_HISTORY` assertion below it: the switch could come back inside a quoted string and the pin would not see it.

The false positive that motivated the strip is a **single line**, and it is a `warn`:

```
scripts/run-e2e.sh:834
  warn "  git -C <clone> checkout ${TARGET_EXPECTED_BRANCH:-release-<cycle>}  # cycle parity, not the commit"
```

I measured it: with the widened regex and comments stripped, that is the **only** line in the file that matches. Dropping message-emitter lines instead of quoted strings removes it and nothing else, because a `warn`/`die`/`info` line cannot be a repository write in the first place:

```diff
-    .map((l) => l.replace(/"[^"]*"/g, '""'))
+    .filter((l) => !/^(warn|info|die|err|echo|printf)\b/.test(l.trim().replace(/^\|\|\s*/, "")))
```

Checked both directions: clean tree 59/59 green, and the `eval`-hidden write above **fails** test 10. That keeps the widening the item asked for without paying for it on the axis the file actually uses.

### 2. Follow-up — `require_flag` covers two of the switches its own criterion names

The comment states the rule as *apply it where a typo'd truthy walks past a fatal gate; skip it where the typo leaves the switch off, which is the safe direction*. That rule is right, and it also selects four switches this PR leaves unvalidated:

- `DRY_RUN=yes` → `[ "$DRY_RUN" = "1" ]` is false → the run does **not** stop after the partition; a dispatch that asked for a dry run gets a full run on the VM.
- `REQUIRE_TARGET_VERSION=yes` → enforcement silently off, on a flag whose own error text argues *"an unperformed check is not a weaker guarantee — it is none"*.
- `CHECK_TARGET_VERSION=yes` / `PREPARE_TARGET=yes` → same shape (and `PREPARE_TARGET` also feeds `STAMP_REQUIRED`).

Not a blocker — #1723 is scoped to the ledger and the fix that shipped is correct for it. But the comment reads as if the ledger pair were the complete set of unsafe-on-typo switches, and it is not. Worth either a sentence naming the four as known-and-deferred, or a follow-up issue that extends `require_flag` to them.

### 3. Nits

- `new RegExp(\`ledger: ${dir}\`)` interpolates a tmp path into a pattern unescaped. Harmless on this machine (`/var/folders/7w/…`), but the assertion has no reason to be a regex — `assert.ok(r.stdout.includes(\`ledger: ${dir}\`), …)` says the same thing and cannot be surprised by a path character.
- The 4th commit (`docs(reports): name the second writer in the Sources table`) is not one of #1723's four items and is not mentioned in the PR body. The row is correct and I'd keep it; it just needs a line in the description so a reviewer knows it was intended rather than carried over.

### Checklist

- Links `Closes #1723`, a review follow-up — an approved exception per `ROADMAP.md` → *Intake*. ✅
- Infra-only change: no spec, no spec doc, no `QA-CHECKLIST.md` bullet required. ✅
- CI green (TypeScript, ESLint, Python units, checklist guard, spec-doc deps); `Run impacted E2E specs` correctly reports `skipping` — nothing under `tests/` imports the VM lane. ✅

Victor-w-Madeira:
Both reviews land on the same line, and they are right: the quote-strip was the worse half of the trade I made.

Reproduced before changing anything — your `eval` case, on the branch as it was:

```
phase_publish() {
+  eval "git -C $REPO_DIR add -A && git -C $REPO_DIR commit -m x && git push"

# tests 59  # pass 59  # fail 0
```

And your measurement holds exactly: with the widened pattern and comments stripped, **1** line in the file matches, it is the `warn` at `run-e2e.sh:834`, and dropping message-emitter lines takes it to **0**. So the fix is yours as written.

**Fixed in 688786e2.** Emitter lines (`warn|info|die|err|echo|printf`, with a leading `||` allowed) are dropped instead of quoted strings. Four spellings injected into `phase_publish`, each caught now, including the two that were blind before:

| injected | before | after |
|---|---|---|
| `eval "git -C … add -A && … commit … && git push"` | green | **fails** |
| `target_ssh "cd … && git commit -am x && git push"` | green | **fails** |
| `bash -c "git push origin main"` | green | **fails** |
| `git -C "$REPO_DIR" add -A` | fails | **fails** |

Clean tree 59/59, `test:scripts` 1143 / 1142 pass / 0 fail.

The residual is written into the comment rather than left to be found: a line that *begins* with an emitter and then also invokes git (`echo x; git push`) is exempt. Nothing in the file is shaped that way, and closing it means going back to reading inside strings. Say the word if you want it tighter.

Copilot's version of the same finding suggested keeping the strip but scoping it to `warn` lines without `$(`/backticks. Took yours instead — it drops the line rather than reading into it, which is one rule instead of two and cannot be argued with about what counts as executable.

---

**Item 2 — the four other switches.** Verified: `DRY_RUN=yes` gives `RUNS THE WHOLE SUITE`, and `REQUIRE_TARGET_VERSION` / `CHECK_TARGET_VERSION` / `PREPARE_TARGET` have the same shape. Kept out of this PR since #1723 is scoped to the ledger, and did both things you offered rather than one: **#1725** tracks the fix, and the comment above `require_flag` now names the four as known-and-deferred so the two calls under it do not read as the complete set.

**Item 3 — nits.** The `new RegExp(\`ledger: ${dir}\`)` is a plain `includes` now. And the PR body has a *Docs* section for the fourth commit — you are right that it was intended and unmentioned; the `reports/README.md` Sources row came from your #1722 review, not from #1723's four items.

Issue #1717, opened 2026-09-04
Nothing pins the VM lane's env against daily-stable.yml's, and the silent half of that gap cannot be found by running it


Victor-w-Madeira:
Follow-up to #1714, found while reviewing #1716.

#1714 was one variable — `LANGFLOW_DEACTIVATE_TRACING` — pinned off in the source starter while `daily-stable.yml` runs it on. It cost a full VM run to surface and one PR to fix. It is an instance of a class, and the rest of that class is still open.

## What is measured

`daily-stable.yml` sets 12 `LANGFLOW_*` variables on its service. The VM path — `scripts/run-e2e.sh` → ssh → `scripts/start-langflow-source.sh` — sets 8. Discounting the ones that cannot apply to a source instance (`LANGFLOW_IMAGE` / `LANGFLOW_VERSION` select an image; `LANGFLOW_WORKERS` the starter carries as `--workers`), **three** are absent from the whole VM path — not in `run-e2e.sh`, not in `start-langflow-source.sh`, not in `start-langflow-pip.sh`:

| Variable | What the divergence costs |
|---|---|
| `LANGFLOW_ALLOW_CUSTOM_COMPONENTS` | **Red.** The product default is `false` (`docs/core-components/full-custom-component.md:125`; `daily-stable.yml`'s own comment records that the nightly image ships `false` deliberately), so every spec declaring it as a precondition — `core-components/full-custom-component`, `security/code-execution-endpoints` and the siblings — cannot pass on the VM. Plausibly the 9th failure of the run #1714 was raised from, which the tracing fix does not address |
| `LANGFLOW_SQLITE_PRAGMAS` (`foreign_keys:ON`) | **False green.** With SQLite foreign keys off, a cascade/orphan defect (the #13955 class) is invisible on the VM and visible on Actions |
| `LANGFLOW_WORKER_TIMEOUT` | Changes wedge/timeout behaviour (the #922/#927 class), so the two lanes attribute flakes differently |

## Why this is worth a mechanism and not three lines

The first one is discoverable by running the lane: it turns red, ~50 minutes per discovery, which is the cycle #1714 just paid.

The second is **not discoverable by running the lane at all** — it produces agreement, not a failure. And this lane exists to *compare verdicts* with the Actions daily, so a green that should not be green corrupts the exact product stage 1 of the migration is built to deliver. No amount of running exposes that half of the class.

## Proposal

[...]

PR #1728, opened 2026-09-05
Compare the Actions and VM daily verdicts, and record the version that makes the comparison legal

Victor-w-Madeira:
Step 14 of the VM migration — the stage's product — is the classified list of differences between the Actions daily and the VM daily. Monday is the first unattended VM run, so this is the instrument that reads what it returns.

## What it is

`scripts/compare-lane-verdicts.mjs`, over `scripts/lib/lane-verdict-diff.mjs`.

```
node scripts/compare-lane-verdicts.mjs              # newest day both lanes recorded
node scripts/compare-lane-verdicts.mjs --date 2026-09-07 --json
```

It reads `daily-history.jsonl` and nothing else. Both lanes already append one row per run to that series — the VM through the ledger, since #1722 — distinguished by `workflow`, and the row carries the totals plus every failure with file, title, tags, attempts and error signature. So a day's comparison is a diff of two rows in one file: no artifact download, no API call, nothing that expires after seven days.

Output is the three buckets the stage asks for — failed on the VM only, failed on Actions only, failed on both — plus flake-only-on-one-side and severity differences, and it always prints the path it read, because reading the wrong series in silence is the failure this lane keeps finding in itself.

## What it refuses to do, which matters more

Three conditions **block** the comparison and produce no list at all — in `--json` as much as in the text, so the two surfaces cannot tell different stories about one run:

- **different Langflow versions.** That list would be the product's changelog wearing the costume of an environment difference, which is the single thing step 14 says must not happen.
- **a lane that reported top-level run errors.** Its row still carries totals, so it looks like a verdict and is not one.
- **a lane with no row for the day.**

That first blocker was dead code when I wrote it. `langflow_image` is a moving tag — two rows both saying `:latest` are two different products on two different days — and neither lane put the resolved version on the row. Actions resolved it and spent it only on the QA Platform payload; `run-e2e.sh` read it off the served instance and wrote it only to `run-metadata.json`. The first commit puts it on the history row, additive to schema v1 like `infra_signature` and `backend` before it, and a test pins **both** writers — scoped to the appending step, because `daily-stable.yml` already sets `LANGFLOW_VERSION` further down and a file-wide grep would stay green with the history step carrying nothing. That is #1717's shadowing shape, one file over.

## The blind spot, stated instead of assumed away

A history row names the tests that failed or went flaky. It never names the ones that passed, and never the ones that were **skipped**. So a spec skipped on one lane and passed on the other is invisible here, and the only signal the substrate offers is the skipped **count** — reported as a warning that narrows the comparison. A comparator printing "no divergences" over two lanes that ran different test sets would be worse than no comparator.

Other warnings, none of them blocking: version parity unverifiable, differing shard counts (Actions runs 4, the VM 1), differing executed totals, duplicate rows for a date, and failures carrying an `infra_signature` — the harness could not reach the backend, so they are not attributable to the spec that reported them.

## Evidence

33 unit tests, and the review-driven shape the lane uses: every assertion is driven by constructed data, and the ones worth naming pin the failures that would make the list *look* right while being wrong — a line-keyed diff reporting a spec both lanes failed identically as one-sided (the lanes can sit one commit apart), a blocked comparison still handing a list to a JSON consumer, an agreed failure being dropped instead of reported.

**One of those tests failed on the first run and was right to.** The blocked path returned an empty list in the report and a populated one in the result object; the fix is in the code, not the expectation.

**Smoke on real data.** Last night's VM run (`step11-n1`, 592 passed / 9 failed / 1 flaky / 14 skipped) through the real appender, compared against the most recent Actions row:

- the nine failures come out as **VM-only**, and they are exactly the tracing family of #1720 — one of them names its own cause in the assertion
- **11 tests skipped on the VM that Actions ran** — the first number anyone has put on the missing provider keys, which until now was described only as "Wednesday is not comparable"
- the shard-count warning fires (Actions 4, VM 1), and the infra-signature warning catches 2 on Actions and 3 on the VM

Dates were aligned by hand for that smoke, since the two rows are from different days.

## Scope

Read-only. Nothing in the daily path changes except one env var reaching an appender that already ran.


Victor-w-Madeira (commented):
## Verdict: good to merge after adjustments — one real reporting bug and one mandatory doc gap

The design holds up: reading only the JSONL, blocking instead of misleading, and a key that excludes `line` are all the right calls, and the two structural guards genuinely catch the mutation they claim to. Nothing below questions the approach; item 1 is the only one that makes the tool emit something wrong.

### What I verified

- `node --test` on both touched test files: **44/44 pass**. Whole lane (`test:scripts`, 1178 tests): **0 failures**.
- **Mutation-tested both structural guards.** Deleting `LANGFLOW_VERSION` from the *Append daily history* step fails the `daily-stable.yml` test; deleting it from `run-e2e.sh` fails the other. And the scoping claim checks out: deleting the **payload** occurrence instead (line 945) leaves the guard green — a file-wide grep really would have missed the gap.
- Every field the comparator reads exists on the appender's row: `run_errors`, `backend.shard_total`, `param`, `infra_signature`, `totals.skipped`.
- `LANGFLOW_VERSION` is a global assigned in `phase_merge` (`run-e2e.sh:1236`), and `main()` always runs `phase_merge` before `phase_publish`, so `${LANGFLOW_VERSION:-}` arrives populated.
- `testKey` joins on `U+0001`, not on the empty string the diff appears to show, so there is no collision between a file name and a title. Checked, false alarm.

---

### 1. A test flaky on BOTH lanes prints under "Failed on BOTH lanes" — bug

`agreed` collects `agreed-failed` **and** `agreed-flaky`, and `renderReport` puts both under a heading that asserts failure. Executed:

```
kind: agreed-flaky
Failed on BOTH lanes (the product, not the environment): 1
  a.spec.ts :: t
```

This is exactly the "the list looks right while being wrong" class the PR uses as its own criterion — a reader at 09:00 counts a flake as a hard failure on both lanes. Splitting the bucket by kind (two headings, or the status next to the name) fixes it, and it is worth a test: all 33 existing ones only ever construct `agreed-failed`.

### 2. `reports/README.md` does not document `langflow_version` — the file's own rule

> *Backwards-compatible additions (new optional fields): no version bump. **Document the addition in this README** and ship.*

The `## Schema (version 1)` block documents `infra_signature`, `param`, `run_errors` and `backend`; the new field is missing from it. Alongside that, the *Files* table still names `daily-stable.yml` as the only source of `daily-history.jsonl`, while the VM lane has been writing the same series through the ledger since #1722 — which is the premise this whole comparator rests on.

### 3. `defaultHistoryPath()` ignores `LEDGER_DIR`

`run-e2e.sh` accepts `LEDGER_DIR` as an override ("the one knob a machine ever needs"); the comparator only derives `XDG_STATE_HOME` / `HOME`. On a machine with a custom `LEDGER_DIR` it falls back to the clone's `reports/daily-history.jsonl` — Actions rows only — and blocks with "no `daily-stable-vm` row". Not silent, since the path is printed, but it costs a Monday morning to work out. Reading `env.LEDGER_DIR` first is one line.

### 4. A warning disappears from the text report under a blocker, but stays in the JSON

With a `run_errors` blocker plus no version on either row:

```
warnings on the object: ["version parity UNVERIFIED: ..."]
text contains UNVERIFIED?  false
```

`renderReport` prints the blockers and returns before the *Narrowed by* section. It is the mirror image of the asymmetry this PR commits to not having ("in `--json` as much as in the text"). Low impact, but the argument is the PR's own.

### 5. Design question: the version blocker has no escape hatch

It is exact string equality. With `PREPARE_TARGET=1` and `REQUIRE_TARGET_VERSION=1` (steps 16/17) the normal case matches, so this is not a daily blocker — but if `:latest` moves between the Actions pull and the VM's resolution, that day produces **no list at all**, and it is precisely a day somebody wants to look at. Worth deciding and recording either way: an `--allow-version-mismatch` that still stamps the report, or blocking on `X.Y.Z` and demoting a `devN` difference to a warning.

### 6. Nits

- The test comment says the payload step is "further down" — it is **above** (945 vs 1270). The scoping is right, the prose is not.
- No npm alias (`tokens:trend` has one), and nothing in the repo tells Monday's operator the tool exists — there is no VM-migration doc under `docs/` or in `CLAUDE.md`.
- `--json` with a missing history file exits 1 emitting no JSON, so a machine consumer gets nothing parseable.
- The row's `date` is UTC. The lanes degrade well when they drift apart (blocks naming the missing lane, lists `datesAvailable`), but a `--vm-date` would avoid aligning dates by hand as the smoke had to.
- `needs.test.outputs.langflow_version` comes from a **matrix** job (4 shards) — the value is whichever shard finished last. That is already how the payload has consumed it, and the worst case is empty → `UNVERIFIED`, never a wrong version, since all four shards test the same image. Recording it, not asking for a change.

### 7. Process

No `Closes #n`, and there is no open issue for step 14. There is precedent (#1722 had none either), but the repo's PR checklist asks for a link to a wave item or an approved exception issue — worth opening the step-14 issue and referencing it, if only so items 1–3 have an owner when they become follow-ups.

Victor-w-Madeira:
## Review addressed — pushed as `c9fb1f26`

Full lane after the changes: **1185 pass, 0 fail** (41 in this file, up from 33). Real-data smoke re-run and unchanged in substance.

Two findings did not survive checking, and I want to be explicit about which and why rather than quietly skip them.

### Fixed

**1 — a flake both lanes saw printed under "Failed on BOTH lanes".** Correct, and it is the one thing that made the tool emit something wrong. Two headings now (`Failed on BOTH` / `Flaky on BOTH`). Your point that all 33 tests only ever constructed `agreed-failed` is exactly why it could be wrong and green; the new test builds `agreed-flaky` and asserts both counts.

**3 — `defaultHistoryPath()` ignored `LEDGER_DIR`.** Fixed, and the reasoning is yours: `run-e2e.sh` calls it "the one knob a machine ever needs", so a machine that sets it has its VM rows *only* there. The test also pins that a declared-but-nonexistent `LEDGER_DIR` must not shadow a ledger that does exist.

**4 — warnings vanished from the text under a blocker but stayed in the JSON.** Fixed. It was the mirror of the asymmetry the PR commits to not having, so the argument was the PR's own.

**5 — the version blocker had no escape hatch.** Took the first option: `--allow-version-mismatch` compares anyway and **stamps** both surfaces — a banner at the head of the report and `versionMismatch: {ci, vm, allowed}` on the object — so the caveat cannot be lost between running the tool and reading it. I deliberately did **not** take the second: demoting a `devN` difference to a warning would soften exactly the case that needs the guard, since `1.13.0.dev3` against `dev4` is a day of commits on the release branch. There is a test pinning that dev3-vs-dev4 still blocks.

**Copilot's two real ones.** The missing-row blocker hardcoded the default workflow ids, so `--vm-workflow` callers were told a row was missing under a name they never used — fixed, ids now flow through. And the unverified-version warning was **inverted**: with one row missing the field it said that row "carries a langflow_version". My test asserted only on the word `UNVERIFIED`, so it passed. Fixed, with a test per lane.

**6 — nits.** `reports/README.md` documents the field; `--json` stays parseable on the missing-file path; `lanes:compare` alias added; the comment that said the payload step is "further down" now says **above**, which is where it is.

**7 — process.** Opened #1729 for step 14 and referenced it below.

### Checked and not changed

**2, second half — the *Files* table.** It already documents the second writer, since #1722. `reports/README.md:13` reads: *"**A second writer exists off this tree**: the VM lane (`scripts/run-e2e.sh`) appends the same schema to a ledger outside the clone … Its rows carry `workflow: "daily-stable-vm"`"*. The *Source* column heading names `daily-stable.yml`, which is what I think you read; the cell itself is current. The first half of your item was right and is fixed.

**Copilot's entrypoint-guard comment is wrong.** Node resolves `process.argv[1]` to an absolute path regardless of how the script was named on the command line. Verified rather than argued — invoked by relative path from the repo root, the CLI printed the full report:

```
$ node scripts/compare-lane-verdicts.mjs --history /tmp/.../smoke-history.jsonl
Lane verdict comparison - 2026-09-05
  Actions  run 33873006780 | 591 passed, 0 failed, 5 flaky, 3 skipped
  ...
```

### Deferred, with the reason

**`--vm-date`.** Both lanes fire at 08:00 UTC and the row's date is UTC, so a drift needs a lane to cross midnight — and when it happens the tool already degrades well: it blocks naming the missing lane and prints `datesAvailable`. The hand-alignment in the smoke was an artifact of comparing rows from two different days on purpose. I would rather add it the first time a real run needs it than carry a second date axis from day one.

**The matrix-output nit** I am recording, not changing, for the reason you gave: worst case is empty, which reads as `UNVERIFIED`, never as a wrong version.

Refs #1729.

Victor-w-Madeira:
## Independent review round 2 — five findings, four fixed, pushed as `9891f268`

I ran a second review that had no knowledge of why I wrote any of this. It found something the first round did not, and it is the one that would have mattered on Monday.

### 1 — HIGH, confirmed: the comparator read one file, and no file holds both lanes

`defaultHistoryPath()` preferred the ledger, and its comment claimed *"the ledger is the superset"*. That is true for exactly one instant. Verified in the code rather than taken on the reviewer's word:

- `ledger_seed()` (`run-e2e.sh:529`) copies the tracked file in **once** — `if [ -e "$ledger" ]; then return 0; fi` — and is the only thing that ever puts an Actions row in the ledger.
- The Actions daily **commits `reports/daily-history.jsonl` back to main** (`daily-stable.yml:1311`), so the tracked file keeps growing and reaches the VM's clone through the mirror and the wrapper's morning `git pull --ff-only`.

So the ledger holds Actions rows frozen at seed time plus every VM row; the tracked file holds Actions rows and never a VM row. `selectRuns` picks "the newest day BOTH lanes recorded", which is permanently the seed day. Depending on whether the VM ran that day, the tool either blocks with *"the Actions lane has nothing to compare against"* on a morning Actions ran perfectly, or compares a week-old day **and exits 0**.

Fixed by reading both and merging, deduplicated on `workflow|date|run_id` — the seeded copy and the tracked original are the same row, and without the dedupe every seeded day would report a re-run that never happened. `--history` is now repeatable. The test is built in the exact shape of the defect (ledger frozen at 09-01 + VM rows through 09-07, tracked Actions rows through 09-07) and asserts that reading the ledger alone still picks 09-01 — so the test fails if anyone reverts to one source.

Smoke on two real files, one lane each: `--history <vm ledger> --history reports/daily-history.jsonl` compares 2026-09-04 and exits 0.

### 2, 5 — fixed

A blocked day computed **only** the version warning: the early return sat above the skip, executed-count, shard and infra-signature ones. The previous round fixed the *printing* and left the computation where it was, which is the sort of half-fix that reads as done. The reader needs the skip delta most precisely when the day is blocked — it is what says whether re-running with `--allow-version-mismatch` buys anything.

And nothing warned when the compared day was not the newest in the series. That needs nothing to be broken: the VM's history append is `|| warn`, non-blocking, so a run that produced a verdict can leave no row, and the tool would answer yesterday to someone asking about today.

### 4 — the finding is wrong, the underlying problem is real

The reviewer read `join("")` off the rendered diff and filed a collision. The bytes say otherwise — the separator was a literal U+0001 control character, which is exactly what @Victor-w-Madeira checked and cleared in round 1 ("not the empty string the diff appears to show").

Two reviewers, opposite conclusions, same invisible byte. So I fixed what actually caused that: it is written as the six-character escape `\u0001` now. Same behaviour; the difference is that the source shows it. A separator nobody can see is a separator nobody can verify. The collision test stays, because the property is worth pinning whatever the byte is.

### 3 — real, filed as #1731, deliberately not fixed here

`daily-stable.yml` resolves `langflow_version` **per shard** (`lfver`, `continue-on-error`, `curl -sf` against that shard's own service container) and exposes it as a **matrix** output, so the last shard to write wins — and a wedged shard writes empty. On a wedge day, which is the failure this whole comparison exists to study, the Actions row can land with `langflow_version: null` and the version blocker degrades to `UNVERIFIED`.

The asymmetry is the tell: `run-e2e.sh:1235-1242` already sweeps *every* shard's version file, with the comment *"avoids ending up without one just because shard 1 was the one that died"*. The Actions side has no equivalent.

Not fixed here because it changes the daily's matrix plumbing three days before the first unattended VM run, and because the degradation is honest — `UNVERIFIED`, never a wrong version. Same argument that kept `SHARDS` at 1 for Monday.

### Verification

- 49 tests in this file (was 41), full lane **1193 pass, 0 fail**, twice in a row.
- One full-lane run mid-session reported a single failure I did not capture; two subsequent full runs are clean and this file is 12/12 across repeated runs, so it was not from here. Recording it rather than calling the lane green without qualification.

Refs #1729, #1731.

Issue #1732, opened 2026-09-05
watch-upstream-areas tests leak a temp git repo per fixture — 2699 of them on one machine


Victor-w-Madeira:
Found by the independent review of PR #1730 (#1182), and confirmed by measurement afterwards. Pre-existing, unrelated to that PR's diff.

## What leaks

Four fixture helpers in `scripts/watch-upstream-areas.test.mjs` create a temp directory and **never remove it** — no `finally`, no `t.after()`:

| Helper / site | Prefix | Call sites per run |
|---|---|---|
| `upstreamFixture()` | `watcher-upstream-` | 8 |
| `docsFixture()` | `watcher-docs-` | 6 |
| the no-remote CLI test | `watcher-noremote-` | 1 |
| the bad-ref CLI test | `watcher-badref-` | 1 |

The first two are **real git repositories** — each runs `git init` plus commits on a trunk and a release branch — so every `npm run test:scripts` leaves 16 of them behind.

Measured on this machine, 2026-09-05:

```
$ ls "$TMPDIR" | grep -cE '^watcher-(upstream|docs|noremote|badref)-'
2699
$ du -sh "$TMPDIR"
1.3G
```

## Why it is worth fixing rather than tolerating

Not for the bytes. Two reasons:

1. **The file already knows how to do this.** The synthetic checkout fixtures in the same file (`buildCheckoutFixture`, the symlink test, the sweep test) all clean up in a `finally`. The four above are the exceptions, so the inconsistency reads as an oversight rather than a decision — which is exactly how the next fixture gets written without cleanup.
2. **CI hides it.** Runners are ephemeral, so the lane is green and stays green; the cost lands entirely on whoever runs the unit lane locally, repeatedly, which is the person iterating on this script. That is the same asymmetry `#1012` keeps naming: a cost nobody's dashboard shows is not a cost nobody pays.

## Not a defect in the tests' verdicts

Every one of the 14 affected tests asserts correctly and passes. This is hygiene only — no assertion depends on the directory surviving, and none is weakened by removing it.

## Done when

- [ ] `upstreamFixture()`, `docsFixture()` and the two inline `mkdtempSync` sites remove their directory — `t.after(() => fs.rmSync(dir, { recursive: true, force: true }))` is the least invasive shape, since the helpers return the path and the callers use it throughout the test body.
- [ ] `npm run test:scripts` twice in a row leaves the `watcher-*` count in `$TMPDIR` unchanged (state it in the PR, measured).
[...]

PR #1736, opened 2026-09-05
Merge shard reports above one shard, and name the failure instead of dying silently

### LINKED ISSUE #1726: run-e2e.sh cannot merge shard reports above SHARDS=1: one working copy per shard means one testDir per blob
`scripts/run-e2e.sh` cannot produce a verdict at any `SHARDS` above 1. The shards run, every blob is written, and the run then dies in `merge-reports` — at the last phase, after the full test time is already spent.

## What happens

`prepare_shard_workdir()` (`scripts/run-e2e.sh:1011`) gives each shard its own working copy of the tree at `runs/<id>/shard-N/`, because `collect-models` writes into it. Playwright records that path as the blob's `testDir`. `merge-reports` then refuses blobs whose `testDir` values differ:

```
    4/4 blobs present
Error: Blob reports being merged were recorded with different test directories, and
merging cannot proceed. ...
Found directories:
/root/e2e-qa/runs/step11-n4-20260905T023736Z/shard-1/tests
/root/e2e-qa/runs/step11-n4-20260905T023736Z/shard-2/tests
/root/e2e-qa/runs/step11-n4-20260905T023736Z/shard-3/tests
/root/e2e-qa/runs/step11-n4-20260905T023736Z/shard-4/tests
```

The script runs under `set -euo pipefail`, so this aborts the run where it stands. No `results.json`, no `run-metadata.json`, no verdict, nothing appended to the ledger — and, because the publish phase never runs, **no Slack message either**. The day ends in silence.

## Evidence

A step 11 measurement run on the QA VM, suite at `5e72e0b1`, `SHARDS=4`:

- all four shards ran to completion — 616 tests, 19.7 min wall clock
- `4/4 blobs present` — integrity guard 1 reported health one line before the failure
- `exit=1`, no merged report

Merging each blob **individually** succeeds, which is the proof of cause: alone, each blob has a single `testDir`. Recovered that way, the run's real verdict was 589 expected / 9 unexpected / 4 flaky / 14 skipped — the 9 being exactly the known tracing family of #1720, with no contention failures added by the four-way concurrency.

## Why this has never been seen

Every run of this lane so far has been `SHARDS=1`: one blob, one `testDir`, merge succeeds. The wrapper on the QA VM pins `SHARDS=1` with a comment inviting the change — *"Raise this once the load measurement lands"* — which walks straight into this.

## Why the Actions lane does not have it

There each shard runs on its own runner, at the same absolute path, so the `testDir` values coincide. The per-shard working copy is a VM-side construct, so this is a divergence in the instrument rather than in the product — the class this migration exists to find, turned against the tooling.

## The shape of the failure is the worst available

It costs the entire test time before it fails, it fails after the guards have reported health, and it fails silently on the surface the lane uses to announce itself. What does catch it is the watchdog's third case ("finished without `run-metadata.json`") — the alarm works, and the day's data is still gone.

## Fix

Playwright's own error names the remedy: pass `merge-reports -c <config>` whose `testDir` points at the real tests location, so the merged report's paths are correct against the repository rather than against a per-shard copy. The fix has to keep the per-shard copies — `collect-models` writing into the tree is why they exist.

Two things worth pinning with the change, because both were true here and neither was reported as a problem:

- a failing merge should be caught and named, not left to `set -e` — the run knows at that moment that it has 4/4 blobs and no report, which is a far better message than a stack trace
- the failure path should still reach the publish phase, or the one surface that would have told a human stays quiet

## What it blocks

Step 11 (shards under real load) cannot conclude while the lane cannot merge, and `SHARDS` cannot be raised above 1 at all. The resource side is already measured and comfortable — at four shards the target peaked at 30% CPU and 9.1 GB of 60, the runner at 54% and 9.6 GB, swap zero on both — so this defect, and the tunnel's four forwarded ports, are what actually cap the shard count today. Neither is CPU.

Victor-w-Madeira:
Closes #1726 — the defect that made the VM lane unable to produce a verdict at any `SHARDS` above 1.

## The fix, in two halves

**The flag.** `prepare_shard_workdir()` gives each shard its own working copy of the tree, because `collect-models` writes into it. Playwright records that copy's path as the blob's `testDir`, and `merge-reports` refuses blobs whose values differ. Passing a merge config is the remedy Playwright's own error text names.

It lives in `playwright.merge.config.ts` rather than being generated at run time, so it is a reviewable file — and a test fails if its `testDir` ever drifts from the suite config's. That drift would not fail loudly; it would relabel every path in the merged report.

**The failure path, which matters more.** Under `set -euo pipefail` a failed merge aborted the run where it stood: the whole test time already spent, no verdict, nothing appended to the ledger, and — because `phase_publish` never ran — no Slack message either. The day ended in silence, one line after the integrity guard had printed `4/4 blobs present`.

It is now caught and named, and it deliberately does **not** return. Guard 2 (`check-run-integrity.mjs`) is written for exactly this state — "there is no readable report" — and letting it run is what turns a failed merge into a red verdict that still reaches publish.

## What the review changed

The first version of this PR got the flag right and the failure path wrong. The review measured, rather than argued, that the abort had only **moved one phase later**, and four other things besides. All six are fixed; the full answer is in [this comment](https://github.com/oriontech-me/langflow-e2e/pull/1736#issuecomment-5555178440), and each is its own commit.

| # | what was wrong | commit |
|---|---|---|
| 1 | the run still died, now at `phase_publish` — `build-run-payload.mjs` exits 1 with no report and was the one step there without a `\|\| warn`, so the new verdict branch was unreachable in production | `ff593196` |
| 2 | the "does not abort" test opened with a blanket `set +e`, which disabled the mechanism under test — a `phase_merge` that really aborts passed it | `1bbc0635` |
| 3 | `MERGE_OK` never left the terminal: the issue and the Slack message still headlined "executed ZERO tests" and told triage to find why nothing ran | `46bfe6d3` |
| 4 | `run-metadata.json` recorded nothing about the merge, so a run whose four shards passed was byte-identical to one where nothing ran | `c21c852b` |
| 5 | the drift test compared the `testDir` *string*, which Playwright resolves against each config's own directory — the two could point at different trees and stay green | `1bbc0635` |
| 6 | **mine, found fixing 1**: the fail-soft turned an unanalysable report into `Green run.` and exit 0 | withdrawn |
| 7 | **mine, found fixing 6**: that branch fired on cases the guards had already rejected, contradicting the branch above it | withdrawn |
| — | 6 and 7 were both the wrong instrument for 1. Reworked to the narrower remedy the review offered first, and the two commits containing them reverted | `a319c8f1` |

Numbers 6 and 7 were both mine, both in the same hunk, and both came out of one choice: the review offered two remedies for finding 1 and I took the second — `|| warn` on the payload build — rather than the first. Two defects in one hunk is not two mistakes; it is the wrong instrument. The defect is that a failed merge leaves no report, which happens on one kind of run, and a fail-soft tolerates payload failures on *every* run.

`a319c8f1` uses the first remedy instead: the build is **skipped** when there is nothing to build from, and left completely alone otherwise.

| situation | now |
|---|---|
| merge failed | build skipped; the notifiers read the guards |
| merge fine, report unreadable | build skipped; the empty branch fails the run |
| merge fine, report readable, build dies | **aborts, exactly as it always has** |
| anything else | byte-for-byte unchanged |

The verdict needs no new branch, and both defects stop being reachable rather than being answered. Net effect on the diff: 78 insertions against 102 deletions — smaller than what it replaces. The whole payload question is now about fifteen lines in one place.

For the record, what 6 was: `check-run-integrity.mjs` reads `stats`; `build-run-payload.mjs` walks `suites`. A report where the first is fine and the second is not passes every guard and still cannot be analysed — and making the build fail-soft turned that from an abort into a clean-looking day with no totals and no QA Platform record. The verdict now answers for it separately from the report's shape, and stays quiet when the merge failed, where the absent payload is the consequence rather than the cause.

## Verified on the blobs that found it

The four real blobs from the `SHARDS=4` measurement are still on the QA VM, so the flag is checked against the exact material that reproduced the defect rather than a fixture:

| | result |
|---|---|
| without `-c` | `exit=1` — *"Blob reports being merged were recorded with different test directories"* |
| with `-c` | `exit=0` |

And the merged report is right, not merely produced: **589 expected / 9 unexpected / 4 flaky / 14 skipped** — matching, count for count, the totals recovered by merging each blob separately. `rootDir` resolves to the repository's `tests` directory, and all 231 spec paths are repository-relative — which is what `compare-lane-verdicts.mjs` keys on, so a per-shard path here would have produced false divergences in the very list stage 1 exists to build.

Independently confirmed against Playwright's own implementation: `--reporter` wins over a config's reporters (`lib/program.js:263-267`), and `rootDirOverride = config.config.rootDir`, derived in `common/config.js:79` as `pathResolve(configDir, userConfig.testDir)` — so `testDir: "./tests"` at the repo root yields exactly those repository-relative paths.

The smoke ran with the config copied to the VM's clone and removed afterwards; the clone's tree was verified clean before and after.

## Measured, not asserted

Every fix above was checked by mutation — the fix reverted, the test expected to fail, and it does:

| mutation | result |
|---|---|
| revert only `run-e2e.sh` | the end-to-end test fails at `REACHED_AFTER_PUBLISH`; output stops at `[payload] no .../results.json` |
| revert only the notifiers | fails on `[Daily Failure] @stable run executed ZERO tests` |
| `npx ...` then `if [ $? -ne 0 ]` — the natural refactor | the "does not abort" test now fails; under the old `set +e` it passed |
| merge config one directory down, same `"./tests"` literal | literal comparison passes, resolved comparison fails |
| restore `\|\| warn` on the payload build | `Green run.` returns, and the abort test fails |
| remove the skip gate | the builder runs against a missing report, and the skip test fails |

## Tests

Eight new, and the lane at **1209 pass / 0 fail** (was 1197):

- the merge passes the config — read from `phase_merge` with comments stripped, because this file explains the flag at length right above the call and a check that a comment mentions it would pass with the flag removed
- the merge and suite configs *resolve* to the same `testDir`
- a failed merge is named, does not abort, and guard 2 still classifies it — with errexit **on**, so an abort is observable as the absence of the marker
- `run-metadata.json` carries `merge_ok=false`
- a failed merge fails the verdict and does not call it "zero tests executed"
- a failed merge reaches **publish**, and the issue and the Slack message name the merge rather than an empty run — the three phases in `main()`'s own order, both notifiers on in dry-run
- the merge-failed shape outranks `empty` in the issue, and absent `MERGE_OK` leaves the Actions body unchanged
- the same for Slack

Each of the five commits was run through `test:scripts` on its own, so the branch bisects.

## What this does NOT change, and what is NOT verified

Monday's scheduled run stays at `SHARDS=1`, where the merge already worked. Raising it is a separate, deliberate decision that wants a confirmation run behind it.

`CREATE_ISSUE`, `NOTIFY_SLACK` and `POST_QA_PLATFORM` are all still `0` by default, so the notifier half of this PR is dormant until steps 08/09.

`merge_ok` has **no reader yet** — `grep -rl run-metadata` finds this script and its test and nothing else. It is evidence kept for the stage that will compare the lanes, and the comments say so rather than claiming a consumer that does not exist.

**The failure path has never run on the VM** — and on inspection that matters less than it sounds, so it is stated rather than held open as a gate. The trigger is already measured in the real environment: against the four real blobs, without `-c`, `merge-reports` exits 1, which is exactly the condition the new guard tests. `$RUN_DIR/logs` exists by then (`run-e2e.sh:710`, in `phase_preflight`), so `merge.log` is writable. And the failure path reads only `$RUN_DIR`, `$REPO_DIR` and `$SHARD_TOTAL` — nothing that behaves differently under the systemd unit. What a VM smoke would add over the tests is proof under the real unit, and this path uses nothing environment-dependent the nightly success path does not already use.

The confirmation run that DOES still matter is the separate one this PR does not do: a real `SHARDS=4` run on the VM before the schedule is raised. That exercises the flag in the lane with consequence.


Victor-w-Madeira (commented):
## Review

The `-c` half is right, and the evidence behind it is the right kind: the real blobs, the recovered totals matching count for count, and `rootDir` checked against what `compare-lane-verdicts.mjs` keys on. `playwright.merge.config.ts` as a reviewable file with a drift test on `testDir` is the correct shape — that value silently relabelling every path is exactly the failure a comment could not prevent.

The second half — "a failed merge is named, does not abort, and still reaches publish" — **does not hold in the real call sequence**, and #1726's second acceptance criterion ("the failure path should still reach the publish phase, or the one surface that would have told a human stays quiet") is not met. Details below, measured rather than argued.

---

### 1. The run still dies before the verdict, one phase later — now at `phase_publish` (blocking)

`phase_merge` no longer aborts. But `main` is `phase_merge; phase_publish; phase_verdict`, and the **first** thing `phase_publish` does is build the payload:

```sh
PLAYWRIGHT_JSON="$RUN_DIR/results.json" ... node scripts/build-run-payload.mjs > "$RUN_DIR/payload.json"
```

`build-run-payload.mjs:13` is `if (!existsSync(reportPath)) { console.error(...); process.exit(1); }`. There is no `|| true` on that line, and the script is `set -euo pipefail`. This PR's own premise is that a failed merge leaves no `results.json` (it is why guard 2 marks the run empty), so that branch is taken by construction.

Measured on this branch's sequence, with a `RUN_DIR` that has no `results.json`:

```
==> Building the run payload
[payload] no <run>/results.json
exit=1        # "REACHED_AFTER_PUBLISH" never printed
```

So on a merge failure the run still ends where it stood: **no verdict** (the new `MERGE_OK` branch in `phase_verdict` never executes), no history append, no token summary, and **no Slack message** — the day still ends in silence. The four new tests do not see this because they call `phase_merge` and `phase_verdict` directly and never traverse `phase_publish`.

One consequence is worse than a no-op: `run-metadata.json` is written at the **end** of `phase_merge`, which now completes. Per #1726, the watchdog's third case is "finished without `run-metadata.json`" — that is the alarm that *did* fire for this defect. After this change the file exists, so that case stops firing while publish still aborts. The silent failure becomes an undetected silent failure.

Suggested shape (either is fine, the second is closer to the PR's own argument):

- gate the payload build on a readable report — `if [ "${MERGE_OK:-true}" = "true" ] && [ "${RUN_UNREADABLE:-}" != "true" ]; then ... fi` — and let the notifier block run regardless; or
- keep the build and make it fail-soft **only** in this state, since `notify-slack.mjs` already handles an absent payload (`[slack] ... is unreadable — reporting what the guards saw instead`, line 119).

Whichever, the end-to-end property is worth a test that runs `phase_merge; phase_publish; phase_verdict` with the stub `npx` already written here — that is the property the PR body claims, and it is currently the one nothing covers.

### 2. The "does not abort" test cannot observe aborting

```js
"set +e",
...
"phase_merge",
'echo "AFTER_PHASE=$? MERGE_OK=$MERGE_OK"',
```
with `assert.match(..., /AFTER_PHASE=0/, "the phase must not abort — guard 2 has to run")`.

Shell options are global, not per-function, so `set +e` disables the very mechanism under test for the whole body. A mutation that **does** abort a real run passes all four assertions — e.g. replacing the `if !` guard with the natural refactor:

```sh
npx playwright merge-reports ... > "$merge_log" 2>&1
if [ $? -ne 0 ]; then MERGE_OK=false; err "merge-reports FAILED ..."; fi
```

Under the harness's `set +e`: `AFTER_PHASE=0`, `MERGE_OK=false`, `merge-reports FAILED`, `different test directories` — green. Under real `set -e`: the run dies at that line. Verified both ways.

The existing `verdict()` helper needs `set +e` for a genuine reason (it asserts on a non-zero return). `phase_merge` returns 0 in this scenario, so it does not: keep `set -e` on and let a failure to reach the `echo` be the failure.

### 3. `MERGE_OK` only exists in the terminal — the published surfaces still say "ZERO tests executed"

The distinction the PR argues for is made in `phase_verdict`'s stderr. Everything downstream still receives `RUN_EMPTY=true` / `RUN_UNREADABLE=true`, and both consumers are hardcoded around that:

- `notify-slack.mjs:230` — `⚠️ Daily @stable executed ZERO tests — <date>`
- `create-failure-issue.mjs:179` — `[Daily Failure] @stable run executed ZERO tests on ...`

Those are the surfaces a human reads at 06:00; the stderr line is the one nobody opens. Once `NOTIFY_SLACK`/`CREATE_ISSUE` turn on at steps 08/09, the "true sentence pointing at the wrong repair" this PR sets out to remove is precisely what gets published. Passing `MERGE_OK` into both (and into `run-metadata.json`, so the comparator and the history can tell a merge failure from an infra abort after the fact) is what makes the fix reach the reader.

### 4. `playwright.merge.config.ts` is typechecked by nothing (minor)

`tsconfig.json`'s `include` names `playwright.config.ts` explicitly and does not glob the root, and `npm run lint` is `eslint tests/`. The new file is in neither, so `defineConfig` here is checked by no gate — same class as #1593. One line in `include` fixes it; the drift test covers the `testDir` value but not the file's validity.

---

### Verified

- `-c` reasoning against `phase_merge` / `prepare_shard_workdir` / `playwright.config.ts:75` (`testDir: "./tests"` on both sides) — consistent.
- `phase_merge` completes on a failed merge: guard 2, outages, version sweep and `run-metadata.json` are all `|| true`-tolerant. Confirmed.
- No repo glob picks the new root config up as a suite-wide surface (`impacted-specs-by-import.mjs` matches `playwright.config.ts` exactly), so the E2E lane `skipping` on this PR is correct.
- Not re-verified: the VM smoke on the four real blobs, and the `1197 pass` lane (CI's TypeScript Check is green).

Findings 1 and 2 are the ones I would block on: 1 because the stated behaviour is not the behaviour, 2 because it is what let 1 through.

Victor-w-Madeira:
## Answering the review

All five findings are addressed, and a sixth turned up in the fix for the first one — that one is mine, and it is the worst of the six. Each item below says what was measured rather than what was intended.

Head is `1baa25be`. `main` was merged in (`8c2a2e8f`) — #1732 landed while this was open and conflicted on one import block; the resolution is in that commit's message.

---

### 1. The run died at `phase_publish` instead — fixed in `ff593196`

Confirmed exactly as reported. `build-run-payload.mjs:13` exits 1 with no `results.json`, and it was the only step in `phase_publish` without a `|| warn`. Reproduced on the branch with a `RUN_DIR` holding no `results.json`:

```
==> Building the run payload
[payload] no <run>/results.json
exit=1        # REACHED_AFTER_PUBLISH never printed
```

The build is now fail-soft and says so, and the QA Platform POST is skipped rather than sent an empty body. Your second suggested shape, for the reason you gave: `notify-slack.mjs:119` is already written for an absent payload.

The end-to-end property has the test you asked for. It drives `phase_merge; phase_publish; phase_verdict` in `main()`'s own order, with the stub `npx` already written here and both notifiers on in dry-run. Measured both directions:

| mutation | result |
|---|---|
| revert only `run-e2e.sh` | fails at `REACHED_AFTER_PUBLISH` — output stops at `[payload] no .../results.json` |
| revert only the notifiers | fails on `[Daily Failure] @stable run executed ZERO tests` |

On the watchdog point — you are right that this made it worse, not neutral. `run-metadata.json` now exists, so "finished without `run-metadata.json`" stops firing while publish still aborts. That is written into the comment above the fix, because it is the reason the fix could not wait.

### 2. The "does not abort" test could not observe aborting — fixed in `1bbc0635`

Your mutation reproduces. Errexit stays on now, and the marker moved after the call, so an abort is observable as its absence:

| `phase_merge` shape | old test | new test |
|---|---|---|
| the `if !` guard (this branch) | pass | pass |
| `npx ...` then `if [ $? -ne 0 ]` | **pass** | **fail** |

The verdict test keeps `set +e` for the reason you allow — `phase_verdict` is expected to return non-zero — but scoped to that one call, the way the `verdict()` helper already scopes it.

### 3. `MERGE_OK` only existed in the terminal — fixed in `46bfe6d3`

Both notifiers now receive it. Your two lines, plus a third: `create-failure-issue.mjs:117` (`### ⚠️ ZERO tests executed — infra abort`) and the closing `Triage this as infrastructure: find why nothing ran`.

Each script grows a fourth shape, ranked ahead of `empty` for the same reason the verdict branch is. It names the merge step, says the zeros are unread rather than measured, and points at `merge.log` and the blobs. Absent `MERGE_OK` is a working merge, so the Actions lane — which never passes it, and has never had this failure mode — renders byte-for-byte what it did. Pinned in both test files.

The comment above the Slack block said the message and the issue are two views of one verdict and must not disagree. It is true again; it says why.

### 4. `merge_ok` in `run-metadata.json` — added in `c21c852b`, and the reason corrected in `1baa25be`

Added. But the reason I gave for it — repeating your "the file stage 1's comparison actually consumes" — does not hold, and I should have checked before writing it down:

```
$ grep -rl run-metadata --include='*.mjs' --include='*.ts' --include='*.sh' --include='*.yml' . | grep -v node_modules
scripts/run-e2e.sh
scripts/run-e2e.test.mjs
```

The writer and its test. `compare-lane-verdicts.mjs` does not open the file; neither does the history append. The field is still right — without it a run whose shards all passed but whose merge failed is byte-identical to one where nothing ran — but it is evidence for the stage that will compare the lanes, not a field with a reader today. Both comments now say that.

### 5. `playwright.merge.config.ts` typechecked by nothing — fixed in `2efc21d0`

One line in `include`. `tsc --noEmit` green with it.

---

### 6. Mine: the fail-soft made a green day out of a run nobody could analyse — fixed in `6221e171`

Found by asking what fix 1 does on runs that are *not* a failed merge.

`check-run-integrity.mjs` reads `stats`; `build-run-payload.mjs` walks `suites`. Different fields of the same file, so a report can pass every guard and still be unanalysable. Before the fail-soft that aborted the run — loud, and at least visible. After it, measured on `{"stats":{"expected":300,...},"errors":[],"suites":42}`:

```
::warning:: the payload build FAILED — the notifiers report what the guards saw instead.
==> Green run.
VERDICT_EXIT=0
```

No totals, no QA Platform record, notifiers silent because the day looked clean. #1012's failure pointed a third way, and worse than the abort it replaced.

The verdict answers for it now, independently of the report's shape — the report can be readable and its *analysis* still fail. Quiet when the merge failed, where the absent payload is the consequence and the branch above already names the cause. Removing the branch and keeping the fail-soft puts `Green run.` back and fails the new test.

---

### Lanes

`test:scripts` 1209 pass / 0 fail (was 1197), `test:units` 1046 pass, `typecheck` and `lint:ci` clean. Every CI check green on `1baa25be`.

Each of the five commits was run through `test:scripts` on its own, so the branch bisects.

### What is NOT verified

The failure path has never run on the VM. The `-c` half was measured against the four real blobs; this half has only ever run against a stub `npx` in a sourced shell. The test proves the wiring, not the environment — a real `merge-reports` failure on the QA VM, checked for `merge_ok=false` in the metadata and a merge verdict rather than "ZERO tests", is the same standard the other half was held to and I would want it before `SHARDS` goes above 1.

Lower stakes than it sounds for today: `CREATE_ISSUE`, `NOTIFY_SLACK` and `POST_QA_PLATFORM` are all still `0` by default, so the whole notifier half is dormant until steps 08/09.

Victor-w-Madeira:
## A seventh, in the fix for the sixth

Re-reading the three points I had asked to be re-read, adversarially rather than to confirm them. Two hold. The third did not.

### `run-e2e.sh:1538` — the branch contradicted the one above it

The condition I shipped in `6221e171` was `PAYLOAD_OK=false && MERGE_OK=true`. That is not the case it was written for. With the merge fine and no `results.json`, the guards report the run empty and unreadable — and the verdict printed, in this order:

```
::error:: ZERO tests executed — an infrastructure abort... find out why nothing ran, not which test broke.
::error:: the run payload could NOT be built from a report the guards ACCEPTED — this run has no analysis...
::error:: Triage: the report itself is readable, so start from scripts/build-run-payload.mjs...
```

Two contradictory triage instructions in one verdict, and the second is false twice: the guards did not accept it, and there is no report to be readable. The defect this PR is about, reintroduced by the fix for it, one commit later.

Fixed in `344facb6`. The branch is gated on the guards having accepted the report. Every excluded case is already red with its own cause named — there the missing payload is a consequence, not the finding, which is the same reason it stays quiet on a failed merge. What is left is the case that would otherwise be green, and that was always the only one that needed a branch.

| mutation | result |
|---|---|
| gate removed (i.e. `6221e171`) | the contradiction test fails; the green-day test still passes |
| branch removed entirely | `Green run.` returns; the green-day test fails |

Lane: 1210 pass / 0 fail.

### The two that hold

**`run-e2e.sh:1356`/`:1369`** — one thing I had not checked and should have: whether a failed build can leave a partially-written `payload.json` that parses. It cannot. `build-run-payload.mjs` ends in a single `process.stdout.write(JSON.stringify(payload))`, so the file is either whole or empty, and `notify-slack.mjs` handles empty. `warn` returns 0, so the `|| { ... }` cannot trip errexit, and `PAYLOAD_OK` is global rather than `local`, so `phase_verdict` sees it.

**`run-e2e.test.mjs:1015`/`:1020`** — errexit is in effect at the call (the sourced script sets it), the marker is after the call, and your `if [ $? -ne 0 ]` mutation fails it. Nothing further found.

---

### What this says about the review state

Two of the seven findings are mine, and both arrived *after* I had said the branch was ready — the second one inside the fix for the first. Ordinary review would probably not have caught either; both needed someone to run the failure path with different inputs than the ones the tests use. That is worth weighing when deciding how much of a re-read this wants.

Head is now `344facb6`.

Victor-w-Madeira:
## Reworked to the first of your two shapes

Findings 6 and 7 were both mine, both in the same hunk, and both came out of the same choice: I took the second remedy you offered (`|| warn` on the payload build) rather than the first (gate the build on there being a report). Two defects in one hunk is not two mistakes, it is the wrong instrument — the defect is that a failed merge leaves no report, which happens on one kind of run, and a fail-soft tolerates payload failures on **every** run.

`a319c8f1` replaces it with your first shape, and reverts the two commits that existed only to contain the wider one.

| situation | now |
|---|---|
| merge failed | build skipped; the notifiers read the guards |
| merge fine, report unreadable | build skipped; the empty branch fails the run |
| merge fine, report readable, build dies | **aborts, exactly as it always has** |
| anything else | byte-for-byte unchanged |

The verdict needs no new branch. Findings 6 and 7 stop being reachable rather than being answered — the "Green run." case cannot arise, because a build that runs and fails still kills the run the way it did before this PR.

Net effect on the diff: **78 insertions against 102 deletions**. The change is smaller than the one it replaces.

Two tests replace the two removed:

| mutation | result |
|---|---|
| restore `\|\| warn` on the build | "Green run." returns; the abort test fails |
| remove the skip gate | the builder runs against a missing report; the skip test fails |

Lanes: `test:scripts` 1210 pass / 0 fail, `test:units` 1046, `typecheck` and `lint:ci` clean.

Head is `a319c8f1`. The whole payload question is now ~15 lines in one place, `phase_publish`.

PR #1749, opened 2026-09-07
Record the mirrored environment that was in force, so a machine-local override stops being invisible

### LINKED ISSUE #1748: check-vm-env-parity.mjs reports OK while a machine-local override changes the target's environment: the guard reads the repository, the exception lives on the VM
`check-vm-env-parity.mjs` answers "does the VM lane account for every environment variable `daily-stable.yml` sets on the Langflow service it tests?" by reading two files in this repository: the workflow, and `run-e2e.sh`'s `mirrored_target_env()`. Both were correct on 2026-09-07, and the guard printed:

```
VM env parity OK — 10 variables on the daily's service: 10 carried to the target, 0 recorded out of scope.
```

The instance under test that morning was nevertheless started with `LANGFLOW_DEACTIVATE_TRACING=true`, against the workflow's `"false"`. The value does not come from either file the guard reads — it comes from an `export` in the daily wrapper that lives on the QA VM, outside any clone. `run-e2e.sh` defaults the variable to `false` and mirrors it, exactly as designed; the wrapper's export wins because it is set in the environment the orchestrator inherits.

The exception itself is deliberate and documented next to the line (it holds #1720's wedge at bay). **This issue is not about the exception. It is about the guard being unable to see that one exists.**

## Why this matters more than a stale verdict

The guard's own failure message states the reason it is fail-closed:

> half of this class produces AGREEMENT, not failure. A variable the daily sets and the VM lane does not carry can make the VM pass a spec the Actions daily would fail — and the migration's whole product is the comparison between those two verdicts, so no run can find it by going red.

That argument holds identically for a variable the VM lane carries and a machine then overrides. The guard closes the repository-shaped half of the hole and leaves the machine-shaped half open — and the open half has the same silent failure mode. While it stays open, "no open divergence can falsify a verdict" (the Phase 2 switch criterion) cannot be asserted honestly, because not every override is visible.

## Measured cost, 2026-09-07

The first real lane comparison produced 20 divergences. **Sixteen of them trace to this single variable** — 9 failures across the traces/observability family plus `credential-secret-exposure`, and 7 of the 14 VM skips in the same family. Establishing that required reading a file on the VM over SSH; nothing in the run's own output named the value that was actually in force.

## Proposed fix

Record the **effective** value of every mirrored variable in `run-metadata.json`, alongside the suite SHA and the resolved Langflow version already there. The orchestrator knows these values at the moment it builds `mirrored_target_env()`; writing them costs nothing and makes the run self-describing.

Two consequences worth having:

- a machine-local override becomes visible in the run's own artifact, without a shell on the machine;
- `compare-lane-verdicts.mjs` gains a source it can check, so a lane whose environment drifted from the workflow's can be reported by the comparator rather than discovered by investigation.

## Properties to pin

- **The recorded value is the effective one, not the default.** A test that exports an override and asserts the metadata reflects it — asserting the default would pass against the very bug this issue describes.
- **Every mirrored name appears.** The enumeration in `mirrored_target_env()` and the set written to metadata cannot drift apart; adding a variable to one without the other should fail the lane, in the same fail-closed spirit as the existing parity guard.

Victor-w-Madeira:
Closes #1748 — the parity guard reports `OK` while a machine-local override changes the target's environment, and nothing the run produces names the value actually in force.

## What the run now records

`run-metadata.json` gains `mirrored_target_env`: the **effective** value of every variable this lane sends to the instance under test, beside the suite SHA and the resolved Langflow version already there.

```json
"mirrored_target_env": {
  "LANGFLOW_DEACTIVATE_TRACING": "true",
  "LANGFLOW_WORKER_TIMEOUT": "120",
  "LANGFLOW_ALLOW_CUSTOM_COMPONENTS": "true",
  "LANGFLOW_SQLITE_PRAGMAS": "{\"synchronous\": \"NORMAL\", …, \"foreign_keys\": \"ON\"}"
}
```

The same values also go to the log, printed from `mirrored_target_env()`'s **own output** rather than rendered a second time — so the line cannot describe an environment other than the one that crossed the ssh boundary.

## Why a field is the right answer, and not a fix to the guard

`check-vm-env-parity.mjs` was **correct** on 2026-09-07. It answers "does the VM lane carry everything `daily-stable.yml` sets on its service" by reading two files in this repository, both of which said what they should. The instance nevertheless ran with `LANGFLOW_DEACTIVATE_TRACING=true` against the workflow's `"false"`, because the override is an `export` in the daily wrapper on the qa VM — **outside every clone**, where a repository-reading guard cannot reach it by construction.

That one variable explained **16 of the day's 20 divergences** — nine failures across the traces/observability family plus `credential-secret-exposure`, and seven of the fourteen VM skips in the same family. Establishing it took a shell on the machine.

A machine *may* hold a measured exception; that is how a VM records one, with the reason written beside it. What it may not do is hold it **invisibly** — and until the run's own artifact answers for its environment, Phase 2's criterion ("no open divergence can falsify a verdict") is not a claim anyone can check.

## The enumeration, and why the issue's guard is not here

The issue asked for the two enumerations — `mirrored_target_env()` and the metadata — to be kept in step by a guard. There is no guard, because **there are no longer two enumerations**:

```bash
MIRRORED_TARGET_VARS="LANGFLOW_DEACTIVATE_TRACING LANGFLOW_WORKER_TIMEOUT LANGFLOW_ALLOW_CUSTOM_COMPONENTS LANGFLOW_SQLITE_PRAGMAS"
```

Both the composer and the record loop that one list. Adding a variable is what carries it *and* what records it; there is no way to do one without the other, so a guard would have nothing to report. Same reasoning as #1724's ledger paths, where the state disappeared rather than gaining a guard.

Behaviour of the composer is byte-identical — same string, same order, same quoting (`refactor` commit, verified against the real file).

### What that cost the parity guard, and what replaced it

`readMirroredNames()` read the names out of `mirrored_target_env()`'s **body**, on the stated grounds that "the file names it somewhere" is not the property: a mention in a string, or one left behind in a stale branch, would satisfy it while nothing reached the target. That reasoning is unchanged; what moved is where the names live.

A list the function loops over **is** the composition — so the reader takes the list *and* pins the shape that makes it evidence. A body that stopped reading `MIRRORED_TARGET_VARS` is refused outright, because at that moment the list becomes precisely the stale mention the old form existed to reject. Driven by a **mutation of the real file**, not a fixture: a fixture is what would be adjusted to match a broken source.

## Measured, not asserted

Every property was checked by mutation — the fix reverted, the test expected to fail, and it does:

| mutation | result |
|---|---|
| `--mirrored` dropped from the metadata call | both metadata tests fail |
| the record writes a literal instead of `${!name}` | "values that were IN FORCE" fails |
| a name removed from `MIRRORED_TARGET_VARS` | 3 orchestrator tests + 4 parity tests fail, and `check-vm-env-parity.mjs` exits 1 naming the variable |
| the composer's loop stops reading the list | `readMirroredNames` throws; the guard reports `unreadable` rather than crashing the lane |

The effective-value test hands the override in **through the environment**, the way the real one arrives, and asserts against the **overridden** value: asserting the default is what would pass against the very bug this PR is about. The untouched names are asserted too, read out of `daily-stable.yml` rather than copied, so the record describes the whole set and not only the half someone overrode.

`npm run test:scripts` — 1233 passing, 0 failing. `typecheck` clean, `lint` 0 errors. Each commit was run through `test:scripts` on its own, so the branch bisects.

## What this does NOT do

`compare-lane-verdicts.mjs` still does not read it. The comparator works off `daily-history.jsonl`, and putting an environment on that row is a question about **both** lanes: the Actions side would have to write it too, or the comparator sees one side and reports a difference it cannot have measured. This is the evidence half; the reading half is its own change.

Said plainly rather than implied, on the model of `merge_ok` (#1736), which says the same about itself instead of claiming a consumer that does not exist.

### DIFF
--- a/scripts/check-vm-env-parity.test.mjs
+++ b/scripts/check-vm-env-parity.test.mjs
@@ -56,13 +56,16 @@ const workflowWith = (envLines, tail = "") =>
     tail,
   ].join("\n");
 
-/** run-e2e.sh's composer, in the shape the reader expects. */
+/** run-e2e.sh's list and its composer, in the shape the reader expects. */
 const orchestratorWith = (names, extra = "") =>
   [
     extra,
+    `MIRRORED_TARGET_VARS="${names.join(" ")}"`,
     "mirrored_target_env() {",
-    "  printf '%s' \\",
-    ...names.map((n) => `    "${n}=$(shq "$${n}") " \\`),
+    "  local name",
+    "  for name in $MIRRORED_TARGET_VARS; do",
+    '    printf \'%s=%s \' "$name" "$(shq "${!name}")"',
+    "  done",
     "}",
   ].join("\n");
 
@@ -161,6 +164,28 @@ test("naming a mirrored variable outside the composer does not count as carrying
   const found = result.findings.find((f) => f.name === "LANGFLOW_DEACTIVATE_TRACING");
   assert.equal(found?.kind, "not-carried");
   assert.match(found.message, /mirrored_target_env\(\)/);
+});
+
+test("a list the composer stopped reading is refused, not believed", () => {
+  // The half of the property that moved when #1748 made MIRRORED_TARGET_VARS the single
+  // enumeration. Reading a list instead of the function body is only equivalent while
+  // the function composes FROM the list — the moment it does not, the list is precisely
+  // the stale mention the test above refuses, and believing it would report a variable
+  // as carried while nothing reached the target.
+  //
+  // Driven by a mutation of the real file rather than a fixture, because the fixture is
+  // what would be adjusted to match a broken source.
+  const real = readFileSync(join(REPO_ROOT, "scripts/run-e2e.sh"), "utf8");
+  assert.ok(readMirroredNames(real).size >= 4, "the real file should still be readable");
+
+  const noLoop = real.replace(/for name in \$MIRRORED_TARGET_VARS; do/, "for name in LANGFLOW_DEACTIVATE_TRACING; do");
+  assert.notEqual(noLoop, real, "the mutation did not apply — the composer's loop was rewritten");
+  assert.throws(() => readMirroredNames(noLoop), /no longer composes MIRRORED_TARGET_VARS/);
+
+  // And the CLI-facing half: the guard reports it rather than crashing out of the lane.
+  const result = check(workflowWith(MINIMAL), { orchestrator: noLoop });
+  assert.equal(result.ok, false);
+  assert.equal(result.findings[0]?.kind, "unreadable");
 });
 
 test("a starter-carried variable its launch block does not set fails", () => {
--- a/scripts/lib/vm-env-parity.mjs
+++ b/scripts/lib/vm-env-parity.mjs
@@ -355,8 +355,9 @@ export function checkVmEnvParity({
     const shadow = launch.assignments.get(name);
 
     if (entry.carrier === "orchestrator") {
-      // Composed into the remote environment — read out of the function, not out of
-      // the file, so a stale mention cannot answer for a live one.
+      // Composed into the remote environment — read out of MIRRORED_TARGET_VARS, and
+      // only after readMirroredNames has checked that the composer still loops it, so
+      // a stale mention cannot answer for a live one.
       if (!mirrored.has(name)) {
         findings.push({
           kind: "not-carried",
--- a/scripts/lib/vm-env-parity.mjs
+++ b/scripts/lib/vm-env-parity.mjs
@@ -135,15 +135,32 @@ export function readStarterLaunchEnv(text) {
 /**
  * The names the orchestrator actually composes into the remote environment.
  *
- * Read out of `mirrored_target_env`'s BODY, not out of the whole file. "The file names
- * it somewhere" was the first version's test and it is not the property: a variable
- * mentioned in a string, or left behind in a stale branch, would satisfy it while
- * nothing reached the target.
+ * Read out of `MIRRORED_TARGET_VARS` — but only after checking that
+ * `mirrored_target_env()` still composes FROM that list. The two halves are the
+ * property together, and neither is it alone.
+ *
+ * The earlier version read the function's body, because "the file names it somewhere"
+ * is not the property: a variable mentioned in a string, or left behind in a stale
+ * branch, would satisfy that while nothing reached the target. That reasoning is
+ * unchanged; what changed is where the names live. #1748 made the list the single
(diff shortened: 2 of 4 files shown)


Victor-w-Madeira:
## Review response — four findings, four fixes

An independent review of the diff, conducted without the reasoning above. All four hold; all four are fixed, one commit each.

**1 · `--mirrored` was a sentinel the data could carry.** `args.indexOf("--mirrored")` matches the first occurrence *anywhere*, values included — and `langflow_prepared_reason` is parsed out of the preparer's output on the other machine, so its content is not this script's to promise. A value equal to the marker truncated the kv list and mis-keyed every field after it, writing a wrong file silently. Now the pairs are **counted**, and a count cannot collide with a value. Pinned by the case that would have failed: a run whose `prepared_reason` *is* `--mirrored`, asserting both that the value survives and that the fields after it are still themselves. (`21163078`)

**2 · The record only existed if the run reached `phase_merge`.** The strongest of the four. A wrong mirrored value is one of the likelier reasons the backend never comes up — so the run that most needed its environment on the record was exactly the run that produced none. The log line moved to the **first line of the first phase**; the metadata write stays in `phase_merge`, which needs the totals. The test drives a real abort (`TARGET_SSH` empty, the earliest `die`) and asserts **position**, not presence: a line printed after the first die would satisfy "it is in the output" and still be missing from every run that fails early. (`4f925ec6`)

**3 · The effective-value test read the ambient environment.** `metadataFrom()` without `BLANKED`, and `sourced()` forwards `process.env` — so the assertion that an *untouched* name still records the workflow default would fail on any machine exporting it. Which is the qa VM: exporting one of them is the situation this test exists to describe. Verified both ways — with `LANGFLOW_SQLITE_PRAGMAS` exported it now passes, and reverted to the ambient form it fails. (`ec11f9eb`)

**4 · A comment described the reader it replaced.** The docstring on `readMirroredNames` was updated; its call site 200 lines down still said "read out of the function, not out of the file". After this PR the names *do* come from a file-level assignment — safe only because the reader first checks the composer still loops it, which is the half the stale comment omitted. In a file whose guards are justified entirely by their comments, a stale rationale is what the next reader acts on. (`37d30a56`)

Two of these are the review reading what I wrote rather than what the code did (3 and 4), and one is a branch no current value reaches (1) — the same shape as the quoting note on `shq`, and the same reason it was written.

`npm run test:scripts` 1234 passing, 0 failing. Each commit runs the lane on its own, so the branch still bisects.

Issue #1176, opened 2026-07-31
A total shard abort opens no umbrella issue and leaves no daily-history entry


Victor-w-Madeira:
> **Rescoped 2026-09-03.** Both defects verified unchanged on today's `main`, at exact lines. One claim in the original text was overstated and is corrected below; one argument the original did not have now raises the stakes; and the migration splits the two fixes by whether they survive it. The original framing is in the edit history.

When every shard aborts before its first test, the daily goes red and **says nothing else**: no umbrella issue, no run-history entry. The failure mode that most needs announcing is the one that announces itself least.

Observed on [run 30624101377](https://github.com/oriontech-me/langflow-e2e/actions/runs/30624101377) (2026-07-31), where all four shards died in a pre-flight step (cause tracked in #1175).

## Still live, verified 2026-09-03

| | Location on `main` | State |
|---|---|---|
| **P1** — umbrella skipped | `daily-stable.yml:1402` — the `if:` carries no status-check function, so an implicit `success()` applies | unchanged |
| **P2** — day vanishes from history | `scripts/append-weekly-history.mjs:70-73` — `process.exit(0)` when `results.json` is absent | unchanged |

The gap in `reports/daily-history.jsonl` is still there: `2026-07-30` is followed directly by `2026-08-03`, so a `jq` query reads them as consecutive weekdays.

## P1 — the umbrella issue is skipped precisely when it is needed

```yaml
- name: Create issue on failure
  if: (needs.test.result == 'failure' || steps.runguard.outputs.empty == 'true') && github.event_name == 'schedule'
```

Both clauses held on 30624101377, yet the step was skipped: with no status function GitHub applies an implicit `success()` over the preceding steps, and `Merge blob reports` had already failed with `No report files found`.

A total shard abort is **guaranteed** to fail `Merge blob reports` — there are no blobs to merge — so on exactly that failure mode the step cannot run.

**Correction to the original text.** It said the `steps.runguard.outputs.empty == 'true'` clause "can never fire". That is too strong, and the distinction scopes the fix: on a run where the shards *do* produce blobs but zero tests match (a bad `--grep`), the merge succeeds, the implicit `success()` holds, and the clause works as #1012 intended. The clause is dead **only** for a total abort — which is the case that also kills the merge.

`daily-stable.yml` already documents this hazard 240 lines above, for a different step:

[...]

PR #1799, opened 2026-09-10
Say which providers a run verified, and fail only when none was usable

### LINKED ISSUE #1456: pr-validation: a provider-health skip reports SUCCESS, indistinguishable from a run that actually covered the provider
Split out of #1450, whose other item — restoring credits on the CI OpenAI account — is an
operational fix and closes on its own. This one is the design decision that would otherwise
die with that closure, and it is the half that outlives any single outage.

## The gap

When `collect-models` probes a provider `inactive`, `providerSkipGate()` / the
provider-parametrized resolvers do the right thing per test: they **skip**, citing the reason
the sweep measured. What they cannot do is change how the *lane* reports.

Measured on PR #1381's E2E lane, run
[31698035402](https://github.com/oriontech-me/langflow-e2e/actions/runs/31698035402)
(2026-08-13): **3 skipped / 5 passed / 1 flaky**, and the check reported **SUCCESS**. The three
skips were the OpenAI targets of `agent-multi-tool-selection.spec.ts` — i.e. the provider
`pr-validation.yml` pins the lane to (#1169). Nothing in the check status distinguishes that
run from one where the same three tests actually ran and passed.

This is the all-skip green that #570 and #1012 exist to prevent, arriving by a route neither
covered: the skips are *correct* (a dead key cannot produce a verdict about Langflow), the
reason *is* printed in the log, and the lane is *right* not to hard-fail — #980's trade says a
drained provider key must not kill the specs that never touch it. So every individual decision
is defensible and the aggregate still reads as coverage that did not happen.

## Why a warning is not the answer

`mode=count` in the daily's prep log was printed on every run for weeks and read by nobody
(#1252). A `::warning::` here would be the same artifact. The signal has to reach something a
human already looks at — the check status, the run summary's headline, or a distinct
conclusion — not a line in a 2000-line log.

## Prior occurrences

Fourth drained-account event on this project: #772 (openai), #1029 (google, capped),
#1169 (anthropic), #1450 (openai). Each was diagnosed from scratch; none of them changed what
the check reports.

## Done when

- [ ] A decision is recorded on how a provider-health skip surfaces on `pr-validation.yml`:
      keep SUCCESS, a distinct neutral conclusion, a summary headline, or a fail-closed gate
      when the pinned provider is the one that skipped
- [ ] Whatever is chosen is covered by a unit test over the verdict, not by a regex over the
      workflow text (#1226's lesson: a guard that pins a spelling does not pin a behaviour)
- [ ] The same question answered for `daily-stable.yml`, where the weekday rotation (#1185)
      means one dead provider silently costs that provider its whole slot — up to 7 days for
      google, its only cron day being Wednesday

Victor-w-Madeira:
## The gap

When `collect-models` probes a provider `inactive`, every spec parametrized on it `test.skip`s with the reason the sweep measured — and every decision on the way there is individually right. The skip is correct (a dead key cannot produce a verdict about Langflow); `collect-models` deliberately downgrades a billing/quota outage to a warning rather than reddening every LLM PR until someone tops up an account (Approach B, #952 — the recurring #915/#910/#911 cost); and the lane is right not to hard-fail (#980).

The aggregate still read as coverage that did not happen. Measured on PR #1381's E2E lane, run [31698035402](https://github.com/oriontech-me/langflow-e2e/actions/runs/31698035402) (2026-08-13): **3 skipped / 5 passed / 1 flaky**, check **SUCCESS**, where the three skips were the OpenAI targets of `agent-multi-tool-selection.spec.ts` — the provider `pr-validation.yml` pins itself to (#1169). Nothing told that run apart from one where those three ran and passed.

## The decision recorded

A `::warning::` was ruled out at the outset: `mode=count` was printed in the daily's prep log every run for weeks and read by nobody (#1252). The verdict goes to the **run summary**, and in one case to the **job status**:

| level | condition | surface |
|---|---|---|
| `covered` | no provider-health skip | silent |
| `degraded` | a provider went unverified, and the **account** still had a usable one | run-summary headline naming the provider and what the sweep measured; **check stays green** (#980) |
| `uncovered` | **no provider was usable at all** | headline + the step **fails**, on both lanes (#570/#1012) |

The policy is a pure function in `scripts/provider-coverage-verdict.mjs` with its own unit tests — pinning it by a regex over the workflow text is what #1226 showed does not work.

## The failing line is "was any provider usable", not "did any test run"

The first version of this PR drew it on the wrong quantity, and the independent review caught it before it ever ran. It counted tests that executed *in the spec files that produced a skip*, which is wrong in both directions:

- **False red.** A run with `rag-pipeline.spec.ts` (gated on google, wholly skipped) beside a passing `openai-provider.spec.ts` scored `uncovered` and printed *"no evidence about `google` or any other provider"* with two openai tests green in the same report, then exited 1. A file that produces no skip is not a gated file, so its passing tests were invisible — and **12 spec files are wholly gated on one provider**, so a PR editing any one of them during a drain went red for a reason its author cannot fix. #980 inverted, on the lane a human is waiting on.
- **False green, on the lane the issue actually asked about.** The daily runs all three `*-provider.spec.ts`, each of which deliberately leaves its FIRST test on the env-presence gate so it still runs on a dry day. That executed test sits inside a gated file, so `uncovered` was unreachable on the daily even with every provider down.

Asking whether the **account** had a usable provider fixes both, and draws the line where a re-run stops being able to help. It cannot come from the report — a healthy provider leaves no trace in it, which is exactly how the first version went wrong — so it is read from `providers.json`, unioned across the daily's shards (one shard reaching a provider proves the account could). The daily's shards now copy theirs onto the tokens artifact, the merge job having none of its own. With no readable file the verdict **degrades and states the gap**: the one place this guard does not fail closed, because `uncovered` fails a lane and a legitimately-absent optional input must not redden a run.

Both cases are pinned as regression tests and re-proved on real Playwright output.

## The two lanes need two mechanisms

Neither can see the other's loss. The report-based verdict answers *what did the skips cost*. The daily's **rotation** deviation is structurally invisible to it: when the rotation advances past a dry provider the specs are pinned to the substitute, **nothing skips**, and the run is honestly `covered` while the day's slot went unrun — up to 4 days for openai/anthropic and **7 for google**, whose only slot is Wednesday. So `select-daily-model-target.mjs` renders its own deviation headline, and `select-pr-model-target.mjs` renders one when the pin declines and the lane falls back to the costlier multi-provider run.

## What counts as a skip

One marker: the string `inactiveReason()` produces, read by both the hardcoded gate (`providerSkipGate`) and the parametrized resolvers (`providerSkipReasons`). That makes it a contract between two files, so `provider-health.test.ts` runs the real producer against the consumer's regex — a reword would otherwise turn the verdict into a permanent `covered`.

`OPENAI_API_KEY required to run this test` is **deliberately excluded**: it fires for keys a lane legitimately does not carry (`composio.spec.ts` skips on `COMPOSIO_API_KEY`, which no workflow sets), so counting it would make every PR touching that spec red. An unset secret is a real hole of the same family and it is #570's, not this one's.

## Measured, not assumed (Playwright 1.58)

A `test.skip(cond, reason)` — in a test body, a `beforeEach`, a `describe` or at file level — lands as `annotations: [{type: "skip", description}]` with `status: "skipped"`; a `test.skip()` with no argument carries no description and is ignored; and `merge-reports --reporter=json` preserves all of it. The PR lane had **no JSON report at all** (`--reporter=github` replaces the config's list) and now runs `--reporter=github,json` with `PLAYWRIGHT_JSON_OUTPUT_NAME` — not optional, since the json reporter prints the whole report to stdout when it has no output file and would bury the github annotations.

## Wiring

On the daily the guard **reports**, like `shardguard` and `runguard`: the script exits 1 on `uncovered` (which is what the PR lane wants), so `continue-on-error` absorbs it there and the **last** step of the merge job fails the run on its output — leaving the history append, the QA POST and the umbrella unskippable by a dry provider. That final step is fail-closed on anything that is neither `covered` nor `degraded`, absence included.

The umbrella gets a **fifth shape**, because `uncovered` is the second way a GREEN `test` job can mean no coverage — without it the run reddens at the last step and no issue names the reason, exactly what #1176 had to fix for the zero-test case. Because the level says nothing about whether tests *also* failed, that shape **carries** the `@stable` auto-removal summary instead of replacing it (an earlier draft asserted "no spec failed and no tag was touched" unconditionally, which would have told the triager the specs were not implicated on a day three tags had just been stripped) and flags the outage as a plausible cause of those failures.

## Validation

Local, on this branch:

```bash
npm run typecheck     # clean
npm run lint          # 0 errors
npm run test:units    # 1316 pass / 0 fail
npm run test:scripts  # 1370 pass / 0 fail / 1 skipped
```

End-to-end against real Playwright reports (direct `--reporter=github,json` and `--reporter=blob` + `merge-reports --reporter=json`, which agree byte-for-byte):

- the two-file false-red case → `degraded`, exit 0, headline naming `google` and stating `openai` was still usable;
- an executed test inside a gated file with a dead account → `uncovered`, exit 1;
- no readable `providers.json` → `degraded`, exit 0, gap stated.

Mutation-checked rather than assumed: `active.length === 0` → `>= 0` fails 2 tests, `known = true` fails the fallback test, and reWording `inactiveReason()` fails the two new cross-file contract tests (and 6 pre-existing ones).

**The workflow YAML itself cannot be proven locally.** Its final proof is the first `pr-validation.yml` run on this PR — watch the new `Provider coverage verdict` step, which on a healthy account should print nothing and exit 0 — and the next scheduled `daily-stable.yml`, where `Guard — which providers this run verified` should report `covered` and `Fail scheduled run that had no usable provider` should be skipped.

Closes #1456


Victor-w-Madeira:
Superseded by #1798, which landed the same mechanism for #1456 while this branch was in flight — the two sessions started ~10 minutes apart and neither had a remote branch until well after both had started.

Closing rather than rebasing: #1798's `lane-coverage-verdict.mjs` is the merged implementation and does not carry the defect this branch's review found (it counts executed tests over the whole run, not over the gated files).

One finding from this branch's review does still apply to the merged code and is filed separately as a follow-up: on the PR lane `uncovered` is decided by "did the run execute anything", which reddens a PR whose whole selection is one wholly-gated spec during a provider drain. The branch is left on the remote for the diff.

Issue #1773, opened 2026-09-09
api-flows-export-import: "no copy was created" asserts a GLOBAL flow count, so a parallel sibling flakes it (and a sibling create+delete pair would hide a real duplicate)


rafaelgiln:
`api-flows-export-import.spec.ts:121` — *"importing an export of an existing flow updates it in place"* (`@stable @api @workspace`) — asserts a **global** flow count and is therefore perturbed by any concurrent worker that creates a flow inside its window.

## Observed

PR #1772's `Run impacted E2E specs` job (run [34304431241](https://github.com/oriontech-me/langflow-e2e/actions/runs/34304431241)), `74 tests using 2 workers` → **73 passed, 1 flaky, 0 failed**. The job is **green**, because Playwright counts a retried pass as non-fatal — the failing attempt only exists in the log:

```
Error: expect(received).toBe(expected)
  Expected: 27
  Received: 29
  > 154 |         expect(await countFlows(request, authToken)).toBe(before);
  at api-flows-export-import.spec.ts:154:54
```

Not caused by that PR: its diff is comments plus one `test()` title, and the mechanism below is structural.

## Mechanism

`countFlows` (same file, ~line 47) is a **global** read with no scoping to the test's own flows:

```ts
const res = await request.get("/api/v1/flows/?header_flows=true", { headers: { Authorization: authToken } });
return (await res.json()).length;
```

The suite runs `fullyParallel` against a **single shared `auto_login` superuser**, so `before` and the post-upload count are two reads of state every other worker also writes. The test's own window spans `POST /api/v1/flows/download/` + `POST /api/v1/flows/upload/` — long enough for a sibling to create flows. In the observed run the lane held 15 other `api/` specs, several of which create flows per test. 29 - 27 = two sibling creations, not a duplicate from the endpoint under test.

This is the **read** half of the hazard this suite documents exhaustively on the **cleanup** side (#553/#515: never a global wipe, because parallel workers share the flow space). A global *read* of that same space is perturbable for exactly the same reason, and the file's own sibling specs already avoid it — `api-folders-crud.md` records "Deletes are targeted by id (never a global wipe)".

## Why the assertion is also weaker than intended

The comment above it states the intent precisely:

> Both halves matter: the preserved id alone would also be satisfied by an endpoint that created a duplicate and echoed the export's id back.

[...]

Issue #616, opened 2026-07-09
settings-message-history broken — ag-grid column col-id='id' gone from Settings → Messages


rafaelgiln:
## Problem

`ui-ux/settings-message-history.spec.ts:48` ("Settings > Messages displays sent messages in correct order with working filters", `@release`, not `@stable`) fails deterministically on clean main (verified 2026-07-09, nightly 1.11.0.dev36):

```
Error: expect(locator).toBeVisible() failed
  Locator: locator('.ag-header-cell[col-id="id"]')
  Timeout: 10000ms — element(s) not found
```

The spec asserts a fixed set of ag-grid column ids on the Settings → Messages table; the `id` column no longer renders on the 1.11 nightly (column set changed upstream). Needs a live scout of the current column ids — and a decision whether asserting exact column ids is the right contract vs. asserting the message rows/ordering only.

## Origin

Found during the #606 consumer-breakage check (batch of 6 broken @release specs).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1818, opened 2026-09-11
The VM twin's shard matrix has no listing-completeness detector — the lane whose product IS the comparison cannot see a file leave it


Victor-w-Madeira:
Sibling of #1812, which the issue itself calls for ("the same check belongs on the VM twin") and which had no issue filed.

#1812 gives the Actions daily a detector for a spec file that leaves the shard matrix without a trace — not skipped, not red, **absent**, with `--pass-with-no-tests` keeping every shard green and the merged report holding no row to be missing from (#1764). The VM twin does not have it.

## What is already true

`scripts/run-e2e.sh:1196` builds its matrix through the same partitioner:

```sh
node scripts/partition-shards.mjs matrix \
  "$RUN_DIR/stable-list.json" "$durations" "$SHARDS" > "$RUN_DIR/matrix.json"
```

No `--declared`, so it prints `listing completeness: UNVERIFIED — no declared-spec set was passed` and **no `::warning::`** — deliberate, since a caller that never asked must not be warned on every run (the `mode=count` lesson, #1252). The comparison itself is pure and lane-agnostic, so this is wiring rather than design.

## Why it matters more here, not less

The VM twin's whole product is the **cross-lane comparison** — which is how #1764 was found in the first place. A file absent from one lane's matrix and present in the other's is exactly the divergence that comparison exists to surface, and today only one of the two lanes can report it. Worse, the two lanes list in different environments: #1813 already records that the collection gate is a property of the listing process's environment, and `scripts/run-e2e.sh` deliberately **dies** on an unresolved gate where the Actions lane degrades (the asymmetry is the point — there the run's product is the test run, here it is the comparison).

## What the wiring needs

1. Derive the declaration next to the listing: `npx ts-node scripts/declared-stable-specs.ts > "$RUN_DIR/declared-specs.json"`. It exits 2 when the derivation breaks (a floor: an empty declaration certifies every possible listing as complete).
2. Pass `--declared "$RUN_DIR/declared-specs.json"` to the `matrix` call.
3. Decide where the verdict is read. On Actions it is the merge job's last step, read fail-closed, chosen because it runs after the history row, the QA POST and the `@stable` auto-removal — the run's whole product survives and only the green is withheld. The VM lane's phase layout differs and it has a Slack surface the Actions lane does not, so the placement is a decision for this issue rather than a transcription.

Two things worth settling while deciding (3):

[...]

PR #1828, opened 2026-09-12
fix(settings-message-history): scope the grid to the test's own conversation (#1778)

### LINKED ISSUE #1778: settings-message-history — the grid reports 22 rows and the DOM carries 18, all from sibling specs; the test's own 4 newest messages are not among them (VM lane, 3/3 attempts, no outage)
Found by the **VM daily lane** (step 14 of the VM migration), run `20260909T080012Z` — suite `5e199bbd`, Langflow `1.13.0.dev7` (source build `711f3e08` of tag `v1.13.0.dev7`). The Actions daily of the same day ([34352667406](https://github.com/oriontech-me/langflow-e2e/actions/runs/34352667406)) ran the **same Langflow version** and this test **passed** there.

Filed against the hard-failure criterion (3/3 attempts). **Note on the lifecycle:** the VM lane does not run `remove-stable-from-failures.ts`, so `@stable` is **still in place** on this test and there is nothing to restore — unlike a `daily-failure` issue spun out of the Actions lane. Whether the tag should come off is part of what this issue decides.

**Upstream:** _not filed_

## Symptom

`tests/tests-automations/regression/ui-ux/settings-message-history.spec.ts:73` — "Settings > Messages displays sent messages in correct order with working filters" (`@stable @release @workspace @api @settings`).

The test creates a Simple Agent flow, sends two prompts in the Playground, opens Settings > Messages, and asserts its own two prompts are present in the grid's `text` column:

```ts
const joinedTexts = allTexts.join(" ");
expect(joinedTexts).toContain(FIRST_MESSAGE);   // "Hello, how are you?"
```

It failed on **all 3 attempts** — `09:17:44` (24.8 s), `09:18:11` (31.8 s), `09:18:46` (26.8 s) UTC — with `"Hello, how are you?"` absent from the collected text and 18 message texts belonging to **other specs** present instead.

The steps before it passed: the grid rendered ≥ 4 rows with parseable timestamps, and the chronological (oldest-first) order assertion held.

## What was measured

From the failure snapshot (`error-context` attachment of attempt 0), the grid's own footer:

```
1 to 22 of 22. Page 1 of 1
```

**22 rows, single page.** The DOM carries **18 data rows**, every one of them attributable to another spec, spanning `08:34:31.345620` → `08:58:07.074004` UTC:

| Rows | Text / session | Origin |
|---:|---|---|
| 6 | `CTXISO-1788942860445-tkrvqf-A-{1,2,3}`, session `CTXISO-…-session` | `agent-context-id-isolation.spec.ts` (sentinel minted at line 367) |
| 12 | `filters-ok-probe`, `filters-error-probe`, `A model selection is required`, sessions `filters-ok-session-…` | `traces-list-filters.spec.ts` — three probe sets, one per attempt (that spec is a 3/3 hard failure on this lane) |

`22 − 18 = 4` rows exist that the DOM does not carry. The test's own footprint is exactly **4** (two prompts plus two agent responses), and on an ascending sort they are the **newest**, i.e. last.

**Not wedge collateral — ruled out by time.** The run was wedged (11 outages, 812 s, 22.6 % of the observed span), so it has to be excluded explicitly. The last outage window closed at `08:52:07`; all three attempts here ran from `09:17:44`, about **25 minutes later**, with **0 % overlap**. The failures are also assertion-shaped at 25–32 s, not the transport timeouts a wedge produces.

**Two shapes fit the measurement, and one observation separates them. Both must be investigated; neither is asserted here.**

- **Product shape.** The feature reports 22 rows and materializes the first 18; the newest 4 are not in the DOM. If they are also **not reachable** by scrolling the grid body, then a user cannot see their most recent messages in Settings > Messages once history accumulates — a user-facing defect in the message-history feature, and this test caught it correctly. The ascending default makes this the reachability question for *new* messages specifically, which is the common case.
- **Test shape.** AG Grid virtualizes rows as designed and the 4 rows are one scroll away; the spec collects only what is materialized (`page.locator('.ag-cell[col-id="text"]')`), a read whose premise is that every message fits one viewport. Then what broke is the assertion's premise, not the feature.

**Discriminator:** on an instance carrying more than one viewport of messages, open Settings > Messages, scroll the grid body to the end, and re-collect `col-id="text"`. Newest rows appear ⇒ test shape. Newest rows unreachable ⇒ product shape.

**Context, not evidence for either shape:** this same file already hit the virtualization axis once. Under #616, upstream added four columns and pushed `id` out of the horizontally-virtualized header, and the fix was to sweep the horizontal scroll collecting every `col-id` before asserting the promised set. The row axis never received the equivalent treatment. That says the axis is known to virtualize; it does **not** say which shape above is the cause.

## Why the Actions lane did not see it

Descriptive, and relevant only to reproduction — not to classification.

The three specs involved sit on **three different shards** in the Actions daily, i.e. three separate Langflow containers, so their messages can never share a grid there:

| Spec | Actions shard |
|---|---:|
| `settings-message-history.spec.ts` | 3 |
| `agent-context-id-isolation.spec.ts` | 4 |
| `traces-list-filters.spec.ts` | 1 |

The VM lane runs all 621 tests against **one** instance, so a spec that reads a global collection sees every sibling's writes for the life of the run.

Two conditions of this particular run raised the foreign row count, and both are known items on the VM lane: the target runs with `LANGFLOW_DEACTIVATE_TRACING=true`, which makes `traces-list-filters` fail 3/3 and emit one probe set per attempt; and the 22.6 % wedge made `agent-context-id-isolation` retry, emitting its sentinels again. This test **passed on the VM lane on 2026-09-07 and 2026-09-08**, when the lane was not on the provider slot that wedges and the foreign row count was lower.

## Investigation directive

1. **Suspect the product first, and rule it out before anything else.** Reproduce with more than one viewport of messages in the instance and answer the discriminator above: are the newest rows reachable by scrolling the grid body? A "test wrong" explanation may be dismissed only after confirming this failure does not mask a real reachability regression in Settings > Messages.
2. If the rows are reachable: establish what the correct read is. The spec already captures the `flow_id` it created (`createdFlowIds`) and already drives the grid's own filter machinery in the `sender` step, so scoping the read is available with mechanisms the file already exercises. Compare that against sweeping the vertical scroll the way #616 swept the horizontal one, and pick on evidence rather than on symmetry.
3. If the rows are **not** reachable: classify as a product regression, file upstream, and keep this issue open per `CONTRIBUTING.md` — a product regression is done when the product is fixed, not when the test is scoped.
4. Check whether any other `@stable` spec asserts on a global collection under the same premise (message history, traces, flows lists). The exposure is a property of the read, not of this file, and the VM lane is the only lane that can surface it.
5. Classify per `references/triage-verdicts.md` (product regression vs. test read strategy vs. environment).

## A constraint on validating the fix

Whatever the fix is, **the Actions lane cannot verify it.** That lane never reproduces this failure — the three specs are in different containers by construction — so a green Actions run after the change is indistinguishable from a green Actions run before it. The evidence has to come from a run where the pollution is present: the VM topology, one instance, the three specs in that order, `--retries=0`.

This also means the fix should **not** be validated against the next scheduled VM daily. The mechanism needs the foreign rows present at the moment this test runs, and that depended on the wedge, which tracks the provider slot. A provoked run is the only place the fix gets a real verdict.

## Deliverables (Done when)

- [ ] The discriminator is answered and recorded here: are the newest rows reachable by scrolling the grid body, on an instance with more than one viewport of messages?
- [ ] Root cause confirmed per `references/triage-verdicts.md` (product regression vs. test read strategy vs. environment), with the regression hypothesis explicitly ruled out or confirmed — not assumed.
- [ ] If a **product regression**: recorded as such here, filed upstream, and this issue stays **open** until the fix lands in `langflowai/langflow-nightly:latest` (or the `release-1.x.x` branch) and is re-validated there.
- [ ] If a **test read-strategy defect**: the corrected read passes on repeated `--retries=0` runs **with the pollution present** (VM topology, the three specs in one instance), and the run that proves it is linked here.
- [ ] Answered: does any other `@stable` spec assert on a global collection under the same premise?
- [ ] `@stable` state on this test decided explicitly — it was never auto-removed, so leaving it in place is a decision, not the default.

---

Provenance: this is the first false-red candidate the VM lane has produced in the parallel-daily comparison (2026-09-07 onward). It is recorded in the migration's comparison ledger for 2026-09-09.

Victor-w-Madeira:
**Fixes #1778.**

## Problem

`settings-message-history.spec.ts` read the **global** messages grid and asserted its own prompts were in it:

```ts
const joinedTexts = allTexts.join(" ");
expect(joinedTexts).toContain(FIRST_MESSAGE);
```

Settings → Messages is a global audit surface — the suite runs `fullyParallel` against one shared superuser on one instance, so every sibling spec's messages land in that table. And AG Grid virtualizes **rows** the way #616 found it virtualizing columns: only what fits the viewport is in the DOM.

## The discriminator #1778 asked for, answered: **test shape**

The issue left two shapes open and named the measurement that separates them — *on an instance carrying more than one viewport of messages, scroll the grid body to the end and re-collect `col-id="text"`.* Measured on `1.13.0.dev8` (nightly image, 50 stored messages from 25 sessions, viewport 1280×720):

| | |
|---|---|
| grid footer | `1 to 50 of 50`, `Page 1 of 1` |
| `.ag-row` in the DOM | **18** |
| body viewport | `clientHeight=324`, `scrollHeight=2100` |
| own (newest) row materialized | **no** |
| own row after sweeping the vertical scroll | **yes** |

The newest rows are **reachable** — one scroll away. So this is not the product shape: no user is losing sight of their most recent messages, and the assertion's premise ("everything materialized is everything there is") is what broke. The reproduction is faithful to the issue's own numbers: 22 reported / 18 in DOM / own rows absent, here 50 / 18 / absent.

## Fix

The grid is **scoped to this test's own conversation before a single row is read**:

- the session id comes from `GET /api/v1/monitor/messages?flow_id=<own flow>` — never off the screen, because the row carrying `FIRST_MESSAGE` is exactly the row virtualization may not have materialized, so reading the scope key from the grid would depend on the defect being absent;
- the `session_id` column's "Equals" filter is applied (measured: that header is in the DOM at `scrollLeft = 0` and carries a filter button);
- **the scope is asserted, not assumed** — every rendered `session_id` must equal it. A filter that silently failed to apply hands the global grid back, and that now *fails* instead of greening.

Every assertion below it — order, direction, senders, content, the `sender` filter — then measures this test's own four rows. The `sender` filter interaction moved into `applyEqualsFilter`, now used twice.

### Rejected alternative

**Sweeping the vertical scroll**, the row-axis twin of the #616 column sweep, which the issue explicitly notes the row axis never received. It collects the rows, but:

- it leaves every assertion measuring *other specs'* messages — the order, sender and content checks stay global;
- its cost grows with the instance's whole message history. The lane that found this serves **654 tests from one instance**, so the sweep is unbounded work in exactly the environment the bug lives in.

Scoping is `O(this test's own rows)`. Same move PR #1779 made for #1773, where a global flow count was scoped to a project the test owns.

## Validation (Langflow `1.13.0.dev8`, nightly image, `--retries=0`)

Against a **deliberately polluted** instance — 50 messages from 25 sibling sessions seeded through `POST /api/v1/run/`, so the defect's precondition is present and deterministic rather than waited for:

| Run | Result |
|---|---|
| Fixed spec, `--repeat-each=3 --trace=on` | **3 passed** (1.0 min) |
| Original spec, same instance | **fails at line 246**, `Expected substring: "Hello, how are you?"` / received 18 `sibling-probe-*` texts — #1778's signature, same line, same shape |
| Force-fail: scope filter pointed at a non-existent session | **fails at the scope assertion**, not green — no false positive |
| Backend errors (`🚨 Backend Error:`) | **0** |
| `tsc --noEmit`, `eslint`, `check:checklist-coverage` | clean (9 pre-existing warnings, 0 errors) |

## The `@stable` decision #1778 lists, answered explicitly

#1778 records "whether the tag should come off" as one of the things it decides. **The tag stays.** The failure was test-shaped — the discriminator above shows the rows are reachable, so the feature is working — and the test's own logic (the column contract, the chronological order, the direction check, the sender filter) was never in question. `@stable` is still on Test 1 here; nothing in this PR removes or restores it.

## The VM lane's own proof — run on 2026-09-12, and it is decisive

The issue requires a run **with the pollution present**, and says the fix cannot be verified on the Actions lane, which never reproduces it. That run is now on record: `langflowqa` → tunnel → `langflowbin`, the daily's own target (published distribution `1.13.0.dev9` in a venv, brought up by this repo's `start-langflow-source.sh` with the mirrored env and **tracing on**), `collect-models` first as the daily does, then 25 sibling sessions seeded through `POST /api/v1/run/` so the precondition is deterministic rather than waited for.

Both halves ran against the **same instance**, `--workers=1 --retries=0`:

| Tree | Result |
|---|---|
| `main` @ `0c914e9b` — spec without the fix | **failed**, line 246: `Expected substring: "Hello, how are you?"` / received 18 `sibling-probe-*` texts — #1778's signature, reproduced on the lane that owns it |
| this PR @ `15ad4411` | **passed** (27.1 s), 100 messages in the backend at that moment |

Target torn down and verified afterwards: `Langflow on port 7862 stopped`, 0 processes, 0 listeners on 786x. The worktree and temporary branch used for the run were removed; the `qa` clone is back on `main` @ `0c914e9b`, tree clean. Evidence kept on the machine in `/root/probe-1778-20260912/` (script, seeder, log).

One honest note from the run: the preflight warned `ANTHROPIC_API_KEY` is set in the environment but not configured as a Langflow global variable — that provider's collector stalled during `collect-models` (the known dead-key symptom on that account). It does not touch this spec, which gates on OpenAI, recorded active as `gpt-4o-mini`.

### DIFF
--- a/docs/ui-ux/settings-message-history.md
+++ b/docs/ui-ux/settings-message-history.md
@@ -1,6 +1,6 @@
 # Settings → Messages — history shows sent messages in order with working filters
 
-**Last validated:** Langflow 1.12.x (nightly `1.12.0.dev6`)
+**Last validated:** Langflow 1.13.x (nightly `1.13.0.dev8`)
 
 ---
 
@@ -28,8 +28,40 @@ conversation, the test validates that:
    note below.
 3. **Content integrity** — both sent prompts appear verbatim in the `text`
    column; `sender` distinguishes `User` from the machine/agent side.
-4. **Column filters work** — filtering `sender` by "Equals User" leaves only
-   User rows; clearing the filter restores the full row set.
+4. **Column filters work** — the `session_id` filter narrows the grid to one
+   conversation and the `sender` "Equals User" filter then leaves only User
+   rows; clearing the value restores that conversation's full row set.
+
+## Scoping note *(what broke in #1778)*
+
+Settings → Messages is a **global** audit surface. The suite runs
+`fullyParallel` against one shared superuser on one instance, so every sibling
+spec's messages land in this same table — and AG Grid virtualizes **rows** the
+way #616 found it virtualizing columns: only what fits the viewport is in the
+DOM. Collecting `.ag-cell[col-id="text"]` off an unscoped grid therefore reads
+*some other spec's* messages and asserts this test's prompts are among them.
+
+Measured on `1.13.0.dev8` with 50 stored messages: the footer reports
+"1 to 50 of 50", the DOM carries **18** rows, and this test's own — the newest,
+and therefore last under the ascending default — are not among them. That is
+#1778: a hard 3/3 failure on the VM lane while the feature was working, because
+the rows were one scroll away.
+
+The test therefore **scopes the grid to its own conversation before reading a
+single row**: it resolves its session id from
+`GET /api/v1/monitor/messages?flow_id=<own flow>` (never off the screen — the
+row carrying the prompt is exactly the row virtualization may not have
+materialized) and applies the `session_id` column's "Equals" filter. The scope
+is then **asserted**, not assumed: every rendered `session_id` must equal it,
+so a filter that silently failed to apply hands back the global grid *and
+fails* instead of greening.
+
+**Rejected alternative — sweeping the vertical scroll**, the row-axis twin of
+the #616 column sweep. It collects the rows, but it leaves every assertion
+measuring other specs' messages, and its cost grows with the instance's entire
+message history — the lane that found this serves 654 tests from one instance.
+Scoping is O(this test's own rows). The same move PR #1779 made for #1773,
+where a global flow count was scoped to a project the test owns.
 
 ## Virtualization note *(what broke in #616)*
 
@@ -77,18 +109,22 @@ burst on nightly `1.12.0.dev6`.
    finish (Stop button appears → hidden) and assert a non-empty response.
 3. Send `What is 2+2?`; same wait + non-empty assertion.
 4. Close the Playground; navigate **Settings → Messages**.
-5. **Column contract:** sweep the grid horizontally collecting every
+5. **Scope:** resolve this conversation's `session_id` from
+   `GET /api/v1/monitor/messages?flow_id=<own flow>` (the flow id tracked for
+   cleanup), apply the `session_id` column's "Equals" filter, and assert every
+   rendered `session_id` equals it. Everything below reads the scoped grid.
+6. **Column contract:** sweep the grid horizontally collecting every
    `.ag-header-cell` `col-id`; assert the collected set contains all 11
    promised columns (superset-tolerant).
-6. **Order:** read all `timestamp` cells (≥ 4 rows expected: 2 user + 2
+7. **Order:** read all `timestamp` cells (≥ 4 rows expected: 2 user + 2
    agent); assert they parse (≥ 4 parseable, so the check is never vacuous)
    and are **monotonically ascending**. Monotonicity alone also holds for a
    reversed grid, so it is paired with a direction-sensitive check: the row
    index of `Hello, how are you?` must be **less than** the row index of
    `What is 2+2?` in the `text` column.
-7. **Content:** `sender` column contains `User` and a machine/agent value;
+8. **Content:** `sender` column contains `User` and a machine/agent value;
    `text` column contains both prompts verbatim.
-8. **Filter:** click the `sender` header's dedicated filter button
+9. **Filter:** click the `sender` header's dedicated filter button
    (`.ag-header-cell-filter-button` — the old `.ag-icon-menu` + "Filter" tab
    flow no longer exists on 1.11); pick "Equals", type `User`; assert every
    remaining row's sender is `User`; clear the value; assert the row count
@@ -98,6 +134,8 @@ burst on nightly `1.12.0.dev6`.
 
 ## Validation criterion *(required)*
 
+- Every row the grid renders after scoping belongs to this test's own session —
+  asserted, so an unapplied filter fails instead of silently widening the read.
 - The collected column-id set ⊇ the 11 promised columns.
 - Timestamps render in ascending (oldest-first) order, with ≥ 4 parseable
   timestamp cells after two exchanges; the first prompt sent renders above the
@@ -121,7 +159,8 @@ force-fail contract: no-op the cleanup and the flow count grows.
 
 - Message editing / deleting through the table (`edit` column actions).
 - Session-scoped views (`session_metadata`, session rename) — covered by the
-  playground session specs.
+  playground session specs. The `session_id` filter is used here as the
+  **instrument** that scopes the read; only its narrowing effect is asserted.
 - Exact set equality of columns (new upstream columns are tolerated by
   design — only removals of promised columns fail).
 
@@ -133,6 +172,8 @@ force-fail contract: no-op the cleanup and the flow count grows.
   history is asserted).
 - `tests/helpers/other/initialGPTsetup.ts` + `resolveGptModel` +
   `data/models.json` (collect-models).
+- `GET /api/v1/monitor/messages?flow_id=` — read once, to resolve this
+  conversation's `session_id` for the grid scope (#1778).
 - AG Grid rendering of the messages table (`.ag-header-cell[col-id]`,
   `.ag-cell[col-id]`, `.ag-center-cols-viewport` for the horizontal sweep,
   column-menu filter UI) — Settings → Messages page
--- a/docs/ui-ux/settings-message-history.md
+++ b/docs/ui-ux/settings-message-history.md
@@ -59,7 +59,9 @@ fails* instead of greening.
 **Rejected alternative — sweeping the vertical scroll**, the row-axis twin of
 the #616 column sweep. It collects the rows, but it leaves every assertion
 measuring other specs' messages, and its cost grows with the instance's entire
-message history — the lane that found this serves 654 tests from one instance.
+message history — the lane that found this serves the whole suite from one
+instance (**621** tests on 2026-09-09, the day #1778 was filed; 654 on
+2026-09-11, after the target change put sixteen skipped tests back).
 Scoping is O(this test's own rows). The same move PR #1779 made for #1773,
 where a global flow count was scoped to a project the test owns.
 
@@ -109,13 +111,18 @@ burst on nightly `1.12.0.dev6`.
    finish (Stop button appears → hidden) and assert a non-empty response.
 3. Send `What is 2+2?`; same wait + non-empty assertion.
 4. Close the Playground; navigate **Settings → Messages**.
-5. **Scope:** resolve this conversation's `session_id` from
-   `GET /api/v1/monitor/messages?flow_id=<own flow>` (the flow id tracked for
-   cleanup), apply the `session_id` column's "Equals" filter, and assert every
-   rendered `session_id` equals it. Everything below reads the scoped grid.
-6. **Column contract:** sweep the grid horizontally collecting every
+5. **Column contract:** sweep the grid horizontally collecting every
    `.ag-header-cell` `col-id`; assert the collected set contains all 11
-   promised columns (superset-tolerant).
+   promised columns (superset-tolerant). This runs **before** the scope on
+   purpose — it reads headers, not rows, and it is the only check that can name
+   `session_id` as missing instead of timing out on a locator that never had a
+   chance.
+6. **Scope:** resolve this conversation's `session_id` from
+   `GET /api/v1/monitor/messages?flow_id=<own flow>` (the flow id tracked for
+   cleanup), apply the `session_id` column's "Equals" filter, assert the set of
+   rendered `session_id` values equals `[own]` — polled as a value, so an empty
+   grid and a grid full of foreign sessions read differently — and only then
+   dismiss the popup. Everything below reads the scoped grid.
 7. **Order:** read all `timestamp` cells (≥ 4 rows expected: 2 user + 2
    agent); assert they parse (≥ 4 parseable, so the check is never vacuous)
    and are **monotonically ascending**. Monotonicity alone also holds for a
--- a/tests/tests-automations/regression/ui-ux/settings-message-history.spec.ts
+++ b/tests/tests-automations/regression/ui-ux/settings-message-history.spec.ts
@@ -26,6 +26,7 @@
 
 import * as dotenv from "dotenv";
 import path from "path";
+import type { APIRequestContext, Locator, Page } from "@playwright/test";
 import { expect, test } from "../../../fixtures/fixtures";
 import { awaitBootstrapTest } from "../../../helpers/other/await-bootstrap-test";
 import { initialGPTsetup } from "../../../helpers/other/initialGPTsetup";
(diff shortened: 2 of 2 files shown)


Victor-w-Madeira (commented):
## Review — verdict: the fix is the right one, findings below are hardening and process, none of them block

The discriminator #1778 asked for was actually answered (rows reachable one scroll away ⇒ test shape, not product shape), the reproduction is deterministic instead of waited for, and — the part that matters most — **the scope is asserted rather than assumed**, so the failure mode of the fix itself is a red test and not a green one. The rejected alternative is argued with the right unit (cost grows with the instance's whole history vs. `O(own rows)`).

Two things worth recording as genuinely better, not just equivalent: after scoping, the direction check (`firstMessageRow < secondMessageRow`) is finally a statement about *this* conversation — before, a sibling spec's rows could satisfy it. And **not** declaring the new `GET /api/v1/monitor/messages` read through `apiCoverage` is correct, per that fixture's own definition of covered ("a spec that merely passes through an endpoint has asserted no contract") — the read is an instrument here, and `api/flows/api-monitor-messages.spec.ts` owns the contract.

---

### 1. The whole scope mechanism sits on the *horizontal* virtualization axis — the one this file already lost once (#616)

Both halves depend on `session_id` being inside the initial viewport at `scrollLeft = 0`: the `page.hover` + filter-button click in `applyEqualsFilter`, **and** the `.ag-cell[col-id="session_id"]` read the scope assertion polls. If the column is ever pushed out, the cells are not in the DOM at all and no auto-scroll recovers them.

Checked upstream rather than assumed, and it is reasonably stable today: `MESSAGES_TABLE_ORDER` (`src/frontend/src/constants/constants.ts`) pins `session_id` 6th in a list of 7 and everything unlisted sorts to the end, and `session-view.tsx` gives only `text` a flex (`flex: 3, minWidth: 320`) — so at 1280 wide it lands just inside, which is what the PR measured.

The cost is not a false green, it is diagnosis: a `page.hover` timeout on a locator that never had a chance, or `expected true / received false`, neither naming virtualization — in the file whose entire history *is* virtualization. Cheap hardening, pick either:

- run the existing horizontal sweep (step 10) **before** the scoping — it already collects every `col-id` and would fail with `column "session_id" missing from the messages grid` before anything depends on it; or
- scroll the header into view inside `applyEqualsFilter` when the cell is not attached.

### 2. The load-bearing assertion is the one with no failure message

```ts
await expect.poll(async () => {
  const sessions = await sessionCells.allTextContents();
  return sessions.length > 0 && sessions.every((s) => s.trim() === ownSessionId);
}, { timeout: 10000 }).toBe(true);
```

This is the assertion the PR body (rightly) calls the guarantee, and on failure it prints `expected true, received false` — it names neither what was rendered nor which of its two clauses broke. Polling the value instead keeps the same guarantee and prints the evidence:

```ts
await expect.poll(
  async () => [...new Set((await sessionCells.allTextContents()).map((s) => s.trim()))],
  { timeout: 10000 },
).toEqual([ownSessionId]);
```

`[]` then separates "the filter narrowed to zero / the column is not materialized" from "the filter did not apply and here are the foreign sessions" — which is exactly the distinction a triager needs at 07:00.

### 3. `resolveOwnSessionId` cannot say "no flow id was ever captured"

If the `page.on("response")` hook never fires (the flow arrives by another route, a batch payload, a non-201), `createdFlowIds` is empty, the loop body never runs and the throw reads `… for the flow(s) this test created []`. Two very different causes — *the conversation was not persisted* (what the message asserts and directs the reader to) and *this test never learned its own flow id* — render as the same sentence. One branch above the loop fixes it.

### 4. `Escape` is pressed inside the text filter's debounce window

`applyEqualsFilter` ends on `fill(value)` and the caller presses `Escape` on the next line; AG Grid's text filter applies on a debounce. It measured green 3/3, and if it ever loses the apply the cost is a red test — but one that reads as "the filter did not apply", pointing at the wrong thing. Asserting the scope **before** dismissing the popup removes the ambiguity for free: the only thing that needs the popup gone is the `sender` hover ~130 lines later, and the column sweep in between is `page.evaluate`-driven (no pointer, no hit-testing), so an open popup does not block it.

### 5. `.first()` over `.ag-filter` is now ambiguous by construction

Both `expect(page.locator(".ag-filter").first())` and `page.locator('.ag-filter input[type="text"]').first()` run a second time, *after* another column's filter popup has been opened and closed in the same test. Today that works because AG Grid detaches the closed popup. The day it keeps it attached-but-hidden, `applyEqualsFilter(page, "sender", "User")` writes `User` into the **session_id** filter and the scope becomes `session_id = "User"`. Loud (the sender poll fails on zero rows), never a false green — but scoping the locator to the open popup drops the dependency on AG Grid's popup lifecycle entirely.

### 6. Prose: one figure disagrees with the issue, and one measurement is duplicated verbatim

- "the lane that found this serves **654 tests** from one instance" appears in both the spec comment and the doc; #1778 says **621**. Both may be true on their own day — that is the point of `5f8cd39` / `eab93ef` from last week: a number in prose is a claim, so either date it or take the issue's figure.
- The `1.13.0.dev8` / 50 rows / 18 in DOM measurement is repeated in full in the spec comment and in the doc's scoping note. The doc is where `Last validated` stamps it; the code comment stays current longer if it carries the *mechanism* and points at the doc for the numbers.

### 7. Process: #1778 has no approved-exception label

It carries `area:ui-ux` + `vm-migration` only. The PR checklist wants a current-wave item or an approved exception (`follow-up` / `daily-failure` / `community`) — and the two precedents this PR itself leans on both have it: #1773 and #1818 are `follow-up` + (for #1818) `vm-migration`. Adding `follow-up` to #1778 makes the trace hold.

### 8. `Fixes #1778` closes the issue while its own proof is still pending

The PR's "Still to do" is the VM lane's directed 1-shard run — the evidence the issue's investigation directive asks for — and it says the window closes when that lane goes to four shards. On merge, `Fixes` closes #1778 and that pending item loses its owner. Either run it before merging, or open the follow-up now and name it in the PR body.

While there: #1778 explicitly lists "whether the tag should come off" as one of its decisions. The PR answers it (the tag stays, because the failure was test-shaped and is fixed) but only implicitly — worth one sentence, so the issue closes with its own question answered on the record.

Victor-w-Madeira:
## Review answered — seven findings applied in `860a66f`, the eighth was overtaken by the clock

All of them stand; none was a false positive. What each one changed:

**1 · The scope sat on the horizontal axis.** The best finding, and the remedy you offered first is free: the scope now runs **after** the column contract. The sweep reads headers, not rows, so the "nothing is read unscoped" property is intact — and a `session_id` pushed off-viewport now fails as `column "session_id" missing from the messages grid` instead of a hover timeout on a locator that never had a chance. The reasoning is written into the comment so the ordering is not re-sorted by accident later.

**2 · The load-bearing assertion had no message.** Rewritten to poll the value. Measured by forcing it — the filter pointed at a session that does not exist now prints:

```
- Array [
-   "1a3ca600-2592-428f-acec-5e7d6e68f718",
- ]
+ Array []
```

which is exactly the distinction you named: `[]` is "narrowed to nothing / column not materialized", a list of foreign sessions is "the filter never applied".

**3 · `resolveOwnSessionId` could not say "no flow id captured".** A branch above the loop, with its own name (`OWN_FLOW_ID_NOT_CAPTURED`) and a pointer at the template load rather than at the message history.

**4 · `Escape` inside the debounce.** The scope is asserted before the popup is dismissed. Nothing else needed the popup gone until the `sender` hover, and the sweep in between is `page.evaluate`-driven.

**5 · `.first()` over `.ag-filter`.** Applied too, since the verification run was happening anyway: `applyEqualsFilter` scopes to the open popup (`.ag-filter:visible`) and drops the dependency on AG Grid's popup lifecycle.

**6 · The figure and the duplicated measurement.** The suite size is dated — 621 on the day #1778 was filed, 654 after the target change put sixteen skipped tests back — and the `dev8` measurement now lives only in the spec doc, where `Last validated` stamps it. The code comment carries the mechanism.

**7 · The label.** `follow-up` added to #1778, so the trace the PR checklist wants holds.

**8 · `Fixes` closing an issue with a pending proof.** Resolved by the clock rather than by argument: the review landed at 12:06 UTC and the VM lane's directed 1-shard run had started at 12:01 and finished at 12:21. The body no longer has a "still to do" — it has the run: spec at `main` fails on line 246 with #1778's signature, this PR passes, same instance, pollution seeded deterministically. **Your second point there did need answering**, and now has its own section: the tag stays, and the PR says so rather than implying it.

## Re-verification after the changes

Same conditions as before, fresh instance, 50 seeded messages from 25 sibling sessions: **3 passed** with `--repeat-each=3 --retries=0 --trace=on`, zero `🚨 Backend Error`, `tsc` and `eslint` clean. The forced failure above is the fourth run.

Issue #1823, opened 2026-09-11
[Daily #1821] the Anthropic credential write on Settings → Model Providers never answers — 2 specs, 3/3 attempts each, first occurrence of the signature


Victor-w-Madeira:
Spun out of daily-failure triage #1821 (run [34599745145](https://github.com/oriontech-me/langflow-e2e/actions/runs/34599745145), 2026-09-11). Filed against the hard-failure criterion: both tests failed 3/3 attempts. The three skips listed below are downstream of the same cause and are recorded here rather than filed separately.

**Upstream:** _not filed_

## Symptom

Two `@stable` tests failed **3 of 3 attempts** on the same action: writing the Anthropic API key on **Settings → Model Providers**. Neither received an HTTP response — not an error status, no response at all.

- `tests/collect-models.spec.ts:39` — the collector stalls with *"no credential write answered within 240s of clicking Save"* and then asserts that `anthropic` was never configured. Three attempts of **342 s / 363 s / 342 s** (12:47:30, 12:53:14, 12:59:19 UTC).
- `anthropic-provider.spec.ts:184` — `page.waitForResponse` gives up after **60 s** waiting for the write issued by the same screen. Three attempts (12:52:02, 12:53:05, 12:54:13 UTC).

The collector's own message rules the account out before anyone can suspect it: *"this is **NOT** a key or account problem, the key was never probed"*. The failure is upstream of any call to Anthropic.

**Blast radius beyond the two failures.** Three of the run's four skips are consequences of this one: `anthropic-provider.spec.ts:251` and `:285` (both gated on provider health), and `core-components/agent-component-regression.spec.ts` → *"selecting a different-provider model swaps the canvas provider icon"*, whose recorded reason is `not configured in this Langflow instance: ANTHROPIC_API_KEY`. So the run lost the whole Anthropic surface, not two tests.

| Spec (line) | Waits for | Signature |
|---|---|---|
| `tests/collect-models.spec.ts:39` ("collect providers status and models from UI") | `the credential write to answer within 240 s of clicking Save on Settings → Model Providers` | `Error: provider(s) this lane cannot run without were never configured by the collector — this is NOT a key or account problem, the key was never probed: anthropic — collector stall: no credential write answered within 240s of clicking Save` |
[...]

PR #1831, opened 2026-09-12
fix(agent-max-iterations): let the failure name the truncation instead of waiting for it (#1830)

### LINKED ISSUE #1830: agent-max-iterations causal control: the agent answer is TRUNCATED mid-UUID on gemini-2.5-flash — backend stores the same cut text, the tool output is complete
> **Rewritten on 2026-09-12 after measurement.** The first version of this issue diagnosed a partially-streamed read and proposed a retrying assertion. That diagnosis was wrong, and [PR #1831](https://github.com/oriontech-me/langflow-e2e/pull/1831), which implements it, is on hold. What follows is what was measured. The original text is in the issue's edit history.

Found by the **VM daily lane** (step 14 of the VM migration), daily of **2026-09-11**, and re-measured directly on 2026-09-12 against target `1.13.0.dev9`, the published distribution the lane serves.

**Upstream:** _not filed yet_ — see "Who owns this" below.

## Symptom

`tests/tests-automations/regression/core-functionality/llm-agents/agent-max-iterations.spec.ts:311` — "causal control — a high max iterations does not hit the limit" (`@stable @regression @agents @playground`), param **`google / gemini-2.5-flash`**.

The UUID assertion fails because the agent's answer is **cut mid-UUID**:

```
Expected pattern: /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/i
Received:  "The UUID value returned is: feb669ec-c77a-4f9"
```

## It is not a read that arrives early — that was ruled out by experiment

Six directed runs on the VM lane, the two halves alternated against the same instance, `--workers=1 --retries=0`:

| | rep 1 | rep 2 | rep 3 |
|---|---|---|---|
| `main` — single `innerText` read | fail | fail | fail |
| PR #1831 — assertion retries for 30 s | fail | fail | fail |

The retrying half fails **after waiting out the full 30 s**, and its failure screenshot shows a finished run — `✓ Finished in`, token count, `FETCH CONTENT` executed — with the answer still cut. Waiting cannot complete text that is not coming.

The three received strings recorded on 2026-09-11 said the same thing and were read wrong: cut at 26, 34 and 38 characters, always mid-UUID. **A race would not fail 9 attempts out of 9.**

## Where the truncation happens — measured

One provocation, three readings of the same message, 25 s after the run finished:

| Reading | Value |
|---|---|
| bubble `innerText` | `The exact "uuid" value returned is: 323e7651-7872-4576-a565-1e09` |
| bubble `textContent` | *identical* |
| backend, `GET /api/v1/monitor/messages` → `text` | *identical* |
| **the tool's own output**, in `content_blocks` | `{"uuid": "323e7651-7872-4576-a565-1e09116a4f67"}` — **complete** |

**So it is not a rendering defect.** The frontend displays exactly what the backend stored. The tool fetched the value correctly and the full UUID is present in the message's `content_blocks`. What is truncated is the model's own answer text, and it is truncated before it is persisted.

Two more fields from the same message, both relevant:

- `properties.state: "complete"` — nothing marks the message as cut.
- `properties.usage: {"input_tokens": 960, "output_tokens": 820, "total_tokens": 1780}` — **820 output tokens for a 63-character answer.**
- `properties.source.source: "gemini-2.5-flash"`.

## Leading hypothesis, and how to settle it

820 output tokens for one short sentence points at **reasoning tokens consuming the output budget**, with the visible answer cut when the cap is reached. That is an inference from the token count, not a measurement — what is measured is everything in the table above.

Discriminator, cheap: provoke the same flow on the same instance against **`gemini-3.5-flash`** (the model the Actions lane settles on) and against `gpt-4o-mini`, and compare `output_tokens` and whether the answer completes. If 2.5-flash truncates and the others do not, the behaviour is the model's, and the remedy is the one already scheduled — see below.

## Why the Actions lane is green

**The two lanes run different models under the same param name.** Actions settles on `gemini-3.5-flash`; the VM's provisional Google key is pinned to `gemini-2.5-flash`, because that model was retired for new accounts. That is condition 2 of environment divergence nº 5, already recorded.

This matters beyond this test: it means the red **is** explained by the axis on which the two lanes cannot be compared today, and not by anything the VM does wrong.

## Who owns this

Three parts, deliberately separated:

1. **The divergence.** The definitive Google key on the `qa` secrets file puts both lanes on the same model and this red disappears on its own. Already queued, already owned, no new work.
2. **The spec.** Whatever the model does, the failure text should say *the answer was truncated* instead of looking like the agent failed to fetch or the cap misfired. As it stands, the artifact the triage reads does **not** carry its own refutation — which is exactly the property the migration's verdict rule asks of a red. It cost a day of wrong diagnosis, in this issue.
3. **The product, possibly.** A message whose text is cut while `state` says `complete`, with no `finish_reason` surfaced anywhere in the payload, is arguably worth filing upstream. Deferred until the discriminator above says whether the cut is the model's doing.

## What this replaces

The `useInnerText` observation from the first version survives as a fact — `toContainText` compares `textContent` by default, and this file's `expectToolLoopEntered` note records that the collapsed "Agent Steps" disclosure is outside `innerText` — but it is no longer load-bearing: the three readings above are identical, so the distinction changes nothing here.

Victor-w-Madeira:
Closes #1830.

> **Rewritten.** The first commit on this branch implemented a diagnosis that measurement refuted; it is reverted in `6521520f`, which is the change being proposed. The [hold comment](https://github.com/oriontech-me/langflow-e2e/pull/1831#issuecomment-5649051530) and the rewritten #1830 carry the evidence.

## What is actually wrong

The agent's answer is **truncated mid-UUID** — and it is truncated before it is stored, not while it is read.

One provocation, three readings of the same message on `1.13.0.dev9`, param `google / gemini-2.5-flash`:

| Reading | Value |
|---|---|
| bubble `innerText` | `The exact "uuid" value returned is: 323e7651-7872-4576-a565-1e09` |
| bubble `textContent` | *identical* |
| backend `GET /api/v1/monitor/messages` → `text` | *identical* |
| the tool's own output, in `content_blocks` | `{"uuid": "323e7651-7872-4576-a565-1e09116a4f67"}` — **complete** |

`properties.state` reads `complete`, and `usage` reports **820 output tokens for a 63-character answer**. Nothing marks the message as cut.

**Rate, measured, because it is not deterministic:** 8 truncated answers in 9 provocations on this lane. The one clean answer is why this PR was re-validated until a failure appeared rather than declared on a single green run.

## Why the previous approach was wrong

Six directed runs, two halves alternated against one instance, `--retries=0`:

| | rep 1 | rep 2 | rep 3 |
|---|---|---|---|
| `main` — single read | fail | fail | fail |
| the reverted commit — retries for 30 s | fail | fail | fail |

Waiting cannot complete text that is not coming.

## What this PR changes

The spec's real defect is not the assertion — it is that **"the agent never fetched" and "the agent fetched and its answer was cut" are indistinguishable from the failure text**, and the first reading is the one triage reaches for. It cost a full day here, and it is exactly the property the migration's verdict rule asks of a red: the artifact should carry its own refutation.

So the assertion now carries its diagnosis, read from the persisted message. Real output, from the VM lane:

```
Error: the answer is TRUNCATED, not wrong: it ends in a 32-character prefix of the UUID
the tool actually fetched. This is neither a max_iterations failure nor a fetch failure (#1830).
  rendered   : "The \"uuid\" value is: 4e279d8b-638a-4ada-8241-39940919"
  persisted  : identical to the rendered text — the cut is upstream of the UI, not a render
  tool output: 4e279d8b-638a-4ada-8241-399409196e7d
  model      : gemini-2.5-flash · usage {"input_tokens":977,"output_tokens":639,"total_tokens":1616}
A model that spends its output budget before finishing the answer produces exactly this,
and lanes that settle a different model do not reproduce it.
```

A truncated answer still **fails** — the spec's premise is that the agent answers with the value, and passing it would hide a real difference between the lanes. What changes is that it fails saying so.

The helper returns early when the pattern matches, so a passing run issues no extra request, and the branch lives in the helper rather than in the test body (no `no-conditional-in-test`).

## What this PR does NOT do

- **It does not make the red go away.** On this lane the test still fails whenever the model truncates. The remedy for that is already queued and owned elsewhere: the definitive Google key puts both lanes on the same model, and the Actions lane's model does not reproduce this.
- **It does not file the product half.** A message whose text is cut while `state` says `complete`, with no `finish_reason` anywhere in the payload, may be worth filing upstream; #1830 holds that open pending one more measurement.
- **No `@stable` change**, no change to Test 1.

## Validation

- `npm run typecheck`, `npx eslint` on the file (0 errors), `npm run check:checklist-coverage` — all clean.
- VM lane, target `1.13.0.dev9`, tracing on, `--workers=1 --retries=0`: the diagnosis above is a real failure from this branch, not a mock-up. Instance torn down and verified afterwards — zero processes, zero listeners.


Victor-w-Madeira:
**Do not merge — the premise of this PR is refuted by measurement.** The gate being green means nothing here: this test has never failed on Actions.

Directed run on the VM lane, target `1.13.0.dev9`, param `google / gemini-2.5-flash` pinned, `--workers=1 --retries=0`, both halves alternated against the same instance:

| | rep 1 | rep 2 | rep 3 |
|---|---|---|---|
| `main` (single read) | fail | fail | fail |
| this branch (retrying assertion) | fail | fail | fail |

This branch fails **after waiting out the full 30 s**, and the failure screenshot shows why: the run is complete (`✓ Finished in`, token count, `FETCH CONTENT` executed) and the rendered answer is cut mid-UUID — `The exact "uuid" value returned is c912190e-5a70-4d4c-`. The `main` half, same condition: `The UUID value returned is: feb669ec-c77a-4f9`.

The reply is not arriving in pieces — it is **truncated and stays truncated**. Waiting cannot complete text that is not coming.

The three received strings recorded on 2026-09-11 said the same thing and were read wrong: cut at 26, 34 and 38 characters, always mid-UUID. A race would not fail 9 out of 9 attempts; truncation fails every time, which is what was measured.

Holding this branch until the discriminator answers where the truncation happens — what the product renders vs. what it stored. #1830 will carry the finding.

Victor-w-Madeira:
**Cross-lane check on this branch, same selection (`@agents` + `max iterations`), `--retries=0`.**

| param | Actions ([34727808003](https://github.com/oriontech-me/langflow-e2e/actions/runs/34727808003)) | VM lane (`1.13.0.dev9`) |
|---|---|---|
| `openai / gpt-4o-mini` | 2 passed (40.9s) | 2 passed (41.7s) |
| `google / gemini-2.5-flash` | cannot run — the lane probed the provider `inactive` (the #1823 credential signature) | 1 failed with the diagnosis, 1 passed (41.0s) |

**Where the lanes are comparable the change is inert** — same outcome, 0.8s apart between a container on Actions and a venv on the VM. Where they are not comparable, the VM red now arrives explained.

**And this settles the attribution.** Both params ran against the *same instance*, minutes apart: `gpt-4o-mini` answers in full, `gemini-2.5-flash` truncates. Substrate, Langflow version and machine state are out of the explanation — it is the model. The failure from that run:

```
Error: the answer is TRUNCATED, not wrong: it ends in a 23-character prefix of the UUID
the tool actually fetched. This is neither a max_iterations failure nor a fetch failure (#1830).
  rendered   : "The \"uuid\" value is 79aa25b3-c7fb-4014-95ce"
  persisted  : identical to the rendered text — the cut is upstream of the UI, not a render
  tool output: 79aa25b3-c7fb-4014-95ce-a6bab7559dad
  model      : gemini-2.5-flash · usage {"input_tokens":952,"output_tokens":479,"total_tokens":1431}
```

Still untested: `gemini-3.5-flash`, the model the Actions lane settles on — the VM's provisional key cannot reach it, which is condition 2 of divergence nº 5 and the reason this param is not comparable today.

Victor-w-Madeira (commented):
## Verdict

**The direction is right; one change I would ask for before merge** — the diagnostic helper can throw and replace the very failure it exists to explain. The other five are improvements.

The central argument holds and is well evidenced: six directed runs refute the previous commit, the truncation is measured through four readings of the same message, and the rate (8 in 9) is stated rather than hidden behind one green run. The cross-lane check — `openai` inert on both lanes, 0.8 s apart — is the right proof that the change does not touch the passing path.

---

### 1. High — the helper can replace the failure it exists to explain

`tests/tests-automations/regression/core-functionality/llm-agents/agent-max-iterations.spec.ts` (helper body, and the call site)

```ts
const bearer = await getAuthToken(request);   // throws, by contract
...
const messages = await res.json();            // throws on a non-JSON 200
```

`getAuthToken` documents it explicitly: *"If the budget runs out, the original error propagates"* — after ~30 s of backoff (2 s + 8 s + 20 s). Because `await explainMissingUuid(...)` is evaluated **before** `expect(...)` is called, a throw there kills the test before the assertion exists:

- **Scenario**: the backend recycling / wedged (#1077, which this repo documents as recurring) at the moment the model truncates its answer.
- **Result**: the artifact reads `apiRequestContext.get: ...` — with neither the diagnosis **nor** the original `expected to match UUID_SHAPE / received ...`. That is strictly less information than `main` produces today.
- **Aggravating**: +30 s on a test that has already spent its run, against the 5-minute cap — the worst case trades a named assertion failure for a bare timeout.

This is the lesson CLAUDE.md records for `tests/fixtures/http-error-body.ts`: *"the helper that exists to keep the evidence trail honest would have been the one thing in it able to redden a run"* — and the remedy there is the same one: **every branch reports**, tolerant of a throw of any type.

```ts
async function explainMissingUuid(...): Promise<string | undefined> {
  if (UUID_SHAPE.test(rendered)) return undefined;
  try {
    ...current body...
  } catch (e: any) {
    return `could not diagnose the missing UUID: ${String(e?.message ?? e).split("\n")[0]}`;
  }
}
```

### 2. Medium — the closing sentence contradicts its own head in 2 of the 3 branches

```
A model that spends its output budget before finishing the answer produces exactly this,
and lanes that settle a different model do not reproduce it.
```

It is printed unconditionally, but it is only true in the `overlap >= 8` branch. In the other two — *"the model answered without using what it fetched"* and *"no UUID appears in the tool output either"* — it asserts a cause the message itself has just ruled out. In a PR whose product is the honesty of the artifact, that is the next day of misread triage. The sentence belongs to the truncation `head`.

### 3. Medium — the overlap computation is case-sensitive while `UUID_SHAPE` is `/i`

```ts
if (fetched.startsWith(rendered.slice(rendered.length - n)))
```

If the model spells the UUID in uppercase (it happens), `overlap` is 0 and the message falls into **"the model answered without using what it fetched"** — precisely the wrong diagnosis this PR exists to eliminate, with nothing saying the comparison is what failed. Lowercasing both sides closes it.

### 4. Medium-low — duplication with `expectToolLoopEntered`

The new helper repeats that function's monitor GET, its `find(m => m.sender === "Machine" && content_blocks.length > 0)`, and its `flatMap(...).filter(c => c.type === "tool_use")`. The two already differ in one respect (one polls, one reads once), which is how copies start drifting. A `fetchAgentMessage(request, flowId)` returning `{ aiMsg, toolContents }` leaves one reader.

### 5. Low — the measurement did not reach the spec doc

`docs/core-functionality/llm-agents/agent-max-iterations.md` is untouched, and its **Tags** section already carries the paragraph that asks for exactly this: *"belongs in this section as a measured rate, not in a new product issue."* This PR measures **8 truncations in 9 provocations** on `google / gemini-2.5-flash`, in a `@stable` test that is consequently red on that lane — the residual-risk shape that paragraph anticipates. `Last validated` also still reads `1.13.0.dev4` against a `dev9` measurement (same cycle, so it does not fail the checklist).

### 6. Low — the renderer is this PR's product and nothing pins it

The message *is* the deliverable, and #1226's lesson in CLAUDE.md is about this: the rendering moved out of the YAML into a script *whose tests assert on its output*. The pure half here (the overlap plus the six-line assembly) would fit in `tests/helpers/` with a `.test.ts` — and would have caught items 2 and 3 on its own.

---

None of this affects the passing path, `@stable`, or Test 1.

Victor-w-Madeira:
**Field revalidation after the review fixes — and it caught the renderer asserting the wrong reading.**

First revalidation returned an answer of `The exact "` — eleven characters, 415 output tokens, cut before it ever reached the UUID. With no prefix to overlap, the message fell through to the second head and said *"the model answered without using what it fetched"*. That is false, and it is precisely the misdiagnosis this change exists to remove. The unit tests missed it because they only modelled a cut INSIDE the value.

`a962257d` makes that branch state **no cause**: it names both readings and hands over the two numbers that separate them, since no fixed threshold would settle it honestly for every model.

Four more provocations then exercised all three heads:

| # | outcome | head |
|---|---|---|
| 1 | red | truncation, 13-character prefix |
| 2 | red | truncation, 12-character prefix |
| 3 | **green** | complete answer — the path that issues no extra request |
| 4 | red | the fixed branch: `"The"`, 3 characters against 419 output tokens, no cause claimed |

`2a717ee7` then removes the rate from the failure text. The running count is **13 truncations in 15 provocations** (2026-09-12), so the `8 of 9` written into the message was already stale — a figure in prose inside the artifact ages in silence, which is the defect the #1828 review named. The message now describes the shape; the doc carries the count, dated, and the widened range: the cut lands anywhere from 3 to 63 characters.

Unit lane: 8 cases on the renderer, 1393 green overall. The two cases added cover the measured 11-character answer and a reading with no usage block, which reports the token count as unknown rather than implying a small one.

Issue #1789, opened 2026-09-09
T1: park groq-provider and mistral-provider — the components are not in the tested image, and that is already decided


rafaelgiln:
> Filed off `docs/triage/inherited-spec-triage.md`, the measured verdict table landed by #1784 (PR #1785). Nine `manual.yml` dispatches, three passes over three shards, `retries=0`, every test observed three times.
>
> **Tier T1**, which is Wave 8's scope (design §6). Wave 8's exit criterion is *"T1 is empty: every one of its 26 specs has reached promote / delete / park"*.

**Outcome: PARK. There is no engineering work in this batch, and that is the finding.**

Both specs skipped in 3/3 runs. Component availability is a packaging decision **per image**, and the standing answer already exists — `docs/component-distribution-policy.md`, decided in #1039 after the question was answered wrongly twice (#898 and #907 both blamed missing `langchain-*` extras when the 1.12 packaging split was at least as much the cause). Restoring `@stable` would put a test that cannot pass into the daily.

Measured against the image the triage ran on, not assumed: the frozen catalog holds neither family, and `globalSetup`'s drift report for those very runs named **only two additions** (`ext:ibm:WatsonxDataMCPComponent`, `ext:ibm:WatsonxDataPrestoComponent`, both in `ibm`) — nothing removed and nothing moved — so the baseline's absence is the image's absence.

Both are **already declared** in `scripts/lib/stable-orphan-exemptions.json` with this exact reason, and #1746's reconciler reports them as `exempt` rather than orphaned. So the park is recorded and verified in both directions already.

What is left is bookkeeping, and it is worth doing because the frozen baseline is a Wave 8 exit criterion ("the frozen baseline has shrunk to the parked set"):

- [ ] confirm each spec's doc states the parked reason and cites #1039 / `docs/component-distribution-policy.md`
- [ ] confirm the `QA-CHECKLIST.md` Part II bullets read as parked, not as pending
- [ ] leave `scripts/lib/stable-orphan-exemptions.json` alone — a second declaration of the same absence is the silent-expiry hazard both mechanisms exist to close

**Do not** open an upstream ticket, and do not add a skip gate. A packaging decision is not a product defect.

## The measured rows

| | Spec | Test | Verdict | Cleanup | Mirrored doc |
|---|---|---|---|---|---|
[...]

PR #1832, opened 2026-09-13
chore(ci): carry GROQ_TEST_MODEL into the lanes that run specs

Victor-w-Madeira:
> **Body rewritten after review.** The three technical decisions stood; the justification did not, in the two places the review names. What follows replaces it, and commit `cf8db14` carries the doc half.

## What this actually fixes, stated narrowly

`groq-provider.spec.ts` has **two** gates, and they are not equal today:

1. `isProviderComponentAvailable(request, "groq")` — the Groq component is not exposed by the tested image ([#1039](https://github.com/oriontech-me/langflow-e2e/issues/1039), closed; the committed catalog baseline carries no `groq` entry). The spec also no longer carries `@stable`, so it is not in the daily's selection at all.
2. The live catalog probe, which reads `GROQ_TEST_MODEL`.

**Gate 1 fires first, so gate 2 is never reached on any lane today, and this PR changes no observable result anywhere.** What it fixes is the *latent* gate — the one that decides the spec's fate on the day the component returns to the image, which is what [#1789](https://github.com/oriontech-me/langflow-e2e/issues/1789) has parked the spec waiting for. The earlier claim that the spec "skips on every lane at once" without this change described an outcome that is already true for a different, already-decided reason. That is the `#1012` class of overstatement and it does not belong in a body this repository keeps.

The measurement behind the pin stands on its own: against this repository's key, `GET https://api.groq.com/openai/v1/models` returns 14 models and **none of the llama-3.1 line** — `openai/gpt-oss-{20b,120b,safeguard-20b}`, `qwen/qwen3.{6,8}-27b`, `groq/compound{,-mini}`, three whisper entries, `allam-2-7b`, two llama-prompt-guard entries, `canopylabs/orpheus-*`. The spec's default is gone from this account, so the latent gate is already armed to skip the day gate 1 stops firing.

## The three decisions, unchanged

**A variable, never a secret.** GitHub masks secret values in logs, so a masked model id renders the skip reason as `model "***" not in the live Groq catalog` — the refutation leaves the artifact triage reads.

**Only the steps that RUN specs** — `Run @stable tests (shard N)`, `Run impacted specs`, `Run destructive lane`. The `Collect models` steps carry `GROQ_API_KEY` but never read the override: Groq is deliberately absent from `providerConfigMap`.

**`||` instead of `??`.** An unset repository variable arrives as an empty string, not undefined; under `??` it would win and the probe would report `model "" not in the live Groq catalog`.

## Corrections the review earned, recorded rather than quietly fixed

**The dropdown is not the account catalog.** It is that catalog **filtered by a live per-model probe** — `groq_model_discovery.py::_test_chat_completion` and `_test_tool_calling` drop entries as `not_supported`, and a discovery failure returns the static `groq_constants.py` list. So "the id the account serves is the id the dropdown gets" was the wrong mechanism. The conclusion survives on different evidence: the real chat completion on `openai/gpt-oss-20b` is exactly what that filter demands.

**The pin is outside the static fallback list, and that is a new failure mode.** `groq_constants.py` carries `llama-3.1-8b-instant` and not `openai/gpt-oss-20b` (only `openai/gpt-oss-safeguard-20b`, flagged `not_supported`). Under a discovery fallback the dropdown offers the retired default and the `toContainText` assertion **fails** rather than skipping — a red whose cause is "discovery fell back", not "the model is gone". Written into the spec doc by `cf8db14`, next to the precondition it qualifies.

**`OLLAMA_TEST_MODEL` is injected, contrary to the second commit's message.** `daily-stable.yml:351`, `manual.yml:237` and `nightly.yml:90` all set it — as a **literal**, so it can never arrive empty and leaving that sibling on `??` is still correct. The commit message says "no workflow injects those", which is wrong as written; it is corrected here rather than by amending, because the branch tip currently also exists on the read-only mirror and a force-push would leave that ref unreachable from the mirror — the divergence class that froze the destination for 36 hours on 12/09.

## Validation

- `GET /openai/v1/models` with the repository key — 200, the 14 entries above, `llama-3.1-8b-instant` absent.
- Real inference on the pin: `openai/gpt-oss-20b` returns the sentinel exactly, non-empty, no reasoning preamble — which is what the spec's hard assertion (`reply.length > 0`) needs and what the discovery filter requires. `qwen/qwen3.6-27b` emits a `<think>` preamble; `groq/compound-mini` is an agentic system. Both would pass; neither is as close to the retired default.
- `npx tsc --noEmit` clean; `eslint` on the touched spec reports only pre-existing warnings, 0 errors.
- `node --test scripts/daily-matrix-provider-keys.test.mjs scripts/ci-change-coverage.test.mjs scripts/check-vm-env-parity.test.mjs` — 55 passed.
- `node scripts/check-vm-env-parity.mjs` — 10 of 10 carried; the change is on a step's env, not the daily's Langflow service.

**What is NOT proven, precisely:** no run executes this spec while #1039 holds, so the pin's end-to-end path through the dropdown is **deferred**, not verified. It becomes verifiable when the component returns to the image (#1789).

## Scope and linkage

Stage 1 of the VM migration, the lane that places the definitive provider keys on both lanes. It does **not** contradict #1789's PARK verdict (*"do not add a skip gate"*): it changes lane wiring and leaves the spec's own verdict logic untouched.

`manual.yml` is out of scope and named rather than fixed: its `GROQ_API_KEY` sits in the `env:` of the `Collect models` step, and a step's `env` does not cross into another step, so the spec skips there on the key — a different gate again.


Victor-w-Madeira:
## Review

**The code is right; the PR body is not, in two places, and the spec doc went stale under it.** The three decisions (`||` over `??`, a variable rather than a secret, only the steps that run specs) all hold — I checked each against the tree rather than reading the rationale. What needs changing is the justification the repository will keep, plus one paragraph in `docs/`.

---

### 1. The spec does not skip on the model today — it skips earlier, and the file this PR edits says so

`groq-provider.spec.ts:106-112` calls `isProviderComponentAvailable(request, "groq")` **before** the Groq API probe, with the comment stating exactly that: *"Runs BEFORE the cloud-API probe: no point checking the key when the component cannot be placed at all."* The Groq component is not in the tested image (#1039; the committed catalog baseline carries no `groq` category; #1789 is open with a PARK verdict).

So on every lane today the skip is the #1039 one, and the catalog probe is never reached. The body's claim —

> Without an override the spec skips on **every lane at once** — a silent loss of coverage

— describes an outcome that is already true for a different, already-decided reason, and `GROQ_TEST_MODEL` changes no observable result on any lane today. What this PR actually fixes is a **second, latent gate**, for the day the component returns to the image. That is worth merging; it is just a different sentence. An overstated justification is the class #1012 is about, and this repository treats it as a defect rather than as marketing.

Same correction in **Validation**: *"the proof is the next run that executes the spec"* — no run will execute this spec while #1039 holds. Better as: the proof is deferred until the component returns (#1789).

### 2. "the id the account serves is the id the dropdown gets" is not the mechanism

`lfx/base/models/groq_model_discovery.py` filters what it discovers: `_test_chat_completion(model_id)` runs **per model** to drop non-chat entries, `_test_tool_calling` runs alongside it, and a model that fails is recorded `not_supported: True` and leaves the list. The dropdown is the account catalog **filtered by a live probe**, not the raw catalog.

The conclusion survives — the real chat completion on `openai/gpt-oss-20b` in Validation is precisely the evidence that filter demands — but the argument as written is not the one carrying it, and this body is what the next reader reaches for when the pin stops working.

### 3. The pinned model is absent from the static fallback list — worth naming

`groq_constants.py` does not contain `openai/gpt-oss-20b` (only `openai/gpt-oss-safeguard-20b`, flagged `not_supported`). `_get_fallback_models()` returns that static list when discovery fails — and the static list **does** contain `llama-3.1-8b-instant`.

So the pin makes the spec strictly dependent on live discovery succeeding: under a discovery fallback the dropdown offers the retired default and the `toContainText` assertion **fails** rather than skipping. Not a reason to change the pin — the account no longer serves llama-3.1, so the old default would die at inference anyway — but it is a new failure mode and belongs in the doc alongside item 4.

### 4. The spec doc went stale under this PR's own measurement

- `docs/core-functionality/model-provider/groq-provider.md:66` still presents `llama-3.1-8b-instant` as the default with no note that the Groq catalog is **per account**.
- Line 184 lists *"If the Groq catalog drops `llama-3.1-8b-instant` (override via `GROQ_TEST_MODEL`)"* under **When to review this test** — as a hypothetical. It has happened; this PR measured it.

`.env.example` was updated, which is the right place for the operator, but the spec doc is the SDD source of truth. One paragraph.

### 5. No issue reference

The PR review checklist asks for a current-wave item or an approved exception issue. The body names the context ("stage 1, the VM daily") without a number and carries no `Closes`. It also touches a spec #1789 has just parked (*"do not add a skip gate"*) — it does not contradict that, because it changes lane wiring rather than the spec's verdict, but saying so explicitly costs one line and saves the next reader the check.

---

### Verified and correct — recorded so this does not round-trip

- **`||` over `??`** — correct. An undefined repository variable reaches the step as an empty string, and `??` would let it win.
- **A variable, not a secret** — correct, and the `***` argument about the skip reason is the right one.
- **Only the steps that run specs** — confirmed: `Run @stable tests (shard N)` (daily), `Run impacted specs` and `Run destructive lane` (pr-validation). No `Collect models` step touched.
- **`manual.yml` named as out of scope** — correct, and for exactly the stated reason: `GROQ_API_KEY` there sits in the `env:` of the **`Collect models` step** (`manual.yml:355`), and a step's `env` does not cross into another step, so the spec skips on the key.
- **Leaving the siblings on `??`** — the reasoning *"no workflow injects those"* is imprecise: `OLLAMA_TEST_MODEL` **is** injected by all three lanes (`daily-stable.yml:351`, `manual.yml:237`, `nightly.yml:90`). It is injected as a **literal** (`llama3.2:1b`), so it can never arrive empty and the conclusion holds. One word.
- **`hasText` + `.first()` against the new id** — safe: no other id in the measured catalog contains `openai/gpt-oss-20b` as a substring (`openai/gpt-oss-120b` and `openai/gpt-oss-safeguard-20b` do not).
- **The variable exists** — `gh variable list` reports `GROQ_TEST_MODEL=openai/gpt-oss-20b`. The change is not a no-op through an unset variable.
- **`check-vm-env-parity` out of scope** — correct; it covers the env of the daily's Langflow *service*. VM-lane parity is carried by the clone's `.env`, because `playwright.config.ts` calls `dotenv.config()` unconditionally even with `CI=true`.

CI: 7/7 green.

---

Items 1, 2 and 4 are what I would want before merge — two paragraphs in the body and one in the doc. Item 3 is cheap and fits with 4. Item 5 is one line.

Victor-w-Madeira:
All five verified against the tree before acting — including the two that say I was wrong.

| # | Verdict | Where it landed |
|---|---|---|
| 1 | Confirmed. `isProviderComponentAvailable` runs before `probeGroq`, the catalog baseline has no `groq`, and the spec's tags are `@components @model-provider @playground` — no `@stable`, so it is not even in the daily's selection. | Body rewritten: this fixes a **latent** gate and changes no observable result today. Validation now says the proof is deferred to #1789, not "the next run". |
| 2 | Confirmed upstream. `_test_chat_completion` / `_test_tool_calling` run per model and drop entries as `not_supported`; `_get_fallback_models()` returns the static list on failure. | Body: the mechanism is restated, and the live inference is named as the evidence the filter actually demands. |
| 3 | Confirmed. `groq_constants.py` carries `llama-3.1-8b-instant` and not `openai/gpt-oss-20b` (only `openai/gpt-oss-safeguard-20b`, `not_supported`). A discovery fallback makes the spec **fail** on `toContainText`, not skip. | `cf8db14`, in Preconditions, next to the precondition it qualifies. |
| 4 | Confirmed, both lines. | `cf8db14` — the per-account fact with its date, and the hypothetical struck through and dated. |
| 5 | Confirmed. | Body: #1789 named with the reason it is not contradicted (lane wiring, not verdict logic), plus the stage-1 linkage. |

And the `OLLAMA_TEST_MODEL` correction is right: all three lanes inject it, as a literal, so it cannot arrive empty and the sibling stays on `??`. The second commit's message says otherwise and is wrong as written. It is corrected in the body rather than by amending: the branch tip also exists on the read-only mirror right now, and a force-push would leave that ref unreachable from it — the divergence class that froze the destination for 36 hours on 12/09.

CI 7/7 green on `cf8db14`, including *Spec-doc dependency paths*.

PR #1833, opened 2026-09-14
feat(prepare-target-dist): bring the published-distribution installer into the repository

Victor-w-Madeira:
## What this is

`scripts/prepare-target-dist.sh`, the sibling of `prepare-target-source.sh`: it puts the **published distribution** of a named Langflow version into the target's venv, or refuses.

The VM lane switched its target to the published distribution on 2026-09-10, and the script that places it has lived at `/root/install-target-dist.sh` on the target ever since — twenty lines, no version control, no review, and its refusals verified only by the person who wrote them. This brings it in, with every refusal pinned by a test.

`prepare-target-source.sh` does not go away and the division is not preference: a `release-*` branch has no published wheel until the nightly builds one, so testing a commit before that point is only possible from source.

## Why it matters more than its size suggests

This is the piece that decides **which product the lane serves**. Getting it wrong does not produce a red run — it produces a green run against the wrong version, which is the one outcome a comparison lane cannot survive. That is the same class `prepare-target-source.sh` already closes for the source clone.

The check that carries the weight is the **cross-registry** one: the expectation comes from the published image (`resolve-target-version.mjs` matches `latest`'s digest against the other published tags), the wheel comes from PyPI, and nothing makes the two agree. When they disagree, this refuses with both numbers on screen rather than letting a comparison of two lanes look like a product difference. **It has never fired on the machine** — today the two registries agree — so the test here is the only thing that has ever executed that branch.

## What changed on the way in

**The frontend gate no longer asks a guessed path.** `lib/python3.14/site-packages/…` stops being true the day the interpreter moves, and the refusal then reads *"the distribution carries no built frontend"* about a distribution that carries one. It now asks the installed package where it lives.

**An empty `TARGET_VERSION` is refused explicitly.** `langflow==` is not a pin — it asks the index for whatever is newest, and it *succeeds*, so nothing downstream would notice the lane stopped testing what was resolved.

**The interface is env vars, not positional arguments.** The orchestrator already invokes the source preparer as `ssh <host> "VAR=x bash -s" < script`; two shapes for two siblings is a difference with no reason behind it.

## The third commit came from the field, not from review

Asked on the target for a version that does not exist, the refusal read *"the index does not carry it. That is the two registries disagreeing"* — a settled diagnosis about a cause the script cannot determine. An unreachable index, a full disk and a yanked wheel all fail that same call, and `> /dev/null 2>&1` had left our sentence as the only account of the failure.

Both `uv` invocations now write to stderr. Measured on the target, uv answers:

```
× No solution found when resolving dependencies:
╰─▶ Because there is no version of langflow==1.13.0.dev999 and you require
    langflow==1.13.0.dev999, we can conclude that your requirements are
    unsatisfiable.
```

That is the sentence that was being thrown away. stdout stays `key=value`, so the caller's contract is unchanged.

## Verification

**Unit** — 11 tests, and each property was *mutated* to confirm it is held rather than merely covered: removing the cross-registry comparison, restoring the guessed frontend path, dropping the venv recreation, and removing the empty-version refusal each turn the lane red. Full lane: `npm run test:scripts` → 1591 passed, 0 failed.

**Field, on the target, against the real `uv` and the real index**, into a separate venv so the lane's own target was never touched:

- `TARGET_VERSION=1.13.0.dev12` → `prepared_version=1.13.0.dev12`, frontend resolved through the installed package, `install_s=2`, stdout carrying only `key=value`
- `TARGET_VERSION=1.13.0.dev999` → refused, `rc=2`, with uv's reason above our message
- `TARGET_VERSION=` → refused before anything reached the index

The validation venv was removed afterwards and the lane's venv confirmed intact at `1.13.0.dev12`.

## Not in this PR

Wiring it into `run-e2e.sh`. The machine wrapper calls the installer before the orchestrator and sets `PREPARE_TARGET=0`; giving the orchestrator a source/dist mode is a larger change with its own design questions, and it is not what makes this script reviewable.


Victor-w-Madeira:
## Independent review — two cold reads, and the second one hurt

Copilot is out of quota on this repository right now (both attempts answered only that), so the review came from two readers given the diff and nothing about why any line was written. Everything below that changes code was re-measured on the target before being acted on.

### The false accusation

`TARGET_VERSION=v1.13.0.dev12` installs the **identical wheel** — PEP 440 accepts the `v` — and `importlib.metadata` reports the normalised form. The raw string comparison called that *"the two registries disagree"*: the one refusal this file says has never fired, spent on a correct placement, sending the operator to compare Docker Hub against PyPI. And the input is one this repository produces itself — `resolve-target-version.mjs` puts the `v` form in its `ref` field.

Measured on the target: `uv pip install langflow==v1.13.0.dev12` → `rc=0`, metadata `1.13.0.dev12`.

### The silent path, which is the one the file exists to close

This verified `langflow`. On the target `langflow` ships **8 files**; `langflow-base` ships **2746**, and it is the one that owns `langflow/frontend/index.html` — I checked which distribution lists it. On the nightly path the two are pinned together; on a stable version the dependency is a **range**, resolved by a third system nobody here controls. Every refusal could pass while the machine served a base the image never contained.

Both are verified now, and a missing base is refused explicitly.

### A justification of mine was false

The header and the refusal text said `langflow==` *"asks the index for whatever is newest, and it SUCCEEDS"*. It does not — `error: Failed to parse: langflow== — Unexpected end of version specifier`. The guard stays (it fails before the venv is destroyed) but is now defended by what actually happens instead of by a scenario that cannot occur.

### And the tests were largely testing themselves

The second reader mutated the script fourteen times: 5 controls caught, **9 survived**. The cause was one design mistake — the venv's `python` was a stub that matched the program *text* and returned a canned answer, so everything inside the two Python one-liners was unverified. Changing `frontend` to `static`, dropping the path join, and reading `langflow-base` where `langflow` was meant **all shipped green**.

The probe now runs under the real `python3` against a fixture with real `.dist-info` metadata, `-S` so the host's site-packages cannot decide which branch runs, and a `langflow/__init__.py` that **prints on stdout** — which is what turns "found without executing the package" into a fact rather than a claim.

Four properties that had no test at all now have one: `uv venv` is asked for the configured interpreter; the uv-missing refusal happens **before** the venv is destroyed; the `export PATH` line is exercised; and both `>&2` redirects are pinned by making the stub chatty on stdout.

One hygiene finding that was a single stray file from silence: with the stub dropped, `parts.join(":")` gave a PATH beginning with an **empty component**, which POSIX reads as the current directory. The reviewer put an executable `uv` at the worktree root and the "uv is absent" test failed.

### Verification

| | |
|---|---|
| Unit | 22 tests, `npm run test:scripts` → **1602 passed, 0 failed** |
| Mutation | all **11** properties the two reviews named are re-run and **caught** |
| Field | invoked the way the machine invokes it — `ssh <bin> "TARGET_VERSION=x bash -s" <` the file, heredoc and all |

Field cases: the pinned version (`prepared_base_version` now reported alongside, `python=3.14` measured rather than echoed), the `v` form now succeeding, and a non-venv directory refused **with its contents intact**.

### Left standing, said rather than fixed

Timing values are assert-*shaped*, not assert-*valued*: hardcoding `INSTALL_S=0` still passes. Nothing downstream reads them, so pinning a duration would cost a flaky test to protect a number nobody consumes.

Issue #1849, opened 2026-09-14
Provider save specs read the `validate-provider` body only after `Promise.all([validate, persist])` — on a refusal that assert never runs (measured on ollama-provider)


rafaelgiln:
Follow-up found while re-verifying `ollama-provider.spec.ts` locally on nightly `1.13.0.dev12` (2026-09-14). Sibling of #1829, whose fix shape for `anthropic-provider` is *"assert the body, not the status"*: four provider specs already do exactly that, and on the path the assertion exists for it **never executes** — it sits behind a `Promise.all` that also waits on the persistence call the panel does not make after a refusal.

Not a Langflow defect: the product correctly refuses to persist a credential or URL that `validate-provider` rejected (settled in #1823).

## Measured — `ollama-provider.spec.ts` Test 1

Instance: `langflowai/langflow-nightly:latest` = `1.13.0.dev12`, started with the env of `scripts/start-langflow-docker.sh` **minus** `LANGFLOW_SSRF_ALLOWED_HOSTS`, so the base URL is refused.

| Observation | Value |
|---|---|
| `POST /api/v1/models/validate-provider` body | `{"valid":false,"error":"Hostname host.docker.internal resolves to blocked IP address(es): 192.168.5.2. To allow this hostname, add it to LANGFLOW_SSRF_ALLOWED_HOSTS environment variable."}` |
| Test 1 result | **failed after 65.8 s** — `TimeoutError: page.waitForResponse: Timeout 60000ms exceeded` |
| Where | [`ollama-provider.spec.ts:195`](https://github.com/oriontech-me/langflow-e2e/blob/b0d5c63f/tests/tests-automations/regression/core-functionality/model-provider/ollama-provider.spec.ts#L195-L226), the **persistence** waiter |
| The `valid === true` assert (`:223`) | never reached |

The spec doc's M1 note says the body assert *"makes it fail fast and name the cause"*. In exactly the scenario that note describes it does neither: the failure is the same unattributed 60 s persistence timeout the assert was added to replace (#931).

## Why

```ts
const [validateResp, persistResp] = await Promise.all([validatePromise, persistPromise]);
// ...body assert on validateResp
```

[...]

Issue #1829, opened 2026-09-12
anthropic-provider Test 1 — a 2xx from validate-provider does not prove the key authenticates, and the no-skip-gate rationale is false for an invalid key


Victor-w-Madeira:
Follow-up to [#1823](https://github.com/oriontech-me/langflow-e2e/issues/1823). Both defects below were **named in that issue's root-cause comment** and the first was recorded in the body of PR #1418 and never filed; this issue is that filing. Neither is a Langflow defect — `#1823` settled that the product correctly refuses to persist a credential the provider rejects.

Spec: `tests/tests-automations/regression/core-functionality/model-provider/anthropic-provider.spec.ts`, **Test 1** — *"Anthropic API key is configured via Settings → Model Providers"*.

## 1 · The assertion does not assert what its comment claims

```ts
// validate-provider 2xx = the key authenticates against Anthropic live;
// POST|PATCH /variables 2xx = the key is persisted globally.
expect(validateResp.ok()).toBe(true);     // line 245
expect(persistResp.ok()).toBe(true);
```

`POST /api/v1/models/validate-provider` answers **`200` for an invalid key** and puts the verdict in the body. Measured on `1.13.0.dev9` in #1823, one call per key on the same instance:

| Provider | body |
|---|---|
| OpenAI | `{"valid":true,"error":null}` |
| Anthropic | `{"valid":false,"error":"Invalid API key for Anthropic"}` |

So `validateResp.ok()` is `true` in both directions and the first assertion cannot fail on a rejected key. The half that does fail is the second one — and only as a **60 s `waitForResponse` timeout**, because the frontend gate (`useProviderConfiguration.ts` → `handleSaveAllVariables`: `const isValid = await validateCredentials(); if (!isValid …) return;`) never issues `POST /api/v1/variables/` at all.

**Consequence.** A definite refusal that the product reaches in ~0.5 s is reported as an instrument timeout with no cause in it — the same attribution defect #1823 identified one instrument earlier, in `collectModelsForProvider`. On the 2026-09-11 daily this shape cost the run 300 s of the collector's sweep budget and an unreviewed `@stable` removal on a correct test.

**Fix shape.** Assert the **body**, not the status: `validate-provider` must answer `200` *and* `{"valid": true}`, so a rejected key fails at the point of refusal with the provider's own message. Then the `/variables/` wait becomes a consequence check rather than the only detector.

## 2 · The documented reason for carrying no `providerSkipGate` is only half true

[...]

Issue #1415, opened 2026-08-11
anthropic-provider and google-provider still gate on env-key presence, not provider health (#1029 gate not applied)


rafaelgiln:
Spun out of #1333 (which fixed the OpenAI one).

## Problem

`providerSkipGate()` (`tests/helpers/provider-setup/provider-health.ts`) is the health gate #1029 built **specifically for specs that hardcode a provider**: it reads the `providers.json` that `collect-models` writes and skips with the measured reason when the provider is recorded `inactive` (drained balance, revoked key, spend cap).

18 specs honour it. The three provider-centric specs — the ones whose whole subject *is* a hardcoded provider — do not. They gate on `hasProviderEnvKeys(PROVIDER)`, which measures **presence of the env var, never its health**:

| Spec | Gate today | Tests making a live completion call |
|---|---|---|
| `core-functionality/model-provider/openai-provider.spec.ts` | ✅ fixed in #1333 (Test 2 gated) | — |
| `core-functionality/model-provider/anthropic-provider.spec.ts` | `hasProviderEnvKeys` only | to be audited (4 `test.skip` sites) |
| `core-functionality/model-provider/google-provider.spec.ts` | `hasProviderEnvKeys` only | to be audited (2 `test.skip` sites) |

## Why it matters

Measured on the OpenAI copy of this bug, twice. On the 2026-08-06 daily ([run 31093877484](https://github.com/oriontech-me/langflow-e2e/actions/runs/31093877484)) the account had no credits — `collect-models` recorded `openai inactive — You have no credits remaining`, and 6 of that run's 9 skips quoted it. The provider-parametrized specs skipped correctly. `openai-provider.spec.ts`'s execution test **ran anyway**: the Agent's run never completed, the Stop button stayed visible past the 120 s wait, no `div-chat-message` ever rendered, all three attempts hard-failed, and the workflow auto-removed `@stable` (`cb3082d`). The same thing had already happened under #772/#775.

So a billing outage is reported as a product regression, costs the tag, and spends a triage cycle — per provider, every time the account drains.

## Proposal

Audit both specs and apply the same asymmetry #1333 settled on:

- **Gate the tests that make a live completion call** on `providerSkipGate(PROVIDER)`.
[...]

PR #1868, opened 2026-09-15
fix(model-provider): read the Google Save through armProviderSave (#1867)

### LINKED ISSUE #1867: google-provider Test 1 — the last provider save asserting a status behind Promise.all; a rejected key still surfaces as a 60 s timeout
Follow-up to [#1849](https://github.com/oriontech-me/langflow-e2e/issues/1849), which measured this shape on `ollama-provider.spec.ts` and landed the shared helper, and to #1829, which applied it to `anthropic-provider.spec.ts`. `google-provider.spec.ts` is the last provider save still carrying the original form — and the weaker of the two variants, since it asserts the **status** rather than the body.

Not a Langflow defect: the product correctly refuses to persist a credential `validate-provider` rejected (settled in #1823).

Spec: `tests/tests-automations/regression/core-functionality/model-provider/google-provider.spec.ts`, **Test 1** — *"Google API key is configured via Settings → Model Providers"* (`@stable @model-provider @settings`).

## The code, at `955dce03`

```ts
const [validateResp, persistResp] = await Promise.all([   // :171
  validatePromise,
  persistPromise,
]);
// validate-provider 2xx = the key authenticates against Google live;
// POST/PATCH /variables 2xx = the key is persisted globally.
expect(validateResp.ok()).toBe(true);                     // :177
expect(persistResp.ok()).toBe(true);                      // :178
```

Two defects, the same pair #1829 fixed for Anthropic:

1. **The first assertion cannot fail on a rejected key.** `POST /api/v1/models/validate-provider` answers **`200` in both directions** and puts the verdict in the body (`{"valid":false,"error":"Invalid API key for …"}`), so `validateResp.ok()` is `true` either way. The comment above it — *"the key authenticates against Google live"* — is not what the line checks.
2. **`Promise.all` settles only when both settle.** The frontend gates the write on that body (`useProviderConfiguration.ts` → `handleSaveAllVariables`: `const isValid = await validateCredentials(); if (!isValid …) return;`), so after a refusal **no** `POST|PATCH /api/v1/variables/` is issued at all — and the first thing to settle is the persistence waiter, by rejecting at its 60 s timeout. The already-resolved verdict, with Google's own message in it, is discarded.

Measured for the same shape elsewhere on `1.13.0.dev12`: ollama 64.6 s, openai 32.6 s, openai-compatible 63.4 s, azure 63.9 s (#1849), and anthropic **62.4 s → 3.0 s** after the fix (#1829). Each was a bare `waitForResponse` timeout with no cause in it.

**Why it is worth filing.** This is exactly what cost the 2026-09-11 daily 300 s of the collector's sweep budget and an unreviewed `@stable` removal (commit `883047fc`) on a test whose logic was correct. Google is the provider with the widest rotation gap in the daily — Wednesday is its only slot, so up to 7 days between runs (#1185) — and it is the provider of #1029 and #963, i.e. the account this repo has actually seen go bad.

## Fix shape

Migrate to the existing helper, as the other five specs now do:

```ts
import { armProviderSave } from "../../../../helpers/provider-setup/provider-panel-save";

const save = armProviderSave(page, { subject: "key" });
await page.getByRole("button", { name: /Save|Replace/i }).first().click();
await save.validated();                       // 200 + {valid:true}, else throws quoting Google
const persistResp = await save.persisted();
expect(persistResp.ok()).toBe(true);
```

Note the helper's `validate-provider` predicate is anchored on the **pathname**, where the current `url().includes(...)` is not — the trap `CLAUDE.md` records for `/build/`. Its default timeout is 30 s; the current site uses 60 s for a cold Google provider, so pass `{ subject: "key", timeout: 60000 }` if that budget is still wanted (the refusal path no longer spends it either way).

The spec doc (`docs/core-functionality/model-provider/google-provider.md`) needs the matching correction: its validation criterion and its false-positive-guard section both describe the assertion as a 2xx check.

**Done when:** against the instance with a deliberately invalid `GOOGLE_API_KEY`, Test 1 fails in about a second with Google's own message in it; the clean run stays green and leaves the global credential intact (the product refuses to persist the bad key); the spec doc says what the assertion now does.

## Not in scope

- Whether a provider Test 1 should gate on provider health — settled per provider in #1415 / #1829, unchanged here.
- The product refusal itself, which is correct.

Victor-w-Madeira:
## Problem

`google-provider.spec.ts` Test 1 was the last provider save still carrying the original form, and the weaker of the two variants: it asserted the **status** behind a `Promise.all`.

```ts
const [validateResp, persistResp] = await Promise.all([validatePromise, persistPromise]);
expect(validateResp.ok()).toBe(true);
expect(persistResp.ok()).toBe(true);
```

Two defects, the pair #1829 fixed for Anthropic:

1. **The first assertion cannot fail on a rejected key.** `POST /api/v1/models/validate-provider` answers **200 in both directions** and puts the verdict in the body (`{"valid":false,"error":"Invalid API key for Google Generative AI"}`), so `validateResp.ok()` is true either way. The comment above it — *"the key authenticates against Google live"* — was not what the line checked.
2. **`Promise.all` settles only when both settle.** The frontend gates the write on that body (`useProviderConfiguration.ts` → `handleSaveAllVariables`: `const isValid = await validateCredentials(); if (!isValid …) return;`), so after a refusal **no** `POST|PATCH /api/v1/variables/` is issued at all — and the first thing to settle was the persistence waiter, rejecting at its 60 s timeout. The already-resolved verdict, with Google's own message in it, was discarded.

Not a Langflow defect: the product correctly refuses to persist a credential `validate-provider` rejected (settled in #1823).

## Fix

- **Read the Save through `armProviderSave`** (#1849), like the five sibling provider specs, instead of a sixth inline copy of the same rule: the `validate-provider` waiter is matched on the **pathname** (`includes()` is the trap `CLAUDE.md` records for `/build/`), the unread waiter is settled at arm time so a refusal cannot leave an unobserved rejection, and the contract is covered by `provider-panel-save.test.ts` — including *"validated() must not wait for the write waiter"*, which is the property this fix exists for.
- **Both waiters are still armed before the click**, so causality is unchanged; they are now read in the order the panel issues them.
- **The 60 s budget is kept explicitly** (`{ subject: "key", timeout: 60000 }`) — Google validation is slow on a cold provider — over the helper's 30 s default. `persistTimeout` defaults to `timeout`, so both waiters get 60 s, and because both are armed at the same instant the worst case stays ≈60 s rather than doubling. The refusal path no longer spends that budget either way.
- **The gate rationale is corrected, not changed.** Test 1 still gates on `hasProviderEnvKeys`, never `providerSkipGate` (#1415) — the issue puts that out of scope. What the comment now also states is the other half: a **drained/spend-capped** key still answers `{valid: true}` and genuinely exercises the save path, while a **rejected** key means the test cannot pass however healthy Langflow is, which is the test working.

## Measurement (nightly `1.13.0.dev12`, `--workers=1 --retries=0`, deliberately invalid `GOOGLE_API_KEY`)

```
before: failed in 1.0 m  TimeoutError: page.waitForResponse: Timeout 60000ms exceeded
                         while waiting for event "response"
after:  failed in 2.9 s  validate-provider rejected the key: Invalid API key for
                         Google Generative AI
```

The rejected run leaves the global credential intact — confirmed by a clean `2 passed` run immediately after it.

## Second commit — findings from the independent review

No behavioural defect was found in the test (the call site, both timeouts and the unobserved-promise paths were verified clean). Five factual corrections were:

- **`## External dependencies` did not declare `modals/modelProviderModal/hooks/useProviderConfiguration.ts`**, although `handleSaveAllVariables` is now the whole justification for the ordered read. That section is the only thing `--mode=check-docs` resolves on every PR, so the test's new correctness argument was the one part with no declared path behind it. The anthropic sibling added exactly this entry.
- **The path that *was* declared does not render what the prose claimed.** Grep on upstream `main`, `release-1.13.0` and `release-1.12.1`: `provider-variable-input-` and `provider-save-button` appear in exactly one file, `modals/modelProviderModal/components/ProviderConfigurationForm.tsx`, and `provider-item-` in `.../ProviderListItem.tsx`; `pages/SettingsPage/pages/ModelProvidersPage/index.tsx` only mounts them. The guard could never catch this — `pages/SettingsPage/` exists for other reasons and resolves green. All four paths verified against the three refs the guard uses.
- **"cost the 2026-09-11 daily 300 s of the collector's sweep budget"** attributed to this spec's `Promise.all` a cost it never caused. The collector is a separate instrument with its own ceiling (`CREDENTIAL_SAVE_TIMEOUT_MS = 240_000`) and its own fix (#1823's remedy (i)); nothing here recovers it. The accurate half — the unreviewed `@stable` removal on the Anthropic sibling, commit `883047fc` — is kept, with the collector named as the separate instrument it is.
- **"a drained key still answers `{valid: true}` and this test still passes"** was a derivation from the shared `validate_model_provider_key` path presented as a Google measurement. #1029 records only that `collect-models` marked google inactive on run 30374528125, never this test's outcome; the measured instance is the Anthropic sibling on the 2026-07-27 daily.
- **The doc's gate note still carried only the DRAINED half** while the spec comment had been rewritten to carry both, so under SDD the source of truth disagreed with the code about the same decision — on exactly the state a reader consults it for. And *"the general remedy is #976"* is false: #976 is closed **not planned**; the standing remedies are #1185 and #1187.

`anthropic-provider.md`'s *"the four sibling provider specs"* is a count this branch invalidated — six specs now read their Save through the helper — so it is corrected to five. Its spec comment is scoped differently (*"the four sibling provider specs #1849 measured the same shape on"*) and stays correct.

## Scope

Test 2 is untouched — no assertion, gate or helper of its changed. It was force-failed anyway, per the validation protocol.

Both of the issue's "not in scope" items are respected: the provider-health gate decision is unchanged, and the product refusal is correct and untouched.

**Deliberate narrower scope, flagged so it is not read as covered:** Test 1 still clicks `getByRole("button", { name: /Save|Replace/i }).first()` and does not call `awaitProviderPanelSettled` or assert the persist verb, the way `openai-provider.spec.ts` does for #1424/#1431. The "click before `GET /api/v1/variables/` resolves → CREATE on an existing name → `400 Variable name already exists`" race therefore still lands as a bare `expect(persistResp.ok()).toBe(true)` false. It matches the anthropic sibling and the issue's own fix shape.

## Validation (nightly `1.13.0.dev12`, `--workers=1 --retries=0`)

- typecheck ✅ · eslint ✅ (only pre-existing warnings) · `check:checklist-coverage` ✅ · `provider-panel-save.test.ts` 11/11 ✅
- clean runs ✅ — `2 passed` in 20.6 s / 22.0 s / 19.8 s / 19.4 s / 22.4 s
- zero `🚨 Backend Error` ✅
- no orphan flows left on the instance ✅ (28 present, all bootstrap templates; 0 created on the day of the runs)
- force-fail ✅ — one verified red per test in the file, each reverted (`git diff | grep -c FF_MUTATION` → 0, then a final clean run):
  - **Test 1, validation verdict** — invalid `GOOGLE_API_KEY` → failed in **2.9 s**, `validate-provider rejected the key: Invalid API key for Google Generative AI`. The same mutation on `main` failed in **1.0 m** with a bare `TimeoutError`.
  - **Test 1, persistence** — `page.route("**/api/v1/variables/**", abort)` → failed in 1.0 m with `no POST/PATCH /api/v1/variables/ followed the Save of the key within 60 s`, i.e. even the "the write never happened" path now names itself instead of arriving as a bare timeout.
  - **Test 2, model selection** — `/gemini|gemma/i` → `/FF_MUTATION_3_no_such_model/i` → failed in 21.0 s, `Received string: "gemini-flash-latest"`.
  - **Test 2, non-empty reply** — `toBeGreaterThan(0)` → `toBeGreaterThan(999999)` → failed in 14.0 s, `Received: 20`.

`QA-CHECKLIST.md` is untouched: both google bullets are already `[x]` and no tag changed, so no generated count moves.

Closes #1867


Victor-w-Madeira (commented):
## Review — approve

Small, correct and well-argued: the `Promise.all([validate, persist])` — which settles only when *both* do — is replaced by `armProviderSave`'s ordered read, the same helper the other five provider specs use. No behavioural defect found. `MERGEABLE`/`CLEAN`, 7/7 checks green.

**I verified the claims rather than accepting them:**

- **The frontend gate is real and is literally the one quoted.** `handleSaveAllVariables` on upstream `main` and `release-1.13.0`: `const isValid = await validateCredentials(); if (!isValid || !canUseCurrentProviderPolicy(providerName)) return;`. A rejected key therefore issues no `/variables/` request at all — the whole premise of the fix holds.
- **The `External dependencies` correction was right.** `provider-variable-input-` and `provider-save-button` are in `ProviderConfigurationForm.tsx`, `provider-item-` in `ProviderListItem.tsx`, on both refs; `pages/SettingsPage/pages/ModelProvidersPage/` only mounts them.
- **The counts check out.** Six specs read their Save through the helper on this branch (five on `main`), so `anthropic-provider.md`'s four → five is right — and the anthropic *spec* comment staying at "four" is also right, since it is scoped to what #1849 measured, which is exactly four (ollama, openai, openai-compatible, azure, per the helper's own docstring).
- `883047fc` is what the PR says it is; `provider-panel-save.test.ts` has 11 tests; `test:units` 1458/1458 on the branch, `tsc` clean, ESLint warnings pre-existing only.
- **The lane actually ran the spec**: `2 passed in 27.8 s`, **zero `🚨 Backend Error`**, and the #1456 coverage verdict reports *"no provider-health skip — all 2 executed test(s) produced a verdict"*. The validation is not local-only.

---

### The one finding, **fixed in `80580cb2`**: the 300 s correction was applied to google only

The second commit establishes that *"cost the 2026-09-11 daily 300 s of the collector's sweep budget"* charges the `Promise.all` shape for something it did not cause, and writes the corrected form into the google spec. The claim was still live in `anthropic-provider.spec.ts:246` and `anthropic-provider.md:296-297` — **a doc this PR edits** — so the two sibling specs told different stories about the same daily, with the accurate one in the newer file.

Two independent reasons it is wrong, and the second was not in the PR:

1. **Causality.** `reports/daily-history.jsonl` for run `34599745145` records **two** hard failures that day — Anthropic Test 1 *and* `collect providers status and models from UI` — with `883047fc` removing `@stable` from both. Their common cause is the dead key. The assertion shape cost the unreviewed tag removal; it cannot spend an instrument it does not run in.
2. **The figure is not that instrument's.** The collector's own ceiling is `CREDENTIAL_SAVE_TIMEOUT_MS = 240_000`. The 300 s in this family is gunicorn's `WORKER TIMEOUT` from run `30374528125` (#1029), a *different* run — as `google-provider.spec.ts:219` says three lines from the text that borrowed it.

Both places now name what the shape did cost, say what it did not, and point at the separate instrument and its own fix. No assertion, tag or helper touched.

### Nits — not fixed, your call

- **`"the five sibling provider specs"` is a count that has now gone stale twice** (it was "four" until this PR) and will go stale on the next adoption. `CLAUDE.md` writes the opposite convention for exactly this reason — *"a count is deliberately not stated — it went stale when `@serving` was added"*. `"every sibling provider spec"` costs nothing and stops requiring an edit in an unrelated PR.
- **The spec imports from `provider-panel-save.ts` while contradicting that module's stated locator rule.** Its `PROVIDER_SAVE_BUTTON` docstring says the control "is located by testid and **never** by role+name", and the click stays `getByRole("button", { name: /Save|Replace/i }).first()`. The #1424/#1431 gap is declared deliberately and I agree with that scoping — but switching *only the locator* is one line, brings none of the settle logic with it, and stops the file disagreeing with the helper it now depends on. The field is currently split: azure and openai use the testid, anthropic and ollama do not.
- **The persist refusal body is still discarded** — `expect(persistResp.ok()).toBe(true)` fails as a bare `false`, while the same module exports `readResponseBodySafely` and the openai sibling classifies the refusal. Declared out of scope, and #1424's `400 Variable name already exists` race lands exactly there. There is **no open issue** tracking adoption of the openai shape across the remaining provider specs; that deserves one rather than a paragraph.

### Out of scope, for a later sweep

`openai-provider.md:289` and `openai-provider.spec.ts:328` still cite **#976** as the remedy (*"funding the account is #976"*) — the same stale citation this PR corrects for google. Not this PR's job; noted so it is not lost.

### On the branch

70 commits behind `main`, but **nothing** merged since the branch point touches `model-provider` or `provider-setup` — so the clean merge is a fact, not just GitHub's arithmetic.

Issue #1862, opened 2026-09-15
Create `templates-registration.spec.ts` — the registered template set, against a committed baseline (§11.1)


rafaelgiln:
**Type:** Create new spec
**Spec file:** `tests/tests-automations/regression/core-functionality/templates/templates-registration.spec.ts`
**Spec doc:** `docs/core-functionality/templates/templates-registration.md`
**Checklist bullets:** `QA-CHECKLIST.md` §11.1 — *Registered set*
**Scoped by:** #1860 → `docs/core-functionality/templates/templates-coverage-scope.md` (planned inventory row **R1**)
**Milestone:** none yet. Templates is a decided pool item in `ROADMAP.md`; this is one of the three first-slice issues (R1, G1, S1) for a roadmap review to date.

This spec needs no key and no egress, and it only reads. It is the smallest of the three, and the one S1 builds on: S1 parametrizes over the baseline this spec commits.

## What the spec must prove

`GET /api/v1/flows/basic_examples/` is the listing the New Flow gallery reads. It must return exactly the template names a committed baseline expects:

- **An undeclared absence fails**, naming the missing template(s). Nothing in the suite gives this signal today. When the image stops registering a template, it leaves the gallery and nothing goes red; #1234 only surfaced because a Loop component spec uses that template as its fixture.
- **A declared absence is checked both ways.** *Research Translation Loop* ships upstream but is not registered, because `ArXivComponent` is not in the image (#1744). The baseline declares that absence with its reason. The day the template comes back, the spec fails, naming the declaration to delete and #1744 as the issue to close. This is #1084's rule: an exemption must not expire silently.
- **An extra template is reported, not failed.** A new upstream template costs nobody a test (#980's trade). It is reported as an annotation, and accepting it is a baseline refresh.

## Measured (nightly `1.13.0.dev12`, 2026-09-15)

- **Upstream vs listing:** upstream ships **27** JSONs under `src/backend/base/langflow/initial_setup/starter_projects/` on `main`, `release-1.13.0` and `release-1.12.1`. The listing returns **26**.
- **The one absence is logged at startup:** `Skipping starter project 'Research Translation Loop'; unavailable components: ArXivComponent`. It comes from `filter_starter_projects_by_available_components` in `src/backend/base/langflow/initial_setup/setup.py`.
[...]

Issue #1860, opened 2026-09-15
Scope: Templates (§11) — map the starter-project gallery and produce named checklist bullets


rafaelgiln:
Pool — **needs scoping first** (`ROADMAP.md` → *Pool — not yet dated*). **Scoping issue — no spec is written under this issue** (Rule 1 pointer; the same treatment #1195 gave A2A).

**Goal:** turn `QA-CHECKLIST.md` §11 (`### core-functionality/templates/`) into named `[ ]` bullets that match the templates the product actually ships, with a depth decision (structural smoke vs. execution) per template, so the area can be dated at a roadmap review.

## Why §11 cannot be scheduled as it stands

Measured on nightly `1.13.0.dev12` (2026-09-15):

- Upstream ships **27** starter projects under `src/backend/base/langflow/initial_setup/starter_projects/` — the same 27 on `main`, `release-1.13.0` and `release-1.12.1`. The image registers **26** (`GET /api/v1/flows/basic_examples/`): *Research Translation Loop* is skipped at startup because `ArXivComponent` is not shipped (`Skipping starter project 'Research Translation Loop'; unavailable components: ArXivComponent` — tracked by #1744).
- **12 of §11's 34 bullets name no template.** Nine names exist on none of the three refs: Invoice Summarizer, Youtube Analysis, Dynamic Agent, Hierarchical Agent, Pokedex Agent, News Aggregator, Prompt Chaining, Decision Flow, Similarity. Three are variants of an existing template, not templates: Basic Prompting (Anthropic), Simple Agent (Anthropic), Simple Agent with memory.
- A 13th bullet, *MCP Server (starter projects)*, is the MCP-server presets surface already recorded in §14.1 (`mcp/server/mcp-server-starter-projects.spec.ts`), not a flow template — it is counted twice.
- The remaining 21 bullets map to real templates; **6 shipped templates have no bullet at all**: Content Aggregator, Multi Agent Flow, Meeting Summary, Hybrid Search RAG, Knowledge Retrieval, Deep Research Agent.
- §11's subsections (Basic / Content Generation / Analysis and Processing / Agent / Advanced) are not the gallery's taxonomy. The templates modal groups by *Use Cases* (Assistants, Classification, Coding, Content Generation, Q&A) and *Methodology* (Prompting, RAG, Agents), and a template belongs to a category only when its `tags` contain that category's id exactly.
[...]

Issue #1764, opened 2026-09-08
The daily's shard matrix is listed without provider keys, so provider-invalid-auth-error.spec.ts never runs in any shard


Victor-w-Madeira:
`provider-invalid-auth-error.spec.ts` carries three `@stable` tests and the daily runs none of them. Not skipped, not red — absent. The report cannot show it, because a spec that never enters the shard matrix has no row to be absent from.

Found while comparing the Actions daily against the VM daily for 2026-09-08 (step 14 of the VM migration): the VM lane ran two of this spec's variants that morning and the Actions lane ran zero, and the reason turned out to belong to this lane.

## Mechanism

The `prep` job computes the shard matrix by listing the suite:

```yaml
- name: Compute duration-balanced shard matrix
  env:
    SHARDS: ${{ inputs.shards || '4' }}
  run: |
    npx playwright test --grep "@stable" --list --reporter=json > /tmp/stable-list.json
```

That step's `env:` block is `SHARDS` and nothing else — no provider keys. Every test in `provider-invalid-auth-error.spec.ts` is generated at collection time from the environment:

```ts
const targets = getProviderTargets();   // keyedProviders.filter(([p]) => hasProviderEnvKeys(p))
for (const { provider, ... } of targets) { test.describe.serial(...) }
```

`hasProviderEnvKeys` is `every(key => !!process.env[key])`. With no keys in the listing process, `targets` is empty, the file yields **zero** tests, and it is therefore not in the list `partition-shards.mjs` partitions. Sharding is file-level (#936) — `matrix.files` is an explicit spec-file list — so a file missing from the listing is passed to no shard and never runs. `--pass-with-no-tests` on the shard command means nothing complains.

Note the failure is confined to files whose *entire* content is key-gated. Other provider-parametrized specs (`agent-*`) have at least one `@stable` test that exists without keys, so their file survives the listing and their variants are generated at run time, with keys present. That is why exactly one file is lost, not a class of them.

## Evidence

**1. Reproduced locally**, same command as the step, CI=true, nothing else:

| listing env | files | `@stable` tests | `provider-invalid-auth-error` |
|---|---:|---:|---|
| no provider keys | 230 | 614 | absent |
| `OPENAI_API_KEY`/`ANTHROPIC_API_KEY`/`GOOGLE_API_KEY` set | 231 | 617 | 3 tests (openai, anthropic, google) |

The delta is exactly one file and exactly three tests.

[...]

PR #1872, opened 2026-09-16
test(templates): assert every registered template instantiates as itself (#1864)

### LINKED ISSUE #1864: Create `templates-instantiate.spec.ts` — every registered template instantiates as itself (§11.2)
**Type:** Create new spec
**Spec file:** `tests/tests-automations/regression/core-functionality/templates/templates-instantiate.spec.ts`
**Spec doc:** `docs/core-functionality/templates/templates-instantiate.md`
**Checklist bullets:** `QA-CHECKLIST.md` §11.2 — all 26 bullets, one per registered template. The two `[~]` there (Basic Prompting, Memory Chatbot) also point at this spec once it lands.
**Depends on: #1862** — this spec runs one test per template in the baseline #1862 commits.
**Scoped by:** #1860 → `docs/core-functionality/templates/templates-coverage-scope.md` (planned inventory row **S1**)
**Milestone:** none yet. Templates is a decided pool item in `ROADMAP.md`; this is one of the three first-slice issues (R1 #1862, G1 #1863, S1) for a roadmap review to date.

No key, no egress. It creates 26 flows and deletes each one by id.

## What the spec must prove

One test per registered template:

1. Pick the template's card from **All templates**.
2. The editor opens (`canvas_controls_dropdown`).
3. `GET /api/v1/flows/{id}` equals that template's entry in `GET /api/v1/flows/basic_examples/` on four things:
   - **component types**, as a multiset (`data.nodes[]` with `type == "genericNode"`, read from `data.type`)
   - **edge count**
   - **note count** (every other node)
   - **name**: the template's own, or the template's plus a ` (N)` suffix, which the backend adds when the name already exists

The expected entry is read from the listing at run time. If upstream edits a template, both sides change together and nothing fails. What fails is an instantiation path that changes the graph: a node type dropped or rewritten, an edge lost, a note discarded.

## Measured (nightly `1.13.0.dev12`, 2026-09-15)

- **Result:** all 26 registered templates were instantiated this way, read back and deleted. **26 of 26 matched** on component types, edges and notes. The editor opened on every one, and none showed an outdated-component or error indicator.
- **Per-template counts:** in the scope doc's *Instantiation* table.
- **Timing:** each pick took 7–10 s, including a 4 s settle; the first, on a cold page, took 14.5 s.

On this build the spec finds no defect. It is a regression detector, so it is only worth shipping with a force-fail that proves the comparison actually fails when the graph differs.

## Build notes

- **Path:** `loadTemplateByName(page, name)` (`tests/helpers/flows/load-template-by-name.ts`) is exactly this journey, hardened for concurrency in #1002:
  - it retries the creation POST on the upstream same-name 500
  - it recovers a lost navigation
  - it returns the created flow's id and deletes every other flow it created

  Its heading match does not use `exact`. On this build no template name is a substring of another (all 26 pairs checked), and asserting the persisted name stops a future collision from passing with the wrong template.
- **Where the test list comes from — this is the trap in this issue; decide it in PLAN.** The list must exist when Playwright collects the tests:
  - **R1's committed baseline (#1862) — recommended.** It is static and reviewable, it is the same set R1 enforces, and declared absences are skipped by construction. `resolveJsonModule` is on in `tsconfig.json`. No spec `import`s a committed JSON asset today, so this would be the first.
  - **Not the live listing at collection time.** Collection must not depend on the environment. A spec whose test list comes from runtime state can drop out of the daily's shard listing and run nowhere while the run stays green. That is #1764, which removed `provider-invalid-auth-error.spec.ts` from the lane entirely.
  - **Either way,** the per-template titles are built from a variable. The `@stable` listing detector reports those under `unresolvedTitles` (#1812). That is accepted for the provider-parametrized specs; say so in the spec doc.
- **Cleanup:** delete the id `loadTemplateByName` returns in `afterEach`, via `deleteFlow`. Show the user-flow count is the same before and after, on a green run and on a forced-red run.
- **Runtime:** 26 flows created through the UI. Measure the file's duration before deciding on `@stable`, because it lands on the daily's shards.
- **Tags** (per the scope doc): `@workspace` `@regression` `@templates`.

## Done when

- [ ] Spec doc `docs/core-functionality/templates/templates-instantiate.md` exists, with every mandatory section and `Last validated: 1.13.x`
- [ ] `templates-instantiate.spec.ts` has one test per registered template, asserting the equality above
- [ ] Force-fails observed red:
  - one edge removed from one template's expected entry fails exactly that template's test; the other 25 stay green
  - one component type swapped fails, naming the type
- [ ] Flow cleanup is proven: the user-flow count matches before and after, on a green run and on a forced-red run
- [ ] No `🚨 Backend Error` logged
- [ ] All 26 §11.2 bullets point at `core-functionality/templates/templates-instantiate.spec.ts`, each with its symbol; the two `[~]` are resolved

Refs: #1860, #1862, #1863, #1002, #1764, #1812

Victor-w-Madeira:
Closes #1864

It covers the **26 `[ ]`/`[~]` bullets** in `QA-CHECKLIST.md` §11.2 — row **S1** of the planned spec inventory in `docs/core-functionality/templates/templates-coverage-scope.md` (the #1860 scoping pass). Builds on **#1862** (merged): this spec parametrizes over the baseline that PR commits.

Picking a template's card must create a flow that **is** that template. One test per registered template — 26 on this image — comparing `GET /api/v1/flows/{id}` with the template's entry in `GET /api/v1/flows/basic_examples/`.

## 1. Covered tests

| 1–26 | `<Template> instantiates with the template's components, edges and notes` | Component-type **multiset** · **edge count** · **note count** · **per-node wiring** · persisted **name** (`T` or `T (N)`). `afterEach` deletes by id and the readback answers **404** |

The expected side is read at **run time**, so an upstream template edit moves both sides and nothing fails. What fails is an instantiation path that changes the graph.

## 2. How the test was built

**Where the test list comes from — #1864 calls this "the trap in this issue".** It must exist when Playwright **collects**, which rules out the live listing: a spec whose test list depends on runtime state can drop out of the daily's shard listing and run nowhere while the run stays green (#1764, which removed `provider-invalid-auth-error.spec.ts` from the lane entirely). So it is R1's committed baseline, read at module scope — static, reviewable, the same set R1 enforces, and skipping declared absences by construction (*Research Translation Loop* is not in `templates[]`).

Anything unusable **throws there** rather than yielding zero tests: zero tests means this file is *absent* from the listing, not red in it, and `--pass-with-no-tests` keeps the lane green. Verified by mutation — empty `templates`, missing `templates`, corrupt JSON, an entry missing `name` — each gives a visible collection error and `exit 1`, including under `--pass-with-no-tests` and in a whole-suite `--list --grep @stable`.

**Two measurements decided the comparison's spelling.** Notes are counted as the complement of `genericNode`, not as `type === "noteNode"`, because *Basic Prompting*'s two notes are **not uniformly shaped** — one carries `data.type: "note"`, the other carries no `data.type` at all, so a shape-keyed rule counts 1 of 2. And component types are a **multiset**: *Multi Agent Flow* has three Agents, so a set comparison passes with two of three silently dropped.

**The wiring is compared, and that is not in #1864's list.** Without it the spec goes green while broken: only `edges.length` was compared, so an edge repointed onto a different node of the **same type** left every count identical — and **7 of the 26** templates repeat a component type. The instantiation path is `updateIds` (`reactflowUtils.ts`), which rewrites every node id and repoints every edge through the id map, so this is a live regression shape. Measured before adopting it:

| | catches a same-type rewire |
|---|---|
| multiset of `sourceType → targetType` pairs | **0 of 7** |
| per-node neighbourhood signature (adopted) | **6 of 7** |

and expected == actual on **26 of 26** templates, so there is no divergence to tolerate. The residual is *Deep Research Agent*, where the rewire lands between two nodes whose one-hop neighbourhoods coincide — named in the spec doc rather than hidden.

**What it deliberately does not compare, now declared.** Node ids (rewritten by `updateIds`, so comparing them fails on every healthy run) and positions. And **component parameter values**: measured **1398 fields across the 26, zero differing**, so it *is* assertable here — declined because `use-add-flow.ts` resolves the project's global variables and passes `unavailableFields` into `updateGroupRecursion`, making values **instance-dependent by design**; an account whose variables differ would go red on a legitimate state. The regression it would catch is real and guarded against upstream, so it is named as deserving its own issue. The §11.2 bullets say **"structure, not parameter values"**.

**Layering.** The comparison is pure and unit-tested in `helpers/flows/template-graph-shape.ts`, not inline — the lesson #1862's review recorded, where decision logic in a spec was logic no unit test covered. `nameMatchesTemplate` compares the stem **literally, never as a regex source**, because template names carry metacharacters.

**Cleanup** is id-scoped (never a wipe, #553), lives in `afterEach` rather than an in-body `finally` — Playwright gives `afterEach` its own budget and runs it after a test **times out**, which is when a leak compounds — and a cleanup failure is surfaced as a warning that **never replaces the product failure**.

## 3. Dependencies

**None** — no provider key, no network egress, no LLM. Creates 26 flows through the UI and deletes each by id. Parallel-safe; not `@destructive`.

## Validation (nightly `1.13.0.dev12`, `--retries=0`)

- `npm run typecheck` ✅ · `npm run lint` ✅ (**0 problems** on the files in scope)
- `npm run test:units` **1490/1490** ✅ (14 new) · `npm run test:scripts` **1648/1648** ✅
- QA-CHECKLIST guard ✅ · `npm run check:checklist-coverage` ✅
- `workers=1` (the daily): **26/26 in 80.5–82.1 s**, ×2 · `workers=2` (this lane): **26/26 in 63.5 s**
- Flow count **27 before, 27 after** ✅ · zero `🚨 Backend Error` ✅

**`@stable` is carried on a measurement, which #1864 asks for**, since the file creates 26 flows and lands on the daily's duration-balanced shards: `workers=1` 1.3–1.4 min over 3 runs with 0 failures; `workers=2` 1.0–1.2 min over 4 runs with 0 failures; per-pick median 3.0 s. A contention flake in the **shared** `open-new-flow-templates-modal` helper appears at `workers=5` — a parallelism no lane uses — in 2 of 4 runs; it is recorded in the spec doc with its conditions and a **filing trigger** rather than filed, because there is no observation under either lane (0 in 7) and what was measured is this file's own concentrated use, not the helper across its 21 callers.

**Force-fails, all executed:**

| Mutation | Observed |
|---|---|
| One edge removed from one template's expected entry | 🔴 **1 failed, 25 passed** — the scoping #1864 requires |
| A component type swapped | 🔴 naming both halves |
| Note count broken | 🔴 |
| One of *Multi Agent Flow*'s three Agents dropped | 🔴 `the template has 2, the created flow has 3` — the case a **set** comparison would miss |
| **An edge moved between two Agents** | 🔴 **on the wiring alone, every count unchanged** |
| Name asserted against the wrong template | 🔴 (and confirmed the real name is `Basic Prompting (1)`, so the ` (N)` rule is exercised) |
| The `afterEach` delete reverted | 🔴 on the 404 readback, naming the surviving flow |
| Comparison **and** cleanup failing together | The product failure is still reported; the cleanup problem is a warning beside it |

Reverted and verified: 0 mutation markers, spec byte-identical to the good copy, baseline untouched.


Victor-w-Madeira (commented):
## Review — approve once CI re-runs green

Solid work, and the parts that are hard to get right are right: the comparison is pure and unit-tested rather than inline, `wiring` closes a real green-while-broken hole (an edge repointed onto a *different node of the same type*), the test list is read at module scope and **throws** instead of collecting zero (#1764), and cleanup is id-scoped. I re-ran everything rather than taking the numbers on trust: `tsc --noEmit` clean, `test:units` **1490/1490**, ESLint 0, both QA-CHECKLIST guards green, and the lane ran the spec **26/26 in 1.6 min**. Nothing in the PR body is inflated.

Three findings, **pushed as 160f6209** so they can be read as a diff rather than as a list of requests. Everything below is verified, and one of my own hypotheses was refuted by measurement — that part is in the commit too.

---

### 1. The run of this PR logged **57 `🚨 Backend Error`**, against a PR body that says zero

Job `Run impacted E2E specs` of run 35049167652 — 57 lines over 17 distinct flows, every one flow-scoped:

```
15 /api/v1/models?flow_id=…        15 /api/v1/custom_component/update?flow_id=…
10 /api/v1/models/enabled_models   7 /api/v1/flows/{id}/events?since=…
 4 /api/v1/flows/{id}              3 /api/v1/variables/?flow_id=…   3 …/note_translations
```

All 404s for flows the `afterEach` had just deleted: the hook deletes over a **still-mounted editor** (the `page` fixture closes *after* the hook), so the SPA keeps asking for a flow that is gone. `api/flows/api-component-regression.spec.ts:148` and the folder specs already answer this with `await page.goto("about:blank")` before the delete (#1023/#1103), with the rationale written out — *"the fixture logs each one as `🚨 Backend Error`, and the deterministic pipeline's VALIDATE gate hard-stops on those"*. The hook here did not even take `page`.

It fails no test (#1084), which is exactly the cost: step 4 of the validation checklist and #1864's *Done when* both call it a gate, the daily's log is read by a human, and this file would add ~57 lines per run to it.

**My own claim about the fix did not survive being measured, and that is in the commit.** I ran the spec against a live `1.13.0.dev12` at `workers=2`, **with and without** the navigation: 26/26 green both times, **0 of that 404 class either way** (3 unrelated 400s on a shared dev instance, identical in both runs). So the CI figure is the observation, the `about:blank` is the suite's convention applied to it, and the *confirmation* is the next lane run — the spec doc now records this criterion as **UNVERIFIED** rather than met, and the comment says the local pair cannot settle it. Same shape as the honest-scope paragraph `api-component-regression.spec.ts` writes about itself.

Fair caveat: **none** of the 15 `loadTemplateByName` callers applies the convention today, so the class is pre-existing — this file is just the largest single concentration of it, and the only one whose PR claims zero.

### 2. Reuse — `describeBaselineDefect` exists, is unit-tested, and names S1 as its consumer

`registered-templates-drift.ts:204` already validates this exact file, and the docstring right beside it says *"S1 (#1864) parametrizes over this module"*. The inline `!entry?.nameKey || !entry?.name` is weaker in three ways that all fail **silently at collection scope**:

- a non-string or whitespace `nameKey` matches no `name_key` in the listing → the test fails with *"this image does not register it"*, i.e. a **baseline** defect reported as a **registration** defect, in R1's area;
- a whitespace `name` → an unusable card locator;
- a **duplicate** `nameKey` → two generated tests with the same title.

Now delegated. Verified at collection, not asserted: each of the three aborts with a named error and **exit 1** including under `--pass-with-no-tests`, and the good baseline still collects 26. `BaselineTemplate` is imported instead of redeclared. I left `BASELINE_PATH` duplicated on purpose — sharing it means editing the merged R1 spec, which is scope this PR should not take.

### 3. The spec doc named the wrong lane condition, and it is the line the `@stable` case rests on

The table read `workers=1 | the daily (PW_SHARD_FILE_LEVEL=1 ⇒ fullyParallel: false)`. But `playwright.config.ts:111` is `workers: SERIAL_LANE ? 1 : process.env.CI ? 2 : undefined`, and `PW_SHARD_FILE_LEVEL` only flips `fullyParallel` to false — **the daily runs two workers**: this file's 26 tests serialize in one of them while a *different* spec file runs concurrently in the other, against the same backend. The `workers=2` row is not that condition either (it overlaps this file with itself, the heavier contention for the shared modal helper).

The conclusion survives — the daily's real condition sits between two measured rows and both were green, and the lane itself ran it 26/26 — but the mapping had to be corrected, because the contention-flake paragraph's *"a parallelism no lane uses"* argument leans on it.

---

### Smaller, in the same commit

- **`getAuthToken` sat outside the cleanup `try`.** It retries a wedged backend for ~30 s and then rethrows (#1086), so a throw there was a cleanup failure replacing the product failure — by the one path this hook was rewritten to close. Moved inside.
- **Orphan `{ }` block** left by the removed `try/finally`, with the whole body indented one level for it. Gone.
- **The `afterEach`-over-`finally` rationale claims more than it buys**, and the comment now says so: the long pole is `loadTemplateByName` (two 30 s gates) and it runs **before** `createdFlowId` is assigned, so in the window where a timeout is actually likely this hook has no id to delete — that window is covered by the helper's own cleanup. The load-bearing gain is the un-masking, which is real.
- **A created flow with zero components was reported as an unreadable body.** On the *expected* side a zero reading is a parse failure (no registered template is empty); on the *actual* side it is the empty-canvas defect — the severest thing this spec can catch. Still red either way, but it was named wrong. Both readings are in the message now, with the body.
- **Doc**: *"on four things"* over a five-row table, and the step-by-step still comparing three (pre-wiring leftovers).

### Left alone, deliberately

- `expect(id).toBeTruthy()` after `loadTemplateByName` is dead against today's contract (the helper throws, and is typed `Promise<string>`) — **kept**, because it is what would catch that contract changing to return `""`, which `afterEach` would then treat as "nothing to clean up".
- `GET /api/v1/flows/basic_examples/` is fetched **26×** (the full graph of every template each time). A `beforeAll` would pay it once and still be a run-time read, but it moves a listing failure from one test's attribution to the whole worker's. Fine as is; worth knowing it is a choice.
- `templates-coverage-scope.md`'s **S1** row still describes the criterion without `wiring`. It is the plan of record, so I did not rewrite it here.

Victor-w-Madeira:
### Follow-up: the lane settled finding 1 — **57 → 0**

The re-run after `160f6209` closes the one thing the local pair could not decide.

| | `🚨 Backend Error` | Result |
|---|---|---|
| Before (`08151942`, run [35049167652](https://github.com/oriontech-me/langflow-e2e/actions/runs/35049167652)) | **57** over 17 flows | 26 passed, 1.6 min |
| After (`160f6209`, run [35166215203](https://github.com/oriontech-me/langflow-e2e/actions/runs/35166215203)) | **0** | 26 passed, 1.5 min |

Same job, same 26 specs, all 7 checks green. So the teardown 404s were real, were the `afterEach` deleting under a mounted editor, and `about:blank` removes them — the criterion is measured now, not expected, and `f100f135` rewrites the spec comment and the doc accordingly.

One thing worth keeping out of that diff and in view: **a local run cannot check this criterion.** With *and* without the navigation, 26/26 against the same image at `workers=2` logged 0 of that class either way — a green laptop run looks exactly like a fix and proves nothing here. Both figures and that caveat are now in *Build notes*, so the next person validating this file on a dev box knows the lane is the only place the question is answerable.

The other two findings stand as reviewed. From my side this is ready to merge.

Issue #1916, opened 2026-09-18
composio.spec.ts is parked: ComposIO is not in the tested image (lfx-bundles not installed) and the surface is out of scope


Victor-w-Madeira:
> Park owner for `core-functionality/llm-agents/composio.spec.ts`, filed out of #1913 (Wave 9, T2). The spec doc is `docs/core-functionality/llm-agents/composio.md`; the measured row is in `docs/triage/inherited-spec-triage.md`. The standing policy is `docs/component-distribution-policy.md` (#1039 / #1040).

`composio.spec.ts` carries an `isProviderComponentAvailable()` gate and skips with an attributed reason on every run — the treatment `docs/component-distribution-policy.md` prescribes, and the one `groq-provider` / `mistral-provider` already carry (#1039). Three facts, each measured on the running nightly `1.13.0.dev15`, and any one of them is decisive.

**1 — The component is not in the catalog.** `GET /api/v1/all` returns **29 categories / 183 component types** (the 30th top-level key is `component_display_names`, a metadata map and not a category — counting it is what doubles the type total, the trap `CLAUDE.md` documents by name); a case-insensitive search for `composio` over every `category/type` pair returns **0 hits**. The only Gmail components present are Google's own (`ext:google:GmailLoaderComponent@official`, `ext:google:GmailSendComponent@official`) — a different component with a different testid, so `getByTestId("composioGmail")` can never resolve.

**2 — The flavour of absence is `lfx-bundles-shim`, so it is packaging, not a rename.** `src/lfx/src/lfx/components/composio/__init__.py` carries the `# lfx-bundles-shim` header, and in the container:

```
>>> import lfx_bundles
ModuleNotFoundError: No module named 'lfx_bundles'
```

Installed distributions: `lfx`, `lfx-amazon`, `lfx-anthropic`, `lfx-azure`, `lfx-cohere`, `lfx-datastax`, `lfx-docling`, `lfx-google`, `lfx-ibm`, `lfx-ollama`, `lfx-openai`, `lfx-openai-compatible`, `lfx-oracle`, `lfx-toolguard`, `lfx-vllm`. No `lfx-bundles`. This is the case `docs/component-distribution-policy.md` is the standing answer for — a decision per image, not a tracker item. (It also answers, for ComposIO specifically, the *"which flavour of absence"* question #1912 raises for its own cluster.)

[...]

Issue #1915, opened 2026-09-18
voice-assistant.spec.ts is parked: the playground's voice button is absent from the shipped bundle, replaced by an untested audio-button


Victor-w-Madeira:
> Park owner for `ui-ux/voice-assistant.spec.ts`, filed out of #1913 (Wave 9, T2). The spec doc is `docs/ui-ux/voice-assistant.md`; the measured rows are in `docs/triage/inherited-spec-triage.md`.

All three tests in `ui-ux/voice-assistant.spec.ts` are `test.fixme` and run in no lane, because the surface they assert is not in the image this suite tests.

## What was measured, on the running nightly `1.13.0.dev15`

Two independent readings, because one of them alone is easy to misread as a mocking problem.

**The served bundle.** `index.html` references a single JS asset and no lazy chunks. Grepped for the literals the spec targets: `voice-button` **0**, `voice_mode_available` **0**, `voice-assistant` **0**. In the same file `playground-btn-flow-io` is present and `data-testid` occurs 649 times, so testids are not stripped and the absence is real.

**The live DOM.** With `voice_mode_available: true` mocked and the route confirmed to fire (logged, once), the open playground reports:

| testid | count |
|---|---|
| `voice-button` | 0 |
| `voice-assistant-container` | 0 |
| `audio-button` | **1** |
| `input-wrapper` | 1 |

## The cause is a component swap, not a feature flag

`ENABLE_VOICE_ASSISTANT` is still `true` in `src/frontend/src/customization/feature-flags.ts`, and the **old** chat input (`src/frontend/src/modals/IOModal/components/chatView/chatInput/components/input-wrapper.tsx`) still gates a `VoiceButton` on `ENABLE_VOICE_ASSISTANT && config?.voice_mode_available`.

The shipped playground renders the **new** chat input (`src/frontend/src/components/core/playgroundComponent/chat-view/chat-input/components/input-wrapper.tsx`), whose button row holds an `AudioButton` (`data-testid="audio-button"`, browser speech-to-text) and no voice button at all. The voice store, the i18n strings (`voice.*`, `voiceAssistant.*`) and `src/backend/base/langflow/api/v1/voice_mode.py` all still ship — only the way in is gone.

## The `3/3 green` row is the finding, not a promotion candidate

[...]

Issue #1784, opened 2026-09-09
Inherited spec triage: land the measured verdicts for the 92 never-validated tests


rafaelgiln:
Task 7 of `docs/triage/inherited-spec-triage-plan.md` — the measurement the instrument was built to make honest — has run. This issue lands its output, `docs/triage/inherited-spec-triage.md`.

It is a separate issue because the runbook's Step 6 tells you to commit the table on issue #1769's branch, and that premise expired: #1769 merged (PR #1780/#1782) before the measurement ran, so there is no open branch for the table to land on. Same reason #1781 existed.

## What was dispatched

Nine `manual.yml` dispatches on the throwaway branch `measure/inherited-triage` (forked from `main`, **not** from #1769's branch — `main` already carries the run-e2e JSON reporter, the frozen baseline and both npm scripts). Three passes over three shards, so **every** test has three observations, green ones included: `retries=0` throughout, because a retry inside one dispatch hides exactly the intermittence the passes exist to measure.

The 7 quarantined declarations inside the backlog were unmuted on that branch — a `test.fixme` records `0/N` and reads as clean — and the branch was deleted afterwards. Their rows carry the `Quarantine` marker, read from `main`'s committed baseline.

Pre-flight, all green before a runner was spent:

- `npm run triage:baseline -- --check` → `in sync: 55 spec(s), 92 test(s)`
- `npm run triage:verify` → `92 wanted / 92 selected / 0 missing / 0 extra; shards 31/31/30`
- each shard fragment passes **verbatim** through `build-grep-filter.mjs` (single fragment, no lookahead composition), verified for all three

## The result

92 rows, 0 warnings, 0 unmatched titles.

| Verdict | Tests |
|---|---|
| green | **59** |
| hard-failure | 19 |
| flaky | 9 |
| skipped | 5 |

At spec granularity: **29 of 55 specs are 100% green**, covering 48 tests; 6 specs are partial (11 green of 20); 20 specs have zero green (24 tests).

Every non-parametrized green is exactly `3/3` — no `2/3` reached a green verdict.

By tier: T1 is the healthier half (30 green / 5 flaky / 3 hard-failure / 3 skipped of 41), T2 the debt (29 green / 4 flaky / 16 hard-failure / 2 skipped of 51).

## Four things worth reading before the batch issues are filed

[...]

Issue #1746, opened 2026-09-07
Reconcile @stable removals against their trackers mechanically — three hand audits in, the check still does not exist


Victor-w-Madeira:
Split out of #1504's **Detection** section, which says in as many words that it may be its own issue and that #1504 should not wait on it. #1504 is now closed by PR #1745; this is the half that would stop the class from recurring.

## The gap

`@stable` removal is **automatic** — `daily-stable.yml` strips the tag on a hard failure and commits straight to `main` — while restoration is **manual** and lives as a checkbox on a dedicated issue. Nothing reconciles *"tags currently missing"* against *"open issues that own a restore"*. So an issue closes with that checkbox unticked and the test silently stays out of the daily, forever, with nobody holding it.

**This is the third recorded instance of the same class**, which is the argument for building the check rather than doing the audit again:

| Audit | Found | Outcome |
|---|---|---|
| #974 (closed 2026-07-27) | 8 orphaned tests | reconciled by hand; the check was not built |
| #1451 (open) | names the absence for *prose* gates | still open |
| #1504 (closed by #1745) | 5 orphaned tests | reconciled by hand; the check was not built |

Between #974 and #1504 the mechanism produced five more. What caught #1460 — a **fourth** case, closed as `COMPLETED` with zero closing PRs and the tag never restored — was [a human happening to read the call sites](https://github.com/oriontech-me/langflow-e2e/issues/1504#issuecomment-5361591202) while working an unrelated PR. That is the thing that does not scale.

## Relationship to #1451 — adjacent, not the same

#1451 is about **prose**: a `[-]` bullet or a doc note saying *"gated on #N"* whose `#N` has closed. Its input is `QA-CHECKLIST.md` and `docs/**` text; its output is "this justification expired".

This is about **state**: a `test()` whose `tag:` array no longer contains `@stable`, with no open issue that owns putting it back. Its input is the spec ASTs, the git history and the GitHub issue state; its output is "this test runs nowhere and nobody owns it".

They share a thesis and nothing else. Build them separately, cross-link the reports; a single guard that tried to do both would be answering two different questions with one heuristic.

## The mechanism, as #1504 executed it by hand

Reproducible, and worth encoding rather than re-deriving:

1. Parse every `test()` / `test.fixme()` under `tests/tests-automations/` for its `tag:` array. Keep the ones **without** `@stable`.
[...]

PR #1917, opened 2026-09-18
test(triage): resolve the T2 no-verdict cluster — one deleted, two parked (#1913)

### LINKED ISSUE #1913: T2: 3 specs that produced no verdict at all — two skipped every run, and a test.skip muting a test measured 3/3 green
> Filed off `docs/triage/inherited-spec-triage.md`, the measured verdict table landed by #1784 (PR #1785). **Tier T2**, which is Wave 9's scope. The population and the guard that reports it are #1769 / #1770; the standing report is #1900.

**What T2 means, and why it is not Wave 8 with different files.** T1 was "has a doc *or* id-scoped cleanup". T2 is the complement: **neither**. Every spec in this batch is missing **two of PROMOTE's four conditions** (design §3) before anything is measured — a doc with the four mandatory sections, and cleanup that deletes only the flow ids it created. So a `3/3 green` row here is *one* condition met, not three: it says the spec works, not that it is promotable. That is the whole reason design §6 sizes T2 as a full wave rather than a few more batches.

Verified when this was filed: all 28 T2 specs carry `hasMirroredDoc: false` and `hasIdScopedCleanup: false` in `tests/assets/triage/inherited-backlog-baseline.json`, and a source spot-check on three of the green ones found no `afterEach` / `deleteFlow` / `cleanAllFlows` at all. Note the flag is a **source grep**, not the audit: a spec that creates no flow satisfies the cleanup condition vacuously, and the flow-list diff below is what decides that — not the flag.

## This cluster

Three specs, five tests, and one thing in common: **not one of them produced a verdict about Langflow.** Two were `skipped in 3/3 run(s)` and three are `test.skip` declarations on `main`. This is #1568/#1569's class — a test that never runs inflates nothing and validates nothing, and it is the state #1010 and #1012 exist to make visible rather than comfortable.

They are grouped because the *decision* is the same even though the mechanism differs, and the decision is cheap: a test that has run zero times in the measurement has no evidence for promote and no reproducible failure for park. Either it is made to run, or it is deleted, or it is parked with a filed reason — and "leave the skip in place" is not one of the three.

| Spec | Mechanism | What must be established |
|---|---|---|
| `composio.spec.ts` | skipped 3/3 at run time | why the gate never opened. If it is the ComposIO component's absence from the image, this belongs with the packaging cluster and is parked under #1040's standing answer, not fixed |
| `refresh-dropdown-list.spec.ts` | skipped 3/3 at run time | same question, different subject — the skip's own reason string is the starting point, and if there is none, that is the first finding |
| `voice-assistant.spec.ts` | **3 × `test.skip` declaration** on `main` | the middle test measured **3/3 green** with the modifier lifted on the throwaway branch, so it works. The other two are 0/3 on `expect(locator).toBeVisible()`. A declaration muting a working test is its own finding |

**The `voice-assistant` row is the one to read carefully.** Its `⏭️`/`✅` split proves the measurement really did unmute the declarations (design §2), so "3/3 green" there is a fact about the test, not about the skip. Promoting one of three tests in a file while parking the other two is a legitimate outcome and is preferable to leaving a working test muted because its neighbours fail.

All three are in the 7 disabled declarations that the pool's *Disabled-test triage* entry hands to this backlog rather than running in parallel.

## The measured rows

| | Spec | Test | Verdict | Quarantine | Backend errors | Failure signature |
|---|---|---|---|---|---|---|
| ⏭️ | `core-functionality/llm-agents/composio.spec.ts` | user should be able to interact with composio component | skipped in 3/3 run(s) |  |  |  |
| ⏭️ | `ui-ux/refresh-dropdown-list.spec.ts` | refresh dropdown list | skipped in 3/3 run(s) |  |  |  |
| ❌ | `ui-ux/voice-assistant.spec.ts` | should able to see and interact with voice assistant | 0/3 green | `test.skip` |  | Error: expect(locator).toBeVisible() failed |
| ✅ | `ui-ux/voice-assistant.spec.ts` | user should not be able to see voice button if voice mode is not available | 3/3 green | `test.skip` |  |  |
| ❌ | `ui-ux/voice-assistant.spec.ts` | user should be able to see voice button if voice mode is available | 0/3 green | `test.skip` |  | Error: expect(locator).toBeVisible() failed |

## The three outcomes (design §3), applied per spec

Evaluated **in this order**, and there is no fourth. A spec fitting none is `UNKNOWN` in the PR with the reason named, never omitted (#1012).

1. **DELETE** — only with a named replacing `@stable` spec **and** the specific test inside it. Without that named replacement the review rejects it. Evaluated first, because measuring and hardening a duplicate is exactly the spend this triage exists to avoid.
2. **PROMOTE** — all four conditions, no exceptions: 3/3 green, id-scoped cleanup, a doc with the four mandatory sections, and a force-fail run. For this tier that means **writing the doc and adding the cleanup is part of the work**, not a precondition someone else met. Green alone never promotes — that is how Wave 4 imported invisible reds.
3. **PARK** — a reproducible hard failure whose cause is the product: an issue filed (here or Jira `LE-####`), `test.fixme` carrying the link, and the reason written into the doc. Out of `@stable` **with an owner** is a legitimate terminal state.

## Definition of done (design §4)

- [ ] every spec here reaches **one** of promote / delete / park
- [ ] its `QA-CHECKLIST.md` Part II bullet is updated — **manual bullets only**, never the generated blocks (#741)
- [ ] a doc with the four mandatory sections exists for every promotion
- [ ] the PR names, **per spec**, the outcome and the `docs/triage/inherited-spec-triage.md` row it came from
- [ ] nothing here re-measures: the evidence ships in this body (§4 forbids re-measuring after the table lands)

## The two facts the table cannot render (design §4 → this body)

Both are read by hand during the work and recorded in the PR, never in the generated file:

- [ ] **does the spec leak a flow?** — read by diffing the instance's flow list around a run, not from any report. This is PROMOTE's cleanup audit, and for this tier it is the *acceptance test* for cleanup that did not exist before.
- [ ] **the duplicate candidate** — DELETE's defining evidence: the replacing spec and the specific test inside it, named, or delete is not an available outcome.

## Force-fail, and one lesson from #1906

A force-fail that mutates an assertion's own expected value reddens it **by construction** and proves nothing. Mutate what the product does and let the assertion catch it. #1906 shipped the tautological version first and had to redo it.

Victor-w-Madeira:
**Closes #1913.**

Wave 9's T2 cluster of three specs that produced **no verdict about Langflow** in the #1784 measurement — two skipped in 3/3 runs, one file of three `test.skip` declarations with the middle one measured `3/3 green`.

| Spec | Outcome | Triage row |
|---|---|---|
| `ui-ux/refresh-dropdown-list.spec.ts` | **DELETE** | 78 |
| `ui-ux/voice-assistant.spec.ts` (3 tests) | **PARK** — #1915 | 90–92 |
| `core-functionality/llm-agents/composio.spec.ts` | **PARK** — #1916 | 24 |

**Nothing here is promotable, and that is the finding rather than an omission.** Verdict is `test-defect` on the deleted row and *product / packaging* on the two parks, so **no `REGRESSIONS.md` entry and no upstream ticket is owed**.

## Problem

**The issue's own reading of the `voice-assistant` row is refuted, and that is the load-bearing part of this PR.** #1913 reads the measurement's `⏭️`/`✅` split as proof that *user should not be able to see voice button if voice mode is not available* works and could be promoted while its siblings are parked. Measured on `1.13.0.dev15`, the surface is **not in the image at all**: the served bundle (`/assets/index-o-RMEvH1.js`, single asset, no lazy chunks) carries `voice-button` **0**, `voice_mode_available` **0**, `voice-assistant` **0**, while `data-testid` occurs 649 times and `playground-btn-flow-io` is present — so testids are not stripped and the absence is real. Live, with `voice_mode_available: true` mocked and the route confirmed to fire once, the open playground reports `voice-button` **0** / `audio-button` **1** / `input-wrapper` **1**.

The cause is a component swap, not a feature flag. `ENABLE_VOICE_ASSISTANT` is still `true` and the **old** chat input still gates a `VoiceButton` on it; the shipped playground renders `components/core/playgroundComponent/chat-view/chat-input`, whose button row holds an `AudioButton` (browser speech-to-text) and no voice button. The voice store, the `voice.*` / `voiceAssistant.*` strings and `api/v1/voice_mode.py` all still ship — only the way in is gone.

So the green row asserts `not.toBeVisible()` on an element that cannot exist under **any** config value. Both mutations were **run, not argued**: with the mock inverted to `true` it passes (5.2 s), and with the route mock deleted outright it passes (5.0 s). Promoting it would have put a test that cannot fail into the daily.

**`refresh-dropdown-list` never asserted anything.** Its closing line was `getByText("Loading Options").isVisible()` — a boolean nobody reads. Proven rather than inferred: with the string replaced by one that can never exist it still passed, 11.8 s. Its `skipped in 3/3` verdict was **a drained Anthropic key at measurement time**, not a product fact — run `34402231653`'s log has `collect-models` recording `anthropic: candidate "claude-haiku-4-5" failed — Your credit balance is too low`, so `providerSkipGate("anthropic")` fired. Anthropic is healthy today and the spec runs green in 11 s.

**`composio` cannot run in any lane, for three independent reasons.** `GET /api/v1/all` returns **29 categories / 183 component types** with **zero** `composio` hits (the 30th top-level key is `component_display_names`, a metadata map and not a category — counting it is what doubles the type total). `src/lfx/src/lfx/components/composio/__init__.py` is an `lfx-bundles-shim` and `import lfx_bundles` raises `ModuleNotFoundError` in the container, so this is #1040 packaging and not drift. And `COMPOSIO_API_KEY` is **not a repository secret** and appears in no workflow, so the spec's own gate has never opened once since it was imported on 2026-03-11.

## Fix

- **`refresh-dropdown-list` — deleted**, with the replacement named per design §3:

| Deleted test | Why it does not hold | Replacing `@stable` test |
|---|---|---|
| refresh dropdown list | no assertion at all — the closing `isVisible()` result is discarded, and the body never clicks a refresh row (its only dropdown activity is `initialGPTsetup` → `setupOpenAI`) | `flow-functionality/run-flow.spec.ts` → *user should be able to use Run Flow without any issues* |

  You cannot lose a failure condition that was never asserted, and the replacement is strictly stronger. **Both of its halves were force-failed rather than assumed:** removing the refresh click alone times out `waitForSelector("text=Loading")` at 30 s, and stripping the target flow from the `POST /api/v1/custom_component/update` response the option list is built from fails the option selection at 20 s. The second mutation also fails **with the refresh click removed**, so the option delivery is that endpoint's rather than the click's — disclosed in the checklist bullet instead of glossed.

- **`voice-assistant` — parked**, all three tests `test.fixme` carrying #1915, with a new doc recording the measurement. The TODO it replaces (*"Need to review the voice assistant vs text to voice"*) is answered: the product chose text-to-voice.

- **`composio` — parked as gate-and-skip, not `test.fixme`.** The standing policy this cites answers a distribution the tested image does not install with *"Gate and skip, with an attributed reason. Do not delete the spec, do not leave it failing"* — `isProviderComponentAvailable()` before the first UI step, the groq/mistral treatment. That gate **self-heals** the day the image installs `lfx-bundles`; a `test.fixme` is inert until a human edits it. Force-failed behaviourally: with a dummy `COMPOSIO_API_KEY` and the gate in place the spec skips naming the distribution; with the gate removed it hard-fails on `locator.hover: Timeout 20000ms — waiting for getByTestId('composioGmail')`, the deep timeout #1039 exists to remove. The probe runs **before** the credential gate so the reason names packaging, not a missing key. `docs/component-distribution-policy.md`'s (b) inventory goes 3 → 4.

**Why the two parks are owned by open issues rather than by `scripts/lib/stable-orphan-exemptions.json`.** #1746's reconciler verifies every declaration in both directions and reports `outcome.kind === "never"` as **expired** — *"this test never carried `@stable` in the walked history, so there is no removal for the declaration to justify"*. Neither spec ever carried the tag (`git log -S`), unlike groq/mistral which were demoted from it. A declaration would be reported stale on the next Monday run: the silent-expiry problem both guards exist to close, arriving through the front door. #1915 and #1916 each name their spec file in title and body, so `check-stable-ownership.ts` keeps both `owned` after this issue closes.

**Marker asymmetry, stated so it is not "fixed" later:** voice moves `[!]` → `[-]` (`[!]` means *covered but flaky*, `QA-CHECKLIST.md:16`; `[-]` is the file's own precedent for a spec parked because the surface is not on the image, § 7.6 groq/mistral), while composio stays `[ ]` with its struck-through out-of-scope wording — `[-]` feeds the generated *Needs validation* column and the Phase 1 *Validate* table, which would contradict the committed § 6.2 decision that the spec **must not be promoted**.

**Rejected explicitly.** Rewriting `refresh-dropdown-list` into a real test: the behaviour is already covered by an `@stable` spec, so building a second one is precisely the spend this triage exists to avoid (design §3 evaluates DELETE first for that reason). And promoting the green `voice-assistant` row: green there measures the absence of a component, not the behaviour of a flag.

## Scope note

No assertion was weakened, removed or loosened anywhere. `run-flow.spec.ts` is **unchanged** — it appears in this PR only as the named replacement, and its two mutations were reverted with `0` `FF-MUTATION` markers left and a final green run. `tests/assets/triage/inherited-backlog-baseline.json` and `docs/triage/inherited-spec-triage.md` keep the deleted spec's historical rows on purpose: the first is the frozen measurement record the PR lane reads from the base ref, the second is generated from it and marked *Do not hand-edit*.

`QA-CHECKLIST.md` carries **manual Part II bullets only** (#741) — § 2.1 gains the dropdown-refresh bullet pointing at the replacement, § 6.2's Composio entry and § 9.3's Voice mode entry record their parks.

Two collateral corrections in docs this change touched: `docs/core-functionality/templates/templates-coverage-scope.md` named the deleted spec as *Portfolio Website Code Generator*'s only exerciser (it is not — `templates-instantiate.spec.ts` opens it as its own subject in an `@stable` test and asserts its graph shape; that paragraph's counts come from a literal-name grep and that spec builds its titles from a variable, which is the trap recorded there now), and its closing clause counted *"the two `[~]` in §11.2"* where §11.2 is now 26 bullets, all `[x]`. `probe-component-available.ts`'s caller enumeration went from two names to four.

## Validation (nightly `1.13.0.dev15`, `--retries=0`)

> **Row numbers are data rows of `docs/triage/inherited-spec-triage.md`** (92 of them), the convention under which the composio citation resolves. The first version of this table cited 77 and 91–93, which land on `ui-ux/integration-side-bar.spec.ts` and on a row that does not exist — corrected above after review.

> **The `check:stable-ownership` figures below were re-read after review.** The script's line on this head is `30 spec(s) with no @stable test: 10 owned, 2 declared, 18 unowned in the baseline, 0 unowned outside it, 0 undecidable — 0 failure(s)`; an earlier version of this section folded the 18 grandfathered rows into "owned". On the tree merged with today's `main` it reads **13 specs — 11 owned, 2 declared, 0 unowned, 0 undecidable**, with the checklist guard, `check:checklist-coverage` and `typecheck` all passing there too. What the gate turns on is unchanged: 0 unowned outside the baseline, and both parks `owned`.

- `typecheck` ✅ · `eslint` **0 errors** (149 warnings, all pre-existing) ✅ · QA-CHECKLIST guard ✅ · `check:checklist-coverage` ✅ · `check:stable-ownership` **0 unowned outside the baseline, 0 undecidable, 0 failures** ✅ · `reconcile-stable-orphans` **0 expired declarations** ✅
- spec-doc dependency guard against `origin/main` + `release-1.13.0` + `release-1.12.3`: **all 9 new `src/…` paths resolve, 0 unresolved, 0 `::notice::`** ✅
- `test:units` 1598/1598 ✅ · `test:scripts` 1695/1695 ✅ · `test:pipeline` 160/160 ✅
- the replacement `run-flow.spec.ts`: green in 33.3 s, zero `🚨 Backend Error` ✅
- the two parked specs: **4 skipped, no error** ✅
- **Flow-leak audit, read by diffing `GET /api/v1/flows/` around every run rather than from any report:** `voice-assistant` leaked **+3 per run** (one per test, no cleanup) and `refresh-dropdown-list` **+1 per run** — recorded in #1915 as a precondition of any un-park. The replacement `run-flow.spec.ts` measured **40 → 40**.
- **Duplicate candidate** (design §4's second hand-read fact): named above for the DELETE; none exists for either park, which is why neither is a deletion.

Force-fail mutations, one line each — all executed, all reverted, `0` markers left:

- remove the refresh click from the replacement → `text=Loading` never appears (30 s)
- strip the target flow from the refresh response → the repopulated list cannot satisfy the option filter (20 s)
- replace the deleted spec's waited-for string with one that cannot exist → **still passes**, which is the defect
- invert the voice mock to `true` → the "not available" test **still passes**
- delete the voice route mock outright → **still passes**
- remove composio's availability gate with a dummy key set → hard-fails on `getByTestId('composioGmail')` instead of skipping with an attributed reason

Reviewed in three independent rounds with clean context; rounds 1 and 2 each found defects that are fixed in `d87bbecc` / `6f7cfb57` / `1e2bd048`, round 3 found the delta clean.


daniellicnerski1 (commented):
Reviewed the full diff (9 files, +368/−69) plus an independent correctness pass on a worktree of this head. **No correctness bugs in the code changes.**

Three of the body's load-bearing claims re-verified independently rather than read: `run-flow.spec.ts:199-209` does click `refresh-dropdown-list-flow_name_selected` and assert the `Loading` → `Select an option` → option-select sequence inside an `@stable` test; §11.2 is 26 bullets all `[x]` against 26 entries in `registered-templates-baseline.json`, *Portfolio Website Code Generator* among them; `check:stable-ownership` keeps both parked specs `owned` (18 owned / 2 declared / 0 unowned outside the baseline), and this PR closes only #1913, so merging does not orphan #1915 or #1916.

The DELETE holds up on reading the removed spec rather than on the summary of it: its closing line discarded the `isVisible()` boolean, it used `waitForTimeout(3000)` as synchronisation, and — despite its name — it never clicked a refresh row at all. The named replacement is strictly stronger. And refusing #1913's instruction to promote the `3/3 green` voice row, with the two mutations run rather than argued, is the right call: `not.toBeVisible()` on an element absent from the bundle cannot fail.

Three findings, all low, none blocking.

## 1 — The triage row citations are off by one for two of the three specs

`docs/triage/inherited-spec-triage.md` holds 92 data rows (`grep -c '^| .*\.spec\.ts'`). Counting data rows only — the convention under which the composio citation is exactly right:

| Spec | Cited | Actual | What sits at the cited row |
|---|---|---|---|
| `composio.spec.ts` | 24 | 24 ✅ | — |
| `refresh-dropdown-list.spec.ts` | 77 | **78** | `ui-ux/integration-side-bar.spec.ts` — a different T2 spec, still open |
| `voice-assistant.spec.ts` | 91–93 | **90–92** | row **93 does not exist** |

#1913's definition of done asks for the row per spec, so this is the traceability half rather than a cosmetic one: a reviewer auditing the evidence behind the DELETE lands on an unrelated spec.

## 2 — The voice bullet takes `[-]` without the disclosure its own cited precedent carries

`QA-CHECKLIST.md:665`. The groq precedent this cites (line 557) defuses `[-]` inside the bullet:

> Known and left standing: `[-]` also feeds the generated *Needs validation* column and the Phase 1 *Validate* table, so the counts still read these two as scheduled work — a terminal marker would take a legend, a policy and a `coverage-summary.ts` change, which is its own issue rather than this one.

The new voice bullet carries no equivalent. Measured by running `npm run coverage:summary` on this head: `core-functionality/playground/` goes `Partial 1 → 0` / `Needs validation 0 → 1`, and the **Phase 1 — Next Delivery** table's `Validate ([-])` column for that module goes `0 → 1`.

This is the same objection the body's own *Marker asymmetry* paragraph uses to keep Composio at `[ ]`. As it stands, on the next merge to `main` the roadmap presents `voice-assistant.spec.ts` as validation work queued for the next delivery — a spec whose three tests are `test.fixme` and whose surface is in no image. Copying the groq clause into the bullet resolves it.

## 3 — The composio skip reason asserts a packaging fact the probe may not have measured

`composio.spec.ts:35`. `isProviderComponentAvailable()` returns `false` on *any* failure — its `catch` swallows a thrown `getAuthToken`, and `!res.ok()` covers a 401/500 — while the new reason states that the `lfx-bundles` distribution is not installed, which is a packaging claim.

Concrete path: once `COMPOSIO_API_KEY` becomes a repository secret, a shard whose shared Langflow is wedged mid-run (#922/#927) reports this spec as *skipped — component not in this build* rather than as an unevaluated run. That is an unknown recorded as a verdict, which is what #1012 exists to prevent.

It is inherited from the groq/mistral call sites and unchanged by this PR, so it is an observation about new code rather than something this PR should have to carry. The fix — a return value distinguishing *probe could not decide* from *probe decided absent* — belongs in its own issue against the helper. Happy to file that if you want it tracked.

Victor-w-Madeira (commented):
## Second pass — independent verification of the diff and of @daniellicnerski1's review

Re-read the full diff on a worktree of this head, re-ran every guard, and re-measured each of the three findings rather than reading them. **No correctness bug in the code changes** — that half of the earlier review holds.

Two of the PR's load-bearing claims were re-derived from the source rather than from the summary of it:

- **The DELETE.** `run-flow.spec.ts:199-209` clicks `refresh-dropdown-list-flow_name_selected`, asserts the `Loading` → `Select an option` sequence, and then selects the target flow out of the repopulated list — inside a test tagged `@stable @release @workspace @api @regression`. The deleted spec's closing line was `getByText("Loading Options").isVisible()`, a discarded boolean, and despite its name it never clicked a refresh row at all. The replacement is strictly stronger and the deletion loses no failure condition.
- **Refusing #1913's instruction to promote the `3/3 green` voice row** is the right call: `not.toBeVisible()` on an element absent from the bundle cannot fail, and both mutations were run rather than argued.

Also verified and clean, not covered by the earlier review: both new docs carry the four mandatory sections (*What this test validates*, *Tags*, *Validation criterion*, *External dependencies*); the probe's caller enumeration (`groq`, `mistral`, `ollama`, `composio`) matches the four real call sites; #1915 and #1916 are open and name their spec file in the title; and this PR closes only #1913, so neither park is orphaned by the merge.

## The three findings — all three confirmed

**1. Row citations.** Confirmed exactly. Counting data rows only (92 in `docs/triage/inherited-spec-triage.md`, the convention under which the composio citation is right): `refresh-dropdown-list` is row **78** (77 is `integration-side-bar.spec.ts`, a different open T2 spec) and `voice-assistant` is **90-92** (93 does not exist). #1913's definition of done asks for the row per spec, so this is traceability. **Fixed in the PR body** — the citations live nowhere else.

**2. The `[-]` marker without its precedent's disclosure.** Confirmed, and it reproduces on the merged state as well as on this head. Running the generators: `core-functionality/playground/` goes `Needs validation [-]` 0 → 1 and `Partial [~]/[!]` 1 → 0, and the **Phase 1 — Next Delivery** table's `Validate ([-])` column for that module goes 0 → 1. On the next merge to `main` the roadmap would present a spec whose three tests are `test.fixme`, and whose surface is in no image, as validation work queued for the next delivery. It is the same objection this PR's own *Marker asymmetry* paragraph uses to keep Composio at `[ ]`, and the groq bullet it cites (QA-CHECKLIST.md:559) defuses it in one clause. **Fixed** — the clause is now in the voice bullet.

**3. The probe reason asserts a packaging fact it may not have measured.** Confirmed. `isProviderComponentAvailable()` returns `false` on any failure — the `catch` swallows a thrown `getAuthToken`, `!res.ok()` covers a 401/500 — while the skip reason states that `lfx-bundles` is not installed. The helper's own comment concedes it (*"Treat an unreachable/erroring registry as 'not available'"*). One refinement to the framing: it is reachable **today**, not only once `COMPOSIO_API_KEY` becomes a repository secret — a wedged backend produces a byte-identical skip reason right now. It is harmless today only because a skip is the expected outcome either way. Correctly scoped out of this PR: inherited from the groq/mistral call sites and unchanged here. **Filed separately.**

## Two things neither pass caught

**This branch is 6 merges behind `main`**, including the sibling T2 batches (#1919 / #1920 / #1922) that touch the same checklist region. GitHub reports `CLEAN`, and a test merge confirms it: on the merged tree, `check:stable-ownership` reads **13 specs with no `@stable` — 11 owned, 2 declared, 0 unowned, 0 undecidable**, and `check:checklist-coverage`, the QA-CHECKLIST guard and `typecheck` all pass. Not a blocker, but the *Validation* section's numbers no longer describe the post-merge state.

**Both passes misread the same output line.** The body says *"28 owned / 2 declared / 0 unowned"*; the earlier review says *"18 owned / 2 declared"*. On this head the script prints:

```
30 spec(s) with no @stable test: 10 owned, 2 declared,
18 unowned in the baseline, 0 unowned outside it, 0 undecidable — 0 failure(s)
```

The body folded the 18 *unowned-in-baseline* into "owned"; the review quoted those same 18 as the owned count. Neither number is produced by any invocation, including `--base-ref`. It changes no verdict — 0 unowned outside the baseline, 0 failures, both parks `owned` — but as written it reads as *everything has an owner* when 18 specs are grandfathered. **Corrected in the body**, with the post-merge figures.

## Disposition

Findings 1 and 2 fixed; finding 3 filed as its own issue against the helper. Merging.

Victor-w-Madeira:
Finding 3 is filed as #1930 (`qa-infra` / `follow-up`) against `probe-component-available.ts`: the helper collapses *undecided* into *absent*, so all four call sites can state a packaging fact the probe did not measure. Findings 1 and 2 are fixed — the row citations in the body above, and `97de3e0d` for the `[-]` disclosure on the voice bullet. All seven checks green on the fixed head. Merging.

Issue #1587, opened 2026-08-25
Two spec docs keep real upstream paths out of External dependencies on a rule #1574 removed


Victor-w-Madeira:
Two spec docs keep a **real upstream module out of their `## External dependencies` list on purpose**, and both justify it with a rule that stopped being true when #1574 shipped: *"the guard resolves paths against `main`"*. Since #1574, `scripts/watch-upstream-areas.mjs --mode=check-docs` resolves every backticked `src/…` token against `origin/main` **plus the two release lines the nightly is cut from**, and a path resolving on only some of them is reported as a `::notice::` — not a failure (`pr-validation.yml`, *Spec-doc dependency paths*: *"A path resolving on any one ref passes and is NAMED, as a `::notice::`, when only some of them satisfied it"*).

So the workaround now costs coverage for no benefit: the module the assertions actually depend on is invisible to the guard, which is the failure mode #1092/#1298 exist to prevent, arrived at from the other side — not a path that resolves nowhere, but a path deliberately never handed to the resolver.

## Measured 2026-08-25 (`repos/langflow-ai/langflow/git/trees/{ref}?recursive=1`)

| path | `main` | `release-1.12.0` | class under the current guard |
|---|---|---|---|
| `src/lfx/src/lfx/exceptions/tweaks.py` | 404 | blob `e21fcaf833e7` | `release-only` → `::notice::`, passes |
| `src/backend/base/langflow/api/v2/workflow_reconstruction.py` | **present** | present | resolves everywhere, passes silently |

The second row is the sharper one: that module resolves on `main` **today**, so the doc declines to declare a path that would not even produce a notice.

## Where

- `docs/security/tweaks-graph-path-floor.md:167` — `TweakRefusedError` lives in `lfx/exceptions/tweaks.py`; its `refused` and `reason` fields are the text the stream-frame assertions match, so it is a first-order dependency of `security/tweaks-graph-path-floor.spec.ts`. Named in prose *"because the guard resolves paths against `main` (#1298)"*. Found reviewing PR #1586; left out of that PR because it is pre-existing and spans a second doc.
- `docs/api/flows/workflows-v2-job-lifecycle.md:177` — `workflow_reconstruction.py` and the `sync_result_storage_enabled` setting, *"read inside the running container rather than from a ref"*, justified with *"the spec-doc dependency guard resolves paths against `main` and would reject them (#1574)"* — citing the very change that removed the constraint.

## Deliverables (Done when)

[...]

Issue #1092, opened 2026-07-29
file-watcher: no monitored area covers src/lfx/ (1233 files), and a missing path is indistinguishable from "nothing changed"


Victor-w-Madeira:
Surfaced while triaging #1091. The change that broke all 6 stdio registrations in
`mcp-server.spec.ts` landed in `src/lfx/src/lfx/base/mcp/security.py` on
2026-07-15 (`f4d6ac4`, upstream `#14073`). `file-watcher.yml` exists precisely to
open an issue for that class of upstream change. It did not fire, and it was
never going to.

## 1. `src/lfx/` is watched by zero of the 13 areas

```
$ grep -c "src/lfx" .github/workflows/file-watcher.yml
0
$ find src/lfx -name "*.py" | wc -l          # in the upstream checkout
1233
```

Every monitored path lives under `src/backend/base/langflow/` or
`src/frontend/src/`. Langflow has been moving backend behavior into the `lfx`
package, so a growing share of what our specs exercise is outside the watcher's
field of view entirely. This is omission, not drift — the paths listed are fine,
the list is incomplete.

## 2. Two areas watch a path whose implementation already left

Measured on the upstream checkout (`3551acb`, 2026-07-27):

| Monitored path | Size | Where the code actually is | Size |
|---|---|---|---|
| `src/backend/base/langflow/base/agents/` | **1 file, 3 lines** | `src/lfx/src/lfx/base/agents/` | 15 files, 3069 lines |
| `src/backend/base/langflow/inputs/` | 6 files, 238 lines | `src/lfx/src/lfx/inputs/` | 5 files, 1645 lines |
| *(MCP)* `api/v1/mcp.py` + `mcp_projects.py` | 378 + 1718 lines — alive | `src/lfx/src/lfx/base/mcp/` **(unwatched)** | 5 files, 3826 lines, incl. `security.py` |

*Agents & Agentic Flows* watches a 3-line shim for its agents leg (its other two
paths are alive), and *MCP Server* watches the live API layer while the security
and protocol base it depends on is invisible — which is exactly how #1091 got
through.

## 3. A missing path is silent by construction

```bash
COMMITS=$(git log --since="$SINCE" --oneline -- $PATH_LIST 2>/dev/null | head -5)
if [ -n "$COMMITS" ]; then ... fi
```

`git log -- <nonexistent-path>` returns nothing, and `2>/dev/null` discards any
[...]

PR #1924, opened 2026-09-19
chore(guard): resolve the upstream modules docs name without their src/ prefix (#1592)

### LINKED ISSUE #1592: Upstream modules named without the src/ prefix are invisible to the spec-doc dependency guard
Sibling of #1587, arrived at from the other side. #1587 covers a path a doc **declines to declare with a stated reason**; all three of those are fixed. This is the larger, silent half: a doc names a real upstream module in its `## External dependencies` section **without the `src/` prefix**, with no reason given and no full path anywhere in the section — so `parseDocDeps` never collects it and the guard never resolves it.

`scripts/watch-upstream-areas.mjs:668` is the whole mechanism:

```js
for (const match of line.matchAll(/`([^`]+)`/g)) {
  const token = match[1].trim();
  if (token.startsWith("src/")) out.push({ token, line: index + 1 });
}
```

A token written `api/v2/hitl.py` instead of `src/backend/base/langflow/api/v2/hitl.py` is not a path that fails to resolve — it is a path the resolver is never handed. That is #1092's defect ("a nonexistent path is silent") with the silence moved one step earlier, and it is invisible to every guard the repo has: the 502→507 count in the PR-validation summary counts what it was given, not what the docs depend on.

## Measured 2026-08-25

Swept `docs/` structurally, replicating `parseDocDeps` and then excluding, per doc: tokens starting with `src/`, this repo's own paths (`tests/`, `scripts/`, `helpers/`, …), our own `.spec.ts` names, HTTP verb+endpoint strings, ellipsis tokens, and — importantly — any bare filename whose full path **is** declared elsewhere in the same section (a sibling abbreviation such as `chat_output.py` next to a declared `.../input_output/chat.py` is not a hole).

What survives: **~35 tokens across 18 docs**. Ten resolved against the real trees, `git ls-tree -r origin/release-1.12.0` / `origin/main`, all present on both:

| doc | written as | real path |
|---|---|---|
| `ui-ux/keyboardComponentSearch.md:141` | `sidebarDraggableComponent.tsx` | `src/frontend/src/pages/FlowPage/components/flowSidebarComponent/components/sidebarDraggableComponent.tsx` |
| `model-provider/groq-provider.md:160` | `lfx_bundles/groq/groq.py` | `src/bundles/lfx-bundles/src/lfx_bundles/groq/groq.py` |
| `model-provider/mistral-provider.md:159` | `lfx_bundles/mistral/mistral.py` | `src/bundles/lfx-bundles/src/lfx_bundles/mistral/mistral.py` |
| `playground/human-input-pause-resume.md:179` | `api/v2/hitl.py` | `src/backend/base/langflow/api/v2/hitl.py` |
| `playground/playground-output-data.md:69` | `api/v1/flows_helpers.py` | `src/backend/base/langflow/api/v1/flows_helpers.py` |
| `core-components/human-input-node-config.md:156` | `NodeOutputParameter/NodeOutputs.tsx` | `src/frontend/src/CustomNodes/GenericNode/components/NodeOutputParameter/NodeOutputs.tsx` |
| `project-management/folder-deletion-integrity.md:437` | `use-get-folders.ts` | `src/frontend/src/controllers/API/queries/folders/use-get-folders.ts` |
| `flow-functionality/api-access-modal-regression.md:84` | `get-python-api-code.tsx` | `src/frontend/src/modals/apiModal/utils/get-python-api-code.tsx` |
| `mcp/server/mcp-server-install.md:265` | `hooks/useMcpServer.ts` | `src/frontend/src/pages/MainPage/pages/homePage/hooks/useMcpServer.ts` |
| `observability-monitoring/traces-delete-cascade.md:54` | `lfx/services/settings/base.py` | `src/lfx/src/lfx/services/settings/base.py` |

The remaining eight docs: `data-operations-component.md`, `data-operations-legacy-link.md`, `model-provider-model-toggle.md`, `azure-ai-foundry-provider-setup.md`, `traces.md`, `traces-latency-tokens.md`, `playground-bulk-delete.md`, `playground-message-logs.md`.

**An abbreviated token is not only invisible, it can be ambiguous.** `services/settings/base.py` matches two real files — `src/backend/base/langflow/services/settings/base.py` and `src/lfx/src/lfx/services/settings/base.py`. `traces-delete-cascade.md` writes `lfx/services/settings/base.py`, which happens to disambiguate; a doc writing the bare `services/settings/base.py` would leave the next reader to guess, and the guard cannot tell them apart either.

## Why this is a follow-up and not part of #1587

#1587's deliverable 3 asked for a sweep for *"a path described in prose with an explicit reason not to declare it"*. That sweep was done and its result is complete: the three docs it found are fixed, and no fourth doc gives a reason. This class carries **no rationale at all** — it is a different defect (an abbreviation, not a decision), it spans 18 docs against 3, and folding it in would have grown that PR well past its issue. Recorded here rather than left as a remark in the PR.

## Deliverables (Done when)

- [ ] Each of the ~35 tokens is triaged: an upstream module that is a genuine first-order dependency gets its **full `src/…` path** declared; a token that is context rather than a dependency stays prose, and nothing is left with a stated reason to omit (that would re-create #1587).
- [ ] Ambiguous abbreviations are resolved to the specific file, not left matching two.
- [ ] The guard is observed green on the changed docs, and the resolved-path count in its summary is reported before and after — the count is the only visible evidence that coverage widened.
- [ ] If the sweep's filter is worth keeping, it lands as a script with unit tests rather than a one-off, so the class cannot silently return. **Not** a change to `parseDocDeps` itself: making it collect non-`src/` tokens would try to resolve every `helpers/foo.ts` in every doc. That trade is worth arguing in the PR, not assuming here.

## Not in scope

- Anything #1587 covered. Those three docs are done.
- This repo's own paths (`tests/…`, `helpers/…`) written as bare filenames. The guard resolves only upstream tokens by design, and our own tree is checked by other means.

## References

- #1587 — the sibling defect, with a stated reason; the sweep that surfaced this class
- #1298 / #1092 — the guard, and "a nonexistent path is silent"
- #1574 — resolution against the release lines

Victor-w-Madeira:
## Type

- [x] `docs` — documentation update

---

- [x] **Exception** — off-wave work backed by an issue: a follow-up. Issue #1592 (`follow-up` label, sibling of #1587)

---

## What this PR does

`scripts/watch-upstream-areas.mjs:668` hands the resolver only the tokens that start with `src/`. A doc writing `api/v2/hitl.py` therefore has **no path that fails to resolve — it has a path the resolver never sees.** That is #1092's *"a nonexistent path is silent"* one step earlier, and it is invisible to every guard in the repo: the `Resolved N dependency path(s)` line counts what the guard was **given**, never what the docs depend on.

Two halves.

### 1. The 52 abbreviations, triaged and declared

Measured 2026-09-18 against `origin/main` + `origin/release-1.13.0` + `origin/release-1.12.3`: **52 upstream modules across 24 docs** named without their prefix, plus **2 abbreviations matching two real files each**. All corrected, which moves the existing guard's count **730 → 782** on the same base — the only visible evidence coverage widened (the issue's third deliverable).

The issue's own table listed 10 of these; the sweep re-run today found the rest, and three turned out to be more than an abbreviation:

| doc | written as | finding |
|---|---|---|
| `playground-bulk-delete.md:64` | `session-selector.tsx` | **Ambiguous** — two live upstream files. It is the `chat-header/` copy: the only one with `showCheckbox` and `data-testid={`session-${session}-checkbox`}`. The `IOFieldView` copy has neither. |
| `human-input-pause-resume.md:170` | `modals/IOModal/.../chat-message.tsx` | **Ambiguous** — two live copies, and the doc's elided form was right. It is `modals/IOModal/components/chatView/chatMessage/chat-message.tsx`, the only one that emits `chat-message-${sender_name}-`; the `playgroundComponent` sibling carries no testid at all. |
| `playground-message-logs.md:59` | `tableComponent/TableOptions/index.tsx` | **Wrong, not just short** — the real path carries one more `components/` segment. Corrected by hand; see *Known limitations*. |

Four tokens stay prose, because the doc names them to say the spec does **not** depend on them — declaring those would couple a spec to a file whose deletion cannot break it. They go in `scripts/lib/doc-dep-abbreviation-declarations.json`: `traces/model.py` (cited to say *no request reaches it as a `422`*) and the three Trace Details files `traces.md` explicitly places out of its own scope. **Deliberately a repo-side declaration, verified in both directions per #1084, and not an in-doc sentence giving a reason to omit** — that in-doc form is exactly the defect #1587 removed, and re-creating it is what the issue's first deliverable rules out.

### 2. The filter as a mechanism, not a one-off

`--mode=check-doc-abbrevs` — a **sibling mode**, for the reason `check-docs` itself is one: it needs the same `git ls-tree` machinery over the same blobless clone. It reports a non-`src/` token that resolves **by suffix to exactly one upstream path and to nothing in this repo**; a token matching several is reported as *ambiguous* rather than guessed. Same severity rule as its sibling (#980/#1012): a finding in a doc the PR changed fails, the rest are reported so drift stays visible.

**Explicitly not a change to `parseDocDeps`**, per the issue: widening it would try to resolve every `tests/helpers/foo.ts`, every `POST /api/v1/flows/` and every npm package name against the Langflow tree and report each as broken.

---

## How it was built

Three clauses are load-bearing rather than tidy, each measured over the whole of `docs/`:

- **Uniqueness is what makes it gateable.** The prose that survives every other clause — `text/event-stream`, `store/tags`, `models/gemini-embedding-001`, `@modelcontextprotocol/server-everything` — resolves to **nothing**. No keyword list is doing that work.
- **Ours-wins, consulted before the upstream index.** Upstream carries `src/frontend/tests/utils/flow/add-component-from-sidebar.ts` and so do we. A suffix match alone rewrites **20** of this repo's own helper references into Langflow paths, which is out of scope by the issue.
- **An ellipsis resolves by its TAIL**, not skipped — the elided prefix is routinely the part that went stale, and two docs carry that shape today.

The shared reader was extracted rather than copied: `parseDocDepTokens` is now the one parser and `parseDocDeps` is its `src/` filter, because the whole class this check exists for is a token the `src/` filter drops — a second scanner that disagreed about where the section begins or ends would report holes in a region the guard never reads.

---

## Dependencies

None new. The CI step reuses the clone, `$RELEASE_CSV` and `changed-docs.txt` the existing `doc-deps-guard` job already produces, and runs **after** the path resolver: a `::error::` about a path that does not exist is the more urgent annotation and must not be pushed down the list by this one.

---

- Anything #1587 covered — those three docs are done.
- This repo's own paths written as bare filenames (`tests/…`, `helpers/…`). The guard resolves only upstream tokens by design, and the *ours-wins* clause keeps them out on purpose.

---

## Known limitations

**An abbreviation that is also WRONG resolves to nothing and stays invisible.** `playground-message-logs.md` wrote `tableComponent/TableOptions/index.tsx` for a path carrying one more `components/` segment, and no clause here reaches it.

Falling back to the **basename** when the written suffix resolves to nothing was measured over the whole of `docs/` and **rejected**: ~130 rows, essentially all HTTP endpoints (`POST /api/v1/flows/` → the seven upstream directories named `flows`), and the one true positive it would have found resolves **456** ways. A check that loud is a check nobody reads — the `mode=count` lesson. That token was corrected by hand and the limit is recorded in the module header rather than worked around.

**One unit test is named for a partial claim, on purpose.** The whole-repo assertion builds a **synthetic** tree from the `src/…` paths the docs already declare, so it needs no checkout — at the cost of seeing only abbreviations whose full path some *other* doc carries. Measured: **24 of the 55** paths this change introduces. Found by force-fail: reverting `hitl.py` alone left it green while reverting `assistant-discovery-storage.ts` reddened it. It is now named `no doc abbreviates a module that ANOTHER doc already names in full`, with the measurement in a comment — a guard that reads as whole-repo coverage and delivers 44 % is #1226's shape. The complete check is the PR lane, which has the real tree.

---

## Related issue

Closes #1592

---

## Validation

**The mechanism observed working, both directions** (the issue's third deliverable):

```
# before, on this base
Resolved 730 dependency path(s) from 320 doc(s) …
# after
Resolved 782 dependency path(s) from 320 doc(s) …          exit 0
Checked 841 non-`src/` token(s) … 4 declared as context.
Every upstream module named in a dependency section is written as a resolvable `src/` path.   exit 0
```

Both guards green with `--changed` scoped to the 28 docs this PR touches.

**Force-fail — executed, not asserted:**

| mutation | result |
|---|---|
| revert one token in a **changed** doc | `exit 1`, names `doc:line`, the token and the path it should be |
| same token, doc **outside** the diff | `exit 0` + `::warning::` (#980's trade, observed) |
| drop the *ours-wins* clause | 1 test red |
| drop *already declared in the same section* | 2 red |
| guess instead of reporting ambiguity | 1 red |
| severity always `warn` | 1 red |
| never report an expired declaration | 2 red |
| a one-segment word becomes a candidate | 2 red |
| ellipsis: resolve the whole token, not its tail | 1 red |
| declarations keyed by token only, ignoring the doc | 1 red |
| accept a declaration with no `reason` | 1 red |

Two findings came out of running those and are already fixed in the branch: the first *declarations-by-token* mutation was a **no-op** (the key it added was never looked up) and had to be rebuilt against the real lookup; and the whole-repo test's vacuity above.

**Gates:**

- `npm run typecheck` ✅ clean
- `npm run lint` ✅ 0 errors
- `npm run test:scripts` ✅ 1714/1714 (16 new)
- `npm run test:units` ✅ 1628/1628
- `npm run test:pipeline` ✅ 160/160
- `npm run check:checklist-coverage` ✅ · `npm run regressions:check` ✅

No spec, helper or Page Object is touched, so no Playwright run, no `QA-CHECKLIST.md` bullet and no `@stable` change applies.


Victor-w-Madeira:
## Review round applied — 9 findings, 2 of them defects

An independent review with clean context went over the 52 corrected paths (against all three refs, and against the upstream *sources*, not just the trees), the pure functions, the CI wiring and the prose. I verified every finding before acting on it rather than taking the report at face value. **No blockers; nothing in the change's premise moved.** Two were real defects, one was a wrong path, and six were claims of mine that were not true.

### The two defects

**1 — The CLI half was pinned by nothing.** Three mutations to `runCheckDocAbbrevs` each left the unit lane at 93/93:

| mutation | before | now |
|---|---|---|
| delete `process.exit(1)` — *the thing that makes this a gate rather than a report* | 93/93 ✅ | 3 red |
| force every annotation to `::warning::` | 93/93 ✅ | 2 red |
| delete the whole expired-declaration report | 93/93 ✅ | 1 red |

A pure function nothing calls is not a gate. The sibling `--mode=check-docs` has a CLI harness for precisely this reason — its own comment records that *"reverting `--releases` to a no-op, deleting the printed report … all leave the suite green"* — and this mode reused its readers but not its harness. Added one (`abbrevCliFixture` / `runAbbrevCli`; the home has to be a real git repo with an index, since `git ls-files` runs there) plus five CLI tests covering the gate, the ambiguity report, the clean verdict, and both refusal paths.

**2 — The stale-declaration escalation was dead in the only lane that runs it.** `changed-docs.txt` is built with a `-- docs README.md` pathspec and then filtered to `.md$`, so `scripts/lib/doc-dep-abbreviation-declarations.json` **could never appear** in the list the severity rule reads. An expired declaration was a `::warning::` on a green job, and the *"verified in both directions (#1084)"* claim in the commit message, the JSON's own `$comment` and `CLAUDE.md` held by accident. The step now appends it — harmless to the sibling step, whose severity keys on a doc path and can never match a `.json`.

Pinned by **executing** that step's `run:` body under `bash` against a synthetic repo, not by grepping the YAML: a regex over workflow text pins a spelling, not a behaviour, and this repo has measured such a guard passing the mutation it existed to catch (#1226). Dropping the appended line now reddens the lane.

### The wrong path — and why it is the interesting one

`traces-delete-cascade.md:54` named `src/lfx/src/lfx/services/settings/base.py` as the owner of `sqlite_pragmas`. It is not there on any of the three refs — the default lives in `.../settings/groups/database.py:48` (`git grep -l sqlite_pragmas` confirms, on all three). **I expanded the abbreviation to the target the issue's own table gave, without re-checking ownership.** That is this PR's own class one layer down, and the cost is concrete rather than cosmetic: `impacted-tests.ts` maps a changed upstream path to specs by prefix, so a change to `groups/database.py` would never have selected this spec.

Which raised the right question — *how many of the other 51 point at a file that does not own what the prose claims?* Swept mechanically: for every introduced path, the backticked identifiers in the **same bullet** against that file's real contents. Six hits, all cross-attribution inside two-file bullets — `tool-status-${status}` is a template in **both** `ToolCallCard.tsx` and `ContentBlockDisplay.tsx`, both `inputComponent` files reference `nodeId`, and `ChatOutput.sender_name` is a qualified name that appears in no file literally. **That was the only one.**

### Five corrections to claims that were not true

- `readDocDepDeclarations`'s JSDoc promised a floor the code does not have (*"both are refused"*). An empty `declarations` array is accepted **on purpose** — it is the legitimate end state, and its failure direction is loud, not silent — which is the opposite of `--min-categories`, whose floor exists because a wrong baseline is permanent. The comment now argues that instead of asserting the reverse.
- The `expired` reason named two of four causes, and the **commonest** — the token was rewritten as a full `src/` path — was not among them.
- `git ls-files` is fail-loud only when it *throws*. Empty stdout returned cleanly and disabled the *ours-wins* clause in silence, turning this repo's own helper references into findings and, in a touched doc, into hard failures. It was the one input to this mode with no floor; it has one now (exit 2).
- The verdict line called a stale declaration *"in a doc"*; its subject is the JSON file.
- The header counted the exempt template the function skips — printing `321` two lines below the sibling's `320 … 1 exempt`.

### Four numbers that were wrong

| claim | was | is | how |
|---|---|---|---|
| docs carrying the 52 | 24 | **28** | `git diff --name-only -- docs \| wc -l`, and 28 distinct docs in the pre-change finding list |
| own-file references *ours-wins* suppresses | 20 | **17 occurrences, 10 distinct tokens across 10 docs** | enumerated |
| the offline test's reach | "24 of the 55" | **18 of 39** | revert one corrected occurrence at a time, rebuild the synthetic tree, re-run the construction. The old figure was reproducible by no construction and mixed two units — modules in one clause, token instances in the other |
| `@modelcontextprotocol/server-everything` as prose *uniqueness* drops | cited 3× | **removed** | `abbreviationTarget()` returns `""` — the `@` rejects it before the tree is consulted, so it was the one example that disproves the point it was quoted for. `text/event-stream` and `store/tags` do survive and do resolve to nothing, so the claim itself stands |

### Not filed

The reviewer flagged, and declined to file, that `traces.md:44` still carries *"The Trace Details modal (…) is not exercised by this spec"* in prose. I agree: that is a **scope statement**, not a reason-to-omit. The #1587 shape is a doc declining to declare a dependency it has; this one says the spec has no such dependency, and the machine-readable exemption now lives in the JSON regardless.

### Gates after the round

`typecheck` ✅ · `lint` ✅ 0 errors · `test:scripts` ✅ **1721/1721** (was 1714; +7) · `test:units` ✅ 1628/1628

Both guards still `exit 0` scoped to the changed files, and the abbreviation header now reads `320 doc(s) … 1 doc(s) exempt`, agreeing with its sibling. Six mutations re-run, all caught.

Issue #1927, opened 2026-09-19
start-langflow-source.sh requires a source clone it never runs


Victor-w-Madeira:
## What

`scripts/start-langflow-source.sh` refuses to start unless `LANGFLOW_SRC_REPO` names an existing
directory, and `cd`s into it before launching the server. When the instance is served from a
**published distribution in a venv** — how the VM lane has served the target since 2026-09-10 —
nothing in that clone is ever executed:

- the dependency sync is skipped: with `LANGFLOW_SRC_RUN_CMD` set, `SYNC_CMD` defaults to empty;
- the UI comes from `LANGFLOW_SRC_FRONTEND_DIR`, not from the clone;
- the build-stamp comparison has no stamp to compare against on a lane that sets
  `LANGFLOW_REQUIRE_BUILD_STAMP=0`;
- the one remaining use is as the launched process's working directory.

So the script demands 4.9 GB of source to satisfy a `cd`.

## Why it matters now

Stage 2 of the VM migration consolidates runner and target onto a single machine, and that machine
has no Langflow source clone. Measured on 2026-09-18 on the QA VM: with `LANGFLOW_SRC_REPO` pointed
at an empty directory, the published distribution installed and the instance answered
`/health_check` in 35 s, then served a full spec run. The empty directory was a workaround for
exactly this check.

## Change

When `LANGFLOW_SRC_RUN_CMD` is set, stop requiring `LANGFLOW_SRC_REPO`. With no clone given, launch
from the state directory, which already exists and already holds the database, the log and the PID
file.

## What must not change

The source path keeps every guard it has today: the clone stays required when the script builds
from source, and the stale-frontend check stays as it is — an old UI against a new backend fails
green, which is the failure that check exists to prevent.

PR #1932, opened 2026-09-20
fix(run-e2e): let the target be this machine, and stop calling that "no tunnel"

### LINKED ISSUE #1931: run-e2e.sh cannot drive a target on the same machine, and the tunnel branch mis-describes that case
## What

`scripts/run-e2e.sh` can only drive a target on **another** machine. Every interaction
goes through `target_ssh()`, which is `ssh "$TARGET_SSH" …`, and `TARGET_SSH` is required
with no default. Stage 2 of the VM migration puts the runner and the target on one
machine, and that machine does not ssh to itself: measured on the QA VM on 2026-09-18,
`ssh localhost` answers `Permission denied (publickey)`.

Giving the machine its own key would work and would hide the topology in how the box was
built. The orchestrator is where it belongs.

## Two things, and they are the same conversation

**1. A target that is this machine.** `target_ssh` should run the command locally instead
of over ssh. The contract is already the right shape: ssh joins its arguments into one
string for a shell on the other side, and `bash -c "$*"` is that same contract with stdin
still flowing, which every `bash -s <script` call site depends on.

**2. The tunnel branch mis-describes the local case.** Today "no tunnel" is treated as
"bind the backend to the target's private address", and the run is told the ten clipboard
specs will fail deterministically, because Chromium treats only `localhost` as a secure
context. With the target on this machine that is false: the backend binds `127.0.0.1` here
and `localhost` **is** the target. Left as it is, consolidating either raises a warning
that does not apply or invites `ALLOW_NO_TUNNEL=1`, which would accept a cost that is not
being paid — and bind the backend to a private address for no reason.

## Also in scope, and small

`start_backend_for_shard` passes `LANGFLOW_SRC_REPO=${LANGFLOW_SRC_REPO:-$HOME/langflow}`
unconditionally. Since #1927 the starter does not need a clone when the caller supplies the
run command, so the orchestrator should stop naming one on that path. Until it does, #1927
is enabling-only for this lane.

## Not in scope

Consolidation itself, and the services the target host also runs (`ollama`, the echo
server). Those are machine work; this issue is the orchestrator's half.

Victor-w-Madeira:
Closes #1931. Follows #1929, which is what makes the clone optional on the starter's side.

## What changed

**1. `TARGET_SSH=local` — the target is this machine.** `target_ssh()` runs the command
here instead of over ssh. ssh joins its arguments into one string for a shell on the
other side, and `bash -c "$*"` is that same contract with stdin still flowing, which
every `target_ssh "… bash -s" < scripts/x.sh` call site depends on. Spelled as a value
of `TARGET_SSH` rather than as a new switch, so there stays exactly one place that
answers *where is the target*, and the requirement that the caller name it is not
weakened. (An ssh alias literally called `local` cannot be reached from here; the header
says so.)

**2. The tunnel branch stops mis-describing that case.** Today "no tunnel" means "bind
the backend to the target's private address", and the run is told the ten clipboard
specs will fail deterministically, since Chromium treats only `localhost` as a secure
context. With the target on this machine that is false — the backend binds `127.0.0.1`
here and `localhost` **is** the target. The check is skipped with the reason printed,
rather than left to `ALLOW_NO_TUNNEL=1`, which would announce a cost nobody is paying
and bind an auto-login instance wider for no reason.

**3. The clone stops being named when it is not needed.** `backend_launch_env` omits
`LANGFLOW_SRC_REPO` when the caller supplies `LANGFLOW_SRC_RUN_CMD`. Until this, #1927
was enabling-only for the lane: the orchestrator kept naming a clone the starter no
longer wants.

The launch environment moved into `backend_launch_env` so both paths send the same
string and it can be read without a machine.

## Behaviour on the current lane

Unchanged. A remote target still goes over ssh, still holds the session open, and still
binds to the private address when there is no tunnel — that last one is pinned by a test
that passes against both versions. The only lane-visible difference is that a target
served from a venv no longer has `LANGFLOW_SRC_REPO` pushed at it, which is what #1927
made safe.

## Tests

Four added, two rewritten, in `scripts/run-e2e.test.mjs`. All six fail against the
previous script. The rewritten pair asserted the mirrored values crossed by grepping the
ssh line, which is now one of two launch paths; they now assert what **both** paths send
and what the composer puts in it. The local-target test runs with a decoy `ssh` earlier
on `PATH` that announces itself, so a regression to the ssh branch fails loudly instead
of passing because ssh happened to work.

`scripts/run-e2e.test.mjs` has one failure on my machine that is **not from this branch**
— "the field names this script parses are the ones the resolver actually emits" fails the
same way on `main`, from `npx ts-node` compiling against a stale local `node_modules`. CI
has the authoritative answer.

## Not in this PR

Consolidation itself, and the services the target host also runs. `ollama` is installed on
the current target and absent from the machine that will survive; the echo server needs
node, which is already there. Those are machine work, and they get validated by running
the lane, not by review.


Victor-w-Madeira:
Review found a defect that would have made the local path unusable, and three smaller things. All four are fixed in `cea98dae`.

**The one that mattered.** The base URL, the readiness probe and the bind override each derived the host from `LANGFLOW_TUNNEL` on their own. Setting it to 0 for a local target fixed the probe and the bind — both special-cased in the first commit — and left `run_shard` pointing `PLAYWRIGHT_BASE_URL` at the private address while the backend bound loopback. Every spec would have failed with connection refused **behind a healthy probe**, which reads as hundreds of product failures, and the ten clipboard specs would have lost the very secure context this change exists to give them. The three now read one helper, `target_reach_host`.

**The other three.** The local launch reads the starter's exit status instead of discarding it — the starter kills the server on its own timeout, so a swallowed status cost a second full `BACKEND_START_TIMEOUT_S` to report something already known. `LANGFLOW_SRC_REPO` is emptied rather than omitted, because ssh forwards no environment while a local child inherits the operator's, and an exported clone path would put the starter back on the clone branch. And a negative assertion gained the status check it was missing.

Two tests added, both failing against the previous commit. 101 of 102 pass locally; the one failure is the `ts-node` one described in the PR body, which fails the same way on `main` here.

Issue #1943, opened 2026-09-21
The tag removal must follow the verdict: daily-stable should stop removing @stable


Victor-w-Madeira:
The cut happened on 2026-09-20: the VM daily opens the umbrella issue and owns the verdict. `daily-stable.yml` still strips `@stable` from its own hard failures and commits that to main.

## Why that contradicts the cut

The tag decides **what the authoritative lane runs**. Two cases:

- **Red on the VM, green on Actions** — nobody removes; the spec keeps failing and the umbrella keeps naming it. Noisy, visible, tolerable.
- **Green on the VM, red on Actions** — Actions removes the tag. A lane whose verdict no longer counts has just **shrunk the source of truth's selection**, and the VM's next verdict is computed over a suite it did not choose. Coverage leaves without the lane that owns the verdict asking for it.

The rule was already written for this stage: the tag removal belongs to exactly one lane. The cut decided which.

## Change

Stop calling `auto-remove-stable` from `daily-stable.yml`, and drop the three env keys that read the removed step's outputs — a dangling reference renders as empty, and empty is what "the removal ran and found nothing" looks like.

The action itself stays: `weekly-stable.yml` calls it, and the VM lane will once task 4 gives it a write path.

## The cost, accepted and dated

Until then **nothing removes a tag automatically**. A genuinely broken spec stays in the selection and keeps the VM red every morning, named by the umbrella, until someone removes the tag in a PR. Noise a human sees, rather than coverage that leaves quietly.

Issue #1822, opened 2026-09-11
Auto-removal of `@stable` reports a removal it never commits — the commit stages one path prefix, so 28 tests in 5 files are unreachable


Victor-w-Madeira:
Found during the triage of daily-failure #1821 (run [34599745145](https://github.com/oriontech-me/langflow-e2e/actions/runs/34599745145), 2026-09-11). Not a Langflow failure — a defect in our own auto-removal path.

## Symptom

The daily's `Auto-remove @stable from hard failures` step reports a removal it does not persist. On run 34599745145 the umbrella issue, the commit message and the script's own JSON all say two tests lost `@stable`; on `main` only one did.

`scripts/remove-stable-from-failures.ts` edited both files and said so (merge job [103272869654](https://github.com/oriontech-me/langflow-e2e/actions/runs/34599745145/job/103272869654), step `Auto-remove @stable from hard failures`):

```json
{"status":"removed","threshold":5,"hardFailures":2,"attributableFailures":2,
 "removed":[
   {"file":"tests/collect-models.spec.ts","title":"collect providers status and models from UI","line":39,"soleTag":false},
   {"file":"tests/tests-automations/regression/core-functionality/model-provider/anthropic-provider.spec.ts","title":"Anthropic API key is configured via Settings → Model Providers","line":184,"soleTag":false}],
 "skipped":[],"exempt":[],"disagreements":[]}
```

The next step then stages a fixed path prefix — `.github/actions/auto-remove-stable/action.yml:112`:

```bash
git add tests/tests-automations/regression QA-CHECKLIST.md
```

`tests/collect-models.spec.ts` is not under that prefix. Its edit was made in the workspace and dropped without ever being staged. Commit `883047fc` touches exactly two files (`QA-CHECKLIST.md` and `anthropic-provider.spec.ts`) while its message reads *"auto-remove @stable from 2 hard-failing test(s)"*, and `tests/collect-models.spec.ts:41` still carries `@stable` on `main` today.

Nothing failed: the step is green, `git commit` had a non-empty staging area from the other file, and no output mentions the discarded edit.

## Why it matters

[...]

Issue #1763, opened 2026-09-08
The infra-signature exemption cannot see an assertion-shaped failure caused by a measured outage — a recurrent flake proposed for quarantine on a shard measured 41% down


Victor-w-Madeira:
Spun out of daily-failure triage #1757 (run [34227075296](https://github.com/oriontech-me/langflow-e2e/actions/runs/34227075296), 2026-09-08). Not a per-spec cause issue — the subject is what the infra-signature exemption can **see**, which is why it carries `qa-infra` and not `daily-failure`.

Adjacent to #1589 and #1626, and **separable from both**; the boundary is drawn explicitly at the end, because the natural reading is that one of them already owns this and neither does.

## What was measured

`agent-system-prompt.spec.ts:213` ("Agent Instructions are respected in the model response", `@stable @release @agents @playground`) came back from the triage dataset as `actionable: true`, `recurrence.count: 2`, `same_signature: true`, `infra_signature: null` — i.e. a recurrent flake requiring **quarantine** under `CONTRIBUTING.md`'s flake criterion. The automated proposal on #1757 listed it for exactly that.

It is wedge collateral. Its two failing attempts on this run:

| Attempt | Span (UTC) | Error | Overlap with a measured outage on its shard |
|---|---|---|---|
| 0 | `12:47:08.490` → `12:49:19.983` (131.5 s) | `expect(received).toBe(expected)` — `Expected: true / Received: false`, `Timeout 30000ms exceeded while waiting on the predicate` | outage `12:47:11`→`12:48:59` — **108 of 131 s (82 %)** |
| 1 | `12:49:21.100` → `12:51:29.976` (128.9 s) | `TimeoutError: page.waitForResponse: Timeout 30000ms exceeded` | outage `12:49:27`→`12:51:19` — **112 of 129 s (87 %)** |

It ran on **shard 2**, and shard 2's service-container log makes the cause exact rather than circumstantial: its **five outage windows contain exactly five gunicorn `WORKER TIMEOUT`s, 1:1** — `12:46:28`, `12:48:44`, `12:51:00`, `12:53:19`, `12:55:35` — spaced a near-constant **~136 s** apart (the pinned `LANGFLOW_WORKER_TIMEOUT: "120"` plus the ~12 s worker restart, with `LANGFLOW_WORKERS: "1"`). The shard spent its entire measured span in a wedge-kill-restart loop, at 41 % down. Both of this test's failing attempts sit inside that loop.

[...]

Issue #1725, opened 2026-09-05
run-e2e.sh: four more 0/1 switches fall the unsafe way on a typo


Victor-w-Madeira:
Follow-up from the review of #1724 (item 2). `require_flag` shipped there for `KEEP_LEDGER` and `USE_LEDGER_DURATIONS`, and the criterion it states — *apply it where a typo'd truthy walks past a fatal gate; skip it where the typo leaves the switch off, which is the safe direction* — selects four more that the PR left unvalidated.

Measured on `main`:

```
$ DRY_RUN=yes bash -c 'source scripts/run-e2e.sh;
    if [ "$DRY_RUN" = "1" ]; then echo "stops after partition"; else echo "RUNS THE WHOLE SUITE"; fi'
RUNS THE WHOLE SUITE
```

| switch | what a typo'd truthy does |
|---|---|
| `DRY_RUN` | the run does **not** stop after the partition — a dispatch that asked for a dry run gets a full ~50 min suite against the target |
| `REQUIRE_TARGET_VERSION` | enforcement silently off, on a flag whose own error text argues *"an unperformed check is not a weaker guarantee — it is none"* |
| `CHECK_TARGET_VERSION` | the version resolution is skipped, and the run reports an unperformed check |
| `PREPARE_TARGET` | the clone is not placed; also feeds `STAMP_REQUIRED`, so the starter stops demanding the build stamp too |

The fix is the same one call per switch that #1724 used, next to the two that already have it. Kept out of that PR because #1723 was scoped to the ledger.

`run-e2e.sh`'s comment above `require_flag` names these four as known and deferred here, so the two calls below it do not read as the complete set.

PR #1946, opened 2026-09-21
feat(run-e2e): give the lane that owns the verdict the write path for @stable

### LINKED ISSUE #1945: VM lane: the @stable removal has no write path, so since the cut nothing removes a tag at all
Since the cut (#1943, PR #1944) **nothing removes a `@stable` tag automatically.** A genuinely broken spec stays in the selection and keeps the VM red every morning, named by the umbrella, until someone removes the tag by hand in a PR. That cost was accepted with a date attached, and this issue is the date: it is the critical path of stage 2.

> **Corrected on 2026-09-20** — the first version of this body claimed this lane writes no corroboration file and that `AUTO_REMOVE_OUTCOME` had left the repo. Both were read off a stale worktree and both are false; the measurement is in the [correction comment](https://github.com/oriontech-me/langflow-e2e/issues/1945#issuecomment-5754641306) and the scope below is the corrected one.

## What already exists, and it is more than the task needs to build

`remove-stable-from-failures.ts` is shared, carries every safeguard — the mass-failure guard, the infra-signature exemption (#1031), the corroborated-earlier-attempt exemption (#1589) — and **pushes nothing**. `format-auto-remove-summary.mjs` renders the umbrella block. `auto-remove-commit-paths.mjs` stages narrowly and verifies the commit against its own report.

And on this lane specifically: the corroboration file is **already written and already read** (`run-e2e.sh:1660` writes it, `:1932` hands it to the history appender), and the umbrella's three readers are live — `create-failure-issue.mjs` consumes `AUTO_REMOVE_STATUS`, `AUTO_REMOVE_SUMMARY` and `AUTO_REMOVE_OUTCOME`, the last one deciding whether a reported removal is relabelled as uncommitted.

## What is missing

**1 · The credential, and it is the only third-party clock.** The write cannot go to the checkout's own remote: the VM reads from the destination mirror, the mirror is `+refs/*:refs/*` with prune, and a real commit there makes the sync guard record `diverged` and stop pushing — worse than reverting, and not fixable by loosening the guard. So the write goes to the **source**, as a second remote (option A, decided 2026-09-18).

A fine-grained token with `Contents: write` on this repo is installed on the `qa` as `SOURCE_PUSH_TOKEN` (mode 600, beside the destination's) and is **pending org approval**: it authenticates, and `git push --dry-run` answers `403 Permission to oriontech-me/langflow-e2e.git denied`. That dry run is also the probe for when approval lands — it writes nothing.

**2 · The call, which genuinely does not exist.** `run-e2e.sh` references neither the remover nor the summary formatter. It goes in the publish phase, **before** the issue step so the umbrella can report the result, default off (`AUTO_REMOVE=0`) and compared strictly against `"1"` — four switches already fall the unsafe way on a typo (#1725), and this one commits to `main`. It passes `OUTAGE_ATTEMPTS` (the file is there) and `BACKEND_WEDGED` gated on `measured`, as the action's input documents.

**3 · The commit and push, and this is where the two lanes must differ.** The coverage regeneration, the narrow staging and the path verification carry over unchanged. The **push** does not: the action pushes bare on purpose, because its sibling history step has just pushed and `main` barely moves in that window. Here the window is a whole sync cycle — the checkout comes from the mirror and lags. So the commit is replayed onto the source's `main`, and a rebase that does not apply cleanly **must not push** and must say so on the umbrella. A silent skip is the same class of silence as the stalled sync.

**4 · The three states, which inline makes easier.** The orchestrator knows the push result directly, where a composite could only infer it from a step outcome that survives an embedded failure — #1822's whole defect. Removed and pushed, removed nothing, removed but could not push: three distinct states, and **empty must never mean failed**.

## How it gets validated

By a provoked hard failure (task 6), **on the branch, before merge**. PR #1932 passed CI and review and still took a defect to `main`; what caught it was running the lane. The rehearsal fetches the branch from the source, so merging is never a prerequisite for validating. Parts 2 to 4 can be built and merged inert while the token waits.

## Out of scope

Option D (the VM fires `repository_dispatch` and a minimal Actions job removes) stays as a net, not a path: it would leave the authoritative lane unable to close its own cycle without Actions. Option B died on 18/09 with the decision to archive the source; option C stays discarded because it puts the laptop back on the consequence path.

Victor-w-Madeira:
Closes #1945 once it is turned on. **Do not merge before the provoked run** — see *The gate* below.

## Why

Since the cut (#1943, PR #1944) **nothing removes a `@stable` tag automatically**. The tag decides what the authoritative lane runs, that lane is the VM, and the VM had no write path. A broken spec now stays in the selection and keeps the morning red until someone takes the tag off by hand.

## What is reused, and it is most of it

| piece | where it comes from |
|---|---|
| the decision (mass-failure guard, #1031 exemption, #1589 corroboration) | `remove-stable-from-failures.ts`, unchanged — it pushes nothing |
| the umbrella block | `format-auto-remove-summary.mjs`, unchanged |
| staging from the report + verifying the commit against it | `auto-remove-commit-paths.mjs`, unchanged (#1822) |
| the per-attempt corroboration file | already written on this lane since #1763 — passed in, because absent degrades to the last-attempt rule, which removes **more** |

## What is genuinely different: the push

The action pushes bare, and its comment says why — its sibling history step has just pushed, so `main` hardly moves in that window. **Here the window is a whole mirror cycle.** This clone reads from the destination, which trails the source by up to an hour, so the commit is replayed onto the source's `main` before it is sent.

And the write goes to the **source**, never to this clone's `origin`: a commit in the destination is content the source does not have, the sync guard records `diverged` and **mirroring stops** — worse than reverting, and not fixable by loosening the guard.

**Fail-closed in both directions, and the branch is restored on every failure after the commit.** Not tidiness: the wrapper runs `git pull --ff-only` the next morning, so a commit that never reached the source would make it refuse and break the daily on a machine nobody is watching. Pinned by two behavioural tests — one that pushes into a real bare repository, one that makes the replay conflict and then asserts `HEAD` and the working tree are exactly as they were.

## The umbrella still hears the truth

`AUTO_REMOVE_STATUS` + `AUTO_REMOVE_OUTCOME` separate **removed and pushed** from **removed nothing** from **removed and could not push** (`arLost` relabels it). Empty still means "not tracked", never "failed". Running inline is what makes this honest: the composite could only infer the push from a step outcome that survives an embedded failure, which was #1822's own defect.

## Two test changes, named rather than buried

- The pin that read *"writing back to the repository is absent, not merely switched off"* was correct until the cut. **Rescoped, not deleted:** exactly one write path, every git-write line inside it, reachable only behind the switch, on a red day, with a reported removal, and with a single caller. `COMMIT_HISTORY` stays absent.
- A test first written as a `set -e` pin **was measured to protect nothing** — `[ … ] && x=…` failing *mid-function* does not abort under errexit, and the test passed with the idiom reintroduced. It now states what it actually covers.

## The gate

`AUTO_REMOVE` defaults to `0` and is compared strictly against `"1"` (#1725 — four switches already fall the unsafe way on a typo, and this one commits to `main`). It is a switch rather than absent code, unlike the ledger commit the header note argues about, because what is missing is a **credential**, days away, not a later etapa's design.

Before this is turned on:

1. The fine-grained token (`Contents: write`) installed on the machine as `SOURCE_PUSH_TOKEN` is **approved** — today it authenticates and `push --dry-run` answers `403`, which is the probe for when that changes.
2. A **provoked hard failure** (task 6) runs the whole path on the branch. PR #1932 passed CI and review and still took a defect to `main`; what caught it was running the lane.

## One decision worth a reviewer's eye

The commit identity is the lane's (`langflow-e2e vm daily`), not a person's, overridable by env. Attributing it to the token's owner is the other legitimate choice.

## Test

`npm run test:scripts` 1745 pass · `npm run test:units` 1649 pass. Both behavioural tests fail against `main` by construction — the functions do not exist there.


Victor-w-Madeira:
Independent review run against this branch before it is turned on. **Eight findings, all eight addressed in `cee506d4`** — and the first one alone paid for the review.

**1 · The replay was unbounded (critical).** `git rebase FETCH_HEAD` replays everything from the merge base, so on a clone whose HEAD the source does not already contain — **precisely the shape the gate schedules**, since the provoked failure runs on a branch — it lifted that entire branch onto the target and the push landed all of it, unreviewed, with `[skip ci]` on the tip so no lane would have looked at it. The commit's parent is `orig_head`, so the source having `orig_head` already is what bounds the push to one commit; anything else is refused by name. `SOURCE_PUSH_BRANCH` arrives with it, because rehearsing the push onto `main` is not a rehearsal.

**2 and 3 · Two exits left the clone dirty**, which is the exact failure the function's own header says the restore discipline exists to prevent. A failed `coverage:summary` returned without restoring, after the remover had rewritten the specs; and a remover that dies part-way — it writes file by file — returned 0 with the edits on disk. Either one makes the next morning's `git pull --ff-only` refuse, on a machine nobody is watching.

**4 · A crashed remover was invisible.** Empty status renders nothing downstream, and empty means "not tracked", never "failed" (#1822). The Actions step turned the job red; this lane is fail-soft, so the only trace was a `warn`. It now reports `error` with a summary the umbrella renders.

**5 · The identity did not survive the rebase**, which re-creates the commit — so the committer reaching the source was the machine's, or, on a machine with none, `unable to auto-detect email address` and a removal failing closed every morning for a reason nobody would connect to git config.

**6 · The credential was on the command line.** `-c http.extraheader=…` sits in `/proc/<pid>/cmdline` for any local user while the call lasts — wider than the URL form the comment rejected. It goes through `GIT_CONFIG_*` now, and the comment says what the code does.

**7 · Errexit is suppressed inside the function**, since it is invoked as `… ||`. An empty `orig_head` would have made all eight restores silent no-ops; an empty count would have shipped `auto-remove @stable from  hard-failing test(s)` to `main` for good. The three assignments check themselves.

**8 · `AUTO_REMOVE=1` with `CREATE_ISSUE=0`** removes tags with nowhere to report it. It warns, and the gate below says the order.

Three behavioural tests came with the fixes, each **verified by reintroducing the defect**: the branch that must not be pushed, the failed regeneration, and the remover that dies half-way. `npm run test:scripts` 1748 pass · `npm run test:units` 1649 pass.

**The gate is unchanged, and now has an order:** token approved → `CREATE_ISSUE=1` (already on) → provoked failure on a branch with `SOURCE_PUSH_BRANCH` pointing at a scratch branch → only then `AUTO_REMOVE=1`.

Victor-w-Madeira:
## Rehearsed on the `qa`, 2026-09-21 — and it cost two defects

Run on a worktree of the production clone checked out at this branch, with a provoked hard failure naming a real `@stable` spec (`tests/fixtures/locale-gate.spec.ts`), `SOURCE_PUSH_BRANCH` pointed at a scratch branch on the source. The production clone was never touched: `HEAD` and the working tree are byte-identical before and after, and the worktree is gone.

**The whole path ran:** the remover decided `removed` with its real safeguards → `coverage:summary` regenerated the checklist → the narrow staging derived from the report → the commit → the two-way verification → the fetch of the source → the ancestor guard → the replay → the push, refused with

```
remote: Permission to oriontech-me/langflow-e2e.git denied to Victor-w-Madeira.
fatal: … The requested URL returned error: 403
```

which is the pending token, exactly as expected — and then **the restore: `HEAD` unchanged, tree clean.**

### Two defects, both found only by running it

**1 · A dirty clone lost the removal at the last step** (`a11fec24`). With one tracked file modified, everything above ran — the remover, the checklist, the staging, the commit, the verification — and `git rebase` then refused with *"cannot rebase: You have unstaged changes"*. All the work, discarded at the end. A clean tree is now a precondition, checked **before** the remover writes anything; refusing costs the day's removal and says so, where `--autostash` would have let the lane juggle someone else's in-flight work and a later `reset --hard` would have thrown it away.

**2 · The umbrella pointed at a surface this lane does not have** (`9cec6a41`). Rendered with `ISSUE_DRY_RUN=1` on the real body, the relabel was correct —

> ### ⚠️ `@stable` auto-removal did NOT reach `main`
> **Nothing was pushed.** … treat the list as what this run TRIED to quarantine rather than as what it did

— and then it told the reader to open the `Auto-remove @stable from hard failures` **step**, which exists only on Actions. The VM runs the removal inline; its cause is in the run log beside the evidence. #1822's defect one level down: the correction only lands if its words survive the lane.

And one incidental confirmation, from the first attempt: the crash path put the clone back so thoroughly that it deleted the copied script the harness was running from. The restore discipline works.

### What is still unproven

**The push that succeeds.** It needs the approved token, and it is the only step left. Re-running is one command, and the evidence and scripts are kept at `/root/rehearsal-1946/` on the `qa` (`run2.sh` the path, `run3.sh` the path plus the rendered umbrella).

PR #1948, opened 2026-09-21
feat(mirror): ask whether the suite this lane checked out is still current

### LINKED ISSUE #1947: A stalled mirror is silent: nothing asks whether the suite the VM checked out is still current
The VM checks out from the destination mirror, and the mirror is pushed from a laptop on a schedule. **When that stops, nothing says so.** The sync does not fail — it just does not push, and the lane keeps running a suite quietly older than `main`.

It has already cost a measured two days. The guard tripped on **43 consecutive runs** from 2026-09-16 09:49 to the night of 18/09; the destination stayed frozen at `7673d123`, and the VM ran a two-day-old suite against a moving comparison — 682 tests here against 723 there on 18/09. Both days are recorded as measured and **not comparable**, and the only reason anyone found out is that someone walked past it. That is the fifth time the same silence has been caught the same way.

The narrow remedy for the *guard's* false positives landed on the machine on 20/09 (orphan refs are pruned and announced instead of aborting). **This is the other half:** the guard now trips far less often, which makes a real stall rarer *and quieter*.

## The question this has to ask

Not *"did the last sync report an error"* — a sync that never ran reports nothing at all, and that is the failure mode that actually happened. The question is **"does the destination's `main` hold what the source's `main` holds?"**

And the age that decides whether a difference is a stall or an ordinary cycle is the age of the **oldest commit the destination is missing**, not the age of its own tip. A mirror whose tip is recent but which has stopped following reads as young under the wrong measure — the end-to-end test caught exactly that in the first draft.

## Where it belongs

On the **`qa`**, not the laptop: it reaches the destination (its own `origin`) *and* the public source — measured on 18/09 — and it is always up, which the laptop is not. Two callers, in order of value:

1. **The daily's preflight**, so a run that is about to compare an old suite says so in the artifact the triage reads, instead of leaving it to be reconstructed two days later.
2. **A timer of its own**, so a stall is caught on the days the daily does not run.

## Fail-closed

If the source cannot be read the answer is UNKNOWN, never "current" — a network blip must not read as a clean bill of health, since "current" is precisely the answer that lets a stale suite run unremarked.

Victor-w-Madeira:
> **Status: ready to merge — held for one reason, and it is a calendar one.**
>
> **Reviewed twice, independently**, and both rounds are answered in the comments below: the first over the checker (7 findings), the second over all of it including the wiring (8 findings). Every one of them is fixed in this branch, and the two that mattered most in each round were verified by reintroducing the defect.
>
> **Exercised on the `qa` itself**, in the production clone, on both paths: tips equal → `ok: the mirror is current at a50cdc3a`, exit 0, **0.96 s**, `FETCH_HEAD` untouched; tips different → `BEHIND: 389 commit(s) behind, tip 48041 min old`, exit 1, **1.09 s**. Working tree clean before and after in both cases.
>
> **What it waits for:** today's daily, 08:00 UTC. That run is already carrying two first-times — the first morning where this lane's verdict has consequence, and the first on the consolidated machine. Landing a third change into the same morning would mean that a divergence, if one appears, has three candidate explanations instead of one. Merging after it costs nothing; merging before it buys nothing.
>
> Installing the timer on the machine is a separate step, and its two commands are in the unit's own header.

Closes #1947 once it is wired. **This commit adds the check and nothing calls it yet** — see *Why nothing calls it* below.

## Why

The lane reads from the destination mirror, and the mirror is pushed on a schedule. When that stops, **nothing says so**: the sync does not fail, it stops pushing, and the run goes on against a suite quietly older than `main`. Two days are already recorded as measured and **not comparable** for exactly this — 43 stalled syncs from 2026-09-16, 682 tests here against 723 there on the 18th — and all five occurrences of this silence were found by someone walking past it.

## The question it asks

Not *"did the last sync report an error"*: a sync that never ran reports nothing, which is the failure mode that happened. It asks **"does the destination's `main` hold what the source's `main` holds"** — an answer with no dependency on the syncing side's own bookkeeping.

## Two things the tests corrected, both worth naming

**The age has to be the age of the divergence, not of the tip.** The first draft measured the destination's own tip and the end-to-end test failed immediately: a mirror whose tip is recent but which has stopped following reads as *young* — the exact state this exists to catch. It now measures the **oldest commit the destination is missing**, because "how long has the source been ahead?" is the question and the tip cannot answer it.

**An unreachable source is UNKNOWN, never current.** A blip must not read as a clean bill of health: "current" is the answer that lets a stale suite run unremarked, and a check that degrades into silence has not removed the silence it was built for.

## The two callers, wired in `01dd6a8b`

**The daily's preflight**, so a run about to compare an old suite says so in the evidence directory instead of leaving it to be reconstructed two days later. ON by default — it changes nothing about the run — and **fail-soft**: a stale suite still produces a valid run of that suite; what is compromised is the *comparison*. Dying there would trade a day of data for a warning.

**An hourly timer**, because the daily does not run at weekends and the stall of 2026-09-16 began on a Wednesday and survived to the Friday. `Persistent=true`, so a machine that was asleep asks once on the way back.

The alarm around it is a **change notifier, not an hourly repeat**: a stall lasts days, and 24 identical messages is how a channel learns to mute the twenty-fifth. It speaks on the transition and again on recovery — the recovery message is what closes the earlier alarm without anyone going to look. `UNKNOWN` counts as not-current, for the same reason the check itself refuses to call it healthy.

Two more defects the tests caught there: a `CHECK_CMD` **string** is word-split when expanded, so a stub carrying quotes arrived mangled and the shell's own parse error travelled into the notification as though it were a verdict (it takes an executable path now); and the first observation on a fresh state file announced itself, which is how an alarm teaches people to dismiss it.

The units live in `ops/systemd/` rather than only in `/root` — the debt task 3 is about. Installing is two commands, in the unit's own header.

## Test

8 tests: every branch of the verdict, including the two that must never read as current, plus an end-to-end pass against two real repositories (equal tips, then a source that moves without the mirror). `npm run test:scripts` 1758 pass.

Verified against the real remotes too, from a clone whose `origin` is the destination: `[mirror] ok: the mirror is current at e5c391fa`, exit 0.


Victor-w-Madeira:
Independent review run before this is wired to anything. **Seven findings, all seven addressed in `4a4ec1ba`** — and the first would have made the check worse than absent.

**1 · It read the writing dates, not the landing dates (high).** `rev-list A..B` includes every commit a **merge** brought in, dated when the branch was written. This repository merges with merge commits, so an ordinary merge of a day-old branch made the mirror read as a stall *the instant it landed*. Measured on the last 25 merges: **12 carried a commit older than the 120 min window** — confirmed on the tip of `main`, where the oldest commit the merge brought in is `02:47` and the merge itself is `03:34`. Half the merges would have declared the day not comparable for an hour, which is precisely how a check teaches everyone to ignore it. `--first-parent` walks what reached `main`, and its dates are landing times.

**2 · The main-module guard was the fragile template, and its failure is silent.** `file://${argv[1]}` stops matching on a percent-encoded path or a symlinked ancestor — and then the script prints nothing and exits 0, which reads exactly like *"the mirror is current"*. Replaced with the `pathToFileURL(realpathSync(...))` form `check-run-integrity.mjs` already documents, and pinned by running the script from a directory with a space in its name.

**3 · A destination that is AHEAD was reported as "0 commit(s) behind, of unknown age"** — wrong, and indistinguishable from "could not walk the history". Divergence is its own verdict now, with its own exit code: it is the state the sync guard exists to catch, not a lag.

**4 · No timeout, no batch-mode env on network git calls.** An unknown host key or a credential prompt blocks on a tty an unattended timer does not have. A check built to remove a silence must degrade to `UNKNOWN`, not hang.

**5 · "Oldest last" assumed an ordering git does not guarantee.** `Math.min` over the dates, so a skewed committer date cannot understate a stall.

**6 · `MAX_LAG_MINUTES` was unvalidated.** `Number("18O")` is NaN, every comparison with it is false, the young-window branch becomes unreachable and the output reads *"past the NaN min window"*. Now named and replaced with the default, out loud.

**7 · The end-to-end test isolated git config for its own commands but not for the code under test**, which shells out with the inherited environment.

Four tests added. Two verified by reintroducing the defect — the merge that must not read as a stall, and the script that must speak from a path needing escaping. `npm run test:scripts` **1762 pass**, and against the real remotes: `[mirror] ok: the mirror is current at e5c391fa`.

The gate is unchanged: nothing calls this yet, and the two callers (the daily's preflight and a timer on the `qa`) are the next commit.

Victor-w-Madeira:
Second independent review, this time over all three commits — the wiring included, which the first review never saw. **Eight findings, all eight in `0da0be32`**, and the first one had hollowed the thing out.

**1 · The state was written before the notification was attempted (high).** One failed POST consumed the transition, and every later run said *"unchanged — not repeating it"*. A Friday-evening stall plus one bad minute of network is exactly the weekend-long silence this issue exists to remove — restored by the mechanism built to remove it. The state file now carries **two** things, what was **observed** and whether it was **announced**: observation alone cannot express *"seen and not said"*, which is the shape both a failed delivery and a first UNKNOWN leave behind.

**2 · `curl -sS` exits 0 for a 404 as happily as for a 200 (medium).** A rotated webhook answering `no_service` looked exactly like a delivered message. The HTTP status decides now.

**3, 4, 5 · Three on the checker, one cause: the walk trusted state it did not control.** `FETCH_HEAD` is shared mutable repo state and the daily fetches `origin` in the same clone — one landing between the fetch and the walk made the destination compare against *itself*, zero commits missing, which reads as DIVERGED: an ordinary stall would have announced *"something wrote to the mirror directly"*. The destination's tip is a local object only while it is an ancestor of the source, so in the one case `EXIT_DIVERGED` exists for, `rev-list` died with *bad revision* and the catch turned it into "behind by an unknown number" — the verdict was **unreachable in the deployed shape**. And a walk that did not happen was reported as a measurement; it is UNKNOWN now. Both tips are fetched and the walk names shas.

**6 · Nothing checked that the two sides are different repositories (low).** A dev clone made straight from the source compares it with itself and answers *"current"* forever — the one answer this module's own header says it must never give by accident.

**7 · One UNKNOWN was announced as staleness (low).** It says nothing now; two consecutive get their own wording, which does not assert what the check refused to measure.

**8 · The unit set `HOME` and not `PATH` (medium).** systemd's default omits `~/.local/bin`, `node` exits 127, the alarm reads any non-zero as not-current, and the channel gets a red siren whose body is *"node: command not found"* — the same shape as the defect the previous commit fixed. Pinned, with this machine's real paths measured.

And one the review found in the wiring: **the preflight asked before the run directory existed**, writing the verdict to the console only, so the evidence directory a triage opens two days later held no answer — the reconstruction this check exists to prevent. It asks after, and tees into `$RUN_DIR/logs/mirror-freshness.log`; captured rather than piped, since through a pipe the exit status survives only by `pipefail` being set three hundred lines away.

`npm run test:scripts` **1772 pass**. Four tests added, and the two that matter were verified by reintroducing the defect: the delivery that must not consume its transition, and the two remotes that must not answer "current".

Victor-w-Madeira:
**Ready to merge; waiting on the clock, not on the code.**

Two independent reviews, both answered above, with every finding fixed on this branch. Exercised on the `qa` in the production clone, both paths, ~1 s each and nothing left behind — the measurements are in the status block at the top of the description.

The hold is deliberate and it is about attribution, not risk: **today's daily already carries two first-times** (first morning with consequence for this lane, first on the consolidated machine). A third change landing in the same run would give any divergence three candidate explanations instead of one. It merges once that run is in.

While waiting it also picked up `f9281f4e`, which fixes a race this branch would otherwise have created in code already on `main`: `auto_remove_commit` reads `FETCH_HEAD` twice, and the timer here becomes the second process fetching into that clone.

PR #1951, opened 2026-09-21
feat(preflight): the umbrella's credential has an expiry date, and now something reads it

### LINKED ISSUE #1950: The umbrella's credential expires on a known date and nothing reads it
The umbrella issue is how a red morning speaks. Since the cut it is born from the VM's run, on the destination, with a token that lives in `/root/.e2e-secrets` — and **that token expires on 2026-12-19 22:56 UTC**, which is known exactly because the GHES API says so in a header:

```
github-authentication-token-expiration: 2026-12-19 22:56:58 UTC
X-OAuth-Scopes: repo
```

**Nothing reads it.** When it expires, `create-failure-issue.mjs` fails and the orchestrator warns — *"issue creation failed (does not fail the run)"* — and the day goes on. The consequence is the shape this lane keeps paying for: a red day whose umbrella never appears looks, to anyone reading the destination, exactly like a day that had nothing to report.

Expiry is a silent failure with a **date on it**, which is the rare kind that can be caught before it happens rather than diagnosed afterwards. The companion already says it should be a preflight check rather than a reminder.

## What to ask, and when

In the preflight, only when `CREATE_ISSUE=1` — the question is about a credential this run intends to use:

- **Rejected** (401/403): the lane cannot deliver the consequence it is about to spend sixteen minutes and real model calls earning. Refuse at the start, with the cause named, rather than discovering it at the end.
- **Expiring soon**: warn with the date and the days left, so it is replaced during a working week rather than on the morning it dies.
- **Unreachable**: UNKNOWN, and a warning — a network blip must not be read as a dead credential, the same rule #1947 landed on.

## Not in scope

Rotating the token. This is the alarm, not the hand.

Victor-w-Madeira:
Closes #1950.

## Why

Since the cut, the umbrella issue is how a red morning speaks — and it is opened with a token that dies on a **known date**. The destination says so outright:

```
github-authentication-token-expiration: 2026-12-19 22:56:58 UTC
```

**Nothing read it.** When it lapses, `create-failure-issue.mjs` fails, the orchestrator warns *"issue creation failed (does not fail the run)"*, and the day goes on. What that leaves is a red morning whose umbrella never appeared — to anyone reading the destination, indistinguishable from a morning with nothing to report. #1012's rule, arriving through the credential instead of through the report.

## Three answers, and the asymmetry between them

| | what it means | what happens |
|---|---|---|
| **USABLE** | the API answered, with the days left | warns early when the end is near, so it is replaced in a working week |
| **REFUSED** (401/403) | the server said **no** | **fatal** when `CREATE_ISSUE=1` |
| **UNKNOWN** | no answer, or any other status | warns only |

Refusal is fatal because the lane would otherwise spend sixteen minutes and real model calls earning a verdict it cannot deliver — and that is knowable at the start, so stopping costs nothing that was going to arrive. UNKNOWN is not, because a blip must not cost a day of data. Same asymmetry #1947 landed on.

## Two details worth a reviewer's eye

The header **is not ISO-8601** — a space instead of `T`, and a trailing ` UTC` — so it is parsed rather than handed to `Date.parse`. And a token with **no expiry** is reported as exactly that, not as unknown: calling a real configuration "unknown" is how a reader learns to skip the line that matters.

The wiring reads `PIPESTATUS[0]`, because the verdict is tee'd into `$RUN_DIR/logs/issue-credential.log` and through a pipe the exit status belongs to `tee`. Read from the pipeline, every refusal would have looked like a success.

## Test

Eight unit tests: every branch of the verdict, the header parsing, the no-token short-circuit (which must not go to the network to learn what it already knows), and an end-to-end through a stubbed API that checks the enterprise path and the bearer header.

**Measured on the `qa`, against the real destination:** `[credential] ok: the credential works and expires in 89 day(s), on 2026-12-19`, **0.6 s**.

Two unrelated timing tests (`start-langflow-source`, `start-ollama-source`) flaked under the full lane and pass on their own — noted, not touched.

## Not in scope

Rotating the token. This is the alarm, not the hand.


Victor-w-Madeira:
Independent review before merge. **Nine findings, all nine in `af05bbcf`** — and the first one is why the review was worth the five minutes.

**1 · The fatal half was dead code (high).**

```bash
node scripts/check-issue-credential.mjs 2>&1 | tee "$log" || true
cred_rc="${PIPESTATUS[0]}"
```

`|| true` **runs a command**, and every command resets `PIPESTATUS`. Pipeline failed → `true` ran → `(0)`. Pipeline succeeded → `true` skipped → `(0 0)`. Either way `cred_rc` was `0` and `die` could never fire. Measured in a shell rather than argued — along with the two forms that do work (`if ! pipeline; then rc=${PIPESTATUS[0]}` does preserve it; the review's note that it would not is the one thing in the report that did not survive checking). The capture-to-a-file form is what shipped, because it removes the question instead of answering it.

**And the test that was supposed to protect this grepped the source for the literal `PIPESTATUS[0]`** — it asserted the text was present, not that the status was read. It is behavioural now, with the check stubbed: refused stops the run; "could not tell" and node-crashed do not; both are audible.

**2 · No token was called REFUSED (high).** `create-failure-issue.mjs` documents and implements a `gh` fallback for exactly that case — so with the fatal path repaired, this would have killed a lane that could still open the umbrella. It is UNKNOWN now.

**3 · Refusal shared exit 1 with node crashing (medium).** A syntax error or an unhandled rejection would have aborted the daily with a message blaming the token. Refusal is **3**; 1 is left to node and lands in the warn branch.

**4 · 404, not 403, is the likeliest refusal (medium).** GitHub answers 404 for a repository a credential cannot see — a token that lost access. It sat in UNKNOWN, warning while the umbrella went missing.

**5 · The precedence was inverted** relative to the consumer (`GITHUB_TOKEN || GH_TOKEN` there, the reverse here): on a lane where the two differ, this validated a credential the creator never uses — the silent failure it exists to close, in its own uniform.

**6 · The catch cleared a status that was already known**, downgrading a real 401 to UNKNOWN if anything threw after the answer arrived. Only the call is guarded now.

**7 · An expired credential still reported USABLE**, printing *"works but expires in -3 day(s)"* under clock skew. It is a refusal, and it says how long ago.

**8 · EXPIRING and UNKNOWN never reached `warn`** — the notice whose only job is being noticed was one stdout line among thousands instead of a `::warning::`.

**9 · The end-to-end test asserted inside the fetch stub**, where `main`'s catch swallows the throw into "no answer": the message naming a broken enterprise URL could never reach a reader. Recorded outside, asserted after.

What this still does not prove is named in the header rather than papered over: `GET /repos/…` establishes **read** access, and opening an issue needs write. Proving write means opening one.

Re-measured on the `qa` after the refactor: `[credential] ok: the credential works and expires in 89 day(s), on 2026-12-19`. `npm run test:scripts` **1764 pass**, and the dead-code defect was reintroduced on purpose to confirm the new test fails on it.

